PR #330, opened 2014-05-06
ENH: new api_methods - total_commission and cancel_all_open_orders 

humdings:
added 2 api_methods
cancell_all_open_orders() - cancels any open orders in the blotter.
total_commission() - gets the sum of the commission costs since the start of the backtest.

added method all_raw_orders() because the raw_orders() method only returns the open orders.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -788,3 +788,25 @@ class TradingAlgorithm(object):
             bar_count, frequency, field, ffill)
         history_spec = self.history_specs[spec_key_str]
         return self.history_container.get_history(history_spec, self.datetime)
+
+    @api_method
+    def cancel_all_open_orders(self):
+        open_orders = self.get_open_orders()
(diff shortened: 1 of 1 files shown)


humdings:
@ssanderson This can probably be closed, I could open a new PR for `cancel_all_open_orders` if we still think it would provide added utility. The commissions portion is probably not worth the computational overhead.

The only issue I could see with it would be in the case of orders that have been partially filled already. If I did `order_target(stock, 0)` and that order was only partially filled when I call `cancel_all_open_orders`, I would be left with a position in that stock. I'm not sure if this would just cause more confusion.

I think a more worthwhile endeavor would be to implement realistic 'time in force' models for orders. Unfilled orders should be cancelled at EOD unless they're tagged with 'good till cancelled' because that's how it works in practice.

twiecki:
Closing this then.

PR #353, opened 2014-06-20
Enable automated testing with Python 3.4

frewsxcv:


dalejung:
I think https://github.com/quantopian/zipline/pull/511 supercedes this now. I need to get 3.4 binaries up to the binstar and then 3.4 should run clean.

PR #398, opened 2014-09-28
Added load_from_google method

MarkDunne:
Adds the option to load finance data from google instead of yahoo

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -16,7 +16,7 @@
 from unittest import TestCase
 from zipline.utils.factory import (load_from_yahoo,
                                    load_bars_from_yahoo,
-                                   load_from_google)
+                                   load_bars_from_google)
 import pandas as pd
 import pytz
 import numpy as np
@@ -58,11 +58,11 @@ class TestFactory(TestCase):
             start=end, end=start
         )
 
-    def test_load_from_google(self):
+    def test_load_bars_from_google(self):
         stocks = ['AAPL', 'GE']
         start = pd.datetime(1993, 1, 1, 0, 0, 0, 0, pytz.utc)
         end = pd.datetime(2002, 1, 1, 0, 0, 0, 0, pytz.utc)
-        data = load_from_google(stocks=stocks, start=start, end=end)
+        data = load_bars_from_google(stocks=stocks, start=start, end=end)
 
         assert data.index[0] == pd.Timestamp('1993-01-04 00:00:00+0000')
         assert data.index[-1] == pd.Timestamp('2001-12-31 00:00:00+0000')
@@ -70,6 +70,6 @@ class TestFactory(TestCase):
             assert stock in data.columns
 
         np.testing.assert_raises(
-            AssertionError, load_from_google, stocks=stocks,
+            AssertionError, load_bars_from_google, stocks=stocks,
             start=end, end=start
         )
--- a/zipline/data/loader.py
+++ b/zipline/data/loader.py
@@ -374,13 +374,22 @@ def load_bars_from_yahoo(indexes=None,
                 panel[ticker][col] *= ratio_filtered
     return panel
 
-def load_from_google(indexes=None,
-                    stocks=None,
-                    start=None,
-                    end=None):
+def load_bars_from_google(indexes=None,
+                         stocks=None,
+                         start=None,
+                         end=None):
     """
-    Loads price data from Google into a dataframe for each of the indicated
-    securities. Google's prices are adjusted for dividends and splits.
+    Loads data from Google into a panel with the following
+    column names for each indicated security:
+
+        - open
+        - high
+        - low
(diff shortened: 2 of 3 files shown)


llllllllll:
@MarkDunne Is this still something that you are planning on working on?

femtotrader:
I wrote some month ago Pandas unofficial DataReaders (with requests-cache to cache queries)
http://pandas-datareaders.readthedocs.org/

There is a DataReader for Google Finance Daily and Google Finance Intraday
https://github.com/femtotrader/pandas_datareaders/blob/master/tests/test_datareader_google_finance_intraday.py
and
https://github.com/femtotrader/pandas_datareaders/blob/master/tests/test_datareader_google_finance_daily.py

Maybe it could help

twiecki:
@femtotrader This is great! I wonder if we should just refer to your library here and add a few examples of how to fetch different data (including minutely) and run zipline backtests.

femtotrader:
Hello,

First

    $ pip install pandas_datareader

or get latest source version 

    $ git clone https://github.com/femtotrader/pandas_datareaders
    $ cd pandas_datareader

You can do

    from pandas_datareaders import DataReader

    expire_after = 60*60 # cache query in second (see requests-cache doc)
    symbol = "..."
    df = DataReader("GoogleFinanceIntra", expire_after=expire_after).get(symbol, start, end)

`df` is a Pandas DataFrame with minutes Google Finance OHLCV data

see `tests` directory
for example : 
https://github.com/femtotrader/pandas_datareaders/blob/master/tests/test_datareader_google_finance_intraday.py

I'm aware that doc need to be improved.

if symbol is a list of symbol DataReader get method will returns a Pandas Panel

feel free to fork and improve it.

Kind regards

Femto

femtotrader:
Hello,

you might also be interested by 
https://github.com/femtotrader/ig-markets-rest-api-python-library

you can fetch OHLCV historical data (ask, bid and last) from IG Market using `fetch_historical_prices_by_epic_and_num_points` or `fetch_historical_prices_by_epic_and_date_range`
It returns a DataFrame with hierarchical columns (ask, bid and last for level 0 and Open High Low Close Volume for level 1)

Doc also needs to be improved but test file can help to understand how it works
https://github.com/femtotrader/ig-markets-rest-api-python-library/blob/master/tests/test_ig_service.py

I should also improve this code to be able to pass a `session` parameter because there is no cache in this version.

Download data by chunk is probably also an other improvement to make.

`fetch_historical_prices_by_epic_and_num_points` or `fetch_historical_prices_by_epic_and_date_range` should also supports `epic`(symbol) as a list (instead of string) and it should returns a Panel of OHLCV historical data

Kind regards

twiecki:
What's also interesting is that your code uses request cache while we are implementing our own caching mechanism which might then become obsolete (let me know if that's true or not).

So one idea would be make certain `pandas_datareaders` functions available through `zipline.utils.data` and add it as an optional dependency. Is it up on pypi?

femtotrader:
it's also using `requests` package.
For now there is no cache mechanism in `ig_service` (`ig-markets-rest-api-python-library`)
but as I'm using `requests`, it's quite easy to add this using `requests-cache`
I'm quite busy now but I will do it friday or next week.
I'm just add it
https://github.com/femtotrader/ig-markets-rest-api-python-library/issues/2
as a reminder

About `pandas_datareaders` that's a pleasure to see that you want to add it as an optional dependency to `zipline`
it "should" be latest version on pypi

twiecki:
Looks like there's good progress in delineating this with `pandas`. Personally, I'd love to defer responsibility of data loaders to your project which also handles caching and make this a required dependency.

femtotrader:
Thanks Thomas

femtotrader:
Cache mechanism for ig-market REST API works now
see https://github.com/femtotrader/ig-markets-rest-api-python-library

you can cache globally passing session to IGService

    service = IGService(config.username, config.password, config.api_key, config.acc_type, session)
    ig_service.create_session()

or only for a specific method

like

    epic = 'CS.D.EURUSD.MINI.IP'
    resolution = 'D'
    start_date = '2014-12-15'
    end_date = '2014-12-20'
    response = ig_service.fetch_historical_prices_by_epic_and_date_range(epic, resolution, start_date, end_date, session)

Travis CI works fine for Python 2.7, 3.3 and 3.4

same for
https://github.com/femtotrader/pandas_datareaders_unofficial/
but I hope to have cache in official Pandas DataReader which is now a separate package
https://github.com/pydata/pandas-datareader

pip packages are updated

FreddieV4:
Hi @MarkDunne I'm going to go ahead and close this PR as it's been a while since any updates. We're removing `load_bars_from_yahoo()` and so a similar method `load_bars_from_google` would need to be a bundle instead. 

Feel free to open comment here if you'd like to still add this sort of functionality.

PR #416, opened 2014-10-22
MAINT: Make beta calculation robust to missing values.

twiecki:
Risk calculations are robust to nans, except for
beta which calls numpy with the complete list of
algorithm_returns. If nans are present the result
of covar will be nan.

This is fixed by filtering out nans in
algorithm_returns.


twiecki:
Merged with https://github.com/quantopian/zipline/commit/6a41faf474350edf8165811b939d08967963747a.

PR #441, opened 2014-11-20
Daily returns

seongsean:
Changes to calculations to Risk Metrics. 
A few months ago, we made changes to how we calculated the cumulative risk metrics for our backtester (https://www.quantopian.com/posts/corrections-to-cumulative-risk-metrics). Frank Grossman recently pointed out that he found a discrepancy between the 12 month Zipline metrics and the overall risk metrics that the backtester was showing. We found that the difference came from how we were calculating average daily returns. Before we were using cumulative returns divided by the total number of trading days. We found that this served well for short periods of time (e.g. 12 months) but often overstated the returns when it came to backtest periods of greater than a year. Now we've switched to using a method widely incorporated throughout the industry and that's to use the arithmetic mean to calculate daily returns, so simply put, we are taking the average of daily returns to give us back the mean daily return used to calculate our risk metrics. The actual calculations of the risk metrics (Sharpe, Sortino, Information) remain the same as it was before in this post (https://www.quantopian.com/posts/corrections-to-cumulative-risk-metrics), only the annualized average returns used for the calculations have changed. 

With the new way of calculating daily mean return() (daily_returns.mean()), every metric associated will have differing values:
Alpha
Sharpe
Sortino
Information
Beta remains the same


ssanderson:
@dunster @seongsean did anything ever come of this?  It's been sitting unmerged for over a year...

ehebert:
I believe that this has been superseded by using `empyrical` for risk metrics.

PR #445, opened 2014-12-01
schedule function zipline api

llllllllll:


llllllllll:
If you are able to run that on master, then I guess I was fixing a non existent bug.

Issue #455, opened 2014-12-24
Allow external setting of variables in TradingAlgorithm


twiecki:
Currently `TradingAlgorithm.__init__()` calls the user-defined `initialize()` function.

This is problematic as sometimes one would like to set a variable of the algorithm externally. For example during parameter optimization. I think currently it's impossible to do parameter optimization if the parameters need to be accessed from inside `initialize` (e.g. window length).

There's a few ways, these I can think of from the top of my head:
1. Allow passing of `*args` and `**kwargs` to `__init__` which get forwarded to `initialize`.
2. Move the call of `initialize()` from `__init__()` to `.run()`. This way we can create the `TradingAlgorithm` object and set variables which can then get accessed from inside `initialize` and `handle_data`.

CC @ssanderson @ehebert.

dalejung:
Personally, I don't like the algoscript/function api. However, subclassing TradingAlgorithm is messy because it does a lot. Any reason we don't support a simple object api like so?

```python
class AwesomeAlgo:
    def __init__(self, bottom=99, top=100, lag=3):
        self.bottom = bottom
        self.top = top
        self.lag = lag

    def initialize(self):
        set_universe(DollarUniverse(bottom, top))

    def handle_data(self, context, data):
        #do stuff
```

llllllllll:
couldn't you just do:
```python
a = AwesomeAlgo()
algo = TradingAlgorithm(initialize=a.initialize, handle_data=a.handle_data)
```

dalejung:
Sure, I generally work with a wrapper like that. I hate putting code in strings. But this is issue is about first order support as parameter sweeping can already be done via wrappers. 

Conceptually, it feels cleaner if TradingAlgorithm was split up as it's a harness to itself.

twiecki:
@dalejung The algoscript is for use with the ipython magic. The rest of the interface is mainly to get compatibility with Quantopian so that code does not have be rewritten.

But we already have 2 ways of writing algos and I don't think the right solution is a 3rd way :).

dalejung:
@twiecki 1 -> 2 is 100%. but 2 -> 3 is only 50%!  And they get cheaper as we go

twiecki:
@dalejung I don't follow. But the 3rd way is already supported as @llllllllll points out but that's not great UX.

Devs can of course figure out different ways but I think we want something that's more userfriendly.

dalejung:
ah, nm. Was making a joke.

[...]

PR #456, opened 2014-12-26
Support for setting parameters externally

twiecki:
The reason for this change is that it was impossible to
use a TradingAlgorithm that passes in an initialize function
in a parameter optimization. This issue is discussed in #455.

It seems easy enough to just forward any left-over *args
and **kwargs to the user-defined initialize function. As these
are only passed when they are defined we do not break backwards
compatibility with this.

### DIFF
--- a/docs/release-notes/zipline-0.7.1.md
+++ b/docs/release-notes/zipline-0.7.1.md
@@ -88,3 +88,24 @@
     > for stock in tech_stocks:
     >    order_target_percent(stock, 1/3, percent_of_fn=tech_filter)
     > ```
+
+* Forward arguments from __init__ to the user-defined initialize().
+  [PR456](https://github.com/quantopian/zipline/pull/456)
+
+  >  If you used the new way of creating an algorithm by defining an
+  `initialize()` and a `handle_data()` function that you pass to
+  `TradingAlgorithm` it was not possible to externally set variables in `initialize`. This is quite an impediment to parameter optimization where you want to be able to run the `TradingAlgorithm` many times passing in different parameter values that you set in `initialize()`.
+
+  > Example:
+  > ```python
+  > def initialize(context, param=0):
+  >     context.param = param
+  > def handle_data(context, data):
+  >     # use param in some way
+  >     ...
+  > # Instantiate algorithm, setting param to 3.
+  > algo = zipline.TradingAlgorithm(initialize,
+                                    handle_data,
+                                    param=3)
+  > perf = algo.run(data)
+  > ```
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -778,21 +778,18 @@ def handle_data(context, data):
         """Test that passing of args and kwargs to initialize via __init__ works
         as expected."""
         test_algo = TradingAlgorithm(
-            'arg',
             script="""
-def initialize(context, arg, kwarg=False):
-    context.arg = arg
+def initialize(context, kwarg=False):
     context.kwarg = kwarg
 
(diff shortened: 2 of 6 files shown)


ssanderson:
@twiecki sorry, I lost track of this one.

I won't have bandwidth to work on an implementation of this until mid-March I think.  In the shorter term I could probably spend a day or two hashing out what I think is the right design (with input from you and @ehebert, @llllllllll and others of course).

I'm *really* hesitant to use the "we need this soon" argument as a reason to change the API in a way that we agree makes it worse, since that's trading a one-time benefit for a compounding cost.  What's the timescale on which you think you need parameter optimization?

ssanderson:
If we truly do need to do something in the short term, I'd prefer it be something that's explicitly marked as provisional and that's easy to remove in the future.  That's probably easier to do with your explicit `initialize_kwargs` proposal, so I'd probably lean toward that if push came to shove.

twiecki:
@ssanderson We just had our kick-off meeting on that so very soon.

Yeah, I think the `initialize_kwargs` is pretty easy to remove afterwards. Alternatively we could also work on this branch but I'm afraid that it creates some good friction elsewhere in the process.

twiecki:
@ssanderson Any update? I don't see us refactoring this soon and I really need this functionality.

c0indev3l:
Hello,

I'm also looking for such a feature.
It will help to have walk forward optimization for example.

I wonder if passing an additional parameter to algo
parameters={"param1": value1, "param2"; value2}

my_algo = TradingAlgorithm(
    initialize=initialize,
    handle_data=handle_data,

    parameters=parameters

)


Parameters could be available through
context.parameters['param1']

It shouldn't be so hard and it shouldn't break anything.


Kind regards


2015-02-12 14:31 GMT+01:00 Thomas Wiecki <notifications@github.com>:

twiecki:
@c0indev3l That's exactly what this PR would allow. It's pretty simple to do actually. I think I'll just merge this as it's critical functionality and easy to remove later when we actually do refactor this.

warren-oneill:
@twiecki @ssanderson Any word on this? Trying to implement optimization techniques in Zipline is very ugly without this PR and it seems like a feature that quite a few people want. Having a backtester without a clean way of optimizing is a bit non-nonsensical imo.

twiecki:
Agree with @warren-oneill. I'm also not convinced anymore that `.run()` should call `initialize()`. The use case of calling `.run()` multiple times with different data would not be supported anymore. A larger refactoring realistically is not going to happen any time soon so we should merge this and unlock a huge number of use cases with a fairly simple code change.

twiecki:
I have a much simpler idea for implementing this:

Instead of changing when initialize gets called or passing through parameters from `__init__`, we could just add a `call_initialize` kwarg to `__init__` that causes the user `initialize()` not get called. Instead, the user can just call it himself before a call to `.run()`.

warren-oneill:
Makes sense but one thing that hasn't been tackled, as far as I'm aware, is a rolling optimization. Or is it possible with this setup? 
I think as well as an ```initialize``` one would need, e.g, ```optimize_daily()``` which would pick the parameters that preformed the best over the last _x_ number of days and then overwrites this parameters in the algo. 
Of course this can also come in as stage 2. What do you think?

twiecki:
That's true that this is not addressed. It's a bit more tricky though as the algorithm state would have to saved (pickled) and different versions be run over the same time-period.

twiecki:
Closing in favor of https://github.com/quantopian/zipline/pull/600. We should port the example from here however.

Issue #188, opened 2013-07-08
Remove position when a stock is delisted


ehebert:
When a stock is delisted it should be removed from the portfolio to reflect the impact had on portfolio value.

dmichalowicz:
This is being addressed here: https://github.com/quantopian/zipline/pull/982

dmichalowicz:
@richafrank Can I close this, having merged https://github.com/quantopian/zipline/pull/982?

richafrank:
Yes!

PR #460, opened 2014-12-31
Add LIQUIDATION events

jlowin:
Securities sometimes stop trading due to delisting, acquisition, maturity, etc. Currently, if users hold positions in those securities, the positions just persist after the security disappears, maintaining their market values. There is no way to close these positions because orders are only executed when trades occur (which, once the security stops trading, is never!). This creates a strong divergence from reality because users can't make use of cash which would be returned to them when the security is liquidated and must use leverage instead.

Proposed solution: a new `DATASOURE_TYPE` called `LIQUIDATION` which looks just like a `TRADE` but differs in two ways:
    1. When `AlgorithmSimulator` encounters a `LIQUIDATION` event, it (a) cancels all open orders for that `sid` on the blotter (because you can't trade it anymore!) and (b) places one more order that closes any open positions.
    2. When `Slippage` encounters a `LIQUIDATION` event, it generates a transaction without regard for commission or volume. (maybe the commission constraint should go?)

Otherwise, liquidation-generated orders/transactions are handled exactly like "normal" ones, to maintain compatibility. The only difference is that when liquidation events appear, all trading is halted and all positions are forcibly closed out.

Usage is simple. Whenever a `DataSource` gets to the last trade of a security, it emits a `LIQUIDATION` event instead of -- or in addition to -- the normal `TRADE` event. Because this is handled on the data source end, users don't need to worry about leaking future knowledge; whenever the security liquidates, the system will process it automatically. Since a LIQUIDATION event can have its own price, a security could be liquidated at 0, par, or some acquisition price regardless of the last observed trade.

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -267,7 +267,8 @@ class PerformanceTracker(object):
     def process_event(self, event):
         self.event_count += 1
 
-        if event.type == zp.DATASOURCE_TYPE.TRADE:
+        if event.type in (
+                zp.DATASOURCE_TYPE.TRADE, zp.DATASOURCE_TYPE.LIQUIDATION):
             # update last sale
             for perf_period in self.perf_periods:
                 perf_period.update_last_sale(event)
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -72,6 +72,10 @@ from . period import PerformancePeriod
 log = logbook.Logger('Performance')
 
 
+TRADE_OR_LIQUIDATION = (
+    zp.DATASOURCE_TYPE.TRADE, zp.DATASOURCE_TYPE.LIQUIDATION)
+
+
 class PerformanceTracker(object):
     """
     Tracks the performance of the algorithm.
@@ -267,8 +271,7 @@ class PerformanceTracker(object):
     def process_event(self, event):
         self.event_count += 1
 
-        if event.type in (
-                zp.DATASOURCE_TYPE.TRADE, zp.DATASOURCE_TYPE.LIQUIDATION):
+        if event.type in TRADE_OR_LIQUIDATION:
             # update last sale
             for perf_period in self.perf_periods:
                 perf_period.update_last_sale(event)
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -65,15 +65,10 @@ import pandas as pd
 from pandas.tseries.tools import normalize_date
 
 import zipline.protocol as zp
-import zipline.finance.risk as risk
-from zipline.finance import trading
+from zipline.finance import trading, slippage, risk
 from . period import PerformancePeriod
 
 log = logbook.Logger('Performance')
-
-
-TRADE_OR_LIQUIDATION = (
-    zp.DATASOURCE_TYPE.TRADE, zp.DATASOURCE_TYPE.LIQUIDATION)
 
 
 class PerformanceTracker(object):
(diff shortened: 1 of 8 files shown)


ssanderson:
Assigning @ehebert here since I think there's some planned refactoring of the event API coming down the pipe.

jlowin:
Also, might solve #188

ehebert:
ACK'd. Looking forward to reading through, but might not get to it today.

zipline/finance/performance/tracker.py
```diff
@@ -267,7 +267,8 @@ def to_dict(self, emission_type=None):
     def process_event(self, event):
         self.event_count += 1
 
-        if event.type == zp.DATASOURCE_TYPE.TRADE:
+        if event.type in (
+                zp.DATASOURCE_TYPE.TRADE, zp.DATASOURCE_TYPE.LIQUIDATION):
```
ehebert: Could be negligible overall, but this is part of the tight loop so shaving nanoseconds can add up.

Defining the allowed values outside of the loop gets the comparison back to equivalent (or better) than the straight `==` check.

```
In [1]:
class Foo(object):
    ONE = 1
    TWO = 2
    THREE = 3

In [2]:
%%timeit

1 == Foo.ONE
10000000 loops, best of 3: 52.6 ns per loop


In [3]:
%%timeit

1 in (Foo.ONE, Foo.TWO)
10000000 loops, best of 3: 121 ns per loop


In [4]:
ALLOWED = (Foo.ONE, Foo.TWO)

In [5]:
%%timeit

1 in ALLOWED
10000000 loops, best of 3: 44.1 ns per loop
```

zipline/gens/tradesimulation.py
```diff
@@ -225,6 +225,20 @@ def _process_snapshot(self, dt, snapshot, instant_fill):
...
+            elif event.type == DATASOURCE_TYPE.LIQUIDATION:
+                # cancel all open orders, if applicable
+                if event.sid in self.algo.blotter.open_orders:
+                    for order in self.algo.blotter.open_orders[event.sid]:
+                        self.algo.blotter.cancel(order.id)
+
+                # create new liquidation order, if applicable
+                if event.sid in self.algo.portfolio.positions:
+                    amt = self.algo.portfolio.positions[event.sid].amount
+                    self.algo.order(event.sid, -1 * amt)
```
ehebert: Doing this as an order gets the right values in the right places, but I worry that by using order there will be unexpected effects at some point, e.g. if this was used in a live context, this would attempt to order post the stocks liquidation event.

Seems like there needs to be a method on the portfolio such as `self.algo.portfolio.clear_out(sid)` or on the blotter class so that there can be hooks for there making sure that the position is cleaned out there.

ehebert:
Really like how this is handled by a data source type, which allows implementations to 'opt-in' to supporting liquidation by choosing to use that event type or not.

I'm not the best citizen always about it either, but I think this could use a unit test, which could probably be added on to one of the other existing suites that exercise tradesimulation.

Two comments, one about the perf impact of creating the tuple in the loops, and one about using order versus other means of closing out the position would be on my list of blockers.

jlowin:
@ehebert I restructured `LIQUIDATION` to avoid creating an order. I manually create a `Transaction` instead and let Zipline process it as it would any other txn. This way, all reporting remains intact and accounts for the liquidation event -- for example, like the `transactions` dataframe in the result object returned by `Algorithm.run()`. I could have also manually adjusted the positions (basically reimplementing the performance tracker's `execute_transaction()` method) but using transactions seems to be as close as Zipline has to an audit trail, so I thought it would be preferable.

jfkirk:
@ehebert @twiecki I recently implemented a similar behavior on my futures-trading branch, not knowing that this branch existed. Would it be valuable to touch up this PR and push it through to get this functionality in separately?

twiecki:
@jfkirk Does your implementation also work for equities?

jfkirk:
My implementation depends on the AssetFinder. At every end-of-day, the Asset objects are checked to see if they are expiring. If they are, a new Transaction is created to liquidate the position and run through the perf tracker. This Transaction event is not seen outside of the perf tracker/position tracker.

twiecki:
How does the `AssetFinder` know that an asset is expiring?

warren-oneill:
@jfkirk @ehebert @jlowin is this solved with #593? One thing I'm missing is the cancelling of open orders and perhaps the test coverage here is better.

twiecki:
This came up here: https://www.quantopian.com/posts/when-a-company-gets-acquired-my-portfolio-still-owns-shares-of-the-original-company-is-that-right

@jfkirk Is this solved just not on Q?

twiecki:
It seems this is now fixed with #982. Feel free to reopen if not.

PR #465, opened 2015-01-09
WIP: Lot (don't merge)

jlowin:
Per conversation yesterday with @ehebert and @twiecki, this is a first cut at implementing a tax `lot` object that sits under a `position`.

As much as possible, this is supposed to be a "drop-in" replacement for the existing `Position` object, for illustrative and compatibility purposes. It _should_ work in most cases. The only non-`Position` code that was modified is the `update_position` method of the `PerformancePeriod` class. As a result there's a lot of `Position` code that is probably unnecessary but is kept for compatibility.

There is support for close instructions but I didn't change `order()` to allow them as an argument, so for the moment the default `FIFO` is always used.

Zipline makes a few assumptions that aren't really possible in a tax lot world (but are in its single-position world) like blanket adjustments of `amount` and `cost_basis`.

We can discuss further offline.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ benchmarks.db
 
 # Vagrant temp folder
 .vagrant
+
+#PyCharm
+.idea*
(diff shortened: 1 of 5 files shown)


FreddieV4:
Going to close this due to inactivity. If you'd like to rebase and open a new PR feel free to go ahead and do that :)

PR #468, opened 2015-01-13
Revamp documentation system

twiecki:
I basically ripped out the old documentation system using sphinx which was more confusing than actually helpful and replaced it with [`mkdocs`](https://github.com/tomchristie/mkdocs). This new system has the following benefits:

* Documentation can be written in `markdown` instead of `rst`.
* Documentation can automatically be published on the gh-pages branch using `mkdocs gh-deploy` where it will be visible through [zipline.io](http://zipline.io)
* Automatic conversion of IPython notebooks: This allows the documentation to be written and the included/published as an IPython notebook.

For the current look, see here: http://zipline.io/

There are still some markdown quirks (https://github.com/tomchristie/mkdocs/issues/310) but I don't think it's a huge deal. I currently link `README.md` to be the title page.

I also removed pretty much everything of the prior documentation just because it was obsolete and I think it's better to have a fresh start.

We should then start writing IPython Notebooks to document zipline features.


twiecki:
@llllllllll Any update?

twiecki:
@llllllllll Ping.

PR #470, opened 2015-01-22
BUG: Changed benchmark returns to only contain market minutes.

CaptainKanuk:
The series was being generated as all minutes between two times.
It should be only the trading minutes.


zipline/finance/performance/tracker.py
```diff
@@ -69,6 +69,8 @@
 from zipline.finance import trading
 from . period import PerformancePeriod
 
+trd_env = trading.TradingEnvironment()
```
llllllllll: Don't do this at module level, it causes all kinds of craziness, I would look for usages of the `with_environment decorator from the trading module.

PR #471, opened 2015-01-23
ENH: risk measures in datapanel

fawce:
cumulative risk measures added to the datapanel produced from perf frames.


twiecki:
Looks harmless enough. :+1:

PR #478, opened 2015-01-29
Dnt list

fawce:
adding support for passing a function to do not order, for dynamic restricted lists.


zipline/finance/controls.py
```diff
@@ -99,6 +99,39 @@ def validate(self,
...
+    """
+
+    def __init__(self, restricted_list):
+        """
+        restricted list can be an iterable, or
+        a callable that returns an iterable for dynamic restrictions.
+        """
+
+        super(RestrictedListOrder, self).__init__()
+        if hasattr(restricted_list, '__call__'):
```
llllllllll:
```python
if callable(restricted_list):
    ...
```

zipline/finance/controls.py
```diff
@@ -99,6 +99,38 @@ def validate(self,
...
+    def __init__(self, restricted_list):
+        """
+        restricted list can be an iterable, or
+        a callable that returns an iterable for dynamic restrictions.
+        """
+
+        super(RestrictedListOrder, self).__init__()
+        # if passed a callable, call it, otherwise
+        # wrap what we were passed in a lambda
+        self.restricted_list = \
```
ssanderson: What benefit is gained by allowing the user to pass a callable here?  Since the function has to take no arguments, the only way this can ever return something different on different calls is if it's closing over global mutable state, which seems like an idiom we want to discourage (in particular, this is the doomsday scenario for serialization).

zipline/utils/security_list.py
```diff
@@ -0,0 +1,134 @@
...
+    def __init__(self, current_date_func, lookup_func=None):
+        if lookup_func is None:
+            self.lookup_func = loopback
+        else:
+            self.lookup_func = lookup_func
+        self.current_date_func = current_date_func
+        self._leveraged_etf = None
+
+    @property
+    def LEVERAGED_ETF_LIST(self):
```
llllllllll: why is this all caps?

zipline/utils/security_list.py
```diff
@@ -0,0 +1,134 @@
...
+        knowledge_dates = sorted(
+            [pd.Timestamp(k) for k in data.keys()])
+        return knowledge_dates
+
+    def __iter__(self):
+        return iter(self.get_restricted_list())
+
+    def __contains__(self, item):
+        rl = self.get_restricted_list()
+        return rl.__contains__(item)
```
llllllllll: `item in self.get_restricted_list()`

zipline/utils/security_list.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def get_restricted_list(self):
```
llllllllll: could this be a property instead of an getter?

zipline/utils/security_list.py
```diff
@@ -0,0 +1,134 @@
...
+        current_date_func: function taking no parameters, returning
+            current datetime
+        """
+        self.lookup_func = lookup_func
+        self.data = data
+        self._cache = {}
+        self._knowledge_dates = self.make_knowledge_dates(self.data)
+        self.current_date = current_date_func
+        self.count = 0
+        self._list = set()
```
llllllllll: I still think it is confusing that `self._list` is not a `list`, could we call this `_security_list` or something that is farther from the type name?

zipline/utils/security_list.py
```diff
@@ -0,0 +1,134 @@
...
+       {add: [symbol list], 'delete': [symbol list]}
+    """
+    data = {}
+    dir_path = SECURITY_LISTS_DIR + "/" + list_name
+    for kd_name in listdir(dir_path):
+        kd = datetime.strptime(kd_name, DATE_FORMAT).replace(
+            tzinfo=pytz.utc)
+        data[kd] = {}
+        kd_path = os.path.join(dir_path, kd_name)
+        for ld_name in listdir(dir_path + '/' + kd_name):
```
llllllllll: `os.path.join`

zipline/utils/test_utils.py
```diff
@@ -1,10 +1,14 @@
 from contextlib import contextmanager
 from logbook import FileHandler
 from zipline.finance.blotter import ORDER_STATUS
+from zipline.utils.security_list import SECURITY_LISTS_DIR
 
 from six import itervalues
 
+import os
+import os.path
```
llllllllll: I believe that importing os is enough.

zipline/utils/security_list.py
```diff
@@ -0,0 +1,134 @@
...
+            ld = datetime.strptime(kd_name, DATE_FORMAT).replace(
+                tzinfo=pytz.utc)
+            data[kd][ld] = {}
+            ld_path = os.path.join(kd_path, ld_name)
+            for fname in listdir(ld_path):
+                fpath = os.path.join(ld_path, fname)
+                with open(fpath) as f:
```
llllllllll: At this point in the function, we are 3 loops in and adding a new indentation block, could we try to refactor out some of the subloops?

zipline/finance/controls.py
```diff
@@ -99,6 +99,35 @@ def validate(self,
...
+class RestrictedListOrder(TradingControl):
+    """
+    TradingControl representing a restricted list of securities that
+    cannot be ordered by the algorithm.
+    """
+
+    def __init__(self, restricted_list):
+        """
+        restricted list can be an iterable, or
+        an object that implements __contains__ for dynamic
```
llllllllll: restricted_list can be an iterable or container for dynamic restrictions.

I think the convention is to say that something that implements `__contains__` is a container; however, this is not a big deal if you think this is more clear.

zipline/utils/security_list.py
fawce: BECAUSE I WAS THINKING OF IT LIKE AN ENUM. But, that's not right, so I can drop the caps :).

PR #482, opened 2015-02-03
Fix zipline magic

ssanderson:
The %%zipline magic is currently broken if you don't pass an algofile, which is what we do in our one example .ipynb file.


twiecki:
I cherry-picked https://github.com/quantopian/zipline/commit/da82cad243185f2d4f9b9d20417d4b3c25d3d9ce. The other part should already be fixed.

PR #483, opened 2015-02-03
PERF: Add a wrapper around Series to speed up perf tracker bottleneck.

ehebert:
Alleviates bottleneck caused re-indexing into a pd.Series during a tight
loop, by keeping track of the index value into the underlying `.values`
in a lookup table.

### DIFF
--- a/zipline/finance/performance/period.py
+++ b/zipline/finance/performance/period.py
@@ -87,6 +87,28 @@ from . position import positiondict
 log = logbook.Logger('Performance')
 
 
+class FastSeries(object):
+    def __init__(self, *args, **kwargs):
+        super(FastSeries, self).__init__(*args, **kwargs)
+
+        self._loc_map = {}
+        self.series = pd.Series([])
+        self.values = self.series.values
+
+    def __setitem__(self, key, value):
+        try:
+            i = self._loc_map[key]
+            self.values[i] = value
+        except (KeyError, IndexError):
+            self.series = \
+                self.series.append(
+                    pd.Series({key: value}))
+            self._loc_map = dict(
+                zip(self.series.index,
+                    range(len(self.series))))
+            self.values = self.series.values
+
+
 class PerformancePeriod(object):
 
     def __init__(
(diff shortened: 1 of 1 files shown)


ehebert:
Opening for discussion, but merging, since I reviewed @twiecki's changes while testing/rebasing.

zipline/finance/performance/period.py
```diff
@@ -131,6 +153,8 @@ def __init__(
             columns=zp.DIVIDEND_PAYMENT_FIELDS,
         )
 
+        self.loc_map = {}
```
richafrank: Unused?

twiecki: Ah, that's a left-over from @ehebert branch this one was based on.

zipline/finance/performance/period.py
```diff
@@ -87,6 +87,28 @@
 log = logbook.Logger('Performance')
 
 
+class FastSeries(object):
```
llllllllll: Should we rename this, when I see `FastSeries`, I think it is an object that is like a series but faster; however, this is actually write-only and does not behave much like a series.

zipline/finance/performance/period.py
```diff
@@ -87,6 +87,28 @@
...
+    def __init__(self, *args, **kwargs):
+        super(FastSeries, self).__init__(*args, **kwargs)
```
llllllllll: Why are we forwarding all arguments to `object.__init__`, this only raises warning on py3 and is a nop.

twiecki: Yep, that's also a left-over. I originally wanted to inherit from `Series`.

zipline/finance/performance/period.py
```diff
@@ -141,19 +165,10 @@ def rollover(self):
         self.orders_by_id = OrderedDict()
 
     def set_position_amount(self, sid, amount):
```
twiecki: Actually this whole function can be removed, as well as the one below.

twiecki:
A bit of context, this was actually not meant as a PR but still WIP. I suppose it makes sense to merge now to get the speed gain but it will have to be cleaned up.

ehebert:
@twiecki agreed, I was a little anxious getting it in. And the benefits I believe outweighed the WIP-ness of it, I'm sure one of us will clean up soon. :-)

twiecki:
I tried addressing all comments here over at #491.

PR #485, opened 2015-02-05
Remove hardcoded volume

jlowin:
This hardcoded `volume` parameter is extremely confining. This is a quick way to let users override it.

### DIFF
--- a/zipline/sources/data_frame_source.py
+++ b/zipline/sources/data_frame_source.py
@@ -23,6 +23,7 @@ from zipline.gens.utils import hash_args
 
 from zipline.sources.data_source import DataSource
 
+params = {'ds_volume': 1000}
 
 class DataFrameSource(DataSource):
     """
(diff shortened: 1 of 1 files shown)


twiecki:
Thanks @jlowin. Do you think we should also increase the volume to something absurdly high? My impression is that if you don't want to pass it in, you don't want it to matter.

jlowin:
@twiecki yes, my preference would be to make it irrelevant but I thought there might be some internal use case that requires/expects 1000, so I exposed it as configurable. If not, just setting it to `1e9` or some huge number would be great! Too often I get confused when algos running on DataFrame data produce very weird transaction histories until I remember they can only buy some fraction of the 1000 available shares.

twiecki:
If we set it to `1e9`, would you still want it to be configurable?

jlowin:
Nope the config was just in case I was messing with someone's required assumption of 1000.

twiecki:
Done: https://github.com/quantopian/zipline/commit/439f404beda103a80d25a704dd481d1e7b5941bc
Thanks!

PR #486, opened 2015-02-05
Update test matrix to test both pandas 0.12.0 and 0.15.2

dalejung:
So this will test a 2x2 matrix.

py 2.7 pandas 0.12.0
py 2.7 pandas 0.15.2
py 3.3 pandas 0.12.0
py 3.3 pandas 0.15.2

It might be overkill to test all 4. Maybe py 2.7/pandas 0.12.0 and py3.3/pandas 0.15.2 would suffice. 

Either way, if we're supporting 0.12.0 it should be in the test matrix somewhere.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: python
 python:
   - "2.7"
   - "3.3"
+env:
+  - PANDAS_VERSION=0.15.2
+  - PANDAS_VERSION=0.12.0
 before_install:
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then wget http://repo.continuum.io/miniconda/Miniconda-3.7.0-Linux-x86_64.sh -O miniconda.sh; else wget http://repo.continuum.io/miniconda/Miniconda3-3.7.0-Linux-x86_64.sh -O miniconda.sh; fi
   - chmod +x miniconda.sh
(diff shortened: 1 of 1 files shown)


dalejung:
ooh, need to change the hardcoded numpy 1.9, since it won't work with pandas 0.12.0

twiecki:
We should have done this a long time ago. Hopefully though we're moving away from pandas 0.12 soon though, so I'm not sure how long this will be relevant. If it's easy though I'm all for merging.

dalejung:
The readme says it supports back to pandas 0.9. So we either need to update the minimum deps or run travis on those older versions. 

Travis is failing on pandas 0.12.0 and numpy 1.7.0. Havent' checked if that's just a travis thing, but if it's legit then previous deps are probably already failing too.

ehebert:
I think it would be OK to update README to reflect the version we had been
using internally. (0.12).

Changing lib requirements usually necessitates a version bump, but not sure
how much we should be worried here since the requirements.txt had had 0.12
for some time anyway.

dalejung:
So the minimum supported version is whatever is run internally? i.e. 0.12.0 gets dropped once 0.15.2 is deployed?

dalejung:
It seems like requirements.txt is being used both for minimum deps and deployment manifest, which can be at odds. `zipline` could support `0.12.0` even if some higher version is used in production.

twiecki:
Agreed @dalejung that it's not the best model for the reasons you mention.

dalejung:
ah, it was failing on ta-lib compiled with mismatched numpy versions. Switch to the quantopian binstar since that had 1.8. Not sure if that means we drop numpy 1.7? If not, we need to put up a build for it.

twiecki:
For my money we'd focus on upgrading to `pandas` 0.15 and dropping 0.12 rather than setting up everything to support and test pandas 0.12. We're pretty close to that I think.

ehebert:
I agree, especially with how long 15 takes without the history fix applied.

dalejung:
making zipline performant for 15 and dropping support for 12 aren't related. are we dropping support for 12? 



—

ehebert:
Ah, I had forgotten that we had been doing 15 already because of the
un-pegged pandas install. Mea culpa.

twiecki:
If this is superseded by #488, can we close this?

dalejung:
I would have preferred to get this merged in separately. #488 just had it included so travis would run the tests.

twiecki:
Is this merged?

dalejung:
nope, i see eddie removed it from my history perf PR.

ehebert:
I think it's clear to merge those CI specific patches now. Was just
reducing the surface area so I could get my head around it.

twiecki:
@dalejung Can you rebase and merge with master?

dalejung:
it’s rebased. want me to merge when it pings green?


—

twiecki:
Sounds good.

dalejung:
boo i dont have write access

PR #491, opened 2015-02-09
REF Clean up FastSeries PR.

twiecki:
We took a short-cut to get the performance improvements of `FastSeries` here: https://github.com/quantopian/zipline/pull/483. This PR was more WIP so here is an attempt to clean it up.

### DIFF
--- a/zipline/finance/performance/period.py
+++ b/zipline/finance/performance/period.py
@@ -87,10 +87,17 @@ from . position import positiondict
 log = logbook.Logger('Performance')
 
 
-class FastSeries(object):
-    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


zipline/finance/performance/period.py
```diff
@@ -87,10 +87,17 @@
...
+    """
+    Series container with fast item access.
+
+    This class stores a pandas.Series object as self.series. It also caches the
+    index so that changing an existing item is faster. For faster access to the
+    underlying data there is also self.values which caches
+    pandas.Series.values.
+    """
 
+    def __init__(self, *args, **kwargs):
```
ssanderson: Why does this take *args and **kwargs?

twiecki:
FastSeries was replaced in the interim.

PR #492, opened 2015-02-09
MAINT: added cutpoint for overriding SecurityList implementation.

fawce:


zipline/utils/security_list.py
```diff
@@ -102,6 +81,30 @@ def update_current(self, effective_date, symbols, change_func):
             change_func(sid)
 
 
+class SecurityListSet(object):
+
+    def __init__(self, current_date_func, lookup_func=None):
+        # provide a cut point to substitute other security
+        # list implementations.
+        self.sl_constructor = SecurityList
```
llllllllll: If the idea is to make this configurable, could we make this an argument that defaults to SecurityList?

fawce: idea is to give subclasses an override - seems confusing to put it in the arg list.

llllllllll: if that is the case, should it be a class level attribute?

zipline/utils/security_list.py
```diff
@@ -102,6 +83,30 @@ def update_current(self, effective_date, symbols, change_func):
             change_func(sid)
 
 
+class SecurityListSet(object):
+
+    def __init__(self, current_date_func, lookup_func=None):
+        # provide a cut point to substitute other security
+        # list implementations.
+        self.security_list_type = SecurityList
```
ssanderson: Agreed with @llllllllll that this would be cleaner as a class-level attribute.

PR #493, opened 2015-02-12
PERF: Avoid using attribute access in inner event processing.

ehebert:
Reduce overhead of the attribute access of grabbing the process_event
and process_trade from both the algorithm object and the perf_tracekr or
blotter, by assigning those functions to a variable once per snapshot.

### DIFF
--- a/zipline/gens/tradesimulation.py
+++ b/zipline/gens/tradesimulation.py
@@ -83,12 +83,11 @@ class AlgorithmSimulator(object):
         return self.EMISSION_TO_PERF_KEY_MAP[
             self.algo.perf_tracker.emission_rate]
 
-    def process_event(self, event):
-        process_trade = self.algo.blotter.process_trade
-        for txn, order in process_trade(event):
-            self.algo.perf_tracker.process_event(txn)
-            self.algo.perf_tracker.process_event(order)
-        self.algo.perf_tracker.process_event(event)
+    def process_event(self, blotter_process_trade, perf_process_event, event):
+        for txn, order in blotter_process_trade(event):
+            perf_process_event(txn)
+            perf_process_event(order)
+        perf_process_event(event)
 
     def transform(self, stream_in):
         """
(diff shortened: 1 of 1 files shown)


ehebert:
cc: @llllllllll

zipline/gens/tradesimulation.py
```diff
@@ -83,12 +83,11 @@ def perf_key(self):
         return self.EMISSION_TO_PERF_KEY_MAP[
             self.algo.perf_tracker.emission_rate]
 
-    def process_event(self, event):
-        process_trade = self.algo.blotter.process_trade
-        for txn, order in process_trade(event):
-            self.algo.perf_tracker.process_event(txn)
-            self.algo.perf_tracker.process_event(order)
-        self.algo.perf_tracker.process_event(event)
+    def process_event(self, blotter_process_trade, perf_process_event, event):
```
llllllllll: because the bound methods are passed in, this should probably become 'private', also the bound methods should probably be private args ('_blotter_process_trade') so that consumers don't mess this up.

zipline/gens/tradesimulation.py
```diff
@@ -210,6 +209,15 @@ def _process_snapshot(self, dt, snapshot, instant_fill):
         if instant_fill:
             events_to_be_processed = []
 
+        # Assign process events to variables to avoid attribute access in
+        # innermost loops.
+        #
+        # Done here, to allow for perf_tracker or blotter to be swapped out
+        # or changed in between snapshots.
+        perf_process_event = self.algo.perf_tracker.process_event
```
llllllllll: we could even cache self.algo between these calls

llllllllll:
This looks correct, I had one comment about naming but functionally this looks good.

zipline/gens/tradesimulation.py
ehebert: Ah, I remember you mentioning this before. Will fixup.

ehebert:
Fixed up and merged with 7059ba1e6a7c36f44226bbaed41d9e2f8a994d6d

PR #494, opened 2015-02-13
PERF: Makes get_datetime not make a copy.

llllllllll:
datetime.datetime objects are immutable.


ehebert:
/signoff

Looks good to me.
Under testing it didn't have a giant impact on performance, but the slight boost is good.
But even better is that we aren't having an extra moving part of copying, when as you point out the datetime is immutable anyway.

llllllllll:
I just want to change the assertion check, I think equality is more correct even at the cost of speed.

PR #496, opened 2015-02-17
BUG: Remove sid entry from open_orders when there are none.

dalejung:
Note, this changes behavior as it removed the entry if there are no open orders. 

Technically speaking, just setting an empty list is faster than deleting the key, but checking key membership is a more common operation and faster than checking for an empty list. AFAIK, most users won't constantly alternate each tick with resting/canceling orders.


ehebert:
This looks good to me, I'm trying to think of any downstream implications of not having the empty list (with this patch being more correct) and not coming up with any that would be harmful.

ehebert:
Merged with b3bc7e166df57b05e0b4764dee218e84dbf51ca9 thanks!

PR #507, opened 2015-02-25
Blotter perf

dalejung:
This puts an upper bound on the perf regression from having many orders open for the same sid.


ehebert:
This looks good to me. Maybe just update the Allow SlippageModels liquidity commit with more details about how it gets into that factorial problem you mentioned and then it would be good to go.

dalejung:
@eddie @twiecki can you click and rebuild this in travis. I don't have the access.

twiecki:
Done.

ehebert:
Merged via 94d25a54c607d0379aa25f80cbbb79617467d583

PR #508, opened 2015-02-25
Siddata transform perf

dalejung:
Performance for rolling transforms. Main culprit now is the actual history call.

### DIFF
--- a/zipline/protocol.py
+++ b/zipline/protocol.py
@@ -15,8 +15,10 @@
 
 from six import iteritems, iterkeys
 import pandas as pd
+import numpy as np
 
 from . utils.protocol_utils import Enum
+from . utils.math_utils import nanstd, nanmean, nansum
 
 from zipline.finance.trading import with_environment
 from zipline.utils.algo_instance import get_algo_instance
@@ -270,7 +272,7 @@ class SIDData(object):
     def __repr__(self):
         return "SIDData({0})".format(self.__dict__)
 
-    def _get_buffer(self, bars, field='price'):
+    def _get_buffer(self, bars, field='price', raw=False):
         """
         Gets the result of history for the given number of bars and field.
 
@@ -287,7 +289,7 @@ class SIDData(object):
             cls._history_cache = {}
 
         if field not in self._history_cache \
-           or bars > len(cls._history_cache[field].index):
+           or bars > len(cls._history_cache[field][0].index):
             # If we have never cached this field OR the amount of bars that we
             # need for this field is greater than the amount we have cached,
             # then we need to get more history.
@@ -297,12 +299,17 @@ class SIDData(object):
             # Assert that the column holds ints, not security objects.
             if not isinstance(self._sid, str):
                 hst.columns = hst.columns.astype(int)
-            self._history_cache[field] = hst
+            self._history_cache[field] = (hst, hst.values, hst.columns)
 
         # Slice of only the bars needed. This is because we strore the LARGEST
         # amount of history for the field, and we might request less than the
         # largest from the cache.
-        return cls._history_cache[field][self._sid][-bars:]
+        buffer_, values, columns = cls._history_cache[field]
+        if raw:
+            sid_index = columns.get_loc(self._sid)
+            return values[-bars:, sid_index]
+        else:
+            return buffer_[self._sid][-bars:]
 
     def _get_bars(self, days):
         """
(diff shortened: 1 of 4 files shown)


dalejung:
well, specifically the ffill pathway. booo ffill

zipline/protocol.py
```diff
@@ -297,12 +299,17 @@ def _get_buffer(self, bars, field='price'):
...
             if not isinstance(self._sid, str):
                 hst.columns = hst.columns.astype(int)
-            self._history_cache[field] = hst
+            self._history_cache[field] = (hst, hst.values, hst.columns)
 
         # Slice of only the bars needed. This is because we strore the LARGEST
         # amount of history for the field, and we might request less than the
         # largest from the cache.
-        return cls._history_cache[field][self._sid][-bars:]
+        buffer, values, columns = cls._history_cache[field]
```
llllllllll: can this be `buffer_` to not trample the builtin

dalejung:
@llllllllll made the buffer change. Do you know if we have any hard requirements stating that if you're not using a str, your object must be int-able and hash to the same value?

llllllllll:
I don't know of a case where securities don't have the property.

zipline/protocol.py
```diff
@@ -270,7 +272,7 @@ def __contains__(self, name):
     def __repr__(self):
         return "SIDData({0})".format(self.__dict__)
 
-    def _get_buffer(self, bars, field='price'):
+    def _get_buffer(self, bars, field='price', raw=False):
```
llllllllll: when do we call this with `raw=False` or the default?

dalejung: afaik, we don't currently. It's more of a backwards compat thing.

dalejung:
meant that more for zipline in general. Not quantopian per se. @twiecki do we have guidelines for what a correct SID type sid?

zipline/protocol.py
```diff
@@ -360,17 +367,27 @@ def minute_get_bars(days, env=None):
...
 
     def vwap(self, days):
         bars = self._get_bars(days)
-        prices = self._get_buffer(bars)
-        vols = self._get_buffer(bars, field='volume')
+        prices = self._get_buffer(bars, raw=True)
+        vols = self._get_buffer(bars, field='volume', raw=True)
+
+        vol_sum = nansum(vols)
+        if vol_sum == 0:
```
llllllllll: if this is a less common case, we might want to put the division in a try/except. EAFP

zipline/protocol.py
```diff
@@ -360,17 +367,27 @@ def minute_get_bars(days, env=None):
         return self._get_bars(days)
 
     def mavg(self, days):
-        return self._get_buffer(self._get_bars(days)).mean()
+        bars = self._get_bars(days)
```
llllllllll: Just a style (and minor perf) comment about these functions, I find caching all these intermediate results in the locals harder to read than just chaining the computations. Chaining the computations also keeps all the data on the python stack instead of bouncing between the fast and the stack constantly.

zipline/protocol.py
llllllllll: I don't think we need to worry about backwards compat with a private method.

llllllllll:
I understood that you meant for zipline not quantopian; however, I still think that securities are either strings or int like. I think we make this assumption in a lot of places.

zipline/protocol.py
dalejung: What is EAFP? Yeah originally I had it in a try/except, forgot why i changed it. Probably being dumb.

llllllllll: https://docs.python.org/3.5/glossary.html#term-eafp

Easier to ask forgiveness than permission.

dalejung:
I agree. It's more that it isn't immediately obvious to all python users that they need their SID objects to hash to the hash of the equivalent integer.

zipline/utils/math_utils.py
```diff
@@ -14,7 +14,22 @@
...
 
 import math
+import numpy as np
 
 
 def tolerant_equals(a, b, atol=10e-7, rtol=10e-7):
     return math.fabs(a - b) <= (atol + rtol * math.fabs(b))
+
+
+nanmean = np.nanmean
```
llllllllll: we should move these assignments into the except handler so that it is more clear which version you are using.

zipline/protocol.py
dalejung: This isn't an internal repo. I get burned all the time when people change "private" methods in open source projects. And those are usually just signature or whatever changes, not return a wholly different data structure.

llllllllll: Sure; however, using a private method is unsupported behaviour, people working with zipline should not be using this method in the first place. Attempting to backwards compat all private methods forever is too much effort for no reward.

zipline/utils/math_utils.py
ehebert: :+1:

ehebert:
This looks good to me, but let's do @llllllllll 's style requests.

dalejung:
hm. carp. bottleneck is kind of 1-to-1 to numpy versions. I wonder if I need to matrix it in travis. 

It might be a good idea to create a travis dir and offload our matrix setup scripts there instead of inside the travis yaml. Like having separate requirements.txt for each harness.

dalejung:
depends on https://github.com/quantopian/zipline/pull/512 which is small one liner.

ehebert:
Merged/include by https://github.com/quantopian/zipline/pull/514

PR #509, opened 2015-02-25
MAINT: Provide a function to create the position calc containers.

ehebert:
For use in a function that wraps de-serialization, to call instead of
creating the OrderedDicts from a module outside of the object.

So that the other module does not need to the internals of this object,
also to ensure that the cythonized OrderedDict is used, when available.

This need should be superseded with serialization versioning.


CaptainKanuk:
Looks good to me. This won't be superseded as much as copied over when I get to this object.

PR #505, opened 2015-02-23
Position tracker serialization

dalejung:
After slogging through the reflocs, found the commits. Refactored Position management out of the PerformancePeriods since the logic was being duplicated. Also added pickle methods for each PerformanceTracker, PerformancePeriod, Postion, PositionTracker.

Also, I hate git.

PR #510, opened 2015-02-25
MAINT: Added pickle protocol methods into zipline.

CaptainKanuk:
For review only. Tests will be added soon.

Added pickle support to many zipline methods. This will enable
them to be serialized.


dalejung:
@CaptainKanuk just wanted to confirm that it's good for review now?

CaptainKanuk:
Yes, I'll just be adding unit tests to this.

CaptainKanuk:
I ended up adding code to period, tracker and position for consistency. It's not much of a change code-lines wise, so it should be okay to rebase over.

zipline/utils/serialization_utils.py
```diff
@@ -0,0 +1,55 @@
...
+    def __setstate__(self, state):
+        """
+        Many objects require only this code.
+        """
+        self.__dict__.update(state)
+
+    # =====================================================
+    # These are helper methods for some problem data types.
+    # =====================================================
+
```
dalejung: remove these helpers and handle them in `__setstate__`

dalejung: Something like this: https://github.com/quantopian/zipline/pull/505/files#diff-f3d976ceea2b1ea116269ac8b7e9ccc8R428

These objects needs to be internally consistent. These objects won't inflate without the internal stuff that knows what those special types mean.

CaptainKanuk: Yeah I see the issue.

zipline/utils/serialization_utils.py
```diff
@@ -0,0 +1,70 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Label for the serialization version field in the state returned by
+# __getstate__.
+VERSION_LABEL = '_stateversion_'
+
+
+class SerializeableZiplineObject(object):
```
dalejung: Why do we need to have this class at all? The getstate method is a perfectly good utility func. I think it was basic_get_state before. There really not a single subclass that doesn't override those methods.

zipline/finance/slippage.py
```diff
@@ -128,6 +132,25 @@ def to_dict(self):
...
+
+        return state_dict
+
+    def __setstate__(self, state):
+
+        OLDEST_SUPPORTED_STATE = 1
+        version = state.pop(VERSION_LABEL)
+
+        if version < OLDEST_SUPPORTED_STATE:
+            raise BaseException("Transaction saved state is too old.")
```
dalejung: can we refactor this logic out into a check_versino function along with a custom error? I don't mind the BaseException but joe will yell at me if I don't ask for a custom exception :/

dalejung:
More of a general comment about the state we are saving. Since we are versioning state, we want to keep the state as concise as possible. Values that are temporal like `_position_store` should not be stored because they don't really reflect the state of the object. 

For the most part, it is better to very explicit in what state you save so that you don't grab hanger ons. While they won't break the serialization, they will change the state schema and thus require different versions if/when we clean up or refactor.

type specific data like defaultdict and what not are example of extraneous information. The object needs to know position values from state, it does not need to know how these positions were stored and where on the object they were stored. It already knows this. The `__dict__` as state thing is more a quirk and dirty hack for pickling, but too brittle for when you care about versioning.

I bring up the defaultdict specifically because I am going to remove them game of thrones style. They are quite literally the devil.

jbredeche:
Inviting the defaultdict to the red wedding?

zipline/utils/serialization_utils.py
CaptainKanuk: Fixed

dalejung:
https://github.com/quantopian/zipline/pull/525 has tests to grab the failures. 

```
FAIL: test_object_serialization_18_Portfolio (tests.test_serialization.SerializationTestCase)
  File "/Users/datacaliber/miniconda/envs/qexec/lib/python2.7/site-packages/nose_parameterized/parameterized.py", line 233, in <lambda>
    standalone_func = lambda *a: func(*(a + p.args), **p.kwargs)
  File "/Users/datacaliber/work/quantopian/zipline/tests/test_serialization.py", line 123, in test_object_serialization
    self.assertEqual(len(bad_dicts), 0, msg)
AssertionError: Only support bare dicts. type(positions) == <class 'zipline.protocol.Positions'>

FAIL: test_object_serialization_5_PerformancePeriod (tests.test_serialization.SerializationTestCase)
  File "/Users/datacaliber/miniconda/envs/qexec/lib/python2.7/site-packages/nose_parameterized/parameterized.py", line 233, in <lambda>
    standalone_func = lambda *a: func(*(a + p.args), **p.kwargs)
  File "/Users/datacaliber/work/quantopian/zipline/tests/test_serialization.py", line 123, in test_object_serialization
    self.assertEqual(len(bad_dicts), 0, msg)
AssertionError: Only support bare dicts. type(orders_by_id) == <type 'cyordereddict._cyordereddict.OrderedDict'>

```

Are the two failures on my machine.

dalejung:
btw, i dunno if you can just merge this PR into your branch. Give it a shot. I'm curious.

CaptainKanuk:
What do you mean by merge this PR into my branch?

dalejung:
#525 has tests to merge. I didn't have commit access. I dunno if you can just merge it into a branch.

dalejung:
so it looks like if I create a PR off of a branch, it'll merge into the branch and not master. whee

CaptainKanuk:
@dalejung  I'm honestly not sure how to deal with the case of your PR being pulled off my branch.

dalejung:
lol neither do i. If you want to just copy the diff over that’ll work. Is it possible to click the PR interface and merge it? and then just pull from github? 


—

CaptainKanuk:
@dalejung I see what you're doing. I'll copy over the diff and incorporate it into my newest code.

dalejung:
cool cool. I think that should be enough to catch the pickle stuff earlier.


—

CaptainKanuk:
@dalejung That test helper stuff you wrote is on point.

llllllllll:
do you think that it would make sense to archive and compress all of the testing save states?  These files make the repo considerably larger and are only for testing; maybe we should rewrite the tests to unzip and unarchive the tests so that they have a smaller presence on disk

CaptainKanuk:
I think that's a good idea. I'll put that on my TODO list.

PR #512, opened 2015-02-26
BUG: Initial digest frame was set with object dtype.

dalejung:
The dtype propogates on actual history frame creation and makes certian older numpy funcs like nanmean barf. Broken in 1.8.0, works in 1.8.1


ehebert:
Double checked to make sure that volumes wouldn't be getting a new type. And they are floats already. Looks good to me. /signoff

ehebert:
Merged via e908fa9f5c84368ade0c2ea2912c5de15d434457

Thanks for breaking it out, made it super easy to test.

PR #513, opened 2015-02-26
MAINT: Print benchmark return value to assertion message.

ehebert:
For when the attempted midnight fails, print more information about the
returns for debugging.

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -319,10 +319,13 @@ class PerformanceTracker(object):
                 midnight = event.dt
 
             if midnight not in self.all_benchmark_returns.index:
-                raise AssertionError(("Date %s not allocated in  "
-                                      "all_benchmark_returns. Calendar "
-                                      "seems to mismatch with benchmark."
-                                      % midnight))
+                raise AssertionError(
+                    ("Date %s not allocated in all_benchmark_returns. "
+                     "Calendar seems to mismatch with benchmark."
+                     "Benchmark container is=%s" %
+                     (midnight,
+                      self.all_benchmark_returns.index)))
+
             self.all_benchmark_returns[midnight] = event.returns
 
     def check_upcoming_dividends(self, midnight_of_date_that_just_ended):


zipline/finance/performance/tracker.py
```diff
@@ -319,10 +319,13 @@ def process_event(self, event):
...
-                raise AssertionError(("Date %s not allocated in  "
-                                      "all_benchmark_returns. Calendar "
-                                      "seems to mismatch with benchmark."
-                                      % midnight))
+                raise AssertionError(
+                    ("Date %s not allocated in all_benchmark_returns. "
+                     "Calendar seems to mismatch with benchmark."
+                     "Last dt in container is=%s" %
+                     (midnight,
+                      self.all_benchmark_returns.index)))
```
llllllllll: isn't this the whole container, not the last dt in the container (`self.all_benchmarks.returns.index[-1]`)

zipline/finance/performance/tracker.py
```diff
@@ -319,10 +319,13 @@ def process_event(self, event):
...
 
             if midnight not in self.all_benchmark_returns.index:
-                raise AssertionError(("Date %s not allocated in  "
-                                      "all_benchmark_returns. Calendar "
-                                      "seems to mismatch with benchmark."
-                                      % midnight))
+                raise AssertionError(
+                    ("Date %s not allocated in all_benchmark_returns. "
+                     "Calendar seems to mismatch with benchmark."
+                     "Benchmark container is=%s" %
```
llllllllll: we either want a newline or whitespace between this line and the previous or the line will be harder to read.

zipline/finance/performance/tracker.py
ehebert: Yes, I neglected to push the fixup before PR'ing. Mea culpa. (Was originally just the last, but figured it was better to get the `__str__` which will show both beginning and end.)

ehebert:
Fixed up the whitespace and merged with faf856a7367f50260103ea4275afbafb9c3c5ad2

PR #514, opened 2015-02-26
Siddata max bars

dalejung:
Interesting. For whatever reasons my setup isn't backfilling and thus I don't get the penalty for the resize. 

So, this will request the max size from the get go thus minimizing any panel size enlargements/backfills. 

The other side of this is that the backfills should never have been occurring since no new data was requested, but that's an internal thing.

Not entirely sure how much we care about not having History specs for 290 bars. It's not as if they are user defined specs to begin with.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -24,3 +24,7 @@ Cython==0.20.1
 
 # faster OrderedDict
 cyordereddict==0.2.2
+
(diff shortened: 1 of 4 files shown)


dalejung:
Note, this is built on top of the other siddata transform but new branch since it's a bigger dealio.

ehebert:
82378f720d08b1e7ec7480a2429559a3d13f994d

PR #520, opened 2015-03-03
Add security class

jfkirk:
Adds a cython Security class. This is the first step to adding more types of assets to Zipline for simulation.

### DIFF
--- a/zipline/assets/_securities.pyx
+++ b/zipline/assets/_securities.pyx
@@ -21,18 +21,12 @@ cimport cython
 import numpy as np
 cimport numpy as np
 
-from qexec.algo.repr_ import (
-    repr_qdb,
-    make_repr_qdb_node,
-    node_subset,
-    update_visited,
-)
-
 
 # IMPORTANT NOTE: You must change this template if you change
 # Security.__reduce__, or else we'll attempt to unpickle an old version of this
 # class
 CACHE_FILE_TEMPLATE = '/tmp/.%s-%s.v3.cache'
+
 
 cdef class Security:
 
@@ -46,7 +40,6 @@ cdef class Security:
     # TODO: Maybe declare as pandas Timestamp?
     cdef readonly object start_date
     cdef readonly object end_date
-    # This attribute is currently set in Algoproxy.lookup_security.
     cdef public object first_traded
 
     cdef readonly object exchange
@@ -93,8 +86,8 @@ cdef class Security:
 
     def __richcmp__(self, other, int op):
         """
-        Cython rich comparison method.  This is used in place of various equality
-        checkers in pure python.
+        Cython rich comparison method.  This is used in place of various
+        equality checkers in pure python.
 
         <	0
         <=	1
@@ -150,25 +143,6 @@ cdef class Security:
         params = ', '.join(strings)
         return 'Security(%d, %s)' % (self.sid, params)
 
-    def _repr_qdb(self, visited=None):
-        visited = update_visited(self, visited)
-        return make_repr_qdb_node(
-            self,
-            visited,
-            node=node_subset(
-                self,
-                visited, {
-                    'sid',
-                    'symbol',
-                    'security_name',
-                    'exchange',
-                    'start_date',
-                    'end_date',
-                    'first_traded',
-                }
-            )
-        )
-
     cpdef __reduce__(self):
         """
         Function used by pickle to determine how to serialize/deserialize this
@@ -186,6 +160,27 @@ cdef class Security:
                                  self.end_date,
                                  self.first_traded,
                                  self.exchange,))
+
+    cpdef to_dict(self):
+        """
+        Convert to a python dict.
+        """
+        return {
+            'sid': self.sid,
+            'symbol': self.symbol,
+            'security_name': self.security_name,
+            'start_date': self.start_date,
+            'end_date': self.end_date,
+            'first_traded': self.first_traded,
+            'exchange': self.exchange,
+        }
+
+    @staticmethod
+    def from_dict(dict_):
+        """
+        Build a Security instance from a dict.
+        """
+        return Security(**dict_)
 
 
 def make_security_array(int size, Security security):
--- a/zipline/assets/_securities.pyx
+++ b/zipline/assets/_securities.pyx
@@ -1,0 +1,194 @@
+#
+# Copyright 2015 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


ehebert:
ACK'd

zipline/assets/_securities.pyx
```diff
@@ -0,0 +1,189 @@
...
+cimport cython
+
+import numpy as np
+cimport numpy as np
+
+
+# IMPORTANT NOTE: You must change this template if you change
+# Security.__reduce__, or else we'll attempt to unpickle an old version of this
+# class
+CACHE_FILE_TEMPLATE = '/tmp/.%s-%s.v3.cache'
```
ehebert: This might be able to be removed?

It's an artifact of how we do caching, but might want to remove it until/unless there is an eventual solution in Zipline.

zipline/assets/_securities.pyx
```diff
@@ -0,0 +1,189 @@
...
+        self.end_date      = end_date
+        self.first_traded  = first_traded
+
+    def __int__(self):
+        return self.sid
+
+    def __hash__(self):
+        return self.sid_hash
+
+    property security_start_date:
```
ehebert: This could be removed, I don't think it is used anywhere internally.

ehebert:
I think this is all that is needed to get the TravisCI build unstuck, which can be done either as part of this PR, or right after.

```
diff --git a/.travis.yml b/.travis.yml
index bcec803..3910918 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ install:
   - grep nose-ignore-docstring== etc/requirements_dev.txt | xargs pip install
   - pip install coveralls
   - pip install nose-timer
+  - python setup.py build_ext --inplace
 before_script:
   - "flake8 zipline tests"
 script:
```

zipline/assets/_securities.pyx
```diff
@@ -0,0 +1,189 @@
...
+
+    @staticmethod
+    def from_dict(dict_):
+        """
+        Build a Security instance from a dict.
+        """
+        return Security(**dict_)
+
+
+def make_security_array(int size, Security security):
```
ehebert: This might be able to be trimmed as well.

zipline/assets/_securities.pyx
```diff
@@ -0,0 +1,189 @@
...
+
+    property security_start_date:
+        """
+        Alias for start_date to disambiguate from other `start_date`s in the
+        system.
+        """
+        def __get__(self):
+            return self.start_date
+
+    property security_end_date:
```
ehebert: Ibid.

zipline/assets/_securities.pyx
```diff
@@ -0,0 +1,189 @@
...
+                  object symbol="",
+                  object security_name="",
+                  object start_date=None,
+                  object end_date=None,
+                  object first_traded=None,
+                  object exchange=""):
+
+        self.sid           = sid
+        self.sid_hash = hash(sid)
+        self.symbol        = symbol
```
ehebert: Something that I had not considered until reading through is that some of these extra fields are extraneous in a pure Zipline context. We could remove the 'extra' fields and have an internal Security class that inherits from this one, with the internal class adding the display related fields.

(That internal class could also handle the custom caching etc.)

However, that is more of an open question, and wouldn't say we would need to block merge on it.

ehebert:
Looks good.

My recommendations:
- Squash all the commits into one.
- Consider removing more of the internal fields/functions, i.e. anything that isn't directly related to Zipline functionality. (i.e. could the Security class just have a sid and hash_sid member without any of the other members like symbol, first_traded etc.) But I could be sold on either way.

zipline/assets/_securities.pyx
ssanderson: This is used internally in Quantopian.  We could move it out of zipline though.

zipline/assets/_securities.pyx
jfkirk: Remove the numpy imports, specifically?

zipline/assets/_securities.pyx
jfkirk: Remove security_start_date?

zipline/assets/_securities.pyx
jfkirk: As we move forward on implementing Futures, these fields may come in to use. Would you be alright with leaving them in with a mark to remove if not needed later?

zipline/assets/_securities.pyx
ehebert: Oh, sorry I meant the `CACHE_FILE_TEMPLATE`.

zipline/assets/_securities.pyx
ehebert: Yes, I think it's ok to leave them.

zipline/assets/_securities.pyx
ehebert: Yes, as well as `security_end_date`.

ssanderson:
> (i.e. could the Security class just have a sid and hash_sid member without any of the other members like symbol, first_traded etc.) 

What would the point be of having the class at all at this point?  If you have a struct that just has an int, and a copy of that int, shouldn't you just use an int?

zipline/assets/_securities.pyx
jfkirk: Will remove

jfkirk:
Merged via 0375b350b881982e158d2caab21e3180d1b8a2cc

twiecki:
Are we forcing SID to be `int` with this?

jfkirk:
Not yet. This addition makes no functional change outside making this class
available. We want to preserve the current functionality which allows data
to be loaded with a symbol/string so, when that data is loaded, an int
would be assigned for the string.

twiecki:
I don't think this should have gotten merged. It broke travis, makes installation harder for users, and does not add any functionality.

ehebert:
@twiecki I see your point, and post-merge did second guess some of my recommendations on this merge, but I think it was worth it.

Merging it exposed some of the wrinkles that we would have to deal with sooner or later, and this is a key piece of incorporating other data sources.

Though we should probably chat about Cython-izing in general, it is in the plans for improving some bottlenecks, but I've neglected to check in with you about that.

Also, I admit may have been lax with considering all different combinations of Travis. I may need to transition my mental model away from it being a nice to have.

dalejung:
I'm ambivalent to the Security stuff, but how was this merged with travis failures?

ehebert:
I signed off on the merge because I didn't notice that the Python 3 one was
gray. Mea culpa.

jikamens:
The python 3 break is fixed on my PyPI update branch? Should we merge it?

Sent from my Android device



-----Original Message-----

twiecki:
I think we should definitely merge the python 3 fix but the pandas upgrade will make zipline non-shippable.

jikamens:
OK, I cherry-picked the Python3 fixes for the Security class. While I was at it I cherry-picked the fix for not repeatedly downloading the benchmark and treasury curves unnecessarily.

PR #521, opened 2015-03-04
problem when backtesting minutes datasource with local timezone index

dyno:
When I am trying to run my algorithm locally, every time,
zipline tried to fetch the data which was not yet available.
The problem is the len() is total numbers, but searchsorted()
returns index which is 0 based.

  Also minor refactor to remove duplicate code block.


freddiev4:
Not sure if we deal with this behavior now, but going to close this b/c it's 3 years old; if someone wants to open a new PR feel free to do so.

PR #522, opened 2015-03-04
Updated Requirements 

warren-oneill:
I have been trying to implement the zipline development environment and I ran into a lot of problems with the packet versions in requirements.txt. The problems were either compilation errors or failing unittests.
The settings which worked for me are:
* pandas==0.14.1
* statsmodels==0.6.1

and everything else as it is in the master branch.

Note: I changed pip to pip3 for the installation but it should also work for pip.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,3 @@ benchmarks.db
 
 # Vagrant temp folder
 .vagrant
-
-#Ta-lib folder
-ta-lib/*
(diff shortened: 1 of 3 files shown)


twiecki:
Newer versions of pandas did not work?

warren-oneill:
They did but then it conflicted with the dateutil version and it was unable to read the data from yahoo. I tried to implement it with few changes as possible.

twiecki:
Thanks, but closing in favor of #526.

warren-oneill:
Newer is better :) Fingers crossed those versions work on my machine

PR #524, opened 2015-03-04
Change SIDData to only store actual data in __dict__, internal attrs use slots

dalejung:


dalejung:
this is going to be deprecated by cython

PR #526, opened 2015-03-05
MAINT: Update PyPI packages to current versions, plus new PEP8 rules

jikamens:


jikamens:
https://travis-ci.org/quantopian/zipline/jobs/53221597

    Error compiling Cython file:
    ------------------------------------------------------------
    ...
                other_as_int = other
            else:
                retvals = [True, True, False, True, False, False]
                return retvals[op]
            compared = cmp(self.sid, other_as_int)
                         ^
    ------------------------------------------------------------
    zipline/assets/_securities.pyx:102:22: undeclared name not builtin: cmp

Apparently you can use cmp in Cython code in Python 2 but not in Python 3?

[This page ](http://docs.cython.org/src/userguide/language_basics.html) claims that cmp is available and says nothing about not being allowed to use it in Python 3.

The function that's using cmp is pretty simple; can't we just implement the logic in-line? In particular, if `a` and `b` are integers, then `cmp(a, b)` can be replaced with `a - b`.

Note that the code in `__richcmp__` in _securities.pyx is buggy. It assumes that `cmp()` returns -1, 0, or 1, which is incorrect. It is only guaranteed to return negative, zero, or positive. So in addition to replacing the call to `cmp()` as described above, the remainder of the function after that will need to be rewritten slightly.

zipline/data/loader.py
```diff
@@ -182,19 +182,13 @@ def load_market_data(bm_symbol='^GSPC'):
     # we have data,then we need to update
     if len(days_up_to_now) - last_bm_date_offset > 1:
         benchmark_returns = update_benchmarks(bm_symbol, last_bm_date)
-        if (
-            benchmark_returns.index.tz is None
-            or
-            benchmark_returns.index.tz.zone != 'UTC'
-        ):
+        if benchmark_returns.index.tz is None or \
```
llllllllll: Is this one really better? what if the first statement was on the same line as the opening paren? Seeing a continuation `\` always alerts me

jikamens: Since this is outside the bounds of PEP8, it's a matter of stylistic opinion which is left to the discretion of the developer. I prefer it this way, and I did the work. So shut yer trap. ;-)

llllllllll: I didn't know if this was a pep8 change or not, since it is still an option I don't mind

ehebert:
For `__richcmp__` I think that `a -b` should work fine.

zipline/data/loader.py
jikamens: The PEP8 complaint was that I "or" was at the beginning of a line rather than the end of the previous line.

So if I had changed it as you proposed, it would have looked like this:

    if (benchmark_returns.index.tz is None or
        benchmark_returns.index.tz.zone != 'UTC'):
        ... stuff ...

Unfortunately, PEOP8 would have complained about that because the second line of the `if` conditional is at the same indent level as the first line of the body.

PEP8 is a harsh mistress.

ehebert:
When running `./zipline_repo/etc/ordered_pip.sh ./zipline_repo/etc/requirements.txt` locally I am hitting.

```
    from . import scimath as emath

ImportError: cannot import name scimath
```

jikamens:
>For `__richcmp__` I think that a -b should work fine.

I don't understand what this means.

ehebert:
Sorry, mistyped `__richcmp__` for `cmp`. I meant I agree your work around seems sound.

zipline/data/loader.py
llllllllll: I agree that breaking a line on a binop should leave the op on the first line; unfortunately, the pep8 tool forgets this line of pep8

> But most importantly: know when to be inconsistent -- sometimes the style guide just doesn't apply. When in doubt, use your best judgment. Look at other examples and decide what looks best. And don't hesitate to ask!

jikamens:
@ehebert can you give me more context for the `scimath` error you're getting? Like, the stack trace?

ehebert:
I am kicking off a run of the internal performance harness using these versions to check for any CPU/memory regressions.

jikamens:
>I am kicking off a run of the internal performance harness using these versions to check for any CPU/memory regressions.

Maybe wait until after I make the `__richcmp__` change?

ehebert:
```
Installing collected packages: scipy
  Found existing installation: scipy 0.12.0
    Uninstalling scipy:
      Successfully uninstalled scipy
  Running setup.py install for scipy
      File "<string>", line 1, in <module>
      File "/home/ehebert/envs/qexec-update-requirements/build/scipy/setup.py", line 249, in <module>
        setup_package()
      File "/home/ehebert/envs/qexec-update-requirements/build/scipy/setup.py", line 237, in setup_package
        from numpy.distutils.core import setup
      File "/home/ehebert/envs/qexec-update-requirements/lib/python2.7/site-packages/numpy/__init__.py", line 170, in <module>
        from . import add_newdocs
      File "/home/ehebert/envs/qexec-update-requirements/lib/python2.7/site-packages/numpy/add_newdocs.py", line 13, in <module>
        from numpy.lib import add_newdoc
      File "/home/ehebert/envs/qexec-update-requirements/lib/python2.7/site-packages/numpy/lib/__init__.py", line 17, in <module>
        from . import scimath as emath
    ImportError: cannot import name scimath
    Complete output from command /home/ehebert/envs/qexec-update-requirements/bin/python -c "import setuptools, tokenize;__file__='/home/ehebert/envs/qexec-update-requirements/build/scipy/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-vKlBPO-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/ehebert/envs/qexec-update-requirements/include/site/python2.7:

  File "<string>", line 1, in <module>

  File "/home/ehebert/envs/qexec-update-requirements/build/scipy/setup.py", line 249, in <module>

    setup_package()

  File "/home/ehebert/envs/qexec-update-requirements/build/scipy/setup.py", line 237, in setup_package

    from numpy.distutils.core import setup

  File "/home/ehebert/envs/qexec-update-requirements/lib/python2.7/site-packages/numpy/__init__.py", line 170, in <module>

    from . import add_newdocs

  File "/home/ehebert/envs/qexec-update-requirements/lib/python2.7/site-packages/numpy/add_newdocs.py", line 13, in <module>

    from numpy.lib import add_newdoc

  File "/home/ehebert/envs/qexec-update-requirements/lib/python2.7/site-packages/numpy/lib/__init__.py", line 17, in <module>

    from . import scimath as emath

ImportError: cannot import name scimath
```

jikamens:
@ehebert I don't understand the import error you are getting, it looks like something is broken in your venv? My numpy installation has scimath.py in the numpy/lib, and "from numpy.distutils.core import setup" works fine for me.

ehebert:
@jikamens my machine had been missing the OpenBLAS devel headers. That unstuck scipy.

warren-oneill:
In setup.py the Cython packet is written as Cython==0.20.1. Shouldn't this be written as Cython==0.22 or simply without version specification?

jikamens:
>In setup.py the Cython packet is written as Cython==0.20.1. Shouldn't this be written as Cython==0.22 or simply without version specification?

I'll remove the hard-coded version from setup.py.

jikamens:
Replaced a different PR.

PR #528, opened 2015-03-08
FIX: Fix Security for Python 3. Since we're just comparing ints, no need

dalejung:
for cmp


jikamens:
Already fixed with a (slightly different) fix that has been merged to master.

dalejung:
suhweet thank you

PR #530, opened 2015-03-11
FIX: Remove conda installing nose if we're just uninstalling and then reinstalling with pinned PIP version

dalejung:
@jikamens silly fix for travis. python packaging win.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -20,7 +20,7 @@ six==1.6.1
 # For fetching remote data
 requests==2.5.1
 
-Cython==0.20.1
+Cython>=0.20.1
 
 # faster OrderedDict
 cyordereddict==0.2.2
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
         'numpy',
         'pandas',
         'six',
-        'Cython==0.20.1'
+        'Cython>=0.20.1'
     ],
     extras_require={
         'talib':  ["talib"],
--- a/zipline/assets/_securities.pyx
+++ b/zipline/assets/_securities.pyx
@@ -97,14 +97,14 @@ cdef class Security:
         elif isinstance(x, int):
             x_as_int = x
         else:
-            return NotImplemented
+            return False
 
         if isinstance(y, Security):
             y_as_int = y.sid
         elif isinstance(y, int):
             y_as_int = y
         else:
-            return NotImplemented
+            return False
 
         compared = x_as_int - y_as_int
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
 install:
   - conda create -n testenv --yes pip python=$TRAVIS_PYTHON_VERSION
   - source activate testenv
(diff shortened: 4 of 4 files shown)


twiecki:
At least travis builds now. Remaining break:
```
ERROR: test_type_mismatch (tests.test_security_object.TestSecurityRichCmp)
  File "/home/travis/build/quantopian/zipline/tests/test_security_object.py", line 32, in test_type_mismatch
    self.assertIsNotNone(Security(3) < 'a')
TypeError: unorderable types: zipline.assets._securities.Security() < str()
```

Should definitely merge nonetheless. /sign-off

dalejung:
kicked off new travis. I don't think that `x` needs to be type checked since it's always Security for richcmp, but I didn't muck withit.

dalejung:
So I unpinned cython to specifically require 0.20.1. Was annoying for zipline to degrade my environment. If we specifically need that version for some reason, we can re-pin. but then we should remove it from the conda install since setup.py will downgrade anyways.

zipline/assets/_securities.pyx
```diff
@@ -97,14 +97,14 @@ cdef class Security:
         elif isinstance(x, int):
             x_as_int = x
         else:
-            return NotImplemented
+            return False
```
jikamens: Do you have reason to believe that "False" rather than "NotImplemented" is correct? The documentation is bad and sparse, but what I could find suggested that NotImplemented was the correct thing to return when the __richcmp__ function doesn't know how to do the comparison. And it passes the test I wrote to confirm that it doesn't crash.

etc/requirements.txt
```diff
@@ -20,7 +20,7 @@ six==1.6.1
 # For fetching remote data
 requests==2.5.1
 
-Cython==0.20.1
+Cython>=0.20.1
```
jikamens: Shouldn't we list the same version here that we list in requirements.txt?

jikamens: We use specific versions in requirements.txt, not >=, so this should not be changed.

setup.py
```diff
@@ -57,7 +57,7 @@
         'numpy',
         'pandas',
         'six',
-        'Cython==0.20.1'
+        'Cython>=0.20.1'
```
jikamens: setup.py should match requirements.txt.

jikamens:
>kicked off new travis. I don't think that x needs to be type checked since it's always Security for richcmp, but I didn't muck withit.

This is not correct. Read the commit in which I made this change.

jikamens:
>So I unpinned cython to specifically require 0.20.1. Was annoying for zipline to degrade my environment.

Then use a virtualenv. Our convention is to peg the versions of all the packages we use.

jikamens:
Hmm, I see Thomas's comment now about the TypeError from the test.

I didn't get that TypeError. I will have to investigate further. Please don't merge your change until I've had a chance to investigate.

dalejung:
this PR is garbage. @jikamens just cherry-pick the nose removal. Test failed due to python 3 comparison stuff.

ehebert:
One of the commits was cherry-picked into d593bde6689a4effb67a57c8ec07680fad3d08b1

PR #533, opened 2015-03-12
BUG: Calculating the time to process user handle_data

lianga888:


zipline/gens/tradesimulation.py
```diff
@@ -22,6 +22,7 @@
     DATASOURCE_TYPE
 )
 from zipline.gens.utils import hash_args
+from qcommons.date_utils import utcnow
```
ehebert: qcommons is not available to Zipline.

This timing might be better done, for now at least, on the internal wrapper of handle_data.

zipline/gens/tradesimulation.py
```diff
@@ -266,11 +268,14 @@ def _call_handle_data(self):
...
         during the call.
         """
+        time_start = utcnow()
         self.algo.event_manager.handle_data(
             self.algo,
             self.current_data,
             self.simulation_dt,
         )
+        time_end = utcnow()
+        self.algo.call_handle_data_duration.append((time_end-time_start).total_seconds()*1000)
```
ehebert: Also, if the seconds in between is what is desired, I'd recommend just using time.time(), avoids the overhead of creating the datetime/Timestamp/timedeltas.

lianga888: If I use time.time(), would I be able to keep the time calculations in zipline instead of needing to move it to the internal wrapper of handle_data (which I assume is call_algorithm_handle_data in algoproxy)?

ehebert: It would make it more feasible.

I would still lean towards putting it in call_algorithm_handle_data, so that those who aren't using this metadata don't need to run it every bar, and also to leave tradesimulation as focused as possible on getting the trade data in the right places.

lianga888: Ok that makes sense -- i'll move it then

ehebert:
@lianga888 going to close because of decision to implement internally.

PR #534, opened 2015-03-13
Dataverse

dalejung:
WIP 
## WIP 
# WIP.

Based off of the position_tracker refactor. Once that gets merged it'll cut down the noise.

Executive Summary: We have array backed datasources. All path-*independent* data should **not** be calculated iteratively. Vectorize and fast-path the execution.

* [ ] Refactor the HistoryChunker into a UniverseChunker. In general, I'd like the BarData to be based off of that.
* [x] Have a MetaDataverse that choses the BaseDataverse / BacktestDataverse depending on what datasource it gets. Right now it's chosen explicitly, which would normally be fine, except the RandomWalkSource balks since it's not an array.
* [ ] Change position tracker to literally only track positions. Position values should be handled higher up and not require update_last_sale.
* [ ] Build out a matrix of the types of TradeEvents/Dataverses we can have. Technically speaking, the DataPanelSource should be allowed to emit WideTradeBars while using the BaseDataverse.
* [ ] Create a DataSource that is compatible with BacktestDataverse, this will only emit TradeEvents that are used to signifiy new data, but won't necessarily carry any data with them, be it dicts or ndarrays.
* [ ] PositionTracker should emit new Position events. Somewhere we need some sort of pubsub so that when Universes get ported, we are not polling the Positions for sids, but instead only updating when positions are created/deleted.
* [ ] Once Magic BacktestBarData is setup, also wire it to expose vectorized computations from Research.
* [ ] `history` should return a DataFrame proxy so that we can fast-path lines like `history(10, '1m', 'price').mean()`. The goal is to not even create the DataFrame in the first place.
* [ ] Outline the require immutability of objects like BarData and HistoryFrame. This can be opt-in for speed.
* [ ] When the chunked historycontainer rolls over, carry over the last known values for ffill.


zipline/sources/data_frame_source.py
```diff
@@ -111,46 +112,43 @@ def __init__(self, data, **kwargs):
...
+        # treat yoself
+        pass
 
     @property
-    def raw_data(self):
-        if not self._raw_data:
-            self._raw_data = self.raw_data_gen()
-        return self._raw_data
+    def mapped_data(self):
+        # currently volume is same dtype as prices, float64
```
ehebert: It's probably better, or at least ok, to have the volumes as float64.

For places where volume needs to be multiplied by the price, it should be faster to do the float*float.

Also, allows for shaping data in both directions, i.e. ohlcv can be a vector... if we care about that.

zipline/gens/tradesimulation.py
```diff
@@ -206,9 +198,6 @@ def _process_snapshot(self, dt, snapshot, instant_fill):
         any_trade_occurred = False
         benchmark_event_occurred = False
 
-        if instant_fill:
```
ehebert: Should we just remove this now?

ehebert:
Did a first pass through and like with where most of this is headed, but need to digest it a bit.
Especially like the widetradbar, and the history proxy.

The post-position tracker rebase will definitely help the read through. I'll ping you as soon as that gets in.

ssanderson:
This seems unlikely to be merged.  Closing.

PR #535, opened 2015-03-16
Refactor Security to Asset

jfkirk:
This commit refactors the Security cython class to Asset, and refactors some fields of the class accordingly. This change is so the terminology is consistent and correct when Asset is extended to asset types that are not securities, such as futures.


llllllllll:
:+1: on clear naming

ehebert:
The tests pass locally for me as well, I just kicked off a new Travis build to see if the error is repeatable.

ehebert:
Besides the tests, this looks good to me.

Perhaps also add Security to the module as a subclass to show how it would be used as a base?

jfkirk:
@ehebert Will do. This PR should not be merged until that has been added.

jfkirk:
@ehebert Added extensions for Equity and Future, as we've discussed. I also added a test for the Future class that checks that the fields are assigned properly and reported through the repr. Thanks so much for all of your help.

PR #523, opened 2015-03-04
More flexible calendar implementation

warren-oneill:
Zipline is currently very North America focused in the implentation of the calendars and it is not as modular as it should be. 
For example, the NYSE calender/times are hardcoded a number of times. Here I have changed the hardcoding of the get_early_closes to use the given calendar and not always the NYSE one.
There are many more points where more flexiblity is need but here is a good starting point.

PR #540, opened 2015-03-20
Sql datasouce

warren-oneill:
I attempted to implement the framework for communicating with an SQL database using SqlAlchemy. In order to do this I added an SqlDataSource class and tested it with similar unittests to the DataFrame unittests. 
There remains some open design questions, namely:
* What variables should SqlDataSource receive? Currently I used the sql engine and the orm object but this limits the method to use only one orm object and in reality we will prob want to use several.
* Where should the SID be saved? At the moment it is a column in the orm object but I'm sure there must be a more efficient way to do this.  

I'm open for suggestions from the Sql gurus out there :)

Note: this PR contains changes from PR #523.


warren-oneill:
sorry for all the commits. prob should have made a fresh branch and added the final changes.

tests/test_sources.py
```diff
@@ -80,6 +82,25 @@ def test_yahoo_bars_to_panel_source(self):
             self.assertEqual(next(stocks_iter), event['sid'])
 
 
+class TestSqlSource(TestCase):
+    def test_sql_source(self):
+        source, df = factory.create_test_sql_source()
+        assert isinstance(source.start, pd.lib.Timestamp)
```
llllllllll: We use the `unittest.TestCase.asssert*` methods to provide nicer messages and ensure that the tests work under `-O`.

warren-oneill: Ok will do but I took that code from TestDataFrameSource so that class should also be changed.

llllllllll: Okay, I didn't realize we had this pattern still. If you don't want to update the older tests, we can still support the cleaner way going forward and change it here.

warren-oneill: Hey, updated all tests.

jfkirk:
@warren-oneill Given the overlap between this and the trading calendar PR, should we focus on the calendar PR first and revisit this once that is merged?

warren-oneill:
@jfkirk yes that would be great. It should also be a much easier task than this one.

jfkirk:
@warren-oneill I am ready to review this PR now. Could you rebase/squash these commits on to master, and I will review and test?

warren-oneill:
@jfkirk @llllllllll I still think this PR has a way to go before its up to scratch so I'll close it for the time being. We are currently implementing a version of this PR for our own DB and I think it makes sense to wait until we've ironed that out and I can then work on making something commitable. It also might be better to wait until the Asset stuff is finished first. What you guys think?

sorying:
@warren-oneill - will you be thinking to re-open this PR? Zipline now use data bundle and would be great if we can utilize SQLAlchemy to read data from DB

PR #541, opened 2015-03-20
Widebar

dalejung:
The Widebar stuff from dataverse. Generally idea is to make bars fat so that their processing can be cythonized. Um, so new flake8 rules? Need to deal with those.

### DIFF
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -56,7 +56,7 @@ class TestDataFrameSource(TestCase):
                 self.assertTrue('sids' in event)
                 self.assertTrue('arbitrary' in event.columns)
                 arbitrary_idx = event.columns.get_loc('arbitrary')
-                self.assertEquals(event.vals[0][arbitrary_idx], 1.)
+                self.assertEquals(event.values[0][arbitrary_idx], 1.)
                 self.assertEquals(event.sids[0], 0)
             else:
                 self.assertTrue('sid' in event)
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -26,6 +26,7 @@ from zipline.sources import (DataFrameSource,
                              DataPanelSource,
                              RandomWalkSource)
 from zipline.utils import tradingcalendar as calendar_nyse
+import zipline.protocol as zp
 
 
 class TestDataFrameSource(TestCase):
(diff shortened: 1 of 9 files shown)


zipline/protocol.py
```diff
@@ -486,9 +524,10 @@ class BarData(object):
     usage of what this replaced as a dictionary subclass.
     """
 
-    def __init__(self, data=None):
+    def __init__(self, data=None, siddata_class=SIDData):
```
ehebert: What would be the non-SIDData option for the class here?

zipline/protocol.py
```diff
@@ -506,6 +545,26 @@ def has_key(self, name):
         """
         return name in self
 
+    def get_default(self, name):
```
ehebert: So this is a `defaultdict` like behavior? The `get_default` name feels a little disjointed, since it doesn't always return the default, maybe something like `get_or_default` or just `get`?

zipline/protocol.py
```diff
@@ -506,6 +545,26 @@ def has_key(self, name):
...
+        try:
+            sid_data = self[name]
+        except KeyError:
+            sid_data = self[name] = self._siddata_class(name)
+        return sid_data
+
+    def update_sid(self, event):
+        if isinstance(event, WideTradeEvent):
```
ehebert: I might be asking for a premature optimization, but `isinstance` checks have been a bottleneck before.

However, I can see why a separate trade event type is not used, which would be the way to avoid doing this check on every minute.

Thoughts about a boolean in tradesimuaion/algorithm of 'use wide bars?'

zipline/protocol.py
llllllllll: +1 on `get`

zipline/finance/blotter.py
```diff
@@ -48,6 +49,11 @@
 )
 
 
+def identity(self, sid):
```
llllllllll: Why are we using `self` as the identifier if this is not a method?

zipline/protocol.py
dalejung: A different sublcass of SIDData, i.e. dataverse SIDData

zipline/protocol.py
dalejung: hm. I got get_default from when I replaced the Positions defaultdict, where just using get felt wrong. Though I suppose `defaultdict.get` does it's business that way too. I want something like `setdefault` but with the hint that we will create an object in the method name and not the class type.

tests/test_sources.py
```diff
@@ -51,14 +52,21 @@ def test_panel_source(self):
...
         for event in source:
-            self.assertTrue('sid' in event)
-            self.assertTrue('arbitrary' in event)
-            self.assertTrue(hasattr(event, 'volume'))
-            self.assertTrue(hasattr(event, 'price'))
-            self.assertEquals(event['arbitrary'], 1.)
-            self.assertEquals(event['sid'], 0)
-            self.assertTrue(isinstance(event['volume'], int))
-            self.assertTrue(isinstance(event['arbitrary'], float))
+            if isinstance(event, zp.WideTradeEvent):
```
ehebert: Should this use `is_wide` as well?

zipline/finance/blotter.py
```diff
@@ -194,37 +200,48 @@ def process_trade(self, trade_event):
...
 
-        if trade_event.sid not in self.open_orders:
-            return
+        try:
+            sids_set = trade_event.sids_set
+            sid_ohlcv = trade_event.sid_ohlcv
+        except:
+            # handle any Event classes
+            sids_set = {trade_event.sid}
+            sid_ohlcv = partial(identity, trade_event)
```
ehebert: Why is the identity wrapper needed?

ssanderson:
I think this has been superceded by the Q 2.0 branch.

PR #549, opened 2015-03-30
MAINT: Added missing self to refer to sim_params and ENH: Keep record of the price effect due to slippage

nrey:
Missing self keyword


nrey:
It fails the test: FAIL: test_object_serialization_13_Transaction tests.test_versioning.VersioningTestCase), in Travis CI but it passes the test in my local copy.

PR #550, opened 2015-03-30
MAINT: Remove left over simple transform code.

ehebert:
Remove pieces that are no longer used now that the simple transforms are
wrappers around history via the SIDData object.

Move window length related pieces into batch_transform, since the rest
of the utils module is no longer used.


ehebert:
cc: @llllllllll

PR #554, opened 2015-04-04
Implementing Relative Strength index (RSI) as another example algorithm.

sebnil:
skarmavbild 2015-04-04 kl 22 20 20


twiecki:
Thanks, this looks like a great example to add. Note that we enforce PEP8 syntax formatting. You can e.g. use `autopep8` to format your code.

sebnil:
I am a bit new to collaborative python projects. The automated tests and getting reports back was kinda cool. 
Anyway, whats left to get these commits merged?

twiecki:
It seems like your example actually doesn't require the changes to the commission model. It's best to have these kind-of things be separate.

sebnil:
My mistake. It should have been in another pull request.

2015-04-05 14:45 GMT+02:00 Thomas Wiecki <notifications@github.com>:

FreddieV4:
Hi @sebnil I'm going to close this PR as it's been a while since it's last update and many API changes have occurred since this was opened.

Also we now have an example that uses the Pipeline API an `RSI` [here](https://github.com/quantopian/zipline/blob/master/zipline/examples/momentum_pipeline.py). If you'd like to add other examples, feel free to open another PR 🙂

PR #555, opened 2015-04-06
BUG: Replaced deprecated operators for pd.Index

lianga888:
Replaced + and - with union and difference, respectively


ssanderson:
@lianga888 do we have a repro case for the algos that this was breaking?  If so can we add a test to verify that this in fact fixes the crash?

llllllllll:
https://www.quantopian.com/posts/function-that-is-not-called-crashes-backtest-returns-will-not-work-either

I think this hits the case

lianga888:
Okay, I can write a test for it

lianga888:
This was the corresponding zipline PR for https://github.com/quantopian/qexec/pull/5999, which is now closed

PR #559, opened 2015-04-07
Handle a ValueError on from_csv calls

doctorpangloss:
The cached market data could be corrupted. Pandas raises a ValueError in that case, and this error handles it.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -15,7 +15,7 @@ patsy==0.3.0
 statsmodels==0.6.1
 
 # Don't use 2.4.0.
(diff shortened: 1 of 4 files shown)


ssanderson:
@doctorpangloss sorry for the long delay on this, I've been away at PyCon.  This looks good to me, but needs a rebase against master.  I can also pull in your commit manually if you'd prefer.

ssanderson:
@doctorpangloss I pulled your commit into https://github.com/quantopian/zipline/pull/563, which has been merged.  Thanks for the help!

Issue #538, opened 2015-03-19
fast_backtest option


florentchandelier:
I have suggested a  very simple fast_backtest option to be passed to TradingAlgorithm()
see: https://github.com/florentchandelier/zipline/tree/wip-ImproveBacktestSpeed

-> I'm not creating a pull request as I'm sure there is a cleaner and more elegant way to do it.

Speed improvements: 15x/20x
Code Change: minimal, I can easily maintain this while being current with official Zipline branches

Enhancement:
Currently, when fast_backtest is active, to get risk analysis and performance after the backtest, I simply track "context.portfolio.portfolio_value" in "def handle_data(context, data)", in a pd.Series  ... which would be a great default behavior for algo.run to return when fast_backtest is active.

ehebert:
@florentchandelier cutting out the intermediate results when the aren't used is a nice win.

A flag name alike to `generate_intermediate_results` might be good in place of `fast_backtest`. (Though that would flip the logic around.)

Adding a compare link https://github.com/florentchandelier/zipline/compare/6cd2f4f1f2f4a4dca74fcfd6484278cc90e6f77a...florentchandelier:wip-ImproveBacktestSpeed for reference.

Assigning this to myself, as I may try to rework this soon.

ehebert:
Put up a WIP branch here https://github.com/quantopian/zipline/tree/no-intermediate-results

florentchandelier:
Hi, 
thanks for the effort. I've tried it but got the error presented below (simple to fix but I believe like before it's better you address this in a manner consistent with Quantopian usage of Zipline): - otherwise speed seems to be similar to what I have with my hack, eventually even a little bit faster. I have a file ready for comparison when this is fixed.

     results = algo.run(data, intermediate_results=False)
  File "/usr/local/lib/python2.7/dist-packages/zipline-0.8.0rc1-py2.7-linux-x86_64.egg/zipline/algorithm.py", line 466, in run
    for perf in self.gen:
  File "/usr/local/lib/python2.7/dist-packages/zipline-0.8.0rc1-py2.7-linux-x86_64.egg/zipline/gens/tradesimulation.py", line 181, in transform
    risk_message = self.algo.perf_tracker.handle_simulation_end()
  File "/usr/local/lib/python2.7/dist-packages/zipline-0.8.0rc1-py2.7-linux-x86_64.egg/zipline/finance/performance/tracker.py", line 492, in handle_simulation_end
    benchmark_returns=bms)
[...]

PR #561, opened 2015-04-09
MAINT: Remove unused returns vector from perf tracker.

ehebert:
The returns Series in the perf tracker appears to be vestigial, so
remove.

Discovered while investigating what needs to be updated during 'fast backtest' mode for https://github.com/quantopian/zipline/issues/538


twiecki:
Agree with you that this looks like an oversight after a refactoring. No user facing code should depend on this so /sign-off once the tests pass.

ehebert:
All tests had passed, merged.

CaptainKanuk:
:+1:

PR #566, opened 2015-04-16
ENH: Add Max Drawdown Duration as a risk parameter

warren-oneill:
As well as _Max Drawdown_ the _Max Drawdown Duration_ is also widely used to evaluate strategies and it would be a great addition to the set of risk parameters. 
To test the cumulative and period calculations of MDD I used `factory.create_returns_from_list` and gave the same array used in `test_risk_period.TestRisk.test_drawdown`. One misgiving I have is that this array is possible too short for a thorough test but unfortunately I don't know exactly how I could add an additional longer answer to the AnswerKey.
Also I have calculated everything in terms of days. Will that still hold when the algo has a frequency of minutes? Are the risk parameters still calculated on a daily basis?
Note: the  _Max Drawdown Duration_ is the duration of the longest drawdown and not the duration of the _Max Drawdown_.

### DIFF
--- a/tests/risk/test_risk_cumulative.py
+++ b/tests/risk/test_risk_cumulative.py
@@ -17,6 +17,7 @@ import unittest
 
 import datetime
 import numpy as np
+import pandas as pd
 import pytz
 import zipline.finance.risk as risk
 from zipline.utils import factory
@@ -116,3 +117,21 @@ class TestRisk(unittest.TestCase):
                 self.cumulative_metrics_06.max_drawdowns[dt],
                 value,
                 err_msg="Mismatch at %s" % (dt,))
+
+    def test_drawdown_duration(self):
+        returns = factory.create_returns_from_list(
+            [1.0, -0.5, 0.8, .17, 1.0, -0.1, -0.45], self.sim_params)
+        # 200, 100, 180, 210.6, 421.2, 379.8, 208.494
+        drawdown_duration = pd.Series(
+            [0, 2, 3, 3, 3, 3, 3], index=returns.index)
+
+        metrics = risk.RiskMetricsCumulative(self.sim_params)
+
+        for dt, value in returns.iteritems():
+            metrics.update(dt,
+                           value,
+                           0,  # benchmark used so set to zero
+                           {'leverage': 0.0})
+
+        for dt, value in drawdown_duration.iteritems():
+            self.assertEqual(metrics.max_drawdown_durations[dt], value)
--- a/tests/risk/test_risk_period.py
+++ b/tests/risk/test_risk_period.py
@@ -102,6 +102,15 @@ class TestRisk(unittest.TestCase):
                                          returns)
         self.assertEqual(metrics.max_drawdown, 0.505)
 
+    def test_drawdown_duration(self):
+        returns = factory.create_returns_from_list(
+            [1.0, -0.5, 0.8, .17, 1.0, -0.1, -0.45], self.sim_params)
+        # 200, 100, 180, 210.6, 421.2, 379.8, 208.494
+        metrics = risk.RiskMetricsPeriod(returns.index[0],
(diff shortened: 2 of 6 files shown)


jfkirk:
Checks out to me - running a perf matrix job to make sure that it doesn't hurt backtest performance

warren-oneill:
@jfkirk any word on this?

jfkirk:
Sorry, got waylaid to handle bugs. Will confirm today and merge if it checks out.

jfkirk:
Hey Warren - this PR checks out through-and-through, but has caused some internal problems for us so I can't merge it quite yet. I'm working to resolve the issue ASAP. I know this is frustrating; I'm very sorry for the delay.

warren-oneill:
Hey no prob at all. Thanks for taking the time to look at it.

jfkirk:
Just to follow-up: we are testing out a process now that will make adding new perf metrics much easier (on our end). Once that is hammered out, I will revisit this PR. Thanks for your patience!

warren-oneill:
@jfkirk :+1:

warren-oneill:
@jfkirk @ssanderson bump!
are the new perf metrics methods added?

warren-oneill:
@jfkirk @ssanderson ping!

warren-oneill:
@twiecki maybe this is something for pyfolio?

twiecki:
@warren-oneill Most certainly, but this PR looks good for zipline as well. For pyfolio it should actually be fairly easy as we already have functionality to extract the top drawdowns and their duration.

twiecki:
@jfkirk This looks good to me, want to merge?

justinlent:
@warren-oneill this is actually already in pyfolio. The right-most column of the numerical table describing the Top Drawdowns.  Here's an example

image

ssanderson:
If this is already in pyfolio, should we close this?  I'm inclined to say that we should avoid adding risk metrics that can be derived from the backtest output.  Adding them to Zipline, at least with its current architecture, means that everyone incurs the cost of computing them, whether they're used or not.  It's much easier to only compute what you care about if you're doing analysis after the fact in PyFolio.

warren-oneill:
@justinlent looks good. what function did you use to produce that table?
@ssanderson by that logic almost all risk metric calculation which is already in pyfolio should be removed from zipline which in general is not a problem but there are times where you need to know the values of certain metrics during a run.

cmorgan:
@warren-oneill pyfolio.plotting.show_worst_drawdown_periods produces the drawdown table

justinlent:
@warren-oneill as well, along what @cmorgan mentioned, you can get the entire dataframe of the data contained in the drawdown table by calling the function (which `plotting.show_worstdrawdown_periods` uses) `timeseries.gen_drawdown_table`

ssanderson:
> @ssanderson by that logic almost all risk metric calculation which are already in pyfolio should be removed from zipline which in general is not a problem but there are times where you need to know the values of certain metrics during a run.

@warren-oneill we're actually looking at pulling most of the existing risk metrics out of both `zipline` and `pyfolio` so that they can share implementations (we've had a couple bugs recently where they were doing the same calculation in subtly different ways).

I think in the very long term it would probably be a good idea remove most of the in-simulation risk calculations, or at least make them optional.  The main reason we do those calculations on the fly is that Quantopian's UI shows a rolling chart of an algorithm's returns while it's simulating.  That's nice from a UX perspective, but if Zipline's output isn't being consumed by a GUI, then we're wasting a lot of CPU cycles doing those calculations every day when they could be done at the end of the backtest in a vectorized fashion, or even deferred until the user actually asks for them.

FreddieV4:
Going to close this due to inactivity, and because we have Pyfolio/Empyrical for risk metrics. If you'd like to rebase and open a new PR/start a new discussion feel free to go ahead and do that :)

PR #568, opened 2015-04-17
Scrubbing known warnings

lianga888:


zipline/finance/risk/cumulative.py
```diff
@@ -462,16 +462,17 @@ def calculate_beta(self):
...
 
         # Drop nans if there are gaps in the data
         algorithm_returns = self.algorithm_returns.dropna()
         benchmark_returns = \
             self.benchmark_returns.loc[algorithm_returns.index]
 
+        # it doesn't make much sense to calculate beta for less than two days,
+        # so return none.
+        if len(algorithm_returns) < 2:
+            return 0.0
```
richafrank: Can you explain this change @lianga888 ?

lianga888: Talked with @ehebert about this earlier. The warning is returned when we try to calculate covariance between the benchmark and algo when there is only one data point of returns. len(self.annualized_mean_returns) <2: return 0.0 is meant to address that case. However, this length is calculated before NaNs are dropped from algorithm_returns and benchmark_returns

richafrank: Ah, awesome - sounds good.

jikamens:
@lianga888 @richafrank @twiecki @ehebert Is this PR ready to merge? Because the corresponding Qexec PR that references it by commitish has already been merged to master, which means that currently Qexec master points to a Zipline commitish that isn't on its master. That's no good and needs to be cleaned up ASAP.

jikamens:
**BEWARE. I have just force-pushed this branch*** to rebase it against Zipline master.

lianga888:
I had @richafrank look over it a few days ago, so I believe it's good to go

lianga888:
Sorry, I just realized I never merged my corresponding zipline and qcommons PR when I merged my catch warnings PR yesterday

jikamens:
For future reference:

- I would like to see more detailed commit messages, especially when checking code into Zipline, which is a public repository.
- We try to use commit type labels in the summary lines of commits in Zipline. See https://github.com/quantopian/zipline/wiki/Zipline-development-guidelines .

PR #570, opened 2015-04-22
Remove leaky defaultdicts

ehebert:
Stop the growth of members of PerformancePeriod every time that `to_dict` is called with `dt` parameter.

Found by running `objgraph.show_growth()` before and after every packet during a minute emission simulation. The show growth was reporting the following per simulation minute.

```
list           25890        +2
Timestamp      62445        +1
OrderedDict     1020        +1
```

After these changes, there is no growth shown between market minutes.

### DIFF
--- a/zipline/finance/performance/period.py
+++ b/zipline/finance/performance/period.py
@@ -167,7 +167,7 @@ class PerformancePeriod(object):
         self.period_cash_flow = 0.0
         self.pnl = 0.0
         self.processed_transactions = defaultdict(list)
-        self.orders_by_modified = defaultdict(OrderedDict)
+        self.orders_by_modified = {}
         self.orders_by_id = OrderedDict()
 
     def handle_dividends_paid(self, net_cash_payment):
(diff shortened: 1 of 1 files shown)


zipline/finance/performance/period.py
```diff
@@ -203,9 +200,12 @@ def calculate_performance(self):
 
     def record_order(self, order):
         if self.keep_orders:
-            dt_orders = self.orders_by_modified[order.dt]
-            if order.id in dt_orders:
-                del dt_orders[order.id]
+            try:
+                dt_orders = self.orders_by_modified[order.dt]
+                if order.id in dt_orders:
+                    del dt_orders[order.id]
```
ssanderson: Why do we bother `del`ing this if we immediately insert it back in on 209?

ehebert: That's a good question. I was trying to keep existing logic, and merely change the lookup behavior.

But you're right, removing the del should be functionally equivalent. I'll tack on a patch addressing that.

ssanderson: The only reason I can think of that this might matter is if we have a custom __hash__ or something strange like that.

ehebert: Also, right below the `del` is used to preserve the FIFO ordering of orders, but that is not relevant with `orders_by_modified` which is just a normal dictionary... but that probably explains why the `del` crept in there.

ehebert: Oh, this still does use an OrderedDict. So should I think we should preserve it, though that behavior appears to be uncovered in the zipline tests. (Was able to remove the `del` and tests still passed.)

dalejung: sometimes the ordering will mess up the test harness. I had something in blotter which didn't need to be ordered, but due to the SlippageModel semantics ended up having different algo exhaust.

ehebert: Going to leave the del to preserve ordering for now, then.

PR #571, opened 2015-04-24
DO NOT MERGE: Futures trading

jfkirk:
** Work in progress **

PR for review and comparison. Please tear it apart.

This branch adds Asset management to Zipline and adds the necessary accounting to allow for trading of Futures.

**Primary impacts:**
**1. There is now a distinction between an Asset's 'identifier', 'sid', and 'symbol'.**
    * 'identifier': the user-provided unique int or str that allows us to marry-up AssetMetaData and DataSource data
    * 'sid': a unique integer that can be used to identify a specific Asset within a given TradingEnvironment
    * 'symbol': non-unique string name for user look-up and ease-of-use, but is not used internally
    * This distinction has not been completely refactored in to the API – many methods still call for an argument 'sid' where they would actually want an Asset object
**2. Zipline user must provide 'identifiers' or 'asset_metadata' arguments to TradingAlgorithm.__init__ so the asset_finder can be built.**
**3. Zipline algorithms must use Asset objects as arguments to all order*() methods.**
    * The sid(int) and symbol(str) methods can be used to look up an Asset
**4. There is now a distinction between 'value' and 'exposure' in the perf tracker and positions tracker.**

** In Progress **
1. Documentation (waiting until impacts are finalized)
2. API method argument classes/naming (SID v Asset)

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -36,6 +36,7 @@ from zipline.errors import (
     RegisterTradingControlPostInit,
     TradingControlViolation,
     AccountControlViolation,
+    SymbolNotFound,
 )
 from zipline.test_algorithms import (
     access_account_in_init,
@@ -54,6 +55,7 @@ from zipline.test_algorithms import (
     TestTargetPercentAlgorithm,
     TestTargetValueAlgorithm,
     SetLongOnlyAlgorithm,
+    SetAssetDateBoundsAlgorithm,
     SetMaxPositionSizeAlgorithm,
     SetMaxOrderCountAlgorithm,
     SetMaxOrderSizeAlgorithm,
@@ -84,6 +86,9 @@ from zipline.sources import (SpecificEquityTrades,
                              DataFrameSource,
                              DataPanelSource,
                              RandomWalkSource)
+from zipline.assets import (
+    Equity, Future
+)
 
 from zipline.finance.execution import LimitOrder
 from zipline.finance.trading import SimulationParameters
@@ -173,12 +178,12 @@ class TestMiscellaneousAPI(TestCase):
             if algo.minute == 0:
 
                 # Should be filled by the next minute
-                algo.order(1, 1)
+                algo.order(algo.sid(1), 1)
 
                 # Won't be filled because the price is too low.
-                algo.order(2, 1, style=LimitOrder(0.01))
-                algo.order(2, 1, style=LimitOrder(0.01))
-                algo.order(2, 1, style=LimitOrder(0.01))
+                algo.order(algo.sid(2), 1, style=LimitOrder(0.01))
+                algo.order(algo.sid(2), 1, style=LimitOrder(0.01))
+                algo.order(algo.sid(2), 1, style=LimitOrder(0.01))
 
                 all_orders = algo.get_open_orders()
                 self.assertEqual(list(all_orders.keys()), [1, 2])
@@ -288,6 +293,58 @@ class TestMiscellaneousAPI(TestCase):
 
         self.assertIs(composer, zipline.utils.events.ComposedRule.lazy_and)
 
+    def test_asset_lookup(self):
+        metadata = {0: {'symbol': 'PLAY',
+                        'asset_type': 'equity',
+                        'start_date': '2002-01-01',
+                        'end_date': '2004-01-01'},
+                    1: {'symbol': 'PLAY',
+                        'asset_type': 'equity',
+                        'start_date': '2005-01-01',
+                        'end_date': '2006-01-01'},
+                    2: {'symbol': 'OMG15',
+                        'asset_type': 'future'}}
+        algo = TradingAlgorithm(asset_metadata=metadata)
+
+        # Test before either PLAY existed
+        algo.datetime = pd.Timestamp('2001-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        with self.assertRaises(SymbolNotFound):
+            algo.symbol('PLAY')
+        with self.assertRaises(SymbolNotFound):
+            algo.symbols('PLAY', 'OMG15')
+
+        # Test when first PLAY exists
+        algo.datetime = pd.Timestamp('2002-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        self.assertEqual(0, algo.symbol('PLAY'))
+        list_result = algo.symbols('PLAY', 'OMG15')
+        self.assertEqual(0, list_result[0])
+        self.assertEqual(2, list_result[1])
+
+        # Test after first PLAY ends
+        algo.datetime = pd.Timestamp('2004-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        self.assertEqual(0, algo.symbol('PLAY'))
+
+        # Test after second PLAY begins
+        algo.datetime = pd.Timestamp('2005-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        self.assertEqual(1, algo.symbol('PLAY'))
+
+        # Test after second PLAY ends
+        algo.datetime = pd.Timestamp('2006-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        self.assertEqual(1, algo.symbol('PLAY'))
+        list_result = algo.symbols('PLAY', 'OMG15')
+        self.assertEqual(1, list_result[0])
+        self.assertEqual(2, list_result[1])
+
+        # Test lookup SID
+        self.assertIsInstance(algo.sid(0), Equity)
+        self.assertIsInstance(algo.sid(1), Equity)
+        self.assertIsInstance(algo.sid(2), Future)
+
 
 class TestTransformAlgorithm(TestCase):
     def setUp(self):
@@ -379,19 +436,36 @@ class TestTransformAlgorithm(TestCase):
         )
         self.assertEqual(algo.sim_params.data_frequency, 'minute')
 
-    def test_order_methods(self):
-        AlgoClasses = [TestOrderAlgorithm,
-                       TestOrderValueAlgorithm,
-                       TestTargetAlgorithm,
-                       TestOrderPercentAlgorithm,
-                       TestTargetPercentAlgorithm,
-                       TestTargetValueAlgorithm]
+    @parameterized.expand([
+        (TestOrderAlgorithm,),
+        (TestOrderValueAlgorithm,),
+        (TestTargetAlgorithm,),
+        (TestOrderPercentAlgorithm,),
+        (TestTargetPercentAlgorithm,),
+        (TestTargetValueAlgorithm,),
+    ])
+    def test_order_methods(self, algo_class):
+        algo = algo_class(
+            sim_params=self.sim_params,
+        )
+        algo.run(self.df)
 
-        for AlgoClass in AlgoClasses:
-            algo = AlgoClass(
-                sim_params=self.sim_params,
-            )
-            algo.run(self.df)
+    @parameterized.expand([
+        (TestOrderAlgorithm,),
+        (TestOrderValueAlgorithm,),
+        (TestTargetAlgorithm,),
+        (TestOrderPercentAlgorithm,),
+        (TestTargetValueAlgorithm,),
+    ])
+    def test_order_methods_for_future(self, algo_class):
+        metadata = {0: {'asset_type': 'future',
+                        'contract_multiplier': 10}}
+
+        algo = algo_class(
+            sim_params=self.sim_params,
+            asset_metadata=metadata
+        )
+        algo.run(self.df)
 
     def test_order_method_style_forwarding(self):
 
@@ -488,7 +562,8 @@ class TestAlgoScript(TestCase):
             self.sim_params
         )
 
-        self.source = SpecificEquityTrades(event_list=trade_history)
+        self.source = SpecificEquityTrades(sids=[133],
+                                           event_list=trade_history)
         self.df_source, self.df = \
             factory.create_test_df_source(self.sim_params)
 
@@ -538,7 +613,8 @@ from zipline.api import (slippage,
                          set_slippage,
                          set_commission,
                          order,
-                         record)
+                         record,
+                         sid)
 
 def initialize(context):
     model = slippage.FixedSlippage(spread=0.10)
@@ -549,7 +625,7 @@ def initialize(context):
 
 def handle_data(context, data):
     if context.incr < context.count:
-        order(0, -1000)
+        order(sid(0), -1000)
     record(price=data[0].price)
 
     context.incr += 1""",
@@ -602,7 +678,7 @@ def handle_data(context, data):
     if context.incr < context.count:
         # order small lots to be sure the
         # order will fill in a single transaction
-        order(0, 5000)
+        order(sid(0), 5000)
     record(price=data[0].price)
     record(volume=data[0].volume)
     record(incr=context.incr)
@@ -871,6 +947,7 @@ class TestGetDatetime(TestCase):
         algo = TradingAlgorithm(
             script=algo,
             sim_params=sim_params,
+            identifiers=[1]
         )
         algo.run(source)
         self.assertFalse(algo.first_bar)
@@ -917,7 +994,7 @@ class TestTradingControls(TestCase):
 
         # Buy one share four times.  Should be fine.
         def handle_data(algo, data):
-            algo.order(self.sid, 1)
+            algo.order(algo.sid(self.sid), 1)
             algo.order_count += 1
         algo = SetMaxPositionSizeAlgorithm(sid=self.sid,
                                            max_shares=10,
@@ -927,7 +1004,7 @@ class TestTradingControls(TestCase):
         # Buy three shares four times.  Should bail on the fourth before it's
         # placed.
         def handle_data(algo, data):
-            algo.order(self.sid, 3)
+            algo.order(algo.sid(self.sid), 3)
             algo.order_count += 1
 
         algo = SetMaxPositionSizeAlgorithm(sid=self.sid,
@@ -938,7 +1015,7 @@ class TestTradingControls(TestCase):
         # Buy two shares four times. Should bail due to max_notional on the
         # third attempt.
         def handle_data(algo, data):
-            algo.order(self.sid, 3)
+            algo.order(algo.sid(self.sid), 3)
             algo.order_count += 1
 
         algo = SetMaxPositionSizeAlgorithm(sid=self.sid,
@@ -949,7 +1026,7 @@ class TestTradingControls(TestCase):
         # Set the trading control to a different sid, then BUY ALL THE THINGS!.
         # Should continue normally.
         def handle_data(algo, data):
-            algo.order(self.sid, 10000)
+            algo.order(algo.sid(self.sid), 10000)
             algo.order_count += 1
         algo = SetMaxPositionSizeAlgorithm(sid=self.sid + 1,
                                            max_shares=10,
@@ -959,7 +1036,7 @@ class TestTradingControls(TestCase):
         # Set the trading control sid to None, then BUY ALL THE THINGS!. Should
         # fail because setting sid to None makes the control apply to all sids.
         def handle_data(algo, data):
-            algo.order(self.sid, 10000)
+            algo.order(algo.sid(self.sid), 10000)
             algo.order_count += 1
         algo = SetMaxPositionSizeAlgorithm(max_shares=10, max_notional=61.0)
         self.check_algo_fails(algo, handle_data, 0)
@@ -971,7 +1048,7 @@ class TestTradingControls(TestCase):
             restricted_list=[self.sid])
 
         def handle_data(algo, data):
-            algo.order(self.sid, 100)
+            algo.order(algo.sid(self.sid), 100)
             algo.order_count += 1
 
         self.check_algo_fails(algo, handle_data, 0)
@@ -982,7 +1059,7 @@ class TestTradingControls(TestCase):
             restricted_list=[134, 135, 136])
 
         def handle_data(algo, data):
-            algo.order(self.sid, 100)
+            algo.order(algo.sid(self.sid), 100)
             algo.order_count += 1
 
         self.check_algo_succeeds(algo, handle_data)
@@ -991,7 +1068,7 @@ class TestTradingControls(TestCase):
 
         # Buy one share.
         def handle_data(algo, data):
-            algo.order(self.sid, 1)
+            algo.order(algo.sid(self.sid), 1)
             algo.order_count += 1
         algo = SetMaxOrderSizeAlgorithm(sid=self.sid,
                                         max_shares=10,
@@ -1001,7 +1078,7 @@ class TestTradingControls(TestCase):
         # Buy 1, then 2, then 3, then 4 shares.  Bail on the last attempt
         # because we exceed shares.
         def handle_data(algo, data):
-            algo.order(self.sid, algo.order_count + 1)
+            algo.order(algo.sid(self.sid), algo.order_count + 1)
             algo.order_count += 1
 
         algo = SetMaxOrderSizeAlgorithm(sid=self.sid,
@@ -1012,7 +1089,7 @@ class TestTradingControls(TestCase):
         # Buy 1, then 2, then 3, then 4 shares.  Bail on the last attempt
         # because we exceed notional.
         def handle_data(algo, data):
-            algo.order(self.sid, algo.order_count + 1)
+            algo.order(algo.sid(self.sid), algo.order_count + 1)
             algo.order_count += 1
 
         algo = SetMaxOrderSizeAlgorithm(sid=self.sid,
@@ -1023,7 +1100,7 @@ class TestTradingControls(TestCase):
         # Set the trading control to a different sid, then BUY ALL THE THINGS!.
         # Should continue normally.
         def handle_data(algo, data):
-            algo.order(self.sid, 10000)
+            algo.order(algo.sid(self.sid), 10000)
             algo.order_count += 1
         algo = SetMaxOrderSizeAlgorithm(sid=self.sid + 1,
                                         max_shares=1,
@@ -1034,7 +1111,7 @@ class TestTradingControls(TestCase):
         # Should fail because not specifying a sid makes the trading control
         # apply to all sids.
         def handle_data(algo, data):
-            algo.order(self.sid, 10000)
+            algo.order(algo.sid(self.sid), 10000)
             algo.order_count += 1
         algo = SetMaxOrderSizeAlgorithm(max_shares=1,
                                         max_notional=1.0)
@@ -1055,7 +1132,7 @@ class TestTradingControls(TestCase):
 
         def handle_data(algo, data):
             for i in range(5):
-                algo.order(self.sid, 1)
+                algo.order(algo.sid(self.sid), 1)
                 algo.order_count += 1
 
         algo = SetMaxOrderCountAlgorithm(3)
@@ -1074,7 +1151,7 @@ class TestTradingControls(TestCase):
     def test_long_only(self):
         # Sell immediately -> fail immediately.
         def handle_data(algo, data):
-            algo.order(self.sid, -1)
+            algo.order(algo.sid(self.sid), -1)
             algo.order_count += 1
         algo = SetLongOnlyAlgorithm()
         self.check_algo_fails(algo, handle_data, 0)
@@ -1083,9 +1160,9 @@ class TestTradingControls(TestCase):
         # should succeed.
         def handle_data(algo, data):
             if (algo.order_count % 2) == 0:
-                algo.order(self.sid, 1)
+                algo.order(algo.sid(self.sid), 1)
             else:
-                algo.order(self.sid, -1)
+                algo.order(algo.sid(self.sid), -1)
             algo.order_count += 1
         algo = SetLongOnlyAlgorithm()
         self.check_algo_succeeds(algo, handle_data)
@@ -1093,7 +1170,7 @@ class TestTradingControls(TestCase):
         # Buy on first three days, then sell off holdings.  Should succeed.
         def handle_data(algo, data):
             amounts = [1, 1, 1, -3]
-            algo.order(self.sid, amounts[algo.order_count])
+            algo.order(algo.sid(self.sid), amounts[algo.order_count])
             algo.order_count += 1
         algo = SetLongOnlyAlgorithm()
         self.check_algo_succeeds(algo, handle_data)
@@ -1102,7 +1179,7 @@ class TestTradingControls(TestCase):
         # Should fail on the last sale.
         def handle_data(algo, data):
             amounts = [1, 1, 1, -4]
-            algo.order(self.sid, amounts[algo.order_count])
+            algo.order(algo.sid(self.sid), amounts[algo.order_count])
             algo.order_count += 1
         algo = SetLongOnlyAlgorithm()
         self.check_algo_fails(algo, handle_data, 3)
@@ -1128,14 +1205,42 @@ class TestTradingControls(TestCase):
         algo.run(self.source)
         self.source.rewind()
 
+    def test_asset_date_bounds(self):
+
+        # Run the algorithm with a sid that ends far in the future
+        df_source, _ = factory.create_test_df_source(self.sim_params)
+        metadata = {0: {'start_date': '1990-01-01',
+                        'end_date': '2020-01-01'}}
+        algo = SetAssetDateBoundsAlgorithm(asset_metadata=metadata,
+                                           sim_params=self.sim_params,)
+        algo.run(df_source)
+
+        # Run the algorithm with a sid that has already ended
+        df_source, _ = factory.create_test_df_source(self.sim_params)
+        metadata = {0: {'start_date': '1989-01-01',
+                        'end_date': '1990-01-01'}}
+        algo = SetAssetDateBoundsAlgorithm(asset_metadata=metadata,
+                                           sim_params=self.sim_params,)
+        with self.assertRaises(TradingControlViolation):
+            algo.run(df_source)
+
+        # Run the algorithm with a sid that has not started
+        df_source, _ = factory.create_test_df_source(self.sim_params)
+        metadata = {0: {'start_date': '2020-01-01',
+                        'end_date': '2021-01-01'}}
+        algo = SetAssetDateBoundsAlgorithm(asset_metadata=metadata,
+                                           sim_params=self.sim_params,)
+        with self.assertRaises(TradingControlViolation):
+            algo.run(df_source)
+
 
 class TestAccountControls(TestCase):
 
     def setUp(self):
         self.sim_params = factory.create_simulation_parameters(num_days=4)
-        self.sid = 133
+        self.sidint = 133
         self.trade_history = factory.create_trade_history(
-            self.sid,
+            self.sidint,
             [10.0, 10.0, 11.0, 11.0],
             [100, 100, 100, 300],
             timedelta(days=1),
@@ -1167,14 +1272,14 @@ class TestAccountControls(TestCase):
 
         # Set max leverage to 0 so buying one share fails.
         def handle_data(algo, data):
-            algo.order(self.sid, 1)
+            algo.order(algo.sid(self.sidint), 1)
 
         algo = SetMaxLeverageAlgorithm(0)
         self.check_algo_fails(algo, handle_data)
 
         # Set max leverage to 1 so buying one share passes
         def handle_data(algo, data):
-            algo.order(self.sid, 1)
+            algo.order(algo.sid(self.sidint), 1)
 
         algo = SetMaxLeverageAlgorithm(1)
         self.check_algo_succeeds(algo, handle_data)
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -20,7 +20,177 @@ Tests for the zipline.assets package
 import sys
 from unittest import TestCase
 
-from zipline.assets._assets import Asset, Future
+from datetime import (
+    datetime,
+    timedelta,
+)
+import pickle
+import pprint
+import pytz
+import uuid
+
+import pandas as pd
+
+from nose_parameterized import parameterized
+
+from zipline.finance.trading import with_environment
+from zipline.assets import Asset, Future, AssetFinder
+from zipline.errors import (
+    SymbolNotFound,
+    MultipleSymbolsFound,
+)
+
+
+class FakeTable(object):
+    def __init__(self, name, count, dt, fuzzy_str):
+        self.name = name
+        self.count = count
+        self.dt = dt
+        self.fuzzy_str = fuzzy_str
+        self.df = pd.DataFrame.from_records(
+            [
+                {
+                    'sid': i,
+                    'file_name':  'TEST%s%s' % (self.fuzzy_str, i),
+                    'company_name': self.name + str(i),
+                    'start_date_nano': pd.Timestamp(dt, tz='UTC').value,
+                    'end_date_nano': pd.Timestamp(dt, tz='UTC').value,
+                    'exchange': self.name,
+                }
+                for i in range(1, self.count + 1)
+            ]
+        )
+
+    def read(self, *args, **kwargs):
+        return self.df.to_records()
+
+
+class FakeTableIdenticalSymbols(object):
+    def __init__(self, name, as_of_dates):
+        self.name = name
+        self.as_of_dates = as_of_dates
+        self.df = pd.DataFrame.from_records(
+            [
+                {
+                    'sid': i,
+                    'file_name':  self.name,
+                    'company_name': self.name,
+                    'start_date_nano': date.value,
+                    'end_date_nano': (date + timedelta(days=1)).value,
+                    'exchange': self.name,
+                }
+                for i, date in enumerate(self.as_of_dates)
+            ]
+        )
+
+    def read(self, *args, **kwargs):
+        return self.df.to_records()
+
+
+class FakeTableFromRecords(object):
+
+    def __init__(self, records):
+        self.records = records
+        self.df = pd.DataFrame.from_records(self.records)
+
+    def read(self, *args, **kwargs):
+        return self.df.to_records()
+
+
+@with_environment()
+def build_lookup_generic_cases(env=None):
+    """
+    Generate test cases for AssetFinder test_lookup_generic.
+    """
+
+    unique_start = pd.Timestamp('2013-01-01', tz='UTC')
+    unique_end = pd.Timestamp('2014-01-01', tz='UTC')
+
+    dupe_0_start = pd.Timestamp('2013-01-01', tz='UTC')
+    dupe_0_end = dupe_0_start + timedelta(days=1)
+
+    dupe_1_start = pd.Timestamp('2013-01-03', tz='UTC')
+    dupe_1_end = dupe_1_start + timedelta(days=1)
+
+    table = FakeTableFromRecords(
+        [
+            {
+                'sid': 0,
+                'file_name':  'duplicated',
+                'company_name': 'duplicated_0',
+                'start_date_nano': dupe_0_start.value,
+                'end_date_nano': dupe_0_end.value,
+                'exchange': '',
+            },
+            {
+                'sid': 1,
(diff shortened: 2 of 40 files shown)


zipline/algorithm.py
```diff
@@ -247,10 +259,14 @@ def __init__(self, *args, **kwargs):
 
         self._most_recent_data = None
 
+        self.initialized = False
+
+        # Register any default TradingControl objects
+        self.register_trading_control(ExpiredSID())
```
jfkirk: @ssanderson Does it make sense to register this trading control in TradingAlgorithm.__init__?

zipline/assets/_assets.pyx
```diff
@@ -21,10 +21,6 @@ cimport cython
 import numpy as np
 cimport numpy as np
 
-cdef enum AssetType:
-    EQUITY = 1
-    FUTURE = 2
-
```
jfkirk: @llllllllll I removed the enum and converted everything to isinstance() checks, as we discussed.

zipline/assets/assets.py
```diff
@@ -0,0 +1,508 @@
...
+            self.spawn_asset(identifier=identifier, **row)
+            counter += 1
+
+        self.shared_caches.update(
+            {'by_sid': self.cache,
+             'by_symbol': self.sym_cache,
+             'fuzzy_match': {}}
+        )
+
+    def spawn_asset(self, identifier, **kwargs):
```
jfkirk: @ssanderson Can you review this method?

This is where we interpret the AssetMetaData fields and, after interpreting them, use them as the kwargs to construct the Asset.

ssanderson:
@jfkirk I'll try to look at this tonight

zipline/sources/data_frame_source.py
```diff
@@ -108,11 +119,21 @@ class DataPanelSource(DataSource):
...
+        # Remap sids based on the trading environment
+        self.identifiers = kwargs.get('sids', self.data.items)
+        trading.environment.update_asset_finder(identifiers=self.identifiers)
+        self.data.items = [
+            trading.environment.asset_finder.lookup_generic(
+                identifier,
+                as_of_date=self.end)[0].sid
+            for identifier in self.data.items
+        ]
+        self.sids = self.data.items
```
jfkirk: @ssanderson @ehebert @twiecki By default, I have it that the various Source classes (DataFrameSource, PanelSource, RandomWalkSource, SpecificEquityTrades) map all of their identifiers to sids during their __init__, and the asset_finder is told to build assets for all identifiers in the DataFrameSource.

It is not necessary for a source to do any mapping if the source already has sids on its events. These sids would need to be fed in to TradingAlgorithm.__init__, though, so the Assets can be constructed.

Is this sloppy, or does this make sense?

tests/test_algorithm.py
```diff
@@ -285,6 +290,58 @@ def nop(*args, **kwargs):
...
+        with self.assertRaises(SymbolNotFound):
+            algo.symbol('PLAY')
+        with self.assertRaises(SymbolNotFound):
+            algo.symbols('PLAY', 'OMG15')
+
+        # Test when first PLAY exists
+        algo.datetime = pd.Timestamp('2002-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        self.assertEqual(0, algo.symbol('PLAY'))
+        list_result = algo.symbols('PLAY', 'OMG15')
```
ssanderson: `self.assertEqual(algo.symbols(['PLAY', 'OMG15']), [0, 2])` might be slightly clearer here.

tests/test_algorithm.py
```diff
@@ -285,6 +290,58 @@ def nop(*args, **kwargs):
...
+        # Test after second PLAY begins
+        algo.datetime = pd.Timestamp('2005-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        self.assertEqual(1, algo.symbol('PLAY'))
+
+        # Test after second PLAY ends
+        algo.datetime = pd.Timestamp('2006-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        self.assertEqual(1, algo.symbol('PLAY'))
+        list_result = algo.symbols('PLAY', 'OMG15')
```
ssanderson: Same note as above.

tests/test_algorithm.py
```diff
@@ -285,6 +290,58 @@ def nop(*args, **kwargs):
...
+        algo.datetime = pd.Timestamp('2001-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
+        with self.assertRaises(SymbolNotFound):
+            algo.symbol('PLAY')
+        with self.assertRaises(SymbolNotFound):
+            algo.symbols('PLAY', 'OMG15')
+
+        # Test when first PLAY exists
+        algo.datetime = pd.Timestamp('2002-12-01', tz='UTC')
+        self.assertEqual(2, algo.symbol('OMG15'))
```
ssanderson: The actual expected object here is an `Asset` right, not an int?

tests/test_algorithm.py
```diff
@@ -387,6 +444,22 @@ def test_order_methods(self):
             )
             algo.run(self.df)
 
+    def test_order_methods_for_future(self):
```
ssanderson: Can we do more than just assert that this doesn't crash?  What do we actually expect to happen here?

tests/test_algorithm.py
```diff
@@ -387,6 +444,22 @@ def test_order_methods(self):
...
+        AlgoClasses = [TestOrderAlgorithm,
+                       TestOrderValueAlgorithm,
+                       TestTargetAlgorithm,
+                       TestOrderPercentAlgorithm,
+                       TestTargetValueAlgorithm]
```
ssanderson: This might be a good use-case for `nose.parameterized`.  Search around for `parameterized.expand` to see how it's used.

tests/test_algorithm.py
```diff
@@ -1156,14 +1249,14 @@ def test_set_max_leverage(self):
 
         # Set max leverage to 0 so buying one share fails.
         def handle_data(algo, data):
-            algo.order(self.sid, 1)
+            algo.order(algo.sid(self.sidint), 1)
```
ssanderson: Any reason this changed to sidint when most other stuff is still sid?

tests/test_assets.py
```diff
@@ -20,7 +20,179 @@
...
+
+from zipline.finance import trading
+from zipline.assets import Asset, Future, AssetFinder, AssetMetaData
+from zipline.errors import (
+    SymbolNotFound,
+    MultipleSymbolsFound,
+)
+
+trading.environment = trading.TradingEnvironment()
+env = trading.environment
```
ssanderson: Why are we constructing a new TradingEnvironment here at module scope?

tests/test_perf_tracking.py
```diff
@@ -1982,15 +1992,112 @@ def test_empty_positions(self):
             self.assertEquals(val, 0)
             self.assertNotIsInstance(val, (bool, np.bool_))
 
-    def test_serializaition(self):
+    def test_update_last_sale(self):
+        metadata = {1: {'asset_type': 'equity'},
+                    2: {'asset_type': 'future',
+                        'contract_multiplier': 1000}}
+        if trading.environment is None:
```
ssanderson: You can just call `TradingEnvironment.instance()` to get (or create) the current TradingEnvironment.

tests/test_perf_tracking.py
```diff
@@ -1982,15 +1992,112 @@ def test_empty_positions(self):
...
+        # Test gross and net exposures
+        self.assertEqual(100 + 200 + 300000 + 400000, pt._gross_exposure())
+        self.assertEqual(100 - 200 + 300000 - 400000, pt._net_exposure())
+
+
+    def test_generate_end_sid_transaction(self):
+        metadata = {1: {'asset_type': 'equity'},
+                    2: {'asset_type': 'future',
+                        'contract_multiplier': 1000}}
+        if trading.environment is None:
```
ssanderson: Same comment as above.

This isn't new to your code, but the way we interact with `TradingEnvironment` is all over the place right now.  What we probably should do IMO is remove the module-scope `environment` global and replace it with a class-level attribute, which is accessed only via `TradingEnvironment.instance()`.

zipline/finance/trading.py
```diff
@@ -149,6 +153,49 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         # stack.
         return False
 
+    def update_asset_finder(self,
```
ssanderson: When would you need to update the asset finder in this manner rather than loading all the metadata in one shot?  If we can avoid mutating shared global state, we should do so.

zipline/sources/data_frame_source.py
```diff
@@ -108,11 +119,21 @@ class DataPanelSource(DataSource):
     def __init__(self, data, **kwargs):
         assert isinstance(data.major_axis, pd.tseries.index.DatetimeIndex)
 
-        self.data = data
+        self.data = data.fillna(method='ffill', axis=0)
```
ssanderson: Why are we `'fillna'`ing everything here?  We probably don't want to be ffilling anything except price.  Also from testing in the shell I don't think passing an axis to ffill does anything for a panel.

zipline/assets/assets.py
```diff
@@ -0,0 +1,540 @@
...
+    def __init__(self,
+                 metadata,
+                 trading_calendar=tradingcalendar,
+                 force_populate=False):
+
+        # Any particular instance of AssetFinder should be
+        # consistent throughout its lifetime, so we grab a reference
+        # to our cache now. That way, if the cache is refreshed later,
+        # our instance will continue to use the old one.
+        self.cache = self.shared_caches['by_sid']
```
ssanderson: This may not be worth changing now, but I don't think we should be caching these values on the class.  If we're only ever going to instantiate one of these per simulation, I'd rather just treat this as a singleton the way we do with TradingEnvironment.

zipline/assets/_assets.pyx
```diff
@@ -209,11 +203,19 @@ cdef class Equity(Asset):
         params = ', '.join(strings)
         return 'Equity(%d, %s)' % (self.sid, params)
 
+    @staticmethod
+    def from_dict(dict_):
```
ssanderson: If we make this a class method, then we can implement it as:

```
@classmethod
def from_dict(cls, dict_):
    return cls(**dict_)
```
and we won't have to re-implement it in each subclass.

zipline/assets/_assets.pyx
```diff
@@ -224,18 +226,70 @@ cdef class Future(Asset):
...
+        """
+        Build a Future instance from a dict.
+        """
+        return Future(**dict_)
+
+
+def make_asset_array(int size, Asset asset):
+    cdef np.ndarray out = np.empty([size], dtype=object)
+    out.fill(asset)
+    return out
```
ssanderson: Missing a newline at the end of the file here.

zipline/assets/_assets.pyx
```diff
@@ -21,9 +21,10 @@ cimport cython
 import numpy as np
 cimport numpy as np
 
-cdef enum AssetType:
-    EQUITY = 1
-    FUTURE = 2
+# IMPORTANT NOTE: You must change this template if you change
+# Asset.__reduce__, or else we'll attempt to unpickle an old version of this
+# class
+CACHE_FILE_TEMPLATE = '/tmp/.%s-%s.v4.cache'
```
ssanderson: This probably belongs in `~/.zipline` now rather than `/tmp`.

zipline/finance/controls.py
```diff
@@ -269,6 +269,29 @@ def validate(self,
...
+                 algo_datetime,
+                 algo_current_data):
+        """
+        Fail if the algo has passed this Asset's end_date.
+        """
+        # No failure if the Asset has no end_date
+        if not sid.end_date:
+            return
+        # Fail if the algo has passed this Asset's end_date
+        if algo_datetime >= sid.end_date:
```
ssanderson: Should this check for ordering before a sid's start_date as well?  In which case this probably gets renamed to "NonexistentSid" or somesuch.

zipline/finance/controls.py
```diff
@@ -269,6 +269,29 @@ def validate(self,
...
+                 sid,
+                 amount,
+                 portfolio,
+                 algo_datetime,
+                 algo_current_data):
+        """
+        Fail if the algo has passed this Asset's end_date.
+        """
+        # No failure if the Asset has no end_date
+        if not sid.end_date:
```
ssanderson: What are the possible empty values here?  Should we be checking explicitly for `None` rather than non-truthy?

Also, this might be slightly cleaner written as:
```
if sid.end_date and algo_datetime >= sid.end_date:
    self.fail(sid, amount)
```
```

File: zipline/finance/performance/period.py [hunk 22]
```diff
@@ -401,6 +435,8 @@ def as_account(self):
                     self.ending_cash + self.ending_value)
         account.total_positions_value = \
             getattr(self, 'total_positions_value', self.ending_value)
+        account.total_positions_value = \
```
ssanderson: @brianpfink why are these all `getattr` calls instead of just `self.attribute`.?

zipline/finance/performance/position_tracker.py
```diff
@@ -104,33 +152,71 @@ def position_values(self):
         self._position_last_sale_prices is changed.
         """
         if self._position_values is None:
-            vals = list(map(mul, self._position_amounts.values(),
```
ssanderson: We're constructing and destroying a lot of potentially long lists here.

I would write this whole block as something like:
```
from six import itervalues
from itertools import izip
iter_amount_price_multiplier = izip(
    itervalues(self._position_amounts),
    itervalues(self._position_last_sale_prices),
    itervalues(self._position_value_multipliers),
)
self._position_values = [
    price * amount * multiplier for
    price, amount, multiplier in iter_amount_price_multiplier
]
```
cc @ehebert do we want to do this pre-emptively as part of your memory profiling work?

zipline/finance/performance/position_tracker.py
```diff
@@ -104,33 +152,71 @@ def position_values(self):
...
 
+    def calculate_positions_exposure(self):
+        if len(self.position_exposures) == 0:
+            return np.float64(0)
+
+        return sum(self.position_exposures)
+
     def _longs_count(self):
-        return sum(map(lambda x: x > 0, self.position_values))
+        return sum(map(lambda x: x > 0, self.position_exposures))
```
ssanderson: Mapping simple lambdas is expensive for large arrays because you incur a python function call overhead every time.  You can write this as:

```
sum(i for i in self.position_exposures if i > 0)
```

ssanderson: On a length 100000 array:
```
In [85]: %timeit sum(map(lambda n: n > 50000, x))
100 loops, best of 3: 7.5 ms per loop

In [86]: %timeit sum(1 for i in x if i > 50000)
100 loops, best of 3: 2.59 ms per loop
```

zipline/finance/performance/position_tracker.py
```diff
@@ -104,33 +152,71 @@ def position_values(self):
...
 
     def _long_exposure(self):
+        return sum(filter(lambda x: x > 0, self.position_exposures))
```
ssanderson: All the methods here would be better written as sum(<generator expression>) rather than `sum(map(lambda`

zipline/finance/performance/position_tracker.py
```diff
@@ -203,20 +290,36 @@ def pay_dividends(self, dividend_frame):
...
         # Add cash equal to the net cash payed from all dividends.  Note that
-        # "negative cash" is effectively paid if we're short a security,
+        # "negative cash" is effectively paid if we're short an asset,
         # representing the fact that we're required to reimburse the owner of
         # the stock for any dividends paid while borrowing.
         net_cash_payment = payments['cash_amount'].fillna(0).sum()
         return net_cash_payment
 
+    def _generate_end_sid_transaction(self, sid, dt):
+        if (sid not in self._position_amounts) or \
```
ssanderson: This is equivalent to
```
if not self._position_amounts.get(sid):
    return None
```

zipline/finance/performance/position_tracker.py
```diff
@@ -203,20 +290,36 @@ def pay_dividends(self, dividend_frame):
...
 
         # Add cash equal to the net cash payed from all dividends.  Note that
-        # "negative cash" is effectively paid if we're short a security,
+        # "negative cash" is effectively paid if we're short an asset,
         # representing the fact that we're required to reimburse the owner of
         # the stock for any dividends paid while borrowing.
         net_cash_payment = payments['cash_amount'].fillna(0).sum()
         return net_cash_payment
 
+    def _generate_end_sid_transaction(self, sid, dt):
```
ssanderson: Where is this called from?

Can we be doing this check elsewhere rather than inside the method (e.g. `if tracker.has_position(sid): tracker._generate_end_sid_transaction)`)?  

If not, can we rename this to something like `_maybe_create_end_sid_transaction` so it's clearer that it might return `None`

zipline/finance/performance/position_tracker.py
```diff
@@ -276,5 +379,9 @@ def __setstate__(self, state):
         # Arrays for quick calculations of positions value
         self._position_amounts = OrderedDict()
         self._position_last_sale_prices = OrderedDict()
+        self._position_value_multiplier = OrderedDict()
```
ssanderson: Seems like these should probably be plural to match the other attributes here?

zipline/finance/performance/tracker.py
```diff
@@ -84,9 +83,7 @@ class PerformanceTracker(object):
     """
     Tracks the performance of the algorithm.
     """
-
-    @with_environment()
-    def __init__(self, sim_params, env=None):
+    def __init__(self, sim_params):
```
ssanderson: Why doesn't this use `with_environment` anymore?  Was it just unused?

I notice we're pulling the environment directly off the `trading` module, which is an idiom I'd like to purge.  I'd rather use the `with_environment` decorator to supply it as an argument and use that argument instead of `trading.environment` attribute accesses.

zipline/finance/performance/tracker.py
```diff
@@ -209,6 +210,21 @@ def set_date(self, date):
             self.saved_dt = date
             self.todays_performance.period_close = self.saved_dt
 
+    def _update_asset(self, sid):
+        if sid in self._known_asset_sids:
+            return
+        self._known_asset_sids.add(sid)
+        # Collect the end date of the asset
+        asset = trading.environment.asset_finder.retrieve_asset(sid)
```
ssanderson: More usage of `trading.environment` that I'd like to get rid of if we can.

zipline/finance/performance/tracker.py
```diff
@@ -209,6 +210,21 @@ def set_date(self, date):
...
+        if asset.end_date:
```
ssanderson: See comment above w/r/t possible null values.

zipline/finance/performance/tracker.py
```diff
@@ -191,6 +188,10 @@ def __init__(self, sim_params, env=None):
         self.account_needs_update = True
         self._account = None
 
+        self._known_asset_sids = set()
```
ssanderson: What are these used for?  It seems like there's a lot of state tracking here that's being duplicated in the `AssetFinder`

zipline/finance/performance/tracker.py
```diff
@@ -345,7 +365,24 @@ def process_event(self, event):
 
             self.all_benchmark_returns[midnight] = event.returns
 
-    def check_upcoming_dividends(self, midnight_of_date_that_just_ended):
+    def _next_trading_day(self, midnight_of_date_that_just_ended):
```
ssanderson: This method (along with a whole host of date manipulation utilities) already exists on `TradingEnvironment`.

ssanderson: ...and it looks like I wrote this method originally. :crying_cat_face:

zipline/finance/performance/tracker.py
```diff
@@ -396,6 +422,27 @@ def check_upcoming_dividends(self, midnight_of_date_that_just_ended):
             # notify periods to update their stats
             period.handle_dividends_paid(net_cash_payment)
 
+    def check_upcoming_asset_ends(self, next_trading_day):
```
ssanderson: This method name doesn't make it obvious that this had an important and nontrivial side-effect.  I would name this something like `close_expired_asset_positions`.

ssanderson: This method name doesn't make it obvious that this had an important and nontrivial side-effect.  I would name this something like `close_expired_asset_positions`.

zipline/sources/data_frame_source.py
ssanderson: @jfkirk I don't fully understand the question being asked here.  Can you come grab me at some point this afternoon and we'll talk about this?

zipline/sources/test_source.py
```diff
@@ -153,10 +150,21 @@ def __init__(self, *args, **kwargs):
                 self.event_list[1].dt - self.event_list[0].dt)
             self.concurrent = kwargs.get('concurrent', False)
 
+            self.identifiers = kwargs.get(
+                'sids',
+                np.unique([event.sid for event in self.event_list]).tolist())
```
ssanderson: If this just needs to be an iterable, this can be written as `set(event.sid for event in self.event_list)`.

zipline/sources/test_source.py
```diff
@@ -153,10 +150,21 @@ def __init__(self, *args, **kwargs):
...
+            env.update_asset_finder(erase_existing=True,
+                                    identifiers=self.identifiers)
+            assets, _ = env.asset_finder.lookup_generic(self.identifiers,
+                                                        as_of_date=self.start)
+            self.sids = [asset.sid for asset in assets]
+            for event in self.event_list:
+                event.sid = env.asset_finder.lookup_generic(event.sid,
```
ssanderson: I don't think we should be using `lookup_generic` for any internal code.  It's meant to be used for convenient user-facing functions, not internal code where we know the types of our inputs.

zipline/test_algorithms.py
```diff
@@ -368,10 +377,17 @@ def handle_data(self, data):
...
-        self.target_shares += np.floor((.001 *
-                                        self.portfolio.portfolio_value) /
-                                       data[0].price)
+        self.order_percent(self.sid(0), .001)
+
+        if isinstance(self.sid(0), Equity):
+            self.target_shares += np.floor(
+                (.001 * self.portfolio.portfolio_value) / data[0].price
+            )
+        if isinstance(self.sid(0), Future):
```
ssanderson: Why does the logic need to be different here for Equities vs Futures?  Would users have to write analogous code in their algorithms?

zipline/test_algorithms.py
```diff
@@ -703,8 +725,8 @@ def handle_data(self, data):
                 self.return_nan.handle_data(data))
         else:
             nan_data = deepcopy(data)
-            for sid in nan_data.iterkeys():
-                nan_data[sid].price = np.nan
+            for sidint in nan_data.iterkeys():
```
ssanderson: Is this just setting a column on a DataFrame to np.nan?  If so, it should be written as `nan_data.price = np.nan`.

ssanderson:
```
In [1]: df
Out[1]:
           a         b   c
0  -1.955521  1.771930   5
1   0.134790  0.096965   5
2   0.889008 -0.061098   5
3   0.698142 -0.849381   5
4   0.120767 -0.691841   5
5  -0.150403 -0.863975  10
6  -0.693147  1.743716  10
7   1.856462  1.393960  10
8  -0.267299 -0.628845  10
9  -0.458319 -0.682851  10
10  1.091492  0.758339  15
11 -0.299518 -0.945327  15
12  1.484762 -0.886816  15
13  1.932633  0.668340  15
14 -0.501742  0.310666  15
15  0.497662  1.102211  15
16  0.188138 -0.642357  15
17 -1.375983 -0.456558  15
18  1.397060 -1.040300  15
19  0.088938  0.964899  15

In [2]: df.b = np.nan

In [3]: df
Out[3]:
           a   b   c
0  -1.955521 NaN   5
1   0.134790 NaN   5
2   0.889008 NaN   5
3   0.698142 NaN   5
4   0.120767 NaN   5
5  -0.150403 NaN  10
6  -0.693147 NaN  10
7   1.856462 NaN  10
8  -0.267299 NaN  10
9  -0.458319 NaN  10
10  1.091492 NaN  15
11 -0.299518 NaN  15
12  1.484762 NaN  15
13  1.932633 NaN  15
14 -0.501742 NaN  15
15  0.497662 NaN  15
16  0.188138 NaN  15
17 -1.375983 NaN  15
18  1.397060 NaN  15
19  0.088938 NaN  15
```

zipline/utils/cli.py
```diff
@@ -99,9 +100,12 @@ def parse_args(argv, ipython_mode=False):
     parser.add_argument('--start', '-s')
     parser.add_argument('--end', '-e')
     parser.add_argument('--capital_base')
-    parser.add_argument('--source', choices=('yahoo',))
+    parser.add_argument('--source', '-d')
```
ssanderson: Why is `choices` no longer specified here?

Also, most of these single-letter entries are non-obvious enough that I suspect we'd be better off not having them at all.

zipline/utils/cli.py
```diff
@@ -188,9 +225,13 @@ def run_pipeline(print_algo=True, **kwargs):
...
                                     capital_base=float(kwargs['capital_base']),
-                                    algo_filename=kwargs.get('algofile'))
+                                    algo_filename=kwargs.get('algofile'),
+                                    asset_metadata=asset_metadata,
+                                    identifiers=symbols,
+                                    start=start,
+                                    end=end)
 
-    perf = algo.run(source)
+    perf = algo.run(source, overwrite_sim_params=False)
```
ssanderson: Is this just explicitly setting the default, or are we changing behavior here?

zipline/utils/cli.py
```diff
@@ -157,10 +161,43 @@ def run_pipeline(print_algo=True, **kwargs):
...
 
     symbols = kwargs['symbols'].split(',')
-
-    if kwargs['source'] == 'yahoo':
+    asset_identifier = kwargs['metadata_index']
+    if asset_identifier is None:
+        asset_identifier = 'symbol'
+
+    # Pull asset metadata
+    asset_metadata = kwargs.get('asset_metadata', None)
```
ssanderson: Why is one of these optional and the other not?

zipline/utils/security_list.py
```diff
@@ -74,13 +69,17 @@ def restricted_list(self):
...
+    @with_environment()
+    def update_current(self, effective_date, symbols, change_func, env=None):
         for symbol in symbols:
-            sid = self.lookup_func(
-                symbol,
-                as_of_date=effective_date
-            )
-            change_func(sid)
+            try:
+                sid = env.asset_finder.lookup_generic(
```
ssanderson: Can we avoid using `lookup_generic` here?

zipline/sources/test_source.py
```diff
@@ -153,10 +150,21 @@ def __init__(self, *args, **kwargs):
                 self.event_list[1].dt - self.event_list[0].dt)
             self.concurrent = kwargs.get('concurrent', False)
 
+            self.identifiers = kwargs.get(
+                'sids',
+                np.unique([event.sid for event in self.event_list]).tolist())
+            env.update_asset_finder(erase_existing=True,
```
ssanderson: I'm not excited about constantly wiping the asset finder like this.  I'd rather just construct a new one and set it.

zipline/protocol.py
```diff
@@ -57,7 +57,12 @@
     'sid',
 ]
 # Expected fields/index values for a dividend payment Series.
-DIVIDEND_PAYMENT_FIELDS = ['id', 'payment_sid', 'cash_amount', 'share_count']
+DIVIDEND_PAYMENT_FIELDS = [
+    'id',
+    'payment_sid',
+    'cash_amount',
+    'share_count'
```
ssanderson: Probably want a trailing comma here.  I've been bitten many times by adding a list of strings like this, and then adding a new entry without adding another comma, which behaves unfortunately...

```
x = [
    'foo',
    'bar',
    'buzz' # Note the missing comma here.
    'bazz',
    'bozz',
]

In [5]: x
Out[5]: ['foo', 'bar', 'buzzbazz', 'bozz']
```

zipline/finance/performance/position_tracker.py
ssanderson: Some more timings, because I found this interesting:

```
In [18]: %timeit len(filter(lambda x: x > 50000, y))
100 loops, best of 3: 4.88 ms per loop

In [19]: %timeit sum(map(lambda x: x > 50000, y))
100 loops, best of 3: 8.22 ms per loop

In [20]: %timeit sum(x > 50000 for x in y)
100 loops, best of 3: 4.08 ms per loop

In [21]: %timeit sum(1 for x in y if x > 50000)
100 loops, best of 3: 2.47 ms per loop
```

tests/test_algorithm.py
jfkirk: Yep. Should I do an instance check as well?

tests/test_algorithm.py
jfkirk: This check is failing for me because the returned list contains Assets, not just ints. Is there an easy way to convert the list, or should I just leave the original test?

zipline/algorithm.py
```diff
@@ -605,17 +651,32 @@ def record(self, *args, **kwargs):
...
 
     @api_method
     def symbols(self, *args):
         """
         Default symbols lookup for any source that directly maps the
-        symbol to the identifier (e.g. yahoo finance).
+        symbol to the Asset (e.g. yahoo finance).
         """
-        return args
+        result = []
```
ehebert: This definitely works. And sometimes this form makes it easier to step through etc.
But this kind of transform lends itself semantically to being a comprehension, e.g. `return [self.symbol(identifier) for identifier in args]`

zipline/algorithm.py
```diff
@@ -899,15 +973,23 @@ def order_target_value(self, sid, target,
         order. If the position does exist, this is equivalent to placing an
         order for the difference between the target value and the
         current value.
+        If the Asset being ordered is a Future, the 'target value' calculated
+        is actually the target exposure, as Futures have no 'value'.
         """
         last_price = self.trading_client.current_data[sid].price
+
+        value_multiplier = 1
```
ehebert: Looks like this logic is repeated in multiple order functions, maybe factor out a `value_multiplier` function?

zipline/assets/_assets.pyx
ehebert: That aligns with `DataFrame.from_dict` etc. so I'm on board with that.

zipline/assets/assets.py
ehebert: I can be sold on that @ssanderson, that way it should be easier to mock out/replace for testing etc.

I think the class level cache was for historical reasons of an easy way to provide a cache across process forks, which is no longer used.

zipline/assets/assets.py
```diff
@@ -0,0 +1,535 @@
...
+        )
+
+    def spawn_asset(self, identifier, **kwargs):
+
+        # Check if the identifier is an int
+        if isinstance(identifier, int) and ('sid' not in kwargs.keys()):
+            kwargs['sid'] = identifier
+
+        # Make sure that the sid exists in the kwargs
+        if 'sid' not in kwargs.keys():
```
ehebert: Not sure how often `spawn_asset` will be called, but in general it's best to avoid creating the list by calling `keys`.

Either,

```
if 'sid' not in kwargs:
   ...
```

or

```
try kwargs['sid']:
    pass
except KeyError:
    kwargs['sid'] = self._assign_sid(identifier)
```

zipline/assets/assets.py
```diff
@@ -0,0 +1,535 @@
...
+            kwargs['sid'] = self._assign_sid(identifier)
+
+        # If the identifier coming in was a string and there is no defined
+        # symbol yet, set the symbol to the incoming sid
+        if isinstance(identifier, str) and ('symbol' not in kwargs.keys()):
```
ehebert: `basestring` would be a more compatible check. 

Same note above about calling `.keys()`

zipline/assets/assets.py
```diff
@@ -0,0 +1,535 @@
...
+            kwargs['symbol'] = identifier
+
+        # If the file_name is in the kwargs, it may be the symbol
+        if 'file_name' in kwargs.keys():
```
ehebert: To avoid mentioning every line, the note about checking `.keys()` applies to the rest of this function.

zipline/assets/assets.py
```diff
@@ -0,0 +1,535 @@
...
+            kwargs['end_date'] = self.trading_calendar.\
+                canonicalize_datetime(pd.Timestamp(kwargs['end_date']))
+        if 'notice_date' in kwargs.keys():
+            kwargs['notice_date'] = self.trading_calendar.\
+                canonicalize_datetime(pd.Timestamp(kwargs['notice_date']))
+        if 'expiration_date' in kwargs.keys():
+            kwargs['expiration_date'] = self.trading_calendar.\
+                canonicalize_datetime(pd.Timestamp(kwargs['expiration_date']))
+
+        asset_type = kwargs.pop('asset_type', None)
```
ehebert: I would recommend the following to make that we use Equity as the default asset type more clear.

```
try:
    asset_type = kwargs['asset_type'].lower()
except KeyError:
    asset_type = 'equity'

if asset_type == 'equity':
   asset = Equity(**kwargs)
if asset_type == 'future':
   asset = Future(**kwargs)
else:
   raise Invalid...
```

tests/test_assets.py
jfkirk: I'll refactor to use the @with_environment pattern

tests/test_algorithm.py
jfkirk: Line 1260 had a conflict when self.sid and algo.sid mapped to the sid() method, rather than the field sid, so I go rid of the conflict by naming the integer sidint.

zipline/assets/assets.py
jfkirk: Sounds good, I will factor-out class-level storing.

zipline/assets/assets.py
jfkirk: basestring fails in python 3: "name 'basestring' is not defined"
should I stick with str, or use something else?

zipline/assets/assets.py
jfkirk: I use a .pop to remove asset_type from the kwargs to be passed to Asset(**kwargs)
how does this sound:

        # Build an Asset of the appropriate type
        asset_type = kwargs.pop('asset_type', 'equity')
        if asset_type.lower() == 'equity':
            asset = Equity(**kwargs)
        elif asset_type.lower() == 'future':
            asset = Future(**kwargs)
        else:
            raise InvalidAssetType(asset_type=asset_type)

zipline/finance/performance/tracker.py
jfkirk: I cache the sids whose assets have already been pulled in order to avoid pulling the same Asset out of the finder. If the sid is in the set _known_asset_sids, I know that the necessary fields have already been populated.

zipline/finance/controls.py
jfkirk: Makes sense - refactored

zipline/finance/performance/position_tracker.py
ssanderson: @jfkirk ping on these comments because we talked about them today.  This is a one line change that will yield a substantial perf improvement on these calculations.

zipline/finance/trading.py
jfkirk: Makes sense - I'll make it all-or-nothing.

zipline/test_algorithms.py
jfkirk: I only did it as assets vs futures so that the same test algorithm could be used for multiple tests. A user would not have to make this kind of differentiation.

tests/test_algorithm.py
jfkirk: Done- this and the test now use parameterized.

zipline/algorithm.py
jfkirk: Done - I made a new method to be used by value-ordering methods:

_calculate_order_value_amount(self, asset, value):

zipline/sources/test_source.py
jfkirk: thanks - made the change

zipline/sources/test_source.py
jfkirk: Thanks, good catch. I've factored it out everywhere besides TradingAlgorithm.symbol(). I added a new retrieve_asset_by_identifier method to handle general identifier-based lookup.

tests/test_algorithm.py
jfkirk: The algorithms themselves contain assertions about the behavior

zipline/sources/data_frame_source.py
jfkirk: Re-wrote it to fill volume with 0's and ffill everything else. Does that make more sense?

zipline/assets/_assets.pyx
jfkirk: Done - thanks

zipline/finance/performance/position_tracker.py
jfkirk: Roger wilco

zipline/finance/performance/position_tracker.py
jfkirk: Makes sense - making this change

zipline/finance/performance/position_tracker.py
jfkirk: Reworked accordingly

zipline/finance/performance/position_tracker.py
jfkirk: Renamed to _maybe_create_end_sid_transaction.

The method is called from the end-of-day routines in PerformanceTracker, much like how dividends are paid-out.

zipline/finance/performance/position_tracker.py
jfkirk: Renamed - thanks

zipline/finance/performance/tracker.py
jfkirk: It was unused before. I have refactored to use with_environment().

zipline/finance/performance/tracker.py
jfkirk: Refactored to use env utilities

zipline/finance/performance/tracker.py
jfkirk: Thanks - refactored accordingly

zipline/sources/test_source.py
jfkirk: Thanks - reworked accordingly

zipline/sources/test_source.py
jfkirk: I factored-out lookup_generic everywhere except TradingAlgorithm.symbol(). I made a new method retrieve_asset_by_identifier() that is used in its place here.

zipline/test_algorithms.py
jfkirk: Makes sense. I didn't touch the logic here, really, just the variable name, but I've put in the change you suggested.

zipline/utils/cli.py
jfkirk: This arg can now be a path to a file or folder, as well as 'yahoo', so I removed the choices.

Should I get rid of the single-letter entries? That may cause some user confusion, but I don't know how widely the CLI is actually used.

zipline/utils/cli.py
jfkirk: Sorry, this was a carry-over from testing. We do now accept start and end dates properly from the CLI, though, so I've modified it to only overwrite if no start and end dates are provided.

zipline/utils/cli.py
jfkirk: The first one, asset_identifier, is only used if a metadata file is provided. I've refactored to set 'symbol' as the arg default, rather than do this none-checking.

zipline/utils/security_list.py
jfkirk: Yep - refactored

zipline/protocol.py
jfkirk: good catch - trailing comma added

zipline/algorithm.py
jfkirk: Nice and clean - thanks

zipline/algorithm.py
jfkirk: I factored out a new function to be used by all the order_value methods to avoid this redundancy.

zipline/assets/assets.py
jfkirk: Sounds good - I've refactored these steps to do try-excepts

zipline/assets/assets.py
jfkirk: @ehebert Ping on this str/basestring clarification

zipline/assets/assets.py
jfkirk: Refactored - thanks

tests/test_perf_tracking.py
jfkirk: Refactored to use with_environment

tests/test_perf_tracking.py
jfkirk: Refactored to use with_environment

zipline/assets/_assets.pyx
jfkirk: Fixed

zipline/assets/_assets.pyx
jfkirk: Changed accordingly.

zipline/finance/controls.py
jfkirk: I refactored to your example. The expected empty value is a None - should I change it to a "if sid.end_date is None"?

zipline/finance/performance/position_tracker.py
jfkirk: Confirmed - reworked using the example you gave me

zipline/finance/performance/tracker.py
jfkirk: Refactored to use with_environment

zipline/finance/performance/tracker.py
jfkirk: Refactored to use with_environment

zipline/finance/performance/tracker.py
```diff
@@ -388,7 +399,29 @@ def check_upcoming_dividends(self, midnight_of_date_that_just_ended):
...
+        Checks for any assets that are ending the next day and closes them
+        out by executing a transaction. The transaction is generated by the
+        position_tracker.
+        """
+        for date, sids in self._asset_ends.items():
+            if date > next_trading_day:
+                continue
+            self._past_asset_end_dates.add(date)
+            for sid in sids:
+                txn = self.position_tracker._maybe_create_end_sid_transaction(
```
warren-oneill: Is it possible to make this call optional? Maybe with a flag passed to ```TradingAlgorithm```?

zipline/finance/performance/tracker.py
```diff
@@ -450,15 +487,18 @@ def handle_market_close_daily(self):
...
         # Roll over positions to current day.
         self.todays_performance.rollover()
         self.todays_performance.period_open = self.market_open
         self.todays_performance.period_close = self.market_close
 
-        self.check_upcoming_dividends(completed_date)
+        next_trading_day = env.next_trading_day(completed_date)
+        if next_trading_day:
+            self.check_upcoming_dividends(next_trading_day)
+            self.close_expired_asset_positions(next_trading_day)
```
warren-oneill: Is it possible to make this call optional? Maybe with a flag passed to ```TradingAlgorithm```?

zipline/assets/assets.py
```diff
@@ -0,0 +1,543 @@
...
+        except KeyError:
+            pass
+        try:
+            kwargs['notice_date'] = pd.Timestamp(self.trading_calendar.\
+                canonicalize_datetime(pd.Timestamp(kwargs['notice_date'])))
+        except KeyError:
+            pass
+        try:
+            kwargs['expiration_date'] = pd.Timestamp(self.trading_calendar.\
+                canonicalize_datetime(pd.Timestamp(kwargs['expiration_date'])))
```
warren-oneill: @jfkirk does the expiration date have to be canonicalized? I would like to be able to use hours and minutes.

jfkirk: No - good catch. I will remove the canonicalization.

jfkirk:
Closing to reopen squashed PR

PR #572, opened 2015-04-26
PERF: Reduce memory usage during risk metric calculations.

ehebert:
Instead of passing a pd.Series to calculations, use a memory view of the
underlying numpy array which is sliced to the exact dt, to prevent
calculation like beta creating some temporary copies of the Series.

The creation of the temporary Series was consuming more memory than
needed, with the index being the high penalty.

## Other Notes

This patch is more triage than a long term fix.
The long term fix actually may be just removing the 'intraday' containers completely, after researching whether or not they are still needed, with the 'cumulative' mode which uses daily returns with the last value being the latest minute covering all the needed cases.
If minute returns it is still needed, then more of the cumulative stats like `self.algorithm_returns` should be ported over to using memory views as well; however that requires touching more parts (the period/risk report depend on the .index of algorithm_returns, may interact with serialization, etc.) so doing this quick fix to get some savings right away.

This patch reduces the memory consumption of a noop algorithm which uses minute emission from,
image

to

image

with a savings of about 100MB at the two month mark.

### DIFF
--- a/zipline/finance/risk/cumulative.py
+++ b/zipline/finance/risk/cumulative.py
@@ -312,10 +312,18 @@ algorithm_returns ({algo_count}) in range {start} : {end} on {dt}"
         self.excess_returns[self.latest_dt] = (
             self.algorithm_cumulative_returns[self.latest_dt] -
             self.treasury_period_return)
-        self.metrics.beta[dt] = self.calculate_beta()
+        dt_index = self.metrics.index.searchsorted(dt)
+        algorithm_returns = self.algorithm_returns.values[:dt_index + 1]
+        benchmark_returns = \
+            self.benchmark_returns.values[:dt_index + 1]
+        mean_returns = self.mean_returns.values[:dt_index + 1]
+        self.metrics.beta[dt] = self.calculate_beta(algorithm_returns,
+                                                    benchmark_returns)
         self.metrics.alpha[dt] = self.calculate_alpha()
         self.metrics.sharpe[dt] = self.calculate_sharpe()
-        self.metrics.downside_risk[dt] = self.calculate_downside_risk()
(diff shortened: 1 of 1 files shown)


zipline/finance/risk/cumulative.py
```diff
@@ -312,10 +312,18 @@ def update(self, dt, algorithm_returns, benchmark_returns, account):
         self.excess_returns[self.latest_dt] = (
             self.algorithm_cumulative_returns[self.latest_dt] -
             self.treasury_period_return)
-        self.metrics.beta[dt] = self.calculate_beta()
+        dt_index = self.metrics.index.searchsorted(dt)
+        algorithm_returns = self.algorithm_returns.values[:dt_index + 1]
```
richafrank: Does this slice into the array produce the same effect as the dropna call in calculate_beta (but without the memory copy)?

If so, is it ok that we're calling calculate_downside_risk with that memview (and similarly with the mean_returns) now, whereas we were not dropping nans in calculate_downside_risk before?

ehebert: I believe so.

IIRC, historically the `dropna` was there both for the righthand trimming of the values, and dealing with possible holes in the data because of not being triggered every bar; however, now that the benchmark has been acting as the 'clock' for triggering this logic that secondary reason shouldn't be needed anymore.

ehebert: On further review, the dropna is no longer doing right hand trimming, because of the earlier use of `[:dt]` when assigning `self.algorithm_returns`.

Looking at removing some extraneous re-slicing.

ehebert:
@richafrank this has been simplified with https://github.com/quantopian/zipline/pull/573

PR #573, opened 2015-04-28
PERF: Reduce memory usage during risk metric calculations.

ehebert:
Instead of using pd.Series calculations, pass the underlying
numpy array which have already been sliced to the exact dt, to prevent
calculation like beta creating some temporary copies of the Series with
an unneeded call to `dropna`.

The creation of the temporary Series was consuming more memory than
needed, with the index being the high penalty.


ehebert:
See #572

ehebert:
Tested this with same memory profiling as #572 and the same memory profile as using the views holds.

image

PR #574, opened 2015-04-30
TST: Don't modify master security lists directory during tests

jikamens:
Rather than drop files temporarily into the master security lists
directory during unit tests, create temporary directories for the
tests. This avoids issues when the tests are being run at the same
time as other code that uses the real security lists data.

### DIFF
--- a/zipline/utils/test_utils.py
+++ b/zipline/utils/test_utils.py
@@ -1,13 +1,15 @@
 from contextlib import contextmanager
 from logbook import FileHandler
+from mock import patch
 from zipline.finance.blotter import ORDER_STATUS
-from zipline.utils.security_list import SECURITY_LISTS_DIR
+from zipline.utils import security_list
 
 from six import itervalues
 
 import os
 import pandas as pd
 import shutil
+import tempfile
 
 
 def to_utc(time_str):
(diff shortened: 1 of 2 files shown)


jikamens:
@ehebert @twiecki this is to fix a problem with parallel qexec builds. It's a pretty straightforward change. Can you take a look?

zipline/utils/test_utils.py
```diff
@@ -115,15 +117,34 @@ def nullctx():
...
+    old_dir = security_list.SECURITY_LISTS_DIR
+    new_dir = tempfile.mkdtemp()
+    try:
+        for subdir in os.listdir(old_dir):
+            shutil.copytree(os.path.join(old_dir, subdir),
+                            os.path.join(new_dir, subdir))
+            with patch.object(security_list, 'SECURITY_LISTS_DIR', new_dir), \
+                    patch.object(security_list, 'using_copy', True,
+                                 create=True):
+                yield
```
twiecki: Haven't seen this before, does it not have to `yield` something?

jikamens: When you decorate a method with `@contextmanager`, everything up until the yield becomes the eqiuvalent of the `__entering__` function for the context manager and everything after the yield becomes the `__exit__` function.

jikamens: https://docs.python.org/2/library/contextlib.html

twiecki: That's cool, thanks for explaining.

twiecki:
Looks good to me :+1:

PR #575, opened 2015-04-30
Reduce memory footprint of minute emission algorithms.

ehebert:
The intraday_risk_metrics is being removed since the values are not
used; cumulative risk metrics with the last value updated to the latest
close has been used for some time.

Before the removal of intraday_risk_metrics, the position trackers
passing of benchmark returns to the cumulative risk metrics needs to no
longer depend on the calculations done by the intraday stats. So instead
use the all_benchmark_returns stored in the tracker directly.

The memory issues could be further tuned ala https://github.com/quantopian/zipline/commit/00ea7b04d10f095d064256f4dedff29c095466f3, but cutting this off at the root is simpler.

Effect shown by logging RSS and VM highwater mark for a noop algo using minute emissions.
(Over two months the usage is ~200MB vs. ~750MB)

image

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -116,7 +116,6 @@ class PerformanceTracker(object):
         if self.emission_rate == 'daily':
             self.all_benchmark_returns = pd.Series(
                 index=self.trading_days)
-            self.intraday_risk_metrics = None
             self.cumulative_risk_metrics = \
                 risk.RiskMetricsCumulative(self.sim_params)
 
@@ -124,8 +123,6 @@ class PerformanceTracker(object):
             self.all_benchmark_returns = pd.Series(index=pd.date_range(
                 self.sim_params.first_open, self.sim_params.last_close,
                 freq='Min'))
-            self.intraday_risk_metrics = \
-                risk.RiskMetricsCumulative(self.sim_params)
 
             self.cumulative_risk_metrics = \
                 risk.RiskMetricsCumulative(self.sim_params,
@@ -281,7 +278,6 @@ class PerformanceTracker(object):
             _dict.update({'daily_perf': self.todays_performance.to_dict()})
         elif emission_type == 'minute':
             _dict.update({
-                'intraday_risk_metrics': self.intraday_risk_metrics.to_dict(),
                 'minute_perf': self.todays_performance.to_dict(self.saved_dt)
             })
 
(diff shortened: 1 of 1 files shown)


ssanderson:
:space_invader:

ehebert:
Huh, noticed that there are two Travis links, one is broken (on a flake8 failer to test_munge, which is not touched here) and one that passed.

jikamens:
I thought I pushed all my Zipline test fixes, including the flake8 fix for test_munge, but I may have forgotten to push. Sorry, I was under time crunch all afternoon trying to get all the tests cleaned up because I have plans tonight and had to leave early to pick up kids. I'll finish cleaning everything up later tonight, I hope.

Sent from my Android device



-----Original Message-----

zipline/finance/performance/tracker.py
```diff
@@ -401,17 +397,16 @@ def handle_minute_close(self, dt):
...
-        bench_since_open = \
-            self.intraday_risk_metrics.benchmark_cumulative_returns[dt]
+        # Duplicate intraday_risk_metrics work of calculating the benchmark
+        # returns since open.
+        # intraday_risk_metrics is marked for removal.
+        #
+        # This redundant work is in anticipation of no longer being able to
+        # depend on the 'since open' calculations in intraday_risk_metrics.
+        bench_returns = self.all_benchmark_returns.loc[todays_date:dt]
+        bench_since_open = (1. + bench_returns).prod() - 1
```
richafrank: Looks like the right behavior to me.  Do you think it would be worth delegating to RiskMetricsCumulative.calculate_cumulative_returns here, either by factoring it out or making it static?  I couldn't tell exactly from the comment whether that function will eventually disappear too.

zipline/finance/performance/tracker.py
```diff
@@ -281,7 +278,6 @@ def to_dict(self, emission_type=None):
             _dict.update({'daily_perf': self.todays_performance.to_dict()})
         elif emission_type == 'minute':
             _dict.update({
-                'intraday_risk_metrics': self.intraday_risk_metrics.to_dict(),
                 'minute_perf': self.todays_performance.to_dict(self.saved_dt)
```
richafrank: Not a big deal, but since this and the dict above both contain only a single item, they could be ```_dict['minute_perf'] = ...``` instead of ```.update({...})```.

ehebert: Noted, will fix in a later patch. But leaving this as the minimal change.

zipline/finance/performance/tracker.py
ehebert: I should (and will) update the comment before merging, since the actual removal is done in this branch.

I think the factoring out is possibly half of one, six of the other kind of call. Since it is a pretty standard one liner for getting the cumulative returns... but can see the benefits of clearly labeling it.

For now, I think I'm going to leave it, but with a `# cumulative returns` comment.

PR #577, opened 2015-05-04
BUG: Correct method for finding security lists directory

jikamens:
The correct thing to look at to figure out where the root of the
zipline tree is, is `zipline.__file__`, not `zipline.__path__`. The
latter could contain multiple directories in it, and is not intended
to be `os.path.join()`ed as the previous code was doing.


jikamens:
Can somebody take a look at this minor fix? Thanks.

richafrank:
Looks right to me @jikamens .

PR #579, opened 2015-05-06
WIP BUG: env.next_trading_day() not Timezone aware when passed market_open/close

warren-oneill:
In certain cases the trading day and market open can fall on different days when the timestamp is converted to utc. What this means is the ```env.next_trading_day()``` can get stuck in a continuous loop due to the timezone change when ```market_open``` is passed.
Till now this has probably been not much of an issue since the NYSE open times remain on the same day when converted to UTC but when used in markets where the difference between local tz and UTC is bigger this can be an issue.
To rectify this I pass the trading day to  ```env.next_trading_day()``` instead of ```market_open```.

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -61,6 +61,7 @@ from __future__ import division
 import logbook
 import pickle
 from six import iteritems
+from datetime import datetime
 
 import numpy as np
 import pandas as pd
@@ -92,9 +93,12 @@ class PerformanceTracker(object):
         self.period_start = self.sim_params.period_start
         self.period_end = self.sim_params.period_end
         self.last_close = self.sim_params.last_close
-        first_day = self.sim_params.first_open
+        first_open = self.sim_params.first_open.tz_convert(
+            trading.environment.exchange_tz)
+        self.day = pd.Timestamp(datetime(first_open.year, first_open.month,
+                                         first_open.day), tz='UTC')
         self.market_open, self.market_close = \
-            trading.environment.get_open_and_close(first_day)
+            trading.environment.get_open_and_close(self.day)
         self.total_days = self.sim_params.days_in_period
         self.capital_base = self.sim_params.capital_base
         self.emission_rate = sim_params.emission_rate
@@ -423,7 +427,7 @@ class PerformanceTracker(object):
         rate.
         """
         self.update_performance()
-        completed_date = normalize_date(self.market_close)
+        completed_date = self.day
         account = self.get_account(True)
 
         # update risk metrics for cumulative performance
@@ -448,7 +452,8 @@ class PerformanceTracker(object):
 
         # move the market day markers forward
         self.market_open, self.market_close = \
-            trading.environment.next_open_and_close(self.market_open)
+            trading.environment.next_open_and_close(self.day)
+        self.day = trading.environment.next_trading_day(self.day)
 
         # Roll over positions to current day.
         self.todays_performance.rollover()


warren-oneill:
Not sure about this one anymore. I've found some other points in the code which need to be timezone-aware. I'll try and make these changes and commit early next week.

ssanderson:
I'm pretty far removed from this area of the codebase.  My guess is that @ehebert is the best person to review this.

warren-oneill:
thanks, any thoughts @ehebert?

ehebert:
ACK'd will look at this today.

zipline/finance/performance/tracker.py
```diff
@@ -423,7 +427,7 @@ def handle_market_close_daily(self):
         rate.
         """
         self.update_performance()
-        completed_date = normalize_date(self.market_close)
+        completed_date = normalize_date(self.day)
```
ehebert: It looks like your change means that `self.day` is always midnight.
So this call to `normalize_date` may now be redundant.

zipline/finance/performance/tracker.py
```diff
@@ -92,9 +93,12 @@ def __init__(self, sim_params, env=None):
         self.period_start = self.sim_params.period_start
         self.period_end = self.sim_params.period_end
         self.last_close = self.sim_params.last_close
-        first_day = self.sim_params.first_open
+        first_open = self.sim_params.first_open
+        self.day = first_open.tz_convert(trading.environment.exchange_tz)
```
ehebert: The assignment of `self.day` twice makes the `__init__` a little harder to read, at least semantically.
I'd suggest changing the above so that they are something like.
```
first_open = self.sim_params.first_open.tz_convert(trading.enviroment.exchange_tz)
```

and then use first_open instead of `self.day` in setting the 'real' self.day.

ehebert:
@warren-oneill the idea looks pretty sound to me.

I think we could remove `normalize_date` both at the line note I mentioned above, and in the `next_trading_day` function in `trading`.

I'll run it through some of our internal testing as is, and with attempting to remove normalize_date as well.

warren-oneill:
@ehebert thanks for the feedback. I removed the ```normalize_date``` from ```next_trading_day``` but it seems to have opened a big can of worms. Sometimes ```next_trading_day``` is passed UTC sometimes local and sometimes market open and sometimes market close and with my addition also day. I'd be in favor of decided and one type of variable to pass or at least implement a method which converts to the desired format (best case correct day in UTC).

warren-oneill:
ok after looking at the tests you can add ```datetime.date``` as a further variable type :)

ehebert:
I agree that it should be standardized.

The two leading candidates appear to be `datetime.date` or `pd.Timestamp` with a midnight UTC timestamp?

One benefit of the pd.Timestamp with midnight UTC is that aligns with how pandas does many things.
However, using datetime.date would remove any possibility for tz murkiness.

warren-oneill:
Just pushed a new commit which checks for variable type before calculating but it's def not the ideal solution. 
I'm leaning towards ```datetime.date``` because with ```pd.Timestamp```it would still be possible to after to UTC conversion to get a ```market_open``` or ```market_close``` which is at midnight and that could lead to some confusion.
How would we go about making sure the function receives the right input? Should it be able to convert the input to the correct format or should it just raise an error?

warren-oneill:
whoops just saw I messed up my if statements. will push the fix shortly.

warren-oneill:
Ok ran out of time. Gonna have to come back to this next week. Let me know what standardization you want to go for and I'll work on implementing it next week.

warren-oneill:
Hey @ehebert I decided to revert to the previous commit. The standardization is too big a task for me atm but it is definitely important. Basically, nearly every instance where the passing of market open or close to functions in ```env``` when day is need will have to be changed but at least here we have a start.
How did it fair in your internal tests?

ehebert:
@warren-oneill the internal tests on based on a branch of your first version did pass, so that means there should be less friction here.

(https://github.com/quantopian/zipline/commits/eb7725b93252ad406dca1013c50dc888c484339d is the commit I tested)

So I think that means we can pull in up to there.

I agree with the standardization being important, and it may be too big for me tackle in the near future as well, but let's talk to @jfkirk and @ssanderson about prioritization of it.

warren-oneill:
Ok perfect, those are the two commits that are currently on this branch. 
Will I do a rebase and squash them into one commit?

ehebert:
Sounds good. Thanks!

warren-oneill:
:+1: done

PR #580, opened 2015-05-07
PERF: Use specific methods for processing events.

ehebert:
By having both the trade simulation main loop route events to "process"
methods based on event type and the process methods also checking event
type, there was some duplicated effort in doing that comparison many
times.

A particular case where this was noted in profiling was for the
`process_event` function which was checking if the type was not a trade
and returning early, when in a larger universe of stocks the value
returned False 99% of the time.

Instead provide separate process functions specific to each type,
e.g. e.g. `process_trade` and `process_transaction` and route traffic to
those functions in tradesimulation.

For a universe of 160 stocks on both no-op algo and an algo that rebuys
its universe every day, saw about a 10% increase locally.

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -66,7 +66,6 @@ import numpy as np
 import pandas as pd
 from pandas.tseries.tools import normalize_date
 
-import zipline.protocol as zp
 import zipline.finance.risk as risk
 from zipline.finance import trading
 from . period import PerformancePeriod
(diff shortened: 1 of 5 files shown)


jfkirk:
:+1: This change makes a lot of sense. In your words "its going to be convoluted either way, but this way its obviously convoluted".

zipline/finance/performance/tracker.py
```diff
@@ -281,62 +281,56 @@ def to_dict(self, emission_type=None):
...
-                    perf_period.handle_cash_payment(leftover_cash)
+        self.txn_count += 1
+        self.position_tracker.execute_transaction(event)
+        for perf_period in self.perf_periods:
+            perf_period.handle_execution(event)
 
-        elif event.type == zp.DATASOURCE_TYPE.ORDER:
-            for perf_period in self.perf_periods:
-                perf_period.record_order(event)
+    def process_dividend(self, dividend):
```
jfkirk: Where is this method called?

ehebert: Looks like it I omitted it without realizing. In part, because we only do `log.info("Ignoring DIVIDEND event.")`, but I'll add it to the tradesimulation loop for completion sake.

ehebert:
@jfkirk I just pushed a squashed update of the branch. Mind doing another round?

This added the handling of dividends and some compatibility with our internal blotter.

jfkirk:
:+1:

PR #587, opened 2015-05-19
MAINT: Group events by type before processing.

ehebert:
Make the ordering in which processing of event types both explicit and
independent of the sort ordering of the incoming sources.

The overhead of creating the list per snapshot and the iterators appears
to be marginal in the minute data case when tested locally.

This patch is intended as part of the path towards making the trade
simulation loop not depend on consuming and tracking every trade event.
The timing of where last_sale_date was needed to be changed was proving
difficult to adapt in the previous model.

Should also allow the removal of sorting of the various source streams.

### DIFF
--- a/zipline/gens/tradesimulation.py
+++ b/zipline/gens/tradesimulation.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 from logbook import Logger, Processor
 from pandas.tslib import normalize_date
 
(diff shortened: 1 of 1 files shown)


ehebert:
Note, merging depends on co-coordination with an internal branch.

zipline/gens/tradesimulation.py
```diff
@@ -256,13 +299,13 @@ def _process_snapshot(self, dt, snapshot, instant_fill):
             # Now that handle_data has been called and orders have been placed,
             # process the event stream to fill user orders based on the events
             # from this snapshot.
-            for event in events_to_be_processed:
-                for txn, order in blotter_process_trade(event):
+            for trade in events_to_be_processed:
+                for txn, order in blotter_process_trade(trade):
```
jfkirk: Is this redundant with line 270 above?

jfkirk:
This change will not impact futures trading. Thanks for the heads up!

zipline/gens/tradesimulation.py
jfkirk: Ah, nevermind, I didn't realize this was the instant_fill handler.

ehebert: This is a quirk of the implementation to support the `instant_fill` behavior, which is provided as an option to have daily mode execute on the same day. However, I don't know how much use that behavior has.

Depending on the value of the flag which is a simulation "constant", either this block, or the one above is executed.

ehebert: No worries, a question could be 'do we need to keep this, considering the confusion that it causes in implementation'.
Not sure how many people are using the feature, we do not use it internally, as far as I know.

zipline/gens/tradesimulation.py
```diff
@@ -206,46 +207,88 @@ def _process_snapshot(self, dt, snapshot, instant_fill):
...
+        # Handle benchmark first.
+        #
+        # Internal broker implementation depends on the benchmark being
+        # processed first so that transactions and commissions reported from
+        # the broker can be injected.
+        if benchmark is not None:
+            benchmark_event_occurred = True
+            perf_process_benchmark(benchmark)
+            for txn, order in blotter_process_benchmark(benchmark):
+                if txn.type == DATASOURCE_TYPE.TRANSACTION:
```
ssanderson: Probably not germane to this PR, but why do we get transactions and commissions based on processing of the benchmark?  Those seem totally unrelated to me...

zipline/gens/tradesimulation.py
jfkirk: It makes sense to me to have both:

1. If instant_fill, the blotter processes all trades on the same cycle after handle_data.
2. If not instant_fill, the blotter processes all trades on the same cycle before handle_data.

Right?

zipline/gens/tradesimulation.py
```diff
@@ -206,46 +207,88 @@ def _process_snapshot(self, dt, snapshot, instant_fill):
...
+        # of the individual sources.
+
+        # There is only one benchmark per snapshot.
+        benchmark = None
+        # trades are initialized as a list since there is often a trade per
+        # snapshot, so better to allocate the list than check for is None
+        # on each event.
+        trades = []
+        # splits list will be allocated when a split occurs in the snapshot.
+        splits = None
```
ssanderson: The extra complexity and boolean checks necessitated by lazily-allocating these seem not worth it to me.

zipline/gens/tradesimulation.py
```diff
@@ -206,46 +207,88 @@ def _process_snapshot(self, dt, snapshot, instant_fill):
...
+            elif event.type == DATASOURCE_TYPE.CUSTOM:
+                if customs is None:
+                    customs = []
+                customs.append(event)
+            elif event.type == DATASOURCE_TYPE.DIVIDEND:
+                if dividends is None:
+                    dividends = []
+                dividends.append(event)
+            else:
+                raise log.warn("Unrecognized event=%s".format(event))
```
ssanderson: Pretty sure raising `log.warn` doesn't work...

zipline/gens/tradesimulation.py
ehebert: I'm not 100% why it was done this. I believe the intent was a way to splice in getting information back from the broker blotter into the tradesimulation loop, @richafrank might be able to give more context.

With the course of changing the tradesimulation loop to be focused on the simulation dt as top loop, we may want to change how the blotter is queried.

This dependency on getting the injection of values from the benchmark is impetus behind this patch, aiming to provide a setup where that timing of asking of the broker-backed blotter can be independent stream ordering.

zipline/gens/tradesimulation.py
ehebert: Agreed. No idea why I typed that.

zipline/gens/tradesimulation.py
richafrank: That's right: once per bar we're checking for broker info, not on every trade.

zipline/gens/tradesimulation.py
ehebert: I'm leaning towards leaving this as is, in the spirit of making the complexity here more explicit.
i.e. the complexity of calling process_snapshot for both the market data and corporate actions.

It is one of the near-mid-term goals to remove `_process_snapshot`, and I'm hoping having the code as such will be a reminder that there will need to be a need to handle/retrieve the corp actions outside/after the market minute range.

PR #589, opened 2015-05-22
DEP: Removes use of 'count'-defined test sources

jfkirk:
Test sources are now defined by the sim_params period_start and period_end, rather than by the period_start and a defined 'count' of bars. This allows us to consider the sim_params.period_end as the canonical definition of the end of a simulation.

### DIFF
--- a/zipline/sources/test_source.py
+++ b/zipline/sources/test_source.py
@@ -19,9 +19,9 @@ A source to be used in testing.
 
 import pytz
 
-from itertools import cycle
-from six.moves import filter, zip
+from six.moves import filter
 from datetime import datetime, timedelta
+import itertools
 import numpy as np
 
 from six.moves import range
@@ -53,9 +53,9 @@ def create_trade(sid, price, amount, datetime, source_id="test_factory"):
 
 
 @with_environment()
-def date_gen(start=datetime(2006, 6, 6, 12, tzinfo=pytz.utc),
+def date_gen(start,
+             end,
              delta=timedelta(minutes=1),
-             count=100,
              repeats=None,
              env=None):
     """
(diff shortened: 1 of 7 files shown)


jfkirk:
@ehebert @ssanderson Mind giving this a review?

zipline/sources/test_source.py
```diff
@@ -88,7 +87,7 @@ def advance_current(cur):
 
     # yield count trade events, all on trading days, and
     # during trading hours.
-    for i in range(count):
+    while cur < end:
```
ehebert: Making sure I'm reading this right. In the case of `start == end` like with the default parameters and no repeats, this will yield no data, correct?

jfkirk: Yes, and I'm just now seeing that, in the case of default args, this would mean no dates. Should I remove the default args?

ehebert: I think that might help reduce some cases where the gen is yielding nothing and causing something to silently not run.

jfkirk: Done - pushed an overwriting commit

ehebert:
Looks good to me. :+1:

jfkirk:
Re-wrote the trade generator to be lazy

jfkirk:
Thanks - waiting on internal compatibility now

zipline/utils/factory.py
```diff
@@ -224,65 +224,60 @@ def create_returns_from_list(returns, sim_params):
...
 
+
+    # If the sim_params define an end that is during market hours, that will be
+    # used as the end of the data source
+    if trading.environment.is_market_hours(sim_params.period_end):
+        end = sim_params.period_end
+    # Otherwise, the last_close after the period_end is used as the end of the
+    # data source
+    else:
+        end = sim_params.last_close
```
jfkirk: @ehebert -- Does this logic make sense? This allows us to define period_ends that are during a day, but if a time outside of the trading day is given, the last_close of that day is used.

PR #590, opened 2015-05-28
pass day instead of market_open to next_open_and_close

warren-oneill:
@jfkirk as in commit 064d973. Was prob overwritten when refactoring.


jfkirk:
Thanks! I couldn't cherry-pick, so I committed the same fix directly to the branch.

PR #591, opened 2015-05-29
history bugfix

llllllllll:


zipline/history/history.py
```diff
@@ -278,6 +285,8 @@ def __init__(self, bar_count, frequency, field, ffill,
         # The frequency at which the data is sampled.
         self.frequency = frequency
         # The field, e.g. 'price', 'volume', etc.
+        if not isinstance(field, Hashable):
```
ssanderson: Not sure if I have a preference if they're a equivalent, but is this functionally different from:
```
try:
    hash(field)
except:
    raise ValueError(...)
```
?

zipline/history/history.py
```diff
@@ -278,6 +285,8 @@ def __init__(self, bar_count, frequency, field, ffill,
...
+            raise ValueError('field must be hashable')
```
ssanderson: This should reference what the actual received field was.

zipline/history/history.py
```diff
@@ -266,6 +270,9 @@ def __init__(self, bar_count, frequency, field, ffill,
                  data_frequency='daily'):
 
         # Number of bars to look back.
+        if not isinstance(bar_count, int):
+            raise ValueError('bar_count must be an int')
```
ssanderson: This should reference the actual received type so the user can do something about it.

zipline/history/history.py
llllllllll: This is only difference if you are a bully and do something like:

```
class C(object):
    @property
    def __hash__(self):
        raise AttributeError()
```

But the abc is cleaner imho

ssanderson: So the abc is equivalent to `hasattr('__hash__')`?

llllllllll: err no, that is `hasattr(type(c))`

llllllllll:
@ssanderson looks good?

zipline/history/history.py
```diff
@@ -266,6 +270,14 @@ def __init__(self, bar_count, frequency, field, ffill,
                  data_frequency='daily'):
 
         # Number of bars to look back.
+        if not isinstance(bar_count, int):
+            raise ValueError(
```
ssanderson: I think this should be a TypeError

zipline/history/history.py
ssanderson: Looking at this again, I think we should just assert that this is in HistoryContainer.VALID_FIELDS.  We might have to move that somewhere shared to fix a circular dependency.

ssanderson: (Where "assert" might mean "check and raise a ValueError".)

jbredeche:
@llllllllll this is likely no longer needed due to the recent history rewrite, what do you think?

PR #594, opened 2015-06-06
PERF: Use tolerant_equals instead of np.allclose in order types.

ehebert:
np.allclose creates a vector of size 1 if passed a scalar, which showed
up as a bottleneck in algorithms with a large number of
order_target_percent calls.


ehebert:
cc: @jfkirk

jfkirk:
A thing of beauty. A work of art. A generational event.

A mere :+1: can not suffice.

PR #595, opened 2015-06-08
Futures trading

jfkirk:
This branch adds Asset management to Zipline and adds the necessary accounting to allow for trading of Futures.

Primary impacts:
1. There is now a distinction between an Asset's 'identifier', 'sid', and 'symbol'.
    * 'identifier': the user-provided unique int or str that allows us to marry-up AssetMetaData and DataSource data
    * 'sid': a unique integer that can be used to identify a specific Asset within a given TradingEnvironment
    * 'symbol': non-unique string name for user look-up and ease-of-use, but is not used internally
    * This distinction has not been completely refactored in to the API – many methods still call for an argument 'sid' where they would actually want an Asset object
2. Zipline user may provide 'identifiers' or 'asset_metadata' arguments to TradingAlgorithm.init so the asset_finder can be built. Alternatively, the sids given to DataFrameSource or PanelSource are used to build Assets (defaulting to Equities).
3. Zipline algorithms must use Asset objects as arguments to all order*() methods.
    * The sid(int) and symbol(str) methods can be used to look up an Asset
4. There is now a distinction between 'value' and 'exposure' in the perf tracker and positions tracker.

Note: There are a number of API methods that take an arg named 'sid,' but what we really expect is an asset object. This should be reconciled.

### DIFF
--- a/docs/release-notes/zipline-0.8.0.md
+++ b/docs/release-notes/zipline-0.8.0.md
@@ -44,11 +44,10 @@
   > running with minute data, then this will calculate the number of minutes in
   > those days, accounting for early closes and the current time and apply the
   > transform over the set of minutes. `returns` takes no parameters and will
-  > return the daily returns of the given security.
+  > return the daily returns of the given asset.
 
   > Example:
 ```python
-# The standard deviation of the price in the last 3 days.
 data[security].stddev(3)
 ```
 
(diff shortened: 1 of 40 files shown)


warren-oneill:
hey @jfkirk could you remove my unused ```tearDown``` in my algo unittest? Thought it would be easier just to push it through in this PR instead of making a new one.

PR #596, opened 2015-06-09
WIP: Add US Equities loader for FFC.

ehebert:
Not ready to merge, requires at unit tests beforehand.

To use this branch locally:
- download bcolz file from internal server
- ln -s $path_to_file .
- `python -m zipline.data.ffc.shims.us_equity_pricing_index`
- `python -m zipline.data.ffc.shims.us_equity_pricing_test`

Implement FFC loader for US Equity daily pricing using a bcolz file.

The bcolz file packs all of history in blocks of equities, which include
0 values representing nan-values.

The loader uses the offset information for each equity to get a start
and end position for each sid then slices the data for each sid, and
puts the result into the output container.

All data is read into memory since in testing reading it all in and then
using indexing into the np vector is faster than repeatedly making the
bcolz logic determine what blocks to load, when using many sids.

Includes a shim `us_equity_pricing_test` which tries an approximate
average use case of a years worth of data and 8000 sids.

In local testing this query takes ~300ms to execute.

Some questions inline with code comments and in README's.

### DIFF
--- a/zipline/data/ffc/datasets/README.md
+++ b/zipline/data/ffc/datasets/README.md
@@ -1,0 +1,5 @@
+Currently USEquityPricing is located in zipline.data.pricing
+
+Should we move that module to zipline.data.ffc.datasets.us\_equity\_pricing?
+
+(Remove this file before merging.)
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,11 @@ ext_modules = [
         ['zipline/data/adjustment.pyx'],
         include_dirs=[np.get_include()],
     ),
+    Extension(
+        'zipline.data.ffc.loaders._us_equity_pricing',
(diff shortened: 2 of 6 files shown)


ehebert:
cc: @ssanderson

zipline/data/ffc/datasets/README.md
```diff
@@ -0,0 +1,5 @@
+Currently USEquityPricing is located in zipline.data.pricing
```
ssanderson: I have this in zipline.data.equities on my current branch, but there's also a zipline.data.ffc submodule that could be more appropriate.

ssanderson:
@ehebert closing this since the core functionality has been merged to my dev branch

PR #593, opened 2015-06-05
ENH: Add CLOSE_POSTION as DataSource event type

warren-oneill:
This PR enables signals to be sent from ```DataSource``` which tells zipline to close any open positions for the given sid.

Note: this PR uses code from #592 and borrowed lines from the __futures-trading__ branch.
cc @ehebert @jfkirk

PR #598, opened 2015-06-10
TST: adds algo unittest for CLOSE_POSITON event type, adds commission …

warren-oneill:
Something I should have added to #593. The test can be expanded later to include all datasource types.


jfkirk:
Merged with a cherry-pick. Thanks, Warren!

PR #600, opened 2015-06-10
ENH: Moves calling of TradingAlgorithm.initialize to _create_generator

jfkirk:
This commit moves the call to the algo code's initialize method out of TradingAlgorithm.__init__. The algo code's initialize method will be called in _create_generator using args and kwargs that are stored from __init__.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -113,10 +113,6 @@ class TradingAlgorithm(object):
 
     """
 
-    # If this is set to false then it is the responsibility
-    # of the overriding subclass to set initialized = true
-    AUTO_INITIALIZE = True
-
     def __init__(self, *args, **kwargs):
         """Initialize sids and other state variables.
 
@@ -248,13 +244,10 @@ class TradingAlgorithm(object):
 
         self._most_recent_data = None
 
-        # Subclasses that override initialize should only worry about
-        # setting self.initialized = True if AUTO_INITIALIZE is
-        # is manually set to False.
+        # Prepare the algo for initialization
         self.initialized = False
-        self.initialize(*args, **kwargs)
-        if self.AUTO_INITIALIZE:
-            self.initialized = True
+        self.initialize_args = args
+        self.initialize_kwargs = kwargs
 
     def initialize(self, *args, **kwargs):
         """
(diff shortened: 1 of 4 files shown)


jfkirk:
@ssanderson Could you review this? If you're overloaded I can find a different reviewer. It was a straightforward change to move calling the algocode's initialize method.

zipline/algorithm.py
```diff
@@ -248,13 +244,10 @@ def __init__(self, *args, **kwargs):
...
 
-        # Subclasses that override initialize should only worry about
-        # setting self.initialized = True if AUTO_INITIALIZE is
-        # is manually set to False.
+        # Prepare the algo for initialization
         self.initialized = False
-        self.initialize(*args, **kwargs)
-        if self.AUTO_INITIALIZE:
-            self.initialized = True
+        self.initialize_args = args
```
ssanderson: What code do we have that depends on `initialize` receiving the extra `*args` and `**kwargs`?  Can we remove this give `TradingAlgorithm.__init__` a fixed signature?  :IWouldBeSoHappy: if we could do that.

zipline/algorithm.py
```diff
@@ -371,6 +364,11 @@ def _create_generator(self, sim_params, source_filter=None):
         processed by the zipline, and False for those that should be
         skipped.
         """
+
+        if not self.initialized:
+            self.initialize(*self.initialize_args, **self.initialize_kwargs)
```
ssanderson: See note above re: these parameters.

zipline/algorithm.py
jfkirk: The initialize methods in AlgorithmProxy and LiveSessionAlgorithmProxy have different signatures. I will chalk it on to the forthcoming Zipline initialization cleanup/refactoring.

jfkirk:
@StewartDouglas @yankees714 For more eyes

twiecki:
Awesome! This enables parameter optimization :+1:

warren-oneill:
@twiecki oh didn't realize that! great!

femtotrader:
About parameter optimization

this https://jug.readthedocs.org/en/latest/ might also help

I have seen it in "building machine learning systems with python - second edition"

PR #606, opened 2015-06-15
ENH: Adds security_start_date and _end_date with deprecation warnings

jfkirk:
These fields are required for compatibility with existing Quantopian algorithms. They are marked for deprecation in the future if possible.

### DIFF
--- a/zipline/assets/_assets.pyx
+++ b/zipline/assets/_assets.pyx
@@ -19,6 +19,7 @@ Cythonized Asset object.
 cimport cython
 
 import numpy as np
+import warnings
 cimport numpy as np
 
 # IMPORTANT NOTE: You must change this template if you change
@@ -66,22 +67,6 @@ cdef class Asset:
 
     def __hash__(self):
         return self.sid_hash
-
-    property asset_start_date:
-        """
-        Alias for start_date to disambiguate from other `start_date`s in the
-        system.
-        """
-        def __get__(self):
-            return self.start_date
-
-    property asset_end_date:
-        """
-        Alias for end_date to disambiguate from other `end_date`s in the
-        system.
-        """
-        def __get__(self):
-            return self.end_date
 
     def __richcmp__(x, y, int op):
         """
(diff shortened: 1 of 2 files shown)


jfkirk:
@richafrank Could you review this?

richafrank:
Sure thing.  We should get @ssanderson 's input too.

zipline/assets/_assets.pyx
```diff
@@ -203,6 +203,26 @@ cdef class Equity(Asset):
         params = ', '.join(strings)
         return 'Equity(%d, %s)' % (self.sid, params)
 
+    property security_start_date:
+        """
+        DEPRECATION: This property should be deprecated and is only present for
```
richafrank: Are we intentionally choosing not to deprecate it now, or can we warn with a DeprecationWarning at runtime instead?  Or maybe this is a cython convention? (I don't know too much cython.)

jfkirk: I'm unfamiliar with deprecation in general - is there some way that we could make calls to security_start_date throw deprecation warnings in to the user log?

richafrank: Yea, you can use [warnings.warn] (https://docs.python.org/2/library/warnings.html#warnings.warn), passing DeprecationWarning.  I think we have examples of this elsewhere in the code.

jfkirk: Cool, thank you. I've added a warning below to both properties.

richafrank: Awesome!  What do you think of DeprecationWarning instead of UserWarning?

jfkirk: Thanks - wasn't aware of that class. Done.

richafrank:
lgtm @jfkirk !  Sound good to you, @ssanderson ?

ssanderson:
I'm not sure I understand why we have "asset/security_start_date" as distinguished from just start_date and end_date...changes like this seem like a strong argument for just calling the field "start_date" or "end_date" so that they don't become retroactively incorrect if we change names.

Is there code that currently depends on the use of `asset_start_date` as opposed to just `start_date`?

richafrank:
Assuming we keep security_start_date for now, should we add a unit test for it?

jfkirk:
I can add a unit test.

@ssanderson I can cut asset_start_date to just start_date, but we will still need to provide this near-deprecated security_start_date so as to not break old algos.

jfkirk:
@richafrank Added a test
@ssanderson I've removed the asset_*_date fields in deference to simply start_date and end_date. I'm checking impact on qexec now.

richafrank:
Awesome - thanks!

PR #609, opened 2015-06-16
ENH: Add US Equities loader for FFC.

ehebert:
Implement FFC loader for US Equity daily pricing using a bcolz file.

The bcolz file packs all of history in blocks of equities, which include
0 values representing nan-values.

The loader uses the offset information for each equity to get a start
and end position for each sid then slices the data for each sid, and
puts the result into the output container.

All data is read into memory since in testing reading it all in and then
using indexing into the np vector is faster than repeatedly making the
bcolz logic determine what blocks to load, when using many sids.

Adjustments are read from a SQLite database with a table each for
mergers, splits, and adjustments; containing columns for sid, ratio and
effective date.

Includes a shim `us_equity_pricing_test` which tries an approximate
average use case of a years worth of data and 8000 sids.

In local testing this query takes under a second to execute.

To use this branch locally:
- download bcolz file from internal server
- ln -s $path_to_file .
- download adjustments.db from internal server
- ln -s $path_to_adjustments.db
- `python -m zipline.data.ffc.shims.us_equity_pricing_test`


ssanderson:
@ehebert closing since this has been merged into the active development branch

PR #610, opened 2015-06-16
MAINT: Remove position proxy.

ehebert:
Use explicit references to the performance tracker instead of the proxy
lookup.

Mainly a putative change, which surfaced when reasoning about places
where the position tracker and period need access to the last sale
price.


ehebert:
Passed internal tests.

Pending results on performance, but since a layer of function calls are removed, it should be somewhere between no regression and a marginal gain.

jfkirk:
How this PR makes me feel: https://www.youtube.com/watch?v=hw1ERPIBbE8

PR #615, opened 2015-06-18
MAINT: Use sqlite parameterization instead of string formatting.

ehebert:
For security best practices, though it is unlikely that a injectable
value would get to this point, use the Python sqlite3 module's
parameterization for providing parameters.

String formatting is still used on the template for providing the table
name and the number of parameter tokens to use for the IN statement,
since those values are not parameterizable. Those format values there
are hardcoded values, i.e. `splits`, `mergers`, `dividends` and the `?`
character.

Also, factor out the shared query logic between splits, mergers, and
dividends, since this change would have changed all three places in the
same way.

### DIFF
--- a/zipline/data/ffc/loaders/_us_equity_pricing.pyx
+++ b/zipline/data/ffc/loaders/_us_equity_pricing.pyx
@@ -29,36 +29,44 @@ from zipline.data.adjusted_array import (
 from zipline.data.adjustment import Float64Multiply
 
 SID_QUERY_TEMPLATE = """
-SELECT DISTINCT sid FROM {0}
-WHERE effective_date >= {1} AND effective_date <= {2}
+SELECT DISTINCT sid FROM {table_name}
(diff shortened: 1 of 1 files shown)


zipline/data/ffc/loaders/_us_equity_pricing.pyx
```diff
@@ -30,13 +30,13 @@ from zipline.data.adjustment import Float64Multiply
 
 SID_QUERY_TEMPLATE = """
 SELECT DISTINCT sid FROM {0}
-WHERE effective_date >= {1} AND effective_date <= {2}
+WHERE effective_date >= (?) AND effective_date <= (?)
```
ssanderson: I think this would be easier to follow if we did the interpolation on these templates by name.

ehebert:
@ssanderson I did a force push of a fixup implementing the template name recommendation.

ssanderson:
LGTM

PR #619, opened 2015-06-23
TEST: Clean AssetFinder initialization and tests.

ssanderson:
- AssetFinder no longer accepts an unused trading_calendar.
- AssetFinder correctly accepts a DataFrame as input.
- Tests for AssetFinder no longer rely on a global trading environment.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -21,12 +21,10 @@ import sys
 from unittest import TestCase
 
 from datetime import (
-    datetime,
     timedelta,
 )
 import pickle
 import pprint
-import pytz
 import uuid
 import warnings
 import pandas as pd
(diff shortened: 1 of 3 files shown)


tests/test_assets.py
```diff
@@ -330,68 +292,85 @@ class AssetFinderTestCase(TestCase):
 
     @with_environment()
     def test_lookup_symbol_fuzzy(self, env=None):
```
jfkirk: Is this env still necessary?

jfkirk:
Besides the one line comment, this looks good to me :+1:

tests/test_assets.py
ssanderson: Nope!  Good catch.

ssanderson:
@jfkirk will merge when Travis gives the green light.

PR #620, opened 2015-06-23
DEP: Removes unnecessary identifier_cache from asset_finder

jfkirk:
The identifier cache's usage was nearly identical to using lookup_generic, so this commit removes identifier-keyed caching and modifies anything that uses it.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -22,6 +22,7 @@ from unittest import TestCase
 
 from datetime import (
     timedelta,
+    datetime
 )
 import pickle
 import pprint
@@ -542,9 +543,12 @@ class AssetFinderTestCase(TestCase):
         # Build a finder that is allowed to assign sids
         finder = AssetFinder(metadata=metadata, allow_sid_assignment=True)
 
-        # Verify that Assets were built
-        play = finder.retrieve_asset_by_identifier('PLAY')
+        # Verify that Assets were built and different sids were assigned
+        play = finder.lookup_symbol('PLAY', datetime.now())
+        msft = finder.lookup_symbol('MSFT', datetime.now())
         self.assertEqual('PLAY', play.symbol)
+        self.assertIsNotNone(play.sid)
+        self.assertNotEqual(play.sid, msft.sid)
 
     def test_sid_assignment_failure(self):
 
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -719,7 +719,7 @@ class AssetFinderTestCase(TestCase):
         # No contracts exist after 12/14/2015, so we should get none
         self.assertIsNone(finder.lookup_future_by_expiration('AD', dt, jan_16))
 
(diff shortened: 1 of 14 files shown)


tests/test_assets.py
```diff
@@ -591,14 +591,14 @@ def test_consume_asset_as_identifier(self):
     def test_sid_assignment(self):
 
         # This metadata does not contain SIDs
-        metadata = {'PLAY': {'symbol': 'PLAY'},
-                    'MSFT': {'symbol': 'MSFT'}}
+        metadata = {'PLAY': {'symbol': 'PLAY', 'sid': 1},
```
ssanderson: @jfkirk isn't the point of this test to exercise how we handle inputs that **don't** provide sid data?

jfkirk: Yes, and I feel incredibly dense now. Thanks!

jfkirk:
@ssanderson Fixed my silly test mistake.

tests/test_assets.py
```diff
@@ -598,8 +598,9 @@ def test_sid_assignment(self):
         finder = AssetFinder(metadata=metadata, allow_sid_assignment=True)
 
         # Verify that Assets were built
-        play = finder.retrieve_asset_by_identifier('PLAY')
+        play = finder.lookup_generic('PLAY', datetime.now())[0]
```
ssanderson: @jfkirk is there a reason we're calling `lookup_generic` here as opposed to `lookup_symbol`?

jfkirk: Nope, hasty habit. Changing now.

jfkirk: @ssanderson Changed to use lookup_symbol

tests/test_assets.py
```diff
@@ -598,8 +598,9 @@ def test_sid_assignment(self):
         finder = AssetFinder(metadata=metadata, allow_sid_assignment=True)
 
         # Verify that Assets were built
-        play = finder.retrieve_asset_by_identifier('PLAY')
+        play = finder.lookup_symbol('PLAY', datetime.now())
         self.assertEqual('PLAY', play.symbol)
+        self.assertIsNotNone(play.sid)
```
ssanderson: Should we also assert that we assign different sids to `'MSFT'` and `'PLAY'`?

jfkirk: Good idea - done.

jfkirk:
This PR is pending some compatibility tests.

zipline/assets/assets.py
```diff
@@ -544,6 +532,30 @@ def lookup_generic(self,
             self._lookup_generic_scalar(obj, as_of_date, matches, missing)
         return matches, missing
 
+    def map_identifier_list_to_sids(self, index, as_of_date):
```
jfkirk: @ssanderson Does this method make sense? Its function is similar to lookup_generic, but also will build the assets that are needed.

PR #621, opened 2015-06-24
Futures additions for Asset and AssetFinder

yankees714:
This branch adds further functionality for futures in zipline.

Asset objects now have a `root_symbol` attribute, and AssetFinder holds an additional cache, `future_cache`, keyed on root_symbol. Also adds the AssetFinder methods `lookup_future_chain` (returns the list of valid contracts for a given root symbol), `lookup_future_in_chain` (finds the ith contract in the chain), and `lookup_future_by_expiration` (finds a contract by its expiration date).

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -230,6 +230,7 @@ class TestFuture(TestCase):
     future = Future(
         2468,
         symbol='OMH15',
+        root_symbol='OM',
         notice_date=pd.Timestamp('2014-01-20', tz='UTC'),
         expiration_date=pd.Timestamp('2014-02-20', tz='UTC'),
         contract_multiplier=500
@@ -244,6 +245,7 @@ class TestFuture(TestCase):
         self.assertTrue("Future" in reprd)
         self.assertTrue("2468" in reprd)
         self.assertTrue("OMH15" in reprd)
+        self.assertTrue("root_symbol='OM'" in reprd)
         self.assertTrue(("notice_date=Timestamp('2014-01-20 00:00:00+0000', "
                         "tz='UTC')") in reprd)
         self.assertTrue("expiration_date=Timestamp('2014-02-20 00:00:00+0000'"
@@ -256,6 +258,7 @@ class TestFuture(TestCase):
 
     def test_to_and_from_dict(self):
         dictd = self.future.to_dict()
+        self.assertTrue('root_symbol' in dictd)
         self.assertTrue('notice_date' in dictd)
         self.assertTrue('expiration_date' in dictd)
         self.assertTrue('contract_multiplier' in dictd)
@@ -263,6 +266,9 @@ class TestFuture(TestCase):
         from_dict = Future.from_dict(dictd)
         self.assertTrue(isinstance(from_dict, Future))
         self.assertEqual(self.future, from_dict)
+
+    def test_root_symbol(self):
+        self.assertEqual('OM', self.future.root_symbol)
 
 
 class AssetFinderTestCase(TestCase):
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -574,3 +574,125 @@ class AssetFinderTestCase(TestCase):
             for warning in w:
                 self.assertTrue(issubclass(warning.category,
                                            DeprecationWarning))
+
+    def test_lookup_future_in_chain(self):
+        metadata = {
+            2: {
+                'symbol': 'ADN15',
+                'root_symbol': 'AD',
+                'asset_type': 'future',
+                'expiration_date': pd.Timestamp('2015-06-15', tz='UTC')
+            },
+            1: {
+                'symbol': 'ADV15',
+                'root_symbol': 'AD',
+                'asset_type': 'future',
+                'expiration_date': pd.Timestamp('2015-09-14', tz='UTC')
+            },
+            0: {
+                'symbol': 'ADF16',
+                'root_symbol': 'AD',
+                'asset_type': 'future',
+                'expiration_date': pd.Timestamp('2015-12-14', tz='UTC')
+            },
+
+        }
+
+        finder = AssetFinder(metadata=metadata)
+        dt = pd.Timestamp('2015-06-19', tz='UTC')
+
+        # Check that the primary and secondary contracts are as expected
+        primary = finder.lookup_future_in_chain('AD', dt, 1)
+        secondary = finder.lookup_future_in_chain('AD', dt, 2)
+        self.assertEqual(primary.sid, 1)
+        self.assertEqual(secondary.sid, 0)
+
+        # Check that we get None for an invalid contract num
+        self.assertIsNone(finder.lookup_future_in_chain('AD', dt, 0))
+        self.assertIsNone(finder.lookup_future_in_chain('AD', dt, -10))
+        self.assertIsNone(finder.lookup_future_in_chain('AD', dt, 10))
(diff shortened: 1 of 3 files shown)


zipline/assets/assets.py
```diff
@@ -233,6 +235,107 @@ def lookup_symbol(self, symbol, as_of_date, fuzzy=None):
                 else:
                     self.fuzzy_match[(symbol, fuzzy, as_of_date)] = None
 
+    def _sort_future_chains(self):
+        """ Sort by increasing expiration date the list of contracts
+        for each root symbol in the future cache.
+        """
+        exp_key = lambda contract: contract.expiration_date
```
ssanderson: `exp_key = attrgetter('expiration_date')`
https://docs.python.org/2/library/operator.html#operator.attrgetter

tests/test_assets.py
```diff
@@ -251,24 +251,28 @@ def test_type_mismatch(self):
...
         self.assertTrue("Future" in reprd)
         self.assertTrue("2468" in reprd)
-        self.assertTrue("OMK15" in reprd)
-        self.assertTrue("notice_date='2014-01-20'" in reprd)
-        self.assertTrue("expiration_date='2014-02-20'" in reprd)
+        self.assertTrue("OMH15" in reprd)
+        self.assertTrue(("notice_date=Timestamp('2014-01-20 00:00:00+0000', "
+                        "tz='UTC')") in reprd)
+        self.assertTrue("expiration_date=Timestamp('2014-02-20 00:00:00+0000'"
+                        in reprd)
```
jfkirk: We should test that the root_symbol makes it in to the repr, as well.

tests/test_assets.py
```diff
@@ -589,3 +594,134 @@ def test_security_dates_warning(self):
...
+                'asset_type': 'future',
+                'notice_date': '2014-01-20',
+                'expiration_date': '2014-02-20',
+                'contract_multiplier': 500
+            }
+        }
+
+        finder = AssetFinder(metadata=metadata)
+        omh15 = finder.retrieve_asset(0)
+        self.assertEqual('OM', omh15.root_symbol)
```
jfkirk: This test could be piggy-backed in to the TestFuture TestCase (~250). I wrote those cases to ensure that the Future-class specific fields survive creation and the various methods.

tests/test_assets.py
```diff
@@ -589,3 +594,134 @@ def test_security_dates_warning(self):
...
+        primary = finder.lookup_future_in_chain('AD', dt, 1)
+        self.assertEqual(primary.sid, 1)
+
+        secondary = finder.lookup_future_in_chain('AD', dt, 2)
+        self.assertEqual(secondary.sid, 0)
+
+        self.assertIsNone(finder.lookup_future_in_chain('AD', dt, 0))
+        self.assertIsNone(finder.lookup_future_in_chain('AD', dt, -10))
+        self.assertIsNone(finder.lookup_future_in_chain('AD', dt, 10))
+        self.assertIsNone(finder.lookup_future_in_chain('CL', dt, 1))
```
jfkirk: Can you add comments on what we are checking for with these assertions (640-649)? Alternatively, you can add an error message to the assertions that would double as a comment.

tests/test_assets.py
```diff
@@ -589,3 +594,134 @@ def test_security_dates_warning(self):
...
+
+        }
+
+        finder = AssetFinder(metadata=metadata)
+        dt = pd.Timestamp('2015-06-19', tz='UTC')
+
+        ad_contracts = finder.lookup_future_chain('AD', dt)
+        self.assertEqual(len(ad_contracts), 2)
+        self.assertEqual(ad_contracts[0].sid, 1)
+        self.assertEqual(ad_contracts[1].sid, 0)
```
jfkirk: Same as above - some comments about what is being asserted

tests/test_assets.py
```diff
@@ -589,3 +594,134 @@ def test_security_dates_warning(self):
...
+        contract = finder.lookup_future_by_expiration('AD', dt, june_15)
+        self.assertEqual(contract.sid, 1)
+
+        contract = finder.lookup_future_by_expiration('AD', dt, sept_15)
+        self.assertEqual(contract.sid, 1)
+
+        contract = finder.lookup_future_by_expiration('AD', dt, dec_15)
+        self.assertEqual(contract.sid, 0)
+
+        self.assertIsNone(finder.lookup_future_by_expiration('AD', dt, jan_16))
```
jfkirk: Same as above - some comments about what is being asserted

zipline/assets/assets.py
```diff
@@ -242,15 +345,20 @@ def populate_cache(self):
         # Wipe caches before repopulating
         self.cache = {}
         self.sym_cache = {}
+        self.future_cache = {}
```
jfkirk: On second thought, what do you think about renaming this future_chains_cache? That may be more clear, seeing as the items are chains, not individual futures?

zipline/assets/assets.py
```diff
@@ -242,15 +345,20 @@ def populate_cache(self):
...
         self.identifier_cache = {}
         self.fuzzy_match = {}
 
         counter = 0
         for identifier, row in self.metadata_cache.items():
-            self.spawn_asset(identifier=identifier, **row)
+            self._spawn_asset(identifier=identifier, **row)
             counter += 1
```
jfkirk: Scott noticed that the counter is totally unused. Could you cut those variables as part of this PR?

zipline/assets/assets.py
```diff
@@ -338,6 +446,8 @@ def spawn_asset(self, identifier, **kwargs):
         self.identifier_cache[identifier] = asset
         if asset.symbol is not '':
             self.sym_cache.setdefault(asset.symbol, []).append(asset)
+        if isinstance(asset, Future) and asset.root_symbol is not '':
+            self.future_cache.setdefault(asset.root_symbol, []).append(asset)
```
jfkirk: To avoid an instance check, we could bump this up to be within the 'elif asset_type.lower() == 'future':' above.

ssanderson: @jfkirk what's the objection to the `isinstance` check here?

yankees714: @jfkirk also don't we want the separate if block, since futures should be added to both caches?

jfkirk: There's not really any problem with the instance check. I was just thinking that, since we already have a flow where we know the Asset is a Future, we can just put this insertion in that flow. We would still insert it in to .cache with all other assets. The current state of the logic is fine.

jfkirk: On second thought - scratch my objection. The perf difference would be minuscule, and the current state is a clearer flow.

ssanderson: Ah, @jfkirk I misunderstood your suggestion.  I thought you were just saying we should change the branch from `if isinstance(asset, Future)` to `if asset_type == 'future'`.  I agree that we should move this branch up so that there's only one place that we have "futures-only" logic.

ssanderson: Though, after re-reading this function, I don't think we should be putting anything in any caches here; we should just create and return the asset, and then `populate_caches`, which is the only place that calls this, should be responsible for adding assets to the caches.

zipline/assets/assets.py
yankees714: Makes sense, renaming

yankees714:
@jfkirk @ssanderson Incorporated your suggested changes.

jfkirk:
:+1: from me

PR #625, opened 2015-06-29
BUG: Fixes issues with AssetFinder future lookups

yankees714:
Contracts must have been trading at the as_of_date to be considered valid, and a contract's position in the chain is now zero-indexed.


yankees714:
@jfkirk These are just some revisions that make sense after our API discussions.

jfkirk:
Thanks! Reviewing now.

jfkirk:
:+1: Good to merge after Jenkins passes

PR #626, opened 2015-06-29
[BUG] For Discussion - Fixes risky risk free rate

humdings:
Our default risk-free-rate is the 10-year yield, which is
considered risky by 99% of investors. 

This is a quick fix to make the default the 3-month T-bill rate, 
which is a widely used risk-free instrument. 

An api function to set the risk free rate within initialize might be a 
worthwhile project, but just changing this one string solves the
bulk of the issue.

### DIFF
--- a/zipline/finance/risk/cumulative.py
+++ b/zipline/finance/risk/cumulative.py
@@ -38,7 +38,7 @@ from . risk import (
 log = logbook.Logger('Risk Cumulative')
 
 
-choose_treasury = functools.partial(choose_treasury, lambda *args: '10year',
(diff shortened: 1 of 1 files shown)


humdings:
I would be curious to hear others input on what sorts of unexpected consequences might show up as a result of swapping out the 10-year yield for the 3-month T-bill as the default risk-free-rate.

The tests that failed are risk metric calculations, which should be expected because the expected outcome used in the test cases are wrong.

I can fix the test cases if we feel changing this default behavior is worth it.

freddiev4:
Going to close this b/c it's 3 years old and not sure where we want this to go.

PR #627, opened 2015-06-30
BUG: makes is_market_hours() tz-aware

warren-oneill:
Timestamp is first converted to local time before being passed to other functions.


zipline/finance/trading.py
```diff
@@ -204,17 +204,18 @@ def normalize_date(self, test_date):
         return pd.tseries.tools.normalize_date(test_date)
 
     def utc_dt_in_exchange(self, dt):
-        return pd.Timestamp(dt).tz_convert(self.exchange_tz)
+        return pd.Timestamp(dt, tz='UTC').tz_convert(self.exchange_tz)
```
ssanderson: @warren-oneill from reading the name, I think the contract of this method is that it expects to be passed a UTC dt.  Is there an existing location where we're passing a possibly not UTC dt here?

zipline/finance/trading.py
```diff
@@ -204,17 +204,18 @@ def normalize_date(self, test_date):
...
 
     def exchange_dt_in_utc(self, dt):
         return pd.Timestamp(dt, tz=self.exchange_tz).tz_convert('UTC')
 
     def is_market_hours(self, test_date):
-        if not self.is_trading_day(test_date):
+        day = pd.Timestamp(self.utc_dt_in_exchange(test_date).date(), tz='UTC')
```
ssanderson: We're doing an awful lot of conversions here...what is this method doing that's currently undesired?

zipline/finance/trading.py
warren-oneill: @ssanderson removed

zipline/finance/trading.py
warren-oneill: what it's currently doing is taking open time and passing the open time to ```is_trading_day``` and ```get_open_and_close``` but open time can be on the previous when in UTC so you will be returned values for the previous day rather than the current day. I made a very similar PR last month #579.

warren-oneill: so basically what I need is the current day in the local time and what its using the is current day in UTC but the reason my solution is a bit hacked is that sometime the function receives market open and sometimes with day time 00:00 in UTC (and perhaps also market close). Otherwise if you were always receiving the same variable it would be easier to construct a neater solution. All of the time/day functions are in desperate need of standardization.

warren-oneill: me and @ehebert discussed this issue a bit in #579.

warren-oneill:
@ssanderson @ehebert any chance of getting this accepted? anything I need to change first?

warren-oneill:
@ssanderson @ehebert ping!

ssanderson:
@warren-oneill can you provide a concrete example of a TradingEnviromment setup and an input to `is_market_hours` that produces an incorrect answer here?  That would be useful both for my understanding of this change, and for ensuring that we don't accidentally break this down the line.

warren-oneill:
Sure, the issue arises when the timestamp passed to ```is_market_hours``` is on different days depending on if its is in UTC or local time.

In my case we have a 24/7 German market. The open time for, e.g., 01.01.2016 is 00:00 01.01.2016 in local time and 23:00 31.12.2015 in UTC since German time is +01:00. This means that when any timestamp within the first hour of trading is passed to ```get_open_close``` we will actually receive the open and close for the previous day and not the day we are interested in. This error comes from the first line of  ```get_open_and_close```:
```
index = self.open_and_closes.index.get_loc(day.date())
```
In UTC ```day.date()``` will give the previous day and as the result the function will also give the open and close for the previous day.

My solution is to first covert the timestamp pass to ```is_market_hours``` to the day in local time and then pass the day to ```get_open_and_close```. This way the open and close is returned for the correct day.

PR #628, opened 2015-06-30
MAINT: Move ZiplineAPI context so that it always wraps main loop.

ehebert:
Move the responsibility of wrapping the main simulation loop in the
ZiplineAPI context from the algorithm modules generator setup to the
main trade simulation loop, so that different methods of invoking loop
do not need to duplicate how the context is set.

To make it easier for internal implementation of handle_data to
transition off of calling the ZiplineAPI every bar, to only invoking the
context once per simulation.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -510,15 +510,14 @@ class TradingAlgorithm(object):
                 self.sim_params.data_frequency,
             )
 
-        with ZiplineAPI(self):
-            # loop through simulated_trading, each iteration returns a
-            # perf dictionary
-            perfs = []
-            for perf in self.gen:
-                perfs.append(perf)
+        # loop through simulated_trading, each iteration returns a
+        # perf dictionary
+        perfs = []
+        for perf in self.gen:
+            perfs.append(perf)
 
-            # convert perf dict to pandas dataframe
-            daily_stats = self._create_daily_stats(perfs)
+        # convert perf dict to pandas dataframe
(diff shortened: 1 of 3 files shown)


ehebert:
TravisCI failed because of missing contextlib2.

zipline/algorithm.py
```diff
@@ -510,15 +510,14 @@ def run(self, source, overwrite_sim_params=True,
...
-            for perf in self.gen:
-                perfs.append(perf)
-
-            # convert perf dict to pandas dataframe
-            daily_stats = self._create_daily_stats(perfs)
+        # loop through simulated_trading, each iteration returns a
+        # perf dictionary
+        perfs = []
+        for perf in self.gen:
+            perfs.append(perf)
```
richafrank: Definitely unrelated to this changeset, but given that we're looking at performance, I notice that this loop might be sped up by not appending each perf individually: ```perfs = list(self.gen)```

ssanderson: @richafrank better yet, just pass the generator straight to _create_daily_stats.

richafrank: :+1:

ehebert: A wrinkle here is that our internal code does not use this method, which is why I'm not looking at it this particular function too closely.

Though that adds another reason to the list for reasons to work on unification of 'how Zipline is run'.

ehebert:
Added contextlib2 to requirements on branch.

PR #629, opened 2015-06-30
MAINT: Consolidates and improves future lookup methods

yankees714:
Removes unused future lookup methods and consolidates everything into lookup_future_chain. Since the FutureChain object will have to hold a root symbol and dates, it should be responsible for cleaning the user input, so this is removed from the lookup method.

Adds knowledge date to future lookups. This makes our definition of valid contracts more flexible. We know about a contract if it starts trading by the knowledge date, and a contract is expired if it expires by the as_of_date.

Also fixes a bug with computing future chains, where contracts were not included in the chain on their expiration date.

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -29,6 +29,7 @@ from zipline.errors import (
     ConsumeAssetMetaDataError,
     InvalidAssetType,
     MultipleSymbolsFound,
+    RootSymbolNotFound,
     SidAssignmentError,
     SidNotFound,
     SymbolNotFound,
@@ -231,19 +232,7 @@ class AssetFinder(object):
         for root_symbol in self.future_chains_cache:
             self.future_chains_cache[root_symbol].sort(key=exp_key)
 
-    def _valid_contracts(self, root_symbol, as_of_date):
-        """ Returns  a list of the currently valid futures contracts
-        for a given root symbol, sorted by expiration date (the
-        contracts are sorted when the AssetFinder is built).
-        """
-        try:
-            return [c for c in self.future_chains_cache[root_symbol]
-                    if c.expiration_date and (c.expiration_date > as_of_date)
-                    and c.start_date and (c.start_date <= as_of_date)]
-        except KeyError:
-            return None
-
-    def lookup_future_chain(self, root_symbol, as_of_date):
+    def lookup_future_chain(self, root_symbol, as_of_date, knowledge_date):
         """ Return the futures chain for a given root symbol.
 
         Parameters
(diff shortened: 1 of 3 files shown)


yankees714:
@StewartDouglas would you be able to take a look at this?

StewartDouglas:
@yankees714 looking now

zipline/assets/assets.py
```diff
@@ -268,9 +268,10 @@ def lookup_future_chain(self, root_symbol, as_of_date):
         """
         root_symbol.upper()
         as_of_date = normalize_date(as_of_date)
-        return self._valid_contracts(root_symbol, as_of_date)
+        return self._valid_contracts(root_symbol, as_of_date, knowledge_date)
 
-    def lookup_future_in_chain(self, root_symbol, as_of_date, contract_num=0):
+    def lookup_future_in_chain(self, root_symbol, as_of_date, knowledge_date,
```
yankees714: Not sure if we should keep this around, since the FutureChain class should only have to call `lookup_future_chain`, and then work from that.

zipline/assets/assets.py
```diff
@@ -268,9 +268,10 @@ def lookup_future_chain(self, root_symbol, as_of_date):
         """
```
StewartDouglas: Might want to add a knowledge_date description to your docstring

zipline/assets/assets.py
StewartDouglas: Agreed that we don't need this since we'll be accessing specific futures using list notation

zipline/assets/assets.py
```diff
@@ -240,19 +240,19 @@ def _sort_future_chains(self):
...
+    def _valid_contracts(self, root_symbol, as_of_date, knowledge_date):
         """ Returns  a list of the currently valid futures contracts
         for a given root symbol, sorted by expiration date (the
         contracts are sorted when the AssetFinder is built).
         """
         try:
             return [c for c in self.future_chains_cache[root_symbol]
-                    if c.expiration_date and (c.expiration_date > as_of_date)
-                    and c.start_date and (c.start_date <= as_of_date)]
+                    if c.expiration_date and (as_of_date <= c.expiration_date)
```
ssanderson: When is it valid for a future to not have a start date or expiration date?

zipline/assets/assets.py
yankees714: Definitely, will do.

zipline/assets/assets.py
StewartDouglas: We talked about this briefly yesterday - the thinking was that if we don't know the expiration date for a future then we simply shouldn't record it (or perhaps not load it into the cache), so the ```c.expiration_date and``` part here shouldn't be necessary. @yankees714 does that sound right to you?

yankees714: That makes sense to me, and I guess the same would apply for `c.start_date and`. So should we be checking these fields as we're building the caches?

StewartDouglas: I would say yes as a future without a start_date or expiration_date isn't particularly helpful to our users. TBD whether I actually record them at all in our database.

jfkirk: The way that the AssetFinder is structured in Zipline (borrowed from Quantopian internal) is that all Asset fields are optional, except for SID (which can be assigned by the finder).

On Stewart's point, internal to Quantopian, we will have internal systems to guarantee that we are providing complete Asset metadata. Zipline, however, is built to function with whatever level of metadata the user provides.

zipline/assets/assets.py
jfkirk: Agreed.

zipline/assets/assets.py
StewartDouglas: Good point @jfkirk, that suggests to me that actually we should keep in the ```expiration_date``` and ```start_date``` clauses.

yankees714: In that case, I'll leave these checks as they are.

StewartDouglas:
Looks good to merge to me

jfkirk:
:+1: once Jenkins passes

PR #630, opened 2015-06-30
FFC API

ssanderson:
**NOTE:** Throughout this description I've added sections marked with **RFC**, meaning "Request For Comments", on locations where I'm particularly interested in feedback.

This patch lays the groundwork for a compute engine designed to facilitate construction of factor-based universe screening and portfolio allocation.

## Goals
1. Provide a high-level, declarative API for representing trailing window computations on large datasets.
2. Provide groundwork for a dynamic portfolio construction API using Filters and Factors.
3. Do 1 and 2 in way that's sufficiently abstracted from the underlying computation process that we can change the internals without breaking public APIs.
4. Be fast enough and memory efficient enough to support computation on universes with thousands of assets.

### `zipline.modelling`

Contains entities that can be used to express computations as dependency graphs.  Each node in such a graph is an instance of the base `Term` class, defined in `zipline.modelling.term`.  Dependency graphs are executed by instances of `FFCEngine`, defined in `zipline.modelling.engine`.  The only interesting implementation is `SimpleFFCEngine`, but there's also a `NoOpFFCEngine` which is used when no FFC Loaders can be found.

There are three important subclasses of `Term`:
* `Factor`, representing the results of numerical-valued computations.  Currently we only support `float64` as a dtype for Factor, but it would be mostly straightforward to add integral types.  Factors can be computed and referenced directly, and factors can be converted into `Filter`s via comparison operators (`>`, `>=`, `==`, etc).  There are also methods for more fine-grained statistical computations like percentile analysis.  In general, these methods work by producing new Term instances.  For example, `some_factor.percentile_between(10, 20)` returns an instance of `PercentileFilter`, which will perform the desired computation when actually evaluated by an engine.  There is experimental support for defining custom factors, and some examples have been added in `zipline.modelling.factors.technical`.
* `Filter`, representing the results of boolean-valued computations.  Filters are generally with method calls on an instance of `Factor`, but there's an experimental `CustomFilter` for direct expression of boolean-valued computations.

Most of the complexity of the `Term` class is dedicated to a mechanism that makes it so that calling a Term constructor twice with the same arguments will return the same instance.  (**RFC:** Does the implementation of this make sense to others.  Most of the explanation for how and why we're doing this is in `Term.__new__` and `Term.static_identity`.) This is valuable because it gives us common subexpression folding for free, but it adds some boilerplate to the construction of new Term subclasses.  This should only affect implementers of term combinators (i.e. functions that take terms and construct new terms) like `Rank` and `PercentileFilter`.  For users who just want to define new Filters/Factors with rolling computations, there are experimental `CustomFilter` and `CustomFactor` classes, which just require that the user implement a `compute` function that will be called on rolling windows of data.

### `zipline.data.ffc`
(**RFC**: I don't like this directory structure right now.  Should we just move it to zipline.modelling?  Somewhere else?)

Contains loaders and dataset definitions for inputs to the modelling API.  The abstract API for a modelling dataloader is defined by `FFCLoader` (**RFC**: I think we should remove references to FFC everywhere.  Thoughts on better names for this?)

New `TradingAlgorithm` api methods: `add_factor`, and `add_filter`.  These methods can only be called from `initialize`, and are used to inform the algorithm that each day it should compute the given terms.
Computed factor results are made available through a new `factors` property of the `data` object in `before_trading_start` and `handle_data`. Computed filter results control which assets are available in the factor matrix on each day.  An asset will only appear if it passed **all** registered filters.  (**RFC:**: is this the right combinatorial behavior for filters?  Should we even allow multiple filters to be registered?  Should `data.factors` be a method rather than a property so that the user can request values that passed a specific filters instead?  That would add a fairly significant performance burden.


ssanderson:
@ehebert @richafrank @llllllllll @jfkirk @jbredeche 

I'm still working on acceptance tests for this, but it's ready for review to be merged into Zipline.  This is a **big** merge, so I'm not sure how we want to do the review for this. I've tagged all the people who I would expect might want to provide input.

zipline/data/ffc/datasets/README.md
```diff
@@ -0,0 +1,5 @@
+Currently USEquityPricing is located in zipline.data.pricing
```
ssanderson: Note to self: delete this.

ssanderson:
Particular notes:

The two most important modules here are [zipline.modelling.term](https://github.com/quantopian/zipline/pull/630/files#diff-d6fad007686285fa3306ae5172444eb8R1) and [zipline.modelling.engine](https://github.com/quantopian/zipline/pull/630/files#diff-854c3b7a046119cd02f916975d0cad83R1).  It'd be great if the people I tagged could read at least those two files so they have a basic sense of how the core of this works.

@ehebert, @jfkirk, and maybe @yankees714 and @StewartDouglas  can you look at the changes in [assets.py](https://github.com/quantopian/zipline/pull/630/files#diff-e38a95b451336da451f69f2b60840da6R73) since you're actively overhauling that system?  The most important function is the new `lifetimes` method, which gives you a matrix of whether all the assets we know about existed between dates x and y.

@llllllllll can you look at [dataset.py](https://github.com/quantopian/zipline/pull/630/files#diff-1410479ae16100ce74322de4d91c739cR1) and take a deeper look at the caching constructor in term.py?

zipline/modelling/term.py
```diff
@@ -0,0 +1,294 @@
...
+
+        Caching previously-constructed Terms is useful because it allows us to
+        only compute equivalent sub-expressions once when traversing an FFC
+        dependency graph.
+
+        Caching previously-constructed Terms is **sane** because terms and
+        their inputs are both conceptually immutable.
+        """
+        if inputs is None:
+            inputs = tuple(cls.inputs)
```
llllllllll: no need to take the tuple of a tuple

.gitignore
```diff
@@ -57,4 +57,4 @@ docs/_build/*
 benchmarks.db
 
 # Vagrant temp folder
-.vagrant
+.vagrant
```
ssanderson: Note to self: fix this.

zipline/modelling/term.py
```diff
@@ -0,0 +1,294 @@
...
+            return cls._term_cache[identity]
+        except KeyError:
+            new_instance = super(Term, cls).__new__(cls)._init(
+                inputs=inputs,
+                window_length=window_length,
+                domain=domain,
+                dtype=dtype,
+                *args, **kwargs
+            )
+            cls._term_cache[identity] = new_instance
```
llllllllll: `new_instance = cls._term_cache[identity] = ...`

zipline/modelling/term.py
```diff
@@ -0,0 +1,294 @@
...
+class Term(object):
+    """
+    Base class for terms in an FFC API compute graph.
+    """
+    inputs = NotSpecified
+    window_length = NotSpecified
+    domain = None
+    dtype = float64
+
+    _term_cache = {}
```
llllllllll: do we care that the `Term` object will never be gc'd? should this maybe be a `WeakValueDict`?

zipline/modelling/term.py
ssanderson: User-defined Terms can put a list in the class scope.

ssanderson:
@richafrank I don't know if I have specific code for you to look at except maybe the sections marked RFC in the long description above.

zipline/modelling/term.py
llllllllll: ah, subclasses would overwrite the NotSpecified, I see.

zipline/modelling/term.py
```diff
@@ -0,0 +1,294 @@
...
+    nan,
+)
+
+from zipline.errors import (
+    InputTermNotAtomic,
+    TermInputsNotSpecified,
+    WindowLengthNotPositive,
+    WindowLengthNotSpecified,
+)
+from zipline.utils.lazyval import lazyval
```
llllllllll: hype

ssanderson:
@jfkirk it might also be good for you to read the through the stuff in zipline.data.ffc.loaders as we think about how we want to publicize Zipline's data ingestion APIs.

zipline/modelling/term.py
```diff
@@ -0,0 +1,294 @@
...
+        raise NotImplementedError()
+
+    def __repr__(self):
+        return (
+            "{type}({inputs}, window_length={window_length})"
+        ).format(
+            type=type(self).__name__,
+            inputs=self.inputs,
+            window_length=self.window_length,
+            domain=self.domain,
```
llllllllll: domain is unused

zipline/modelling/term.py
twiecki: Should these be relative imports? https://docs.python.org/3/reference/import.html#submodules

ssanderson: @twiecki what would you see as the benefit of relative imports here? IMO, relative imports are useful when you expect that any time you move a module, you'd also move the module from which you're importing.  e.g.  If i have
```
spam/
    __init__.py
    foo.py
    bar.py
```
and I expect that `foo.py` and `bar.py` will always be next to each other, then it's useful in `foo` to do something like:
```
from .bar import ...
```
because then if I rename the whole directory, I don't have to change foo.py at all.

Here, on the other hand, I don't think there's any expected relative relationship between `zipline.modelling.term` and `zipline.utils.lazyval`.  The most likely change is that term.py gets moved somewhere else, which would almost certainly break any relative import.

zipline/modelling/term.py
ssanderson: Hmm...I'm not sure.  The terms themselves have very little data on them, and their expected lifetime is more or less the lifetime of a process as well.  I'll think about it though.

zipline/modelling/engine.py
```diff
@@ -0,0 +1,443 @@
...
+            The first date of the matrix.
+        end : datetime
+            The last date of the matrix.
+
+        Returns
+        -------
+        matrix : pd.DataFrame
+            A matrix of factors
+        """
+        raise NotImplementedError()
```
llllllllll: I prefer to put the name of the not implemented function here to aid debugging. This is style.

zipline/modelling/engine.py
```diff
@@ -0,0 +1,443 @@
...
+        # We only need filters and factors to compute the final output matrix.
+        raw_filters = [lifetimes_between_dates.values]
+        raw_factors = []
+        factor_names = []
+        for name, term in iteritems(terms):
+            extra = extra_row_counts[term]
+            if isinstance(term, Factor):
+                factor_names.append(name)
+                # Remove extra rows.
+                if extra:
```
ssanderson: Welp, this is a dumb block...

zipline/modelling/engine.py
```diff
@@ -0,0 +1,443 @@
...
+    ]
+
+    def __init__(self, loader, calendar, asset_finder):
+        self._loader = loader
+        self._calendar = calendar
+        self._finder = asset_finder
+
+    def factor_matrix(self, terms, start_date, end_date):
+        """
+        Compute a factor matrix.
```
ssanderson: This probably needs a long description of the algorithm being implemented here.

zipline/modelling/engine.py
```diff
@@ -0,0 +1,443 @@
...
+            before `start_date`, continuing through to `end_date`.  The
+            returned frame contains as columns all assets in our AssetFinder
+            that existed for at least one day between `start_date` and
+            `end_date`.
+        """
+        calendar = self._calendar
+        finder = self._finder
+        start_idx, end_idx = self._calendar.slice_locs(start_date, end_date)
+        if start_idx < extra_rows:
+            # TODO: Use NoFurtherDataError from trading.py
```
llllllllll: should this be done?

zipline/modelling/engine.py
ssanderson: (This is supposed to remove the `[extra:]` slice so we don't make an extraneous copy when extra is 0.)

ssanderson:
More bikeshedding questions: should we continue to call the DataSet `USEquityPricing` as opposed to just `EquityPricing`.  I suspect that they'll always be the same data types, and baking `US` into the name seems like we're asking for a headache down the road.  @jfkirk?

zipline/assets/assets.py
```diff
@@ -264,25 +264,39 @@ def populate_cache(self):
         Populates the asset cache with all values in the assets
         collection.
         """
-
         # Wipe caches before repopulating
         self.cache = {}
```
ehebert: @ssanderson I think you need to rebase against master.
The `populate_cache` method has been removed.

zipline/modelling/expression.py
```diff
@@ -0,0 +1,310 @@
...
+
+    Returns the new tuple and the index of elem in the new tuple.
+    """
+    try:
+        return tup, tup.index(elem)
+    except ValueError:
+        return tuple(chain(tup, (elem,))), len(tup)
+
+
+def bad_op(op, left, right):
```
llllllllll: could this not be a subclass of `TypeError`?

ssanderson: Meaning something like:

```
class BadOp(TypeError):
    def __init__(self, op, left, right):
        super(BadOp, self).__init__("Can't compute...".format(...))
```
I could see that

zipline/algorithm.py
```diff
@@ -1193,6 +1216,49 @@ def set_long_only(self):
...
+    def add_filter(self, filter):
+        name = "anon_filter_%d" % len(self._filters)
+        self._filters[name] = filter
+
+    # Note: add_classifier is not yet implemented since you can't do anything
+    # useful with classifiers yet.
+
+    def _all_terms(self):
+        # Merge all three dicts.
+        return dict(
```
ehebert: Is this equivalent to the following?

```
all_terms = {}
all_terms.update(self._filters)
all_terms.update(self._factors)
all_terms.update(self._classifiers)
return all_terms
``

ssanderson: Yes.

twiecki:
:+1:

ehebert:
I second that :+1:

PR #631, opened 2015-06-30
PERF: Reduce time spent indexing in risk cumulative update.

ehebert:
Instead of using the pandas.Series datetime index for every single
vector, get the index at the beginning of the update loop based on the
dt and then use that index to set the values.

Also, since the dt lookup is no longer needed, store the values as numpy
arrays, which are more lightweight.

Locally, this patch cuts out about 60% of the time spent in the update
method.

### DIFF
--- a/zipline/finance/risk/cumulative.py
+++ b/zipline/finance/risk/cumulative.py
@@ -290,9 +290,10 @@ algorithm_returns ({algo_count}) in range {start} : {end} on {dt}"
             raise Exception(message)
 
         self.update_current_max()
-        self.metrics.benchmark_volatility.iloc[dt_loc] = \
+        metrics = self.metrics
+        metrics.benchmark_volatility.iloc[dt_loc] = \
             self.calculate_volatility(self.benchmark_returns)
-        self.metrics.algorithm_volatility.iloc[dt_loc] = \
+        metrics.algorithm_volatility.iloc[dt_loc] = \
             self.calculate_volatility(self.algorithm_returns)
 
         # caching the treasury rates for the minutely case is a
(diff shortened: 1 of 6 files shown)


ehebert:
@ssanderson could you look at this one?

zipline/finance/risk/cumulative.py
```diff
@@ -141,15 +141,17 @@ def __init__(self, sim_params,
             cont_index = self.get_minute_index(sim_params)
 
         self.cont_index = cont_index
+        self.cont_len = len(self.cont_index)
```
ssanderson: Returns frequency should never be 'minute' (noticed by eddie over my shoulder).

zipline/finance/risk/cumulative.py
```diff
@@ -141,15 +141,17 @@ def __init__(self, sim_params,
...
 
-        self.algorithm_returns_cont = pd.Series(index=cont_index)
-        self.benchmark_returns_cont = pd.Series(index=cont_index)
-        self.algorithm_cumulative_leverages_cont = pd.Series(index=cont_index)
-        self.mean_returns_cont = pd.Series(index=cont_index)
-        self.annualized_mean_returns_cont = pd.Series(index=cont_index)
-        self.mean_benchmark_returns_cont = pd.Series(index=cont_index)
-        self.annualized_mean_benchmark_returns_cont = pd.Series(
-            index=cont_index)
+        empty_cont = np.empty(self.cont_len) * np.nan
```
ssanderson: `np.full`

zipline/finance/risk/cumulative.py
```diff
@@ -329,24 +329,29 @@ def to_dict(self):
         Returns a dict object of the form:
         """
         dt = self.latest_dt
+        dt_loc = self.latest_dt_loc
         period_label = dt.strftime("%Y-%m")
```
ssanderson: Caching self.metrics here would save us a dozen or so attribute accesses.

zipline/finance/risk/cumulative.py
```diff
@@ -309,19 +308,20 @@ def update(self, dt, algorithm_returns, benchmark_returns, account):
...
+        self.excess_returns[dt_loc] = (
+            self.algorithm_cumulative_returns[dt_loc] -
             self.treasury_period_return)
-        self.metrics.beta[dt] = self.calculate_beta()
-        self.metrics.alpha[dt] = self.calculate_alpha()
-        self.metrics.sharpe[dt] = self.calculate_sharpe()
-        self.metrics.downside_risk[dt] = self.calculate_downside_risk()
-        self.metrics.sortino[dt] = self.calculate_sortino()
-        self.metrics.information[dt] = self.calculate_information()
+        self.metrics.beta.values[dt_loc] = self.calculate_beta()
```
ssanderson: We might want to cache self.metrics here as well.

zipline/finance/risk/cumulative.py
```diff
@@ -291,9 +290,9 @@ def update(self, dt, algorithm_returns, benchmark_returns, account):
             raise Exception(message)
 
         self.update_current_max()
-        self.metrics.benchmark_volatility[dt] = \
+        self.metrics.benchmark_volatility.values[dt_loc] = \
```
ssanderson: My guess is this will be faster as a .iloc __setitem__.

zipline/finance/risk/cumulative.py
ehebert: Converting over.

zipline/finance/risk/cumulative.py
ehebert: I"ll do that, but in a separate patch.

zipline/finance/risk/cumulative.py
ehebert: You're right, as well as prevent the case of accidentally creating copy as we had discussed.

ehebert:
@ssanderson I force pushed your recommendations and am merging.

PR #633, opened 2015-07-01
Move fill_from_saved_state to Blotter

lianga888:


lianga888:
Moving the method to AlgorithmProxy instead

PR #635, opened 2015-07-06
TST: Remove asset finder test for NASDAQ collisions.

ehebert:
The asset finder retrieved from the test environment is empty, so the
test does not end up testing anything, since the test cases loop over
the empty list of sids in the asset finder.

Remove to possibly be added back in and re-implemented after a larger
refactoring of the module.

# Other Notes

Discovered this broken test while working on a sqlite backed AssetFinder.

@jfkirk what do you think about removing this test for now, and then fixing it (or not adding it back in if we think the fuzzy tests suffice) after the change over?

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -25,14 +25,12 @@ from datetime import (
     datetime
 )
 import pickle
-import pprint
 import uuid
 import warnings
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


ehebert:
Just force pushed a fixup for unused imports the linter found.

jfkirk:
:+1:

PR #637, opened 2015-07-07
ENH: Futures API

yankees714:
Implements a user facing futures API for zipline.

Adds the `FutureChain` object, which allows users to access future contracts though underlying queries to the AssetFinder. Adds the API function `future_chain`, which returns `FutureChain` objects.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -37,6 +37,7 @@ from zipline.errors import (
     TradingControlViolation,
     AccountControlViolation,
     SymbolNotFound,
+    RootSymbolNotFound,
 )
 from zipline.test_algorithms import (
     access_account_in_init,
@@ -334,6 +335,74 @@ class TestMiscellaneousAPI(TestCase):
         # Test lookup SID
         self.assertIsInstance(algo.sid(0), Equity)
         self.assertIsInstance(algo.sid(1), Equity)
+
+    def test_future_chain(self):
+        """ Tests the future_chain API function.
+        """
+
+        metadata = {
+            0: {
+                'symbol': 'CLG06',
+                'root_symbol': 'CL',
+                'asset_type': 'future',
+                'start_date': pd.Timestamp('2005-12-01', tz='UTC'),
+                'notice_date': pd.Timestamp('2005-12-20', tz='UTC'),
+                'expiration_date': pd.Timestamp('2006-01-20', tz='UTC')},
+            1: {
+                'root_symbol': 'CL',
+                'symbol': 'CLK06',
+                'asset_type': 'future',
+                'start_date': pd.Timestamp('2005-12-01', tz='UTC'),
+                'notice_date': pd.Timestamp('2006-03-20', tz='UTC'),
+                'expiration_date': pd.Timestamp('2006-04-20', tz='UTC')},
+            2: {
+                'symbol': 'CLQ06',
+                'root_symbol': 'CL',
+                'asset_type': 'future',
+                'start_date': pd.Timestamp('2005-12-01', tz='UTC'),
+                'notice_date': pd.Timestamp('2006-06-20', tz='UTC'),
+                'expiration_date': pd.Timestamp('2006-07-20', tz='UTC')},
+            3: {
+                'symbol': 'CLX06',
+                'root_symbol': 'CL',
+                'asset_type': 'future',
+                'start_date': pd.Timestamp('2006-02-01', tz='UTC'),
+                'notice_date': pd.Timestamp('2006-09-20', tz='UTC'),
+                'expiration_date': pd.Timestamp('2006-10-20', tz='UTC')}
+        }
+
+        algo = TradingAlgorithm(asset_metadata=metadata)
+        algo.datetime = pd.Timestamp('2006-12-01', tz='UTC')
+
+        # Check that the fields of the FutureChain object are set correctly
+        cl = algo.future_chain('CL')
+        self.assertEqual(cl.root_symbol, 'CL')
+        self.assertEqual(cl.as_of_date, algo.datetime)
+
+        # Check the fields are set correctly if an as_of_date is supplied
+        as_of_date = pd.Timestamp('1952-08-11', tz='UTC')
+
+        cl = algo.future_chain('CL', as_of_date=as_of_date)
+        self.assertEqual(cl.root_symbol, 'CL')
+        self.assertEqual(cl.as_of_date, as_of_date)
+
+        cl = algo.future_chain('CL', as_of_date='1952-08-11')
+        self.assertEqual(cl.root_symbol, 'CL')
+        self.assertEqual(cl.as_of_date, as_of_date)
+
+        # Check that weird capitalization is corrected
+        cl = algo.future_chain('cL')
+        self.assertEqual(cl.root_symbol, 'CL')
+
+        cl = algo.future_chain('cl')
+        self.assertEqual(cl.root_symbol, 'CL')
+
+        # Check that invalid root symbols raise RootSymbolNotFound
+        with self.assertRaises(RootSymbolNotFound):
+            algo.future_chain('CLZ')
+
+        with self.assertRaises(RootSymbolNotFound):
+            algo.future_chain('')
 
 
 class TestTransformAlgorithm(TestCase):
--- a/zipline/assets/futures.py
+++ b/zipline/assets/futures.py
@@ -1,0 +1,169 @@
+from pandas import Timestamp, Timedelta
+from pandas.tseries.tools import normalize_date
+
+
+class FutureChain(object):
+    """ Allows users to look up future contracts.
+
+    Parameters
+    ----------
+    asset_finder : AssetFinder
+        An AssetFinder for future contract lookups, in particular the
+        AssetFinder of the TradingAlgorithm instance.
+    get_datetime : function
+        A function that returns the simulation datetime, in particular
+        the get_datetime method of the TradingAlgorithm instance.
+    root_symbol : str
+        The root symbol of a future chain.
+    as_of_date : pandas.Timestamp, optional
+        Date at which the chain determination is rooted. I.e. the
+        existing contract whose notice date is first after this date is
+        the primary contract, etc. If not provided, the current
+        simulation date is used as the as_of_date.
+
+    Attributes
+    ----------
+    root_symbol : str
+        The root symbol of the future chain.
+    as_of_date
+        The current as-of date of this future chain.
+
+    Methods
+    -------
+    as_of(dt)
+    offset(time_delta)
+
+    Raises
+    ------
+    RootSymbolNotFound
+        Raised when the FutureChain is initialized with a root symbol for which
+        a future chain could not be found.
+    """
+    def __init__(self, asset_finder, get_datetime, root_symbol,
+                 as_of_date=None):
+        self.root_symbol = root_symbol
+
+        # Reference to the algo's AssetFinder for contract lookups
+        self._asset_finder = asset_finder
+        # Reference to the algo's get_datetime to know the current dt
+        self._algorithm_get_datetime = get_datetime
+
+        # If an as_of_date is provided, self._as_of_date uses that
+        # value, otherwise None. This attribute backs the as_of_date property.
+        if as_of_date:
+            self._as_of_date = normalize_date(Timestamp(as_of_date, tz='UTC'))
+        else:
+            self._as_of_date = None
+
+        # Attribute to cache the most up-to-date chain, and the dt when it was
+        # last updated.
+        self._current_chain = []
+        self._last_updated = None
+
+        # Get the initial chain, since self._last_updated is None.
+        self._maybe_update_current_chain()
+
+    def __repr__(self):
+        # NOTE: The string returned cannot be used to instantiate this
+        # exact FutureChain, since we don't want to display the asset
+        # finder and get_datetime function to the user.
+        if self._as_of_date:
+            return "FutureChain(root_symbol='%s', as_of_date='%s')" % (
+                self.root_symbol, self.as_of_date)
+        else:
+            return "FutureChain(root_symbol='%s')" % self.root_symbol
+
+    def _get_datetime(self):
+        """
+        Returns the normalized simulation datetime.
+
+        Returns
+        -------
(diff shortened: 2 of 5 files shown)


yankees714:
@jfkirk I still need to flesh out some of the tests on this, but it'd be great to have your input on where things are right now.

tests/test_algorithm.py
```diff
@@ -1351,3 +1352,212 @@ def check_algo_positions(self, results, expected_positions):
...
+
+        algo = TradingAlgorithm(asset_metadata=self.metadata,
+                                initialize=initialize,
+                                handle_data=handle_data,
+                                sim_params=self.sim_params_50)
+        algo.run(self.source_50)
+
+    def test_root_symbols(self):
+        """ Test that different variations on root symbols are handled
+        as expcted.
```
jfkirk: typo

tests/test_algorithm.py
```diff
@@ -1351,3 +1352,212 @@ def check_algo_positions(self, results, expected_positions):
...
+        """
+        def initialize(algo):
+            algo.cl = algo.future_chain('CL')
+
+            # Test that offset sets as_of_date as expected
+            self.assertEqual(
+                algo.cl.offset('3 days').as_of_date,
+                (pd.tseries.tools.normalize_date(algo.sim_params.period_start)
+                 + pd.Timedelta('3 days'))
+            )
```
jfkirk: Should we test that offset works both forwards and backwards in time?

tests/test_algorithm.py
```diff
@@ -1351,3 +1352,212 @@ def check_algo_positions(self, results, expected_positions):
...
+        self.source_50 = factory.create_minutely_trade_source(
+            sids,
+            sim_params=self.sim_params_50,
+            concurrent=True,
+        )
+
+    def tearDown(self):
+        teardown_logger(self)
+
+    def test_len(self):
```
ssanderson: @yankees714 why do we need to be running full simulations to test this functionality?

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+    as_of_date : pandas.Timestamp
+    """
+
+    def __init__(self, trading_algorithm, root_symbol='', as_of_date=None):
+        self.root_symbol = root_symbol.upper()
+
+        # If an as_of_date is provided, self._as_of_date uses that
+        # value, otherwise None. This attribute backs the as_of_date property.
+        if as_of_date:
+            self._as_of_date = (TradingEnvironment.instance().normalize_date(
```
ssanderson: You don't need the TradingEnvironment to do this.  TradingEnvironment.normalize_date just does:
```
    def normalize_date(self, test_date):
        test_date = pd.Timestamp(test_date, tz='UTC')
        return pd.tseries.tools.normalize_date(test_date)
```

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+    """ Allows users to look up future contracts.
+
+    Parameters
+    ----------
+    trading_algorithm : TradingAlgorithm
+    root_symbol : str, optional
+    as_of_date : pandas.Timestamp
+    """
+
+    def __init__(self, trading_algorithm, root_symbol='', as_of_date=None):
```
jfkirk: When would root_symbol ever be ''?

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+        try:
+            dt = self._trading_algorithm.get_datetime()
+        except AttributeError:
+            dt = self._trading_algorithm.sim_params.period_start
+
+        # If the algo does not yet have a datetime, use the period start
+        if dt is None:
+            dt = self._trading_algorithm.sim_params.period_start
+
+        return TradingEnvironment.instance().normalize_date(dt)
```
ssanderson: c.f. comment above

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+        # value, otherwise None. This attribute backs the as_of_date property.
+        if as_of_date:
+            self._as_of_date = (TradingEnvironment.instance().normalize_date(
+                                as_of_date))
+        else:
+            self._as_of_date = None
+
+        # Maintain a reference to the TradingAlgorithm which holds this
+        # FutureChain instance, so we can get its datetime.
+        self._trading_algorithm = trading_algorithm
```
jfkirk: This may get real interesting with serialization. I don't know how we can improve this situation, but if there is one thing I'd like to change, it is how we hook-in to the algo's dt.

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+
+            Returns
+            -------
+            [Future]
+                The up-to-date current chain.
+        """
+        if (self._last_updated is None or
+           (self._last_updated != self._get_datetime())):
+
+            asset_finder = TradingEnvironment.instance().asset_finder
```
ssanderson: Can we just pass in the asset_finder on construction rather than pulling it out of a global every time this is called?

tests/test_algorithm.py
yankees714: @ssanderson I think it makes sense to do at least some of it this way, since a lot of things are dependent on dates getting set (e.g. caching the current chain, getting the initial date from the trading algorithm). But I agree that a lot of this could be done outside of a simulation

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+            dt = self._trading_algorithm.sim_params.period_start
+
+        # If the algo does not yet have a datetime, use the period start
+        if dt is None:
+            dt = self._trading_algorithm.sim_params.period_start
+
+        return TradingEnvironment.instance().normalize_date(dt)
+
+    @property
+    def as_of_date(self):
```
ssanderson: Having both an as_of_date property and an as_of method is a little confusing.

zipline/assets/futures.py
ssanderson: I'm **strongly** of the opinion that we shouldn't be passing around full and storing a full TradingAlgorithm instance here.  Doing so makes this class much harder to test, and effectively makes all of Zipline a dependency.

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Parameters
+    ----------
+    trading_algorithm : TradingAlgorithm
+    root_symbol : str, optional
+    as_of_date : pandas.Timestamp
+    """
+
+    def __init__(self, trading_algorithm, root_symbol='', as_of_date=None):
+        self.root_symbol = root_symbol.upper()
```
ssanderson: This conversion should probably be done by the context that constructs the FutureChain, not by the chain itself.

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class FutureChain(object):
+    """ Allows users to look up future contracts.
+
+    Parameters
+    ----------
+    trading_algorithm : TradingAlgorithm
+    root_symbol : str, optional
+    as_of_date : pandas.Timestamp
```
jfkirk: Can you add quick descriptions of the parameters?

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+        # last updated.
+        self._current_chain = []
+        self._last_updated = None
+
+        # Get the initial chain, since self._last_updated is None.
+        self._maybe_update_current_chain()
+
+    def __repr__(self):
+        if self._as_of_date:
+            return 'FutureChain(root_symbol=\'%s\', as_of_date=\'%s\')' % (
```
ssanderson: Rather than escape the single quotes here, you should just use double quotes around the outer string.

```
In [6]: "This is a string with 'quotes'"
Out[6]: "This is a string with 'quotes'"
```

zipline/assets/futures.py
```diff
@@ -0,0 +1,128 @@
...
+                self.root_symbol, self.as_of_date)
+        else:
+            return 'FutureChain(root_symbol=\'%s\')' % self.root_symbol
```
ssanderson: ditto

tests/test_algorithm.py
```diff
@@ -1351,3 +1352,212 @@ def check_algo_positions(self, results, expected_positions):
...
+        self.sim_params_2 = factory.create_simulation_parameters(num_days=2)
+        self.source_2 = factory.create_minutely_trade_source(
+            sids,
+            sim_params=self.sim_params_2,
+            concurrent=True,
+        )
+
+        # Params and trade source for a 50 day simulation
+        self.sim_params_50 = factory.create_simulation_parameters(num_days=50)
+        self.source_50 = factory.create_minutely_trade_source(
```
ssanderson: Independently of whether we decide to test FutureChain by running full simulations (I'm not convinced that's necessary), a 50 day simulation at minutely frequency is **WAY** overkill.  The speed of the test suite matters a lot for developer productivity; we should be doing the minimum amount of work required to fully test the desired functionality.

zipline/assets/futures.py
yankees714: @jfkirk I think we had discussed at some point that root_symbol should be a keyword arg. Maybe it should just be positional, since there's really no sensible default case?

tests/test_algorithm.py
yankees714: I'm making most of these directly test the FutureChain object, outside of simulation. Then we'll just need a few tests for the actual future_chain API function.

zipline/assets/futures.py
jfkirk: It should be a kwarg, as it is, but I don't think it should have a default value. Failing to pass a root_symbol should be a failure case.

yankees714:
@jfkirk @ssanderson I cleaned things up a bunch based on your suggestions, also improved the tests. Mind giving this a look again?

zipline/assets/futures.py
```diff
@@ -0,0 +1,165 @@
...
+        -------
+        pandas.Timestamp
+            The normalized datetime of FutureChain's TradingAlgorithm.
+        """
+        return normalize_date(
+            Timestamp(self._algorithm_get_datetime(), tz='UTC')
+        )
+
+    @property
+    def as_of_date(self):
```
yankees714: Do we think having this property and the `as_of(time_delta)` method is confusing enough to merit a name change? I think we should keep the `as_of` syntax  for the user i.e. `future_chain('CL').as_of(...)`.

jfkirk: Not to back-track the conversation, but I'm not totally clear on why we support this pattern:
```future_chain('CL').as_of(dt)```

When, it my mind, this pattern is clearer to me:
```future_chain('CL', as_of=dt)```

Comment (yankees714) [inline zipline/assets/futures.py]: I think the idea is that you'd be able to do something like:
```python
def initialize(context):
    context.cl = future_chain('CL')

    # ...
    cl.as_of(...)...
```

zipline/assets/futures.py
```diff
@@ -0,0 +1,165 @@
...
+        Date at which the chain determination is rooted. I.e. the
+        existing contract that expires first after (or on) this date is
+        the primary contract, etc. If not provided, the current
+        simulation date is used as the as_of_date.
+
+    Attributes
+    ----------
+    root_symbol : str
+        The root symbol of the future chain.
+    as_of_date
```
jfkirk: unfinished comment?

yankees714: Since this is a property and has it's own docstring the numpy conventions suggest to just include it by name
https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#class-docstring

zipline/assets/futures.py
yankees714: cc: @ssanderson

zipline/assets/futures.py
ssanderson: I think a one-line summary would be fine here too.

zipline/assets/futures.py
```diff
@@ -0,0 +1,170 @@
...
+
+    Methods
+    -------
+    as_of(dt)
+    offset(time_delta)
+
+    Raises
+    ------
+    RootSymbolNotFound
+        If the FutureChain is initialized with a root symbol for which
```
ssanderson: Copywriting nit: I think I have a slight preference for "Raised when <condition>" rather than "If <condition" for the opening of the sentence.

:+1: for the completeness of this docstring!

zipline/assets/futures.py
```diff
@@ -0,0 +1,170 @@
...
+        # Attribute to cache the most up-to-date chain, and the dt when it was
+        # last updated.
+        self._current_chain = []
+        self._last_updated = None
+
+        # Get the initial chain, since self._last_updated is None.
+        self._maybe_update_current_chain()
+
+    def __repr__(self):
+        """ NOTE: This is really only a "pseudo" repr, in that the
```
ssanderson: This probably should be a comment, not a docstring.  
FWIW, the property that `eval(repr(obj)) == obj` isn't so ubiquitous that it's a big deal for it not to hold here.

zipline/assets/futures.py
```diff
@@ -0,0 +1,170 @@
...
+        else:
+            return self._get_datetime()
+
+    def _maybe_update_current_chain(self):
+        """ Updates the current chain if it's out of date, then returns
+            it.
+
+            Returns
+            -------
+            [Future]
```
ssanderson: I'm a little confused by the brackets here.  Is this indicating "List of Future"?

zipline/assets/futures.py
ssanderson: @jfkirk the example @yankees714 posted is what I had in mind.  Being able to say "interpret FutureChain X in another `as_of`" seems like a useful idiom.

zipline/assets/futures.py
yankees714: Agreed, that's a bit nicer.

zipline/assets/futures.py
yankees714: Yeah, I'll just change the type here to `list` and indicate in the description that it's a list of futures.

zipline/assets/futures.py
jfkirk: Yep - makes sense. Thanks!

jfkirk:
Great work! :+1: Once the Jenkins gods have been appeased.

PR #638, opened 2015-07-07
BUG: Sets the datetime of TradingAlgorithm initially to the simulation period start

yankees714:
The datetime attribute of TradingAlgorithm was initially None, so calling get_datetime in initialize was causing an unhandled exception. This commit addresses that issue by calling on_dt_changed when initializing the TradingAlgorithm, to force all datetimes to be in line with the period_start.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -156,8 +156,6 @@ class TradingAlgorithm(object):
                 Any asset identifiers that are not provided in the
                 asset_metadata, but will be traded by this TradingAlgorithm
         """
-        self.datetime = None
-
         self.sources = []
 
         # List of trading controls to be used to validate orders.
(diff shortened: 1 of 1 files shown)


zipline/algorithm.py
```diff
@@ -883,7 +883,12 @@ def get_datetime(self, tz=None):
         """
         Returns the simulation datetime.
         """
-        dt = self.datetime
+        # If datetime is not yet set, use the period start.
+        if self.datetime is not None:
```
ssanderson: Can we just set self.datetime to period_start at construction time so that this is never None?

yankees714: @ssanderson good point, makes sense to me.

zipline/algorithm.py
```diff
@@ -208,6 +208,9 @@ def __init__(self, *args, **kwargs):
         if not self.blotter:
             self.blotter = Blotter()
 
+        # Set the dt initally to the period start by forcing it to change
+        self.on_dt_changed(self.sim_params.period_start)
+
```
jfkirk: This makes the self.datetime assignment on line 160 redundant, so you can delete it.

yankees714: @jfkirk removed that initial assignment.

jfkirk:
:+1: Once Jenkins passes

PR #640, opened 2015-07-08
ENH: Adds the full Asset objects to the kwargs of MultipleSymbolsFound

jfkirk:
It is not necessary to pass the possible matches as a str, so this change passes the list as-is.


jfkirk:
Merged!

PR #641, opened 2015-07-08
Remove uses of lookup_generic

ehebert:
Changes two uses of lookup_generic to use the exact lookup in the context of each call.

1) TradingAlgorithm.symbol e.g. `order(symbol('FOO`), 100)`

2) SpecificEquitySource's use of the asset_finder to ensure that sids have been inserted into the asset finder.

The main goal here is to reduce some complexity before a larger wholesale change to the class, and bring usage of lookup_generic back down to just the consumers that require the multiple lookup.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -597,13 +597,19 @@ class TestAlgoScript(TestCase):
 
     def test_api_get_environment(self):
         platform = 'zipline'
+        metadata = {0: {'symbol': 'TEST',
+                        'asset_type': 'equity'}}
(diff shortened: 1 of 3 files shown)


ehebert:
cc: @jfkirk and @ssanderson

jfkirk:
:+1:

ehebert:
I"m investigating the TravisCI errors now.

ehebert:
@jfkirk turns out I had only run on `tests_algorithm.py`. I'm going to remove the second patch for now, since `test_security_list` currently depends on lookup_generic since SpecficEquitySource does use symbols.

PR #642, opened 2015-07-08
MAINT: Move possible side assignment to insert metadata method.

ehebert:
The previous workflow was:

- insert basic metadata for all assets

- iterate over all metadata entries. Create a sid if one does not
exist

- call insert_metadata again to update the assigned sid value.

Instead, create a sid, if missing on the initial metadata assignment and
remove the second pass of calling insert_metadata.

Also, since the sid assignment code is only done in this one context,
inline the sid assignment code so that there is less code step through
while looking for where sid assignment is done.

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -85,19 +85,6 @@ class AssetFinder(object):
 
         self.populate_cache()
 
-    def _next_free_sid(self):
-        if len(self.cache) > 0:
-            return max(self.cache.keys()) + 1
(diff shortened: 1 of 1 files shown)


ehebert:
cc: @jfkirk 

This is chunked off from a larger change to the dichotomy of 'using a pre-defined asset finder' versus 'assigning sids on the fly.'.

jfkirk:
:+1: Makes sense

PR #643, opened 2015-07-08
Reduce surface area of AssetFinder behavior

ehebert:
In preparation of the port of the AssetFinder over to being backed by a sqlite file, remove some methods of AssetFinder, so that they do not need to be ported.

These are not absolute requirements to that port, but should help future changes be simpler, especially in the case of making the 'add assets from the DataFrame index' separate from the rest of the class.

Methods removed:
- consume_identifiers, insert_metadata and consume_metadata encompasses this behavior.
- map_identifier_index_to_sids, changed to a function outside of the AssetFinder which takes an asset finder as a parameter. It is a convenience function, but not core functionality, so moving it away from the core pieces before those pieces are broken up into a AssetSQLiteWriter and AssetFinder.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -152,9 +152,6 @@ class TradingAlgorithm(object):
                 If an object with a 'read' property is provided, 'read' must
                 return rows containing at least one of 'sid' or 'symbol' along
                 with the other metadata fields.
-            identifiers : List
-                Any asset identifiers that are not provided in the
-                asset_metadata, but will be traded by this TradingAlgorithm
         """
         self.sources = []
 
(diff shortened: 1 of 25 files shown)


ehebert:
cc: @jfkirk and @ssanderson 

As mentioned in the PR commit, this does introduce a little bit of jankiness around having to call populate_cache, but that should go away with the bigger change that is building on top of this.

zipline/algorithm.py
```diff
@@ -464,25 +464,38 @@ def run(self, source, overwrite_sim_params=True,
...
-            copy_frame.columns = \
-                self.asset_finder.map_identifier_index_to_sids(
-                    source.columns, source.index[0]
-                )
+            asset_sids = populate_finder_from_identifier_index(
+                self.asset_finder, source.columns, source.index[0])
+            copy_frame.columns = asset_sids
+            for sid in copy_frame.columns:
+                self.asset_finder.insert_metadata(sid)
+            self.asset_finder.populate_cache()
```
jfkirk: Why is it necessary to both run pop_finder_from_ident_index as well as insert_metadata?

ehebert:
Closing this since the changes for the SQLite ended up not needing these changes.

PR #644, opened 2015-07-08
BUG: Prevents duplicate dividends from being added to the perf tracker

jfkirk:
At the end of a call to update_dividends(), the perf tracker now performs a drop_duplicates on itself. Duplicates are identified by the 'sid' and 'pay_date' fields.


jfkirk:
@brianpfink For awareness - this may impact your work.

ehebert:
Looks good to me. /signoff

PR #645, opened 2015-07-09
BUG: Determine valid future contracts with notice date

yankees714:
Since most brokers will cease accepting trades by the notice date, contracts should not be considered valid after the notice date. This commit adjusts the lookup_future_chain method to consider all contracts with notice dates on or following the current date invalid.


yankees714:
@jfkirk This makes the switch to using notice date in the futures functionality that's already been added, would you be able to review?

jfkirk:
:+1: After Jenkins passes

PR #646, opened 2015-07-10
TST: Remove test for lookup of future contract by symbol.

ehebert:
The lookup of future contract by individual symbol is a constraint on
incoming changes of changing how the asset finder stores data.

(i.e. the asset finder is changing so that there are separate tables for
both futures and equities.)

Since this lookup is not yet fully supported, we can add it back in on
top of the new asset finder.

## Other Notes

@jfkirk this is the small reduction in features of the asset finder that we discussed in person. It should be possible to add the lookup back in, but since `symbol('AFUTURE')` is not yet used, it was easier to do the port of the underlying engine without supporting the feature, but shouldn't be too much work to add back in once that port is finished.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -304,49 +304,38 @@ class TestMiscellaneousAPI(TestCase):
                     1: {'symbol': 'PLAY',
                         'asset_type': 'equity',
                         'start_date': '2005-01-01',
-                        'end_date': '2006-01-01'},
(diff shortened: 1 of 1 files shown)


jfkirk:
Makes sense to me. I expect that, at some juncture, we will want to support future look-up by symbol, but it is not critical ATM.

:+1:

ehebert:
Thanks for the quick review! Merging.

PR #647, opened 2015-07-10
PERF: Change asset finder to be backed by sqlite3.

ehebert:
Attack the startup bottleneck of creating the asset finders caches for a
large universe, which was between 1-2 seconds on development and
production machines.

Instead, allow the AssetFinder to be passed a sqlite3 file that has
already been populated and then hydrate asset objects only when an
equity is referenced for the first time.

To create aforementioned sqlite3, create an AssetFinder with an db_path
and `create_table` set to True. If `create_table` is set to False, the
prepopulated data in the sqlite file found at db_path will be used.

Default behavior is to use an in memory database.

Behavior that changes:

- Fuzzy lookup now only works on one character, that character needs to be
specified at write/metadata consumption time, since the fuzzy lookup key
is created by dropping the character from each symbol.

- Overwriting partially written metadata is no longer
  supported. i.e. some unit tests allowed for inserting just the identifier,
  and then later updating the symbol, end_date, etc.

  Instead of building an upsert behavior at this time, this patch
  changes the unit tests so that the data for each asset is only
  inserted once.

Other notes:

- populate_cache is now removed, since there is no longer a two step
  process of inserting metadata and then realizing that metadata into
  assets. _spawn_asset is rolled into insert_metadata, so that a call to
  insert_metadata both converts the metadata and makes it available in
  the data store.

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -14,12 +14,12 @@
 # limitations under the License.
 
 from abc import ABCMeta
-from itertools import chain
 from numbers import Integral
 import numpy as np
-import operator
+import sqlite3
 import warnings
 
+from functools32 import lru_cache
 from logbook import Logger
 import pandas as pd
 from pandas.tseries.tools import normalize_date
@@ -29,7 +29,6 @@ from zipline.errors import (
     ConsumeAssetMetaDataError,
     InvalidAssetType,
     MultipleSymbolsFound,
-    RootSymbolNotFound,
     SidAssignmentError,
     SidNotFound,
     SymbolNotFound,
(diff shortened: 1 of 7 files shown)


ehebert:
A few more thoughts, I had initially conceived of breaking out an AssetSQLWriter from the AssetFinder, but left the class as unit that both reads and writes the asset data in large part for easier compatibility with both the the "create assets from DataFrame index" workflow and the reliance on insert_metadata and other functions in many unit tests.

We may want to splice the behavior apart after this patch.

zipline/assets/assets.py
```diff
@@ -96,47 +223,46 @@ def retrieve_asset(self, sid, default_none=False):
...
-        # If multiple SIDs exist for symbol, return latest start_date with
-        # end_date as a tie-breaker
-        if len(candidates) > 1:
-            best_candidate = sorted(
-                candidates,
-                key=lambda x: (x.start_date, x.end_date)
-            )[-1]
-            return best_candidate, True
+    @lru_cache(maxsize=None)
+    def equity_for_id(self, sid):
```
ssanderson: Are the names here and above intentionally using `id` sometimes and `sid` other times?

zipline/assets/assets.py
```diff
@@ -63,32 +62,160 @@
...
 
-        # The AssetFinder also holds a nested-dict of all metadata for
-        # reference when building Assets
-        self.metadata_cache = {}
-        if metadata is not None:
-            self.consume_metadata(metadata)
-
-        self.populate_cache()
-
+        if allow_sid_assignment:
```
jfkirk: Do we have any contexts in-which this is False? We have it as True by default. The variable itself may just be a crufty artifact.

zipline/assets/assets.py
```diff
@@ -186,38 +357,34 @@ def lookup_symbol(self, symbol, as_of_date, fuzzy=None):
...
-    def _sort_future_chains(self):
-        """ Sort by increasing expiration date the list of contracts
-        for each root symbol in the future cache.
-        """
-        notice_key = operator.attrgetter('notice_date')
-
-        for root_symbol in self.future_chains_cache:
-            self.future_chains_cache[root_symbol].sort(key=notice_key)
+            c = self.conn.cursor()
+            c.row_factory = dict_factory
```
ssanderson: Is there a reason we're using this over [sqlite3.Row](https://docs.python.org/2/library/sqlite3.html#row-objects)?

zipline/assets/assets.py
```diff
@@ -96,47 +223,46 @@ def retrieve_asset(self, sid, default_none=False):
...
+        if data:
+            if data['start_date']:
+                data['start_date'] = pd.Timestamp(
+                    data['start_date'], tz='UTC')
+            if data['end_date']:
+                data['end_date'] = pd.Timestamp(
+                    data['end_date'], tz='UTC')
+            return Equity(**data)
+
+    def futures_contract_for_id(self, contract_id):
```
jfkirk: What does contract id represent? SID?

zipline/assets/assets.py
```diff
@@ -239,121 +406,26 @@ def lookup_future_chain(self, root_symbol, as_of_date, knowledge_date):
...
-            raise InvalidAssetType(asset_type=asset_type)
-
-        return asset
+        c = self.conn.cursor()
+        t = {'root_symbol': root_symbol,
+             'as_of_date': as_of_date.value,
+             'knowledge_date': knowledge_date.value}
+        c.execute("""
+        select sid from futures
+        where root_symbol=:root_symbol
```
ssanderson: I didn't know about the :name syntax.  :+1:

zipline/assets/assets.py
ehebert: IIRC, it was because of the existing passing to the Event object as **kwargs.
I'll will double check to see if Row supports __dict__, or could just do a more verbose, `Event(sid=data['sid'], ...)`

zipline/assets/assets.py
ehebert: Not sure, will investigate.

zipline/assets/assets.py
ehebert: Looking at it again, this should use `sid` throughout.

zipline/assets/assets.py
ehebert: Yes, I will change both the function name and parameter.

StewartDouglas:
At no point do you explicitly close the SQLite connection. Are you relying on program exit to implicitly close the connection? I'm interested to know what's best practice here.

ssanderson:
@StewartDouglas a SQLite "Connection" is actually just an open file descriptor, so that's guaranteed to be reclaimed by the OS when the process exits.  CPython will also call the `__del__` of the connection when it's garbage collected or on interpreter shutdown, which I believe closes the connection.

Generally-speaking, if you expect to hold a connection for the lifetime of a program, it's not especially important to close it explicitly before exit, because Python is going to close it for you anyway.

StewartDouglas:
@ssanderson thanks. More seriously considering removing the context manager in my own AssetMaster class, as you originally suggested.

zipline/assets/assets.py
ehebert: @ssanderson I looked at this again, and I'm pushing a few commits to change this over to using the `sqlite3.Row` attribute lookup.

Those patches change the creation of the Equity to look like.

```
            return Equity(sid=data['sid'],
                          symbol=data['symbol'],
                          asset_name=data['asset_name'],
                          start_date=start_date,
                          end_date=end_date,
                          first_traded=data['first_traded'],
                          exchange=data['exchange'])
```

ehebert: Scratch that, there is no attribute lookup on `sqlite3.Row`, had tricked myself into thinking so because I had accidentally left the row_factory on the cursor.

ssanderson: When you say attribute-based lookup, do you mean accesses like `row.attrname`?  `sqlite3.Row` definitely supports the `__getitem__` access pattern in your example above.

ssanderson:
```
In [8]: from sqlite3 import Connection, Row

In [9]: conn = Connection('/dataloader/output/adjustments.db')

In [10]: conn.row_factory = Row

In [11]: row = list(conn.execute('SELECT * from mergers LIMIT 1'))[0]

In [12]: row
Out[12]: <sqlite3.Row at 0x7f5dbbded690>

In [13]: def foo(**kwargs):
   ....:     return kwargs
   ....:

In [14]: foo(**row)
Out[14]: {'effective_date': 1009929600, 'index': 0, 'ratio': 0.538063, 'sid': 2893}
```

ehebert: Aha, I had not understood that `sqlite3.Row` was a factory, thinking you meant the default return type.
I'll switch over to using `Row`.

ehebert:
@jfkirk @ssanderson closing this in favor of https://github.com/quantopian/zipline/pull/649 which is the same basic implementation, but had enough changes squashed in that a separate PR that doesn't obliterate comment history on this one is warranted.

PR #649, opened 2015-07-13
PERF: Change asset finder to be backed by sqlite3.

ehebert:
Attack the startup bottleneck of creating the asset finders caches for a
large universe, which was between 1-2 seconds on development and
production machines.

Instead, allow the AssetFinder to be passed a sqlite3 file that has
already been populated and then hydrate asset objects only when an
equity is referenced for the first time.

To create aforementioned sqlite3, create an AssetFinder with an db_path
and `create_table` set to True. If `create_table` is set to False, the
prepopulated data in the sqlite file found at db_path will be used.

Default behavior is to use an in memory database.

Behavior that changes:

- Fuzzy lookup now only works on one character, that character needs to be
specified at write/metadata consumption time, since the fuzzy lookup key
is created by dropping the character from each symbol.

- Overwriting partially written metadata is no longer
  supported. i.e. some unit tests allowed for inserting just the identifier,
  and then later updating the symbol, end_date, etc.

  Instead of building an upsert behavior at this time, this patch
  changes the unit tests so that the data for each asset is only
  inserted once.

Other notes:

- populate_cache is now removed, since there is no longer a two step
  process of inserting metadata and then realizing that metadata into
  assets. _spawn_asset is rolled into insert_metadata, so that a call to
  insert_metadata both converts the metadata and makes it available in
  the data store.

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -14,10 +14,10 @@
 # limitations under the License.
 
 from abc import ABCMeta
-from itertools import chain
 from numbers import Integral
 import numpy as np
-import operator
+import sqlite3
+from sqlite3 import Row
 import warnings
 
 from logbook import Logger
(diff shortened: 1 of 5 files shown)


zipline/assets/assets.py
```diff
@@ -63,32 +63,175 @@
...
+EQUITY_BY_SID_QUERY = 'select {0} from equities where sid=?'.format(
+    ", ".join(EQUITY_TABLE_FIELDS))
+
+
 class AssetFinder(object):
 
-    def __init__(self, metadata=None, allow_sid_assignment=True):
+    def __init__(self,
+                 metadata=None,
+                 allow_sid_assignment=True,
```
jfkirk: Is allow_sid_assignment used anywhere? I don't recall specifics, but it may be cruft.

zipline/assets/assets.py
```diff
@@ -506,14 +627,17 @@ def map_identifier_index_to_sids(self, index, as_of_date):
...
         """
         Inserts the given metadata kwargs to the entry for the given
         identifier. Matching fields in the existing entry will be overwritten.
         :param identifier: The identifier for which to insert metadata
         :param kwargs: The keyed metadata to insert
         """
-        entry = self.metadata_cache.get(identifier, {})
+        if identifier in self.metadata_cache:
+            # Multiple pass insertion no longer supported.
+            return
```
jfkirk: Should we throw an error/warning on this to avoid confusion?

zipline/assets/assets.py
```diff
@@ -545,6 +669,136 @@ def insert_metadata(self, identifier, **kwargs):
...
+        try:
+            entry['expiration_date'] = entry.pop('expiration_date_nano')
+        except KeyError:
+            pass
+
+        # Process dates to Timestamps
+        try:
+            entry['start_date'] = pd.Timestamp(entry['start_date'], tz='UTC')
+        except KeyError:
+            entry['start_date'] = pd.Timestamp(0, tz='UTC')
```
jfkirk: So we are enforcing a default value for start_date and end_date? Is this to ensure that the rows in the DB are complete for searching?

zipline/assets/assets.py
```diff
@@ -545,6 +669,136 @@ def insert_metadata(self, identifier, **kwargs):
...
+        except KeyError:
+            pass
+        try:
+            entry['expiration_date'] = pd.Timestamp(entry['expiration_date'],
+                                                    tz='UTC')
+        except KeyError:
+            pass
+
+        # Build an Asset of the appropriate type, default to Equity
+        asset_type = entry.pop('asset_type', 'equity')
```
jfkirk: Do we need to have a default type anymore, since everything comes out of explicit tables?

zipline/assets/assets.py
```diff
@@ -545,6 +669,136 @@ def insert_metadata(self, identifier, **kwargs):
...
+            expiration_date,
+            contract_multiplier)
+            VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)""", t)
+
+            t = (asset.sid,
+                 'future')
+            c.execute("""INSERT INTO asset_router(sid, asset_type)
+            VALUES(?, ?)""", t)
+        else:
+            raise InvalidAssetType(asset_type=asset_type)
```
jfkirk: If the asset types are explicit because of the table structure, this error may not be necessary anymore.

zipline/assets/assets.py
ehebert: It is still used in `_insert_metadata` and tests that use a DataFrame source still rely on that behavior.

I think we could push this further and make the writing of the assets more explicit with that functionality broken out into another class, but I think that would be moving too many pieces at once with this patch.

jfkirk:
:+1: from me but, full disclosure, I'm less familiar with SQLLite and practices around maintaining/using those connections. It may be worth getting another pair of eyes that are more familiar with that.

zipline/assets/assets.py
ehebert: We should, but that would currently be incompatible with the tests in test_algorithm, which insert multiple empty values for sid 133 on data source creation.

We could and should do another round of clean up there. I'll add our desire to raise an Exception in the comment here.

zipline/assets/assets.py
ehebert: That is correct.

Issue #636, opened 2015-07-06
AssetFinder does an is check against strings


ssanderson:
https://github.com/quantopian/zipline/blob/master/zipline/assets/assets.py#L605
https://github.com/quantopian/zipline/blob/master/zipline/assets/assets.py#L280
https://github.com/quantopian/zipline/blob/master/zipline/assets/assets.py#L283

This is almost certainly not what we want to be doing. The `is` operator means "is the same PyObject pointer".  Though strings are generally interned, it's not guaranteed that two strings that compare equal are actually the same object.  Consider:
```
In [9]: "None" == "None"
Out[9]: True

In [10]: "None" is "None"
Out[10]: True

In [11]: "None" == ''.join(['N', 'o', 'n', 'e'])
Out[11]: True

In [12]: "None" is ''.join(['N', 'o', 'n', 'e'])
Out[12]: False
```

ssanderson:
cc @ehebert @jfkirk.  This is a time-bomb waiting to go off if we haven't hit it already.  cc @jdricklefs also in case this is related to some of the internal asset mapping failures we've seen recently.

jfkirk:
Oh gosh, that is clearly wrong. Self-assigning.

ehebert:
Looks like the simple fix is to change to `!=`.
(Or we could remove placeholder support?)

jfkirk:
PR for a fix: https://github.com/quantopian/zipline/pull/651

@ssanderson - I could only find the one bad check in the current Zipline master. Are there others that I missed?

jfkirk:
Fix merged - closing.

PR #651, opened 2015-07-14
BUG: Replaces is-check against string

jfkirk:
Replaces an " is 'None' " check with " != None "

Addresses https://github.com/quantopian/zipline/issues/636


ssanderson:
LGTM, and I think the other two cases were already removed.

PR #652, opened 2015-07-14
Improve risk update speed and remove remaining uses of minute mode risk.

ehebert:
Remove the DataFrame of headline risk metrics, in favor of a numpy array
for each metric, like the underlying vectors.

This patch gains several hundred milliseconds on a year long backtest.

Also included a patch which removes the last remains of the 'minute risk' mode, since that test would have had to been changed to use `crm.alpha` instead of `crm.metrics.alpha`, so completely removing those unused piece instead.


jfkirk:
:+1: Love that negative line count.

Issue #632, opened 2015-07-01
Symbol look-up dates should not be dynamic


jfkirk:
Symbols should be resolved with a fixed, defined look-up date, not the algo's current datetime.

As @ssanderson put it, there are almost no cases where someone would want to order AAPL Apple Computers for part of a simulation, and AAPL Madrid Apple Orchards for another part of the simulation.

Self-assigning.

PR #653, opened 2015-07-14
BUG: Symbol look-up dates should not be dynamic

jfkirk:
Resolves https://github.com/quantopian/zipline/issues/632

@ssanderson I whipped this up as an easy example fix, but I'm thinking this is a good argument for moving the set_symbol_lookup method from Quantopian internal to Zipline. What do you think?

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -308,27 +308,27 @@ class TestMiscellaneousAPI(TestCase):
         algo = TradingAlgorithm(asset_metadata=metadata)
 
         # Test before either PLAY existed
-        algo.datetime = pd.Timestamp('2001-12-01', tz='UTC')
+        algo.sim_params.period_end = pd.Timestamp('2001-12-01', tz='UTC')
         with self.assertRaises(SymbolNotFound):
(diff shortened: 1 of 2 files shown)


ssanderson:
@jfkirk I'd be very :+1: on porting set_symbol_lookup_date.  Do you have a clear sense of what's involved in that?

ssanderson:
FWIW, this looks like an improvement over the previous behavior as well

ssanderson:
Can you add a whatsnew entry for this?  We haven't been doing a great job of adding them, but we should get in the habit.

ssanderson:
ping @jfkirk, what's the status on this?

jfkirk:
@StewartDouglas This PR is ready to merge, but I'm not sure it is the best solution (it was a quick fix). What do you think?

StewartDouglas:
Using the end of the simulation period seems like a sensible option to me (it's more likely that this will result in a resolution to the asset the user really wants to trade versus using the start date, for example.) Merging as this is an improvement. Adding the porting of set_symbol_lookup_date to our to-do list.

PR #656, opened 2015-07-21
BUG: Aligns performance packet generation between minute and daily modes

jfkirk:
This PR re-aligns the end-of-day logic between 'minute' and 'daily' emission modes in the PerformanceTracker. As part of this, the trade simulation is retooled to handle the perf packets.

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -279,8 +279,11 @@ class PerformanceTracker(object):
         Creates a dictionary representing the state of this tracker.
         Returns a dict object of the form described in header comments.
         """
-        if not emission_type:
+
+        # Default to the emission rate of this tracker if no type is provided
+        if emission_type is None:
             emission_type = self.emission_rate
+
         _dict = {
             'period_start': self.period_start,
             'period_end': self.period_end,
@@ -294,6 +297,8 @@ class PerformanceTracker(object):
         elif emission_type == 'minute':
             _dict['minute_perf'] = self.todays_performance.to_dict(
                 self.saved_dt)
+        else:
+            raise BaseException("Invalid emission type: %s" % emission_type)
 
         return _dict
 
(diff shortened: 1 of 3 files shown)


zipline/gens/tradesimulation.py
```diff
@@ -356,9 +341,9 @@ def on_dt_changed(self, dt):
         if self.algo.datetime != dt:
             self.algo.on_dt_changed(dt)
 
-    def get_message(self, dt):
+    def generate_messages(self, dt):
         """
-        Get a perf message for the given datetime.
+        Generator that yields perf messages for the given datetime.
```
yankees714: Remind me again of the benefit of making this a generator? It's not immediately clear to me.

jfkirk: This is a different one than the one we talked about. 

This method used to have clunky logic for handling minute/daily differences, and was reliant on there being exterior logic to create daily perf packets in minutely backtests. I made it a generator because it is where minute and daily perf reporting converge, so by using it as a generator, we know we are getting a stream out regardless of emission rate.

yankees714: Ah okay cool, I'm seeing where this gets called now.

yankees714:
Looks good to me!

zipline/finance/performance/tracker.py
```diff
@@ -294,6 +297,8 @@ def to_dict(self, emission_type=None):
         elif emission_type == 'minute':
             _dict['minute_perf'] = self.todays_performance.to_dict(
                 self.saved_dt)
+        else:
+            raise BaseException("Invalid emission type: %s" % emission_type)
```
ssanderson: Why are we raising a BaseException here?

zipline/gens/tradesimulation.py
```diff
@@ -114,29 +114,20 @@ def transform(self, stream_in):
...
+                    messages = self._process_snapshot(
                         date,
                         snapshot,
                         self.algo.instant_fill,
                     )
                     # Perf messages are only emitted if the snapshot contained
                     # a benchmark event.
-                    if message is not None:
-                        yield message
+                    if messages is not None:
```
richafrank: The generator itself won't ever be None, right?

PR #657, opened 2015-07-21
DEV: Uses ValueError when PerformanceTracker.to_dict receives invalid…

jfkirk:
… emission type

@ssanderson Is this alright?

@richafrank Also takes the opportunity to remove in 'is not None' check in the trade simulation.

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -298,7 +298,7 @@ class PerformanceTracker(object):
             _dict['minute_perf'] = self.todays_performance.to_dict(
                 self.saved_dt)
         else:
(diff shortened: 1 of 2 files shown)


zipline/finance/performance/tracker.py
```diff
@@ -298,7 +298,7 @@ def to_dict(self, emission_type=None):
             _dict['minute_perf'] = self.todays_performance.to_dict(
                 self.saved_dt)
         else:
-            raise BaseException("Invalid emission type: %s" % emission_type)
+            raise ValueError("Invalid emission type: %s" % emission_type)
```
ssanderson: :+1:

PR #658, opened 2015-07-21
MAINT: Remove unused `assets` property.

ssanderson:
It references a self.cache attribute that no longer exists.

ping @jfkirk


StewartDouglas:
Was puzzled earlier as to why this still existed. PR looks good to me.

jfkirk:
Yep, almost definitely cruft. :+1:

PR #661, opened 2015-07-28
Create AssetDBWriter class

StewartDouglas:
DO NOT MERGE

@jfkirk @yankees714 @ssanderson @ehebert 

I'm opening this PR to get additional eyes on the AssetDBWriter class and its subclasses.
The AssetDBWriter and its subclasses will be responsible for writing data to SQLite. Each subclass
is implemented to handle a specific data type, e.g. dictionary, pandas.DataFrame etc.

In the long term we will want to completely decouple the writing and reading functionality, 
with the the reading being taken care of by the AssetFinder. 

A rough example of how our code will typically operate in the future is:

```python
# Connect to SQLite
conn = sqlite3.connection(db_path)
# Instantiate a writer with the data we want to feed it 
asset_writer = AssetDBWriterSubclass(relevant_data)
# Write the data to the database.
asset_writer.write_all(conn)
# Build the caches using the database data.
asset_finder = AssetFinder(conn)
```

However, completely decoupling the reading and writing functionality will mean that we have to 
update a significant number of tests and the wider zipline codebase. In the short term I may not 
have the capacity to implement this due to competing priorities.

As a workaround I have created the subclass AssetDBWriterLegacy which borrows heavily from our
existing code base, and implements functionality such as a ```consume_identifiers``` method, the logic of which will be handled differently in the future. This class and its subclasses will be responsible for writing to the database in the short term, but will ultimately be deprecated. The most significant difference
between AssetDBWriter and AssetDBWriterLegacy is that the later returns a generator from its ```gen_data``` method. This is because previously we implemented an ```_insert_metadata(identifier, **kwargs)``` method which would insert the relevant data for a single sid. In the future AssetDBWriter's ```gen_data``` method will return a pandas DataFrames, as can be seen in the ```AssetDBWriterFromDictionary``` class.  

The current state of testing for the update is the following:
- Passes all the tests after tweeking two in test_assets.py to account for the fact that the AssetFinder no longer has a metadata_cache. 
- Returns the same results in our default AAPL daily and minutely backtests. 
- The performance matrix has not yet been run.

Note also that the SQL schema is still a work in progress.

### DIFF
--- a/zipline/assets/__init__.py
+++ b/zipline/assets/__init__.py
@@ -20,6 +20,14 @@ from ._assets import (
     make_asset_array,
     CACHE_FILE_TEMPLATE
 )
+from .asset_writer import (
+    AssetDBWriterFromDictionary,
+    NullAssetDBWriterLegacy,
+    AssetDBWriterLegacyFromList,
+    AssetDBWriterLegacyFromDictionary,
+    AssetDBWriterLegacyFromDataFrame,
+    AssetDBWriterLegacyFromReadable
+)
 from .assets import (
     AssetFinder,
     AssetConvertible
@@ -29,6 +37,12 @@ __all__ = [
     'Asset',
     'Equity',
     'Future',
+    'AssetDBWriterFromDictionary',
+    'NullAssetDBWriterLegacy',
+    'AssetDBWriterLegacyFromList',
+    'AssetDBWriterLegacyFromDictionary',
+    'AssetDBWriterLegacyFromDataFrame',
+    'AssetDBWriterLegacyFromReadable',
     'AssetFinder',
     'AssetConvertible',
     'make_asset_array',
--- a/zipline/assets/_assets.pyx
+++ b/zipline/assets/_assets.pyx
@@ -232,6 +232,7 @@ cdef class Future(Asset):
     def __cinit__(self,
                   int sid, # sid is required
                   object symbol="",
+                  object root_symbol_id = "",
                   object root_symbol="",
                   object asset_name="",
                   object start_date=None,
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -163,7 +163,13 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
 
     def _write_futures(self, futures, db_conn):
 
+        # Retrieve the data to add to the futures_contracts table
         data = [tuple(x) for x in futures.to_records()]
+
+        # Retrieve the data to add to the asset_router table
+        sids = [x[0] for x in data]
+        futs = ['future'] * len(sids)
+        to_insert = zip(sids, futs)
 
         c = db_conn.cursor()
         c.executemany("""
@@ -174,6 +180,12 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
             VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
         """, data)
 
+        c.executemany("""
+            INSERT INTO asset_router
+            ('sid', 'asset_type')
+            VALUES(?,?)
+        """, to_insert)
+
     def _write_equities(self, equities, db_conn):
 
         # Apply fuzzy matching.
@@ -181,7 +193,13 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
             equities['fuzzy'] = equities['symbol'].str.\
                 replace(self.fuzzy_char, '')
 
+        # Retrieve the data to add to the equitites table
         data = [tuple(x) for x in equities.to_records()]
+
+        # Retrieve the data to add to the asset_router table
+        sids = [x[0] for x in data]
+        eqs = ['equity'] * len(sids)
+        to_insert = zip(sids, eqs)
 
         c = db_conn.cursor()
         c.executemany("""
@@ -190,6 +208,12 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
              'end_date', 'first_traded', 'exchange', 'fuzzy')
             VALUES(?, ?, ?, ?, ?, ?, ?, ?)
         """, data)
+
+        c.executemany("""
+            INSERT INTO asset_router
+            ('sid', 'asset_type')
+            VALUES(?,?)
+        """, to_insert)
 
     def init_db(self,
                 db_conn,
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -105,7 +105,7 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
             If True, create SQL ForeignKey and Index constraints.
 
         """
-
+        self.fuzzy_char = fuzzy_char
         self.allow_sid_assignment = allow_sid_assignment
         if allow_sid_assignment:
             ts = normalize_date(pd.Timestamp('now', tz='UTC'))
@@ -139,6 +139,7 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
     def _write_exchanges(self, exchanges, db_conn):
 
         data = [tuple(x) for x in exchanges.to_records()]
+
         c = db_conn.cursor()
         # The OR IGNORE syntax means we do not insert data
         # which would violate an SQL constraint.
@@ -175,7 +176,13 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
 
     def _write_equities(self, equities, db_conn):
 
+        # Apply fuzzy matching.
+        if self.fuzzy_char:
+            equities['fuzzy'] = equities['symbol'].str.\
+                replace(self.fuzzy_char, '')
+
         data = [tuple(x) for x in equities.to_records()]
+
         c = db_conn.cursor()
         c.executemany("""
             INSERT OR IGNORE INTO equities
@@ -256,6 +263,9 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
             asset_type text
         )""")
 
+        # Note: Would be optimal to use INTEGER PRIMARY KEY here, but SQLite
+        # tables cannot be modified after creation. Using CREATE UNIQUE INDEX
+        # only marginally less performant.
         if constraints:
 
             c.execute('CREATE UNIQUE INDEX IF NOT EXISTS ix_equities_sid '
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -123,32 +123,66 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
         # Get the data to add to SQL
         equities, futures, exchanges, root_symbols = self.load_data()
 
-        # Write to the SQL tables
+        c = db_conn.cursor()
+        # Write to the SQL tables, using the raw SQL driver instead
+        # of the pandas.DataFrame.to_sql method, as the former
+        # allows us to create an SQL transaction
+        c.execute('BEGIN')
+        # Everything between here and the db_conn.commit()
+        # will be part of of the same SQL transaction.
         self._write_exchanges(exchanges, db_conn)
         self._write_root_symbols(root_symbols, db_conn)
         self._write_futures(futures, db_conn)
         self._write_equities(equities, db_conn)
+        db_conn.commit()
 
     def _write_exchanges(self, exchanges, db_conn):
 
-        exchanges.to_sql('futures_exchanges', db_conn, if_exists='replace',
-                         index=True, index_label='exchange_id')
+        data = [tuple(x) for x in exchanges.to_records()]
+        c = db_conn.cursor()
+        # The OR IGNORE syntax means we do not insert data
+        # which would violate an SQL constraint.
+        c.executemany("""
+            INSERT OR IGNORE INTO futures_exchanges
+            ('exchange_id', 'exchange', 'timezone')
+            VALUES (?, ?, ?)
+            """, data)
 
     def _write_root_symbols(self, root_symbols, db_conn):
 
-        root_symbols.to_sql('futures_root_symbols', db_conn,
-                            if_exists='replace', index=True,
-                            index_label='root_symbol_id')
+        data = [tuple(x) for x in root_symbols.to_records()]
+
+        c = db_conn.cursor()
+        c.executemany("""
+            INSERT OR IGNORE INTO futures_root_symbols
+            ('root_symbol_id', 'root_symbol', 'sector',
+             'description', 'exchange_id')
+            VALUES(?, ?, ?, ?, ?)
+        """, data)
 
     def _write_futures(self, futures, db_conn):
 
-        futures.to_sql('futures_contracts', db_conn, if_exists='append',
-                       index=True, index_label='sid')
+        data = [tuple(x) for x in futures.to_records()]
+
+        c = db_conn.cursor()
+        c.executemany("""
+            INSERT OR IGNORE INTO futures_contracts
+            ('sid', 'symbol', 'root_symbol', 'asset_name',
+             'start_date', 'end_date', 'first_traded', 'exchange',
+             'notice_date', 'expiration_date', 'contract_multiplier')
+            VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+        """, data)
 
     def _write_equities(self, equities, db_conn):
 
-        equities.to_sql('equities', db_conn, if_exists='append',
-                        index=True, index_label='sid')
+        data = [tuple(x) for x in equities.to_records()]
+        c = db_conn.cursor()
+        c.executemany("""
+            INSERT OR IGNORE INTO equities
+            ('sid', 'symbol', 'asset_name', 'start_date',
+             'end_date', 'first_traded', 'exchange', 'fuzzy')
+            VALUES(?, ?, ?, ?, ?, ?, ?, ?)
+        """, data)
 
     def init_db(self,
                 db_conn,
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -1,0 +1,711 @@
+import pandas as pd
+import numpy as np
+from pandas.tseries.tools import normalize_date
+from six import with_metaclass, string_types
+from abc import (
+    ABCMeta,
+    abstractmethod,
+)
+from zipline.errors import (
+    ConsumeAssetMetaDataError,
+    InvalidAssetType,
+    SidAssignmentError,
+)
+from zipline.assets import (
+    Asset, Equity, Future
+)
+
+ASSET_FIELDS = [
+    'sid',
+    'asset_type',
+    'symbol',
+    'asset_name',
+    'start_date',
+    'end_date',
+    'first_traded',
+    'exchange',
+    'notice_date',
+    'root_symbol',
+    'expiration_date',
+    'contract_multiplier',
+    # The following fields are for compatibility with other systems
+    'file_name',  # Used as symbol
+    'company_name',  # Used as asset_name
+    'start_date_nano',  # Used as start_date
+    'end_date_nano'  # Used as end_date
+]
+
+# Expected fields for an Asset's metadata
+ASSET_TABLE_FIELDS = [
+    'sid',
+    'symbol',
+    'asset_name',
+    'start_date',
+    'end_date',
+    'first_traded',
+    'exchange'
+]
+
+# Expected fields for an Asset's metadata
+FUTURE_TABLE_FIELDS = ASSET_TABLE_FIELDS + [
+    'root_symbol_id',
+    'notice_date',
+    'expiration_date',
+    'contract_multiplier',
+]
+
+EQUITY_TABLE_FIELDS = ASSET_TABLE_FIELDS
+
+EXCHANGE_TABLE_FIELDS = [
+    'exchange_id',
+    'exchange',
+    'timezone'
+]
+
+ROOT_SYMBOL_TABLE_FIELDS = [
+    'root_symbol_id',
+    'root_symbol',
+    'sector',
+    'description',
+    'exchange_id'
+]
+
+
+class AssetDBWriter(with_metaclass(ABCMeta)):
+    """
+    Class used to write arbitrary data to SQLite database.
+    Concrete subclasses will implement the logic for a specific
+    input datatypes by implementing the load_data method.
+
+    Methods
+    -------
+    write_all(db_conn, fuzzy_char=None, allow_sid_assignment=True,
+              constraints=False)
+        Write the data supplied at initialization to the database.
+    init_db(db_conn, constraints=False)
+        Create the SQLite tables (called by write_all).
+    load_data(self)
+        Returns data in standard format.
+
+    """
+
+    def write_all(self, db_conn, fuzzy_char=None, allow_sid_assignment=True,
+                  constraints=False):
+        """ Write pre-supplied data to SQLite.
+
+        Parameters
+        ----------
+        db_conn: sqlite3.Connection
+            A connection to a SQLite database.
+        fuzzy_char: string
+            A string for use in fuzzy matching.
+        allow_sid_assignment: boolean
+            If True then the class can assign sids where necessary.
+        constraints: boolean
+            If True, create SQL ForeignKey and Index constraints.
+
+        """
+
+        self.allow_sid_assignment = allow_sid_assignment
+        if allow_sid_assignment:
+            ts = normalize_date(pd.Timestamp('now', tz='UTC'))
+            # Store as seconds since UNIX Epoch for compatibility
+            # with SQL.
+            self.end_date_to_assign = (ts.value // 10 ** 9)
+
+        # Store a nested-dict of all metadata for
+        # reference when building Assets
+        self.metadata_cache = {}
+
+        # Create SQL tables
+        self.init_db(db_conn, constraints)
+
+        # Get the data to add to SQL
+        equities, futures, exchanges, root_symbols = self.load_data()
+
+        # Write to the SQL tables
+        self._write_exchanges(exchanges, db_conn)
+        self._write_root_symbols(root_symbols, db_conn)
+        self._write_futures(futures, db_conn)
+        self._write_equities(equities, db_conn)
+
+    def _write_exchanges(self, exchanges, db_conn):
+
+        exchanges.to_sql('futures_exchanges', db_conn, if_exists='replace',
+                         index=True, index_label='exchange_id')
+
+    def _write_root_symbols(self, root_symbols, db_conn):
+
+        root_symbols.to_sql('futures_root_symbols', db_conn,
+                            if_exists='replace', index=True,
+                            index_label='root_symbol_id')
+
+    def _write_futures(self, futures, db_conn):
+
+        futures.to_sql('futures_contracts', db_conn, if_exists='append',
+                       index=True, index_label='sid')
+
+    def _write_equities(self, equities, db_conn):
+
+        equities.to_sql('equities', db_conn, if_exists='append',
+                        index=True, index_label='sid')
+
+    def init_db(self,
+                db_conn,
+                constraints=False):
+        """Connect to database and create tables.
+
+        Parameters
+        ----------
+        db_conn: sqlite3.Connection
+            A connection to a SQLite database.
+        constraints: boolean
+            If True, create SQL ForeignKey and Index constraints.
+        """
+
+        c = db_conn.cursor()
+
+        c.execute("""
+            CREATE TABLE IF NOT EXISTS equities (
+            sid INTEGER NOT NULL,
+            symbol TEXT,
+            asset_name TEXT,
+            start_date INTEGER DEFAULT 0,
+            end_date INTEGER,
+            first_traded INTEGER,
+            exchange TEXT,
+            fuzzy TEXT
+        )""")
+
+        c.execute("""
+            CREATE TABLE IF NOT EXISTS futures_exchanges (
+            exchange_id INTEGER NOT NULL,
+            exchange TEXT,
+            timezone TEXT
+        )""")
+
+        c.execute("""
+            CREATE TABLE IF NOT EXISTS futures_root_symbols (
+            root_symbol_id INTEGER NOT NULL,
+            root_symbol TEXT,
+            sector TEXT,
+            description TEXT,
+            exchange_id INTEGER{fk}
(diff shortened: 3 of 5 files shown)


zipline/assets/__init__.py
```diff
@@ -20,6 +20,15 @@
...
     CACHE_FILE_TEMPLATE
 )
+from .asset_writer import (
+    NullAssetDBWriterLegacy,
+    AssetDBWriterLegacyFromList,
+    AssetDBWriterLegacyFromDictionary,
+    AssetDBWriterLegacyFromDataFrame,
+    AssetDBWriterLegacyFromReadable,
+    FUTURE_TABLE_FIELDS,
+    EQUITY_TABLE_FIELDS
```
jfkirk: Why do these FIELDS lists need to be imported? Are the fields referenced outside of the writer/finder?

zipline/assets/_assets.pyx
```diff
@@ -240,7 +241,8 @@ cdef class Future(Asset):
                   object expiration_date=None,
                   object first_traded=None,
                   object exchange="",
-                  int contract_multiplier=1):
+                  int contract_multiplier=1,
+                  object fuzzy=""):
```
jfkirk: Why does the fuzzy have to be a field of the Asset, rather than just having fuzzies at look-up in the AssetFinder?

zipline/assets/__init__.py
StewartDouglas: Yup, they're now defined in the writer (asset_writer.py) but also used in the reader (assets.py), hence the reader needs to import them.

zipline/assets/assets.py
```diff
@@ -27,112 +25,87 @@
...
-# Expected fields for an Asset's metadata
-FUTURE_TABLE_FIELDS = ASSET_TABLE_FIELDS + [
-    'root_symbol',
-    'notice_date',
-    'expiration_date',
-    'contract_multiplier',
-]
-
-EQUITY_TABLE_FIELDS = ASSET_TABLE_FIELDS
+from zipline.assets import (
```
jfkirk: Is this a circular import, as this module is imported by zipline.assets?

zipline/assets/assets.py
```diff
@@ -27,112 +25,87 @@
...
 # repeatedly.
-FUTURE_BY_SID_QUERY = 'select {0} from futures where sid=?'.format(
+FUTURE_BY_SID_QUERY = 'select {0} from futures_contracts where sid=?'.format(
     ", ".join(FUTURE_TABLE_FIELDS))
 
 EQUITY_BY_SID_QUERY = 'select {0} from equities where sid=?'.format(
     ", ".join(EQUITY_TABLE_FIELDS))
 
 
+def get_relevant_writer(metadata):
```
jfkirk: If this is creating the writers, should it be named create_relevant_writer? 'get_' makes me think that it is storing writers for the different types.

It may not be worth changing, depending on how soon we plan to deprecate this.

zipline/assets/assets.py
```diff
@@ -406,10 +358,10 @@ def lookup_symbol(self, symbol, as_of_date, fuzzy=False):
             c = self.conn.cursor()
             fuzzy = symbol.replace(self.fuzzy_char, '')
             t = (fuzzy, as_of_date.value, as_of_date.value)
-            query = ("select sid from equities "
-                     "where fuzzy=? " +
-                     "and start_date<=? " +
-                     "and end_date>=?")
+            query = ("SELECT sid FROM EQUITIES "
```
jfkirk: should the table name be capitalized?

zipline/assets/assets.py
StewartDouglas: create would probably be a better name - will change.

zipline/assets/_assets.pyx
StewartDouglas: Ideally wouldn't be part of the Future object. Investigating.

zipline/assets/__init__.py
jfkirk: Why not have the assets.py import them directly from asset_writer.py? That would prevent a circular import.

zipline/assets/assets.py
StewartDouglas: I don't think so, as zipline.assets is the *package*. Within this package zipline/assets/assets.py imports from zipline/assets/asset_writer.py, and zipline/assets/asset_writer.py imports from zipline/assets/_assets.pyx.

zipline/assets/assets.py
StewartDouglas: nope, my bad. good catch. just a style thing, but worth being consistent.

zipline/assets/_assets.pyx
ehebert: @jfkirk fwiw I've been re-thinking the fuzzy lookup change that I made. I think we may be able to support the interpolation at runtime, the 'fuzzy' value which was the symbol name with the fuzzy token dropped may be not worth the complexity/rigidity.

zipline/assets/__init__.py
StewartDouglas: Gonna hold off making a call on this one till I've done some initial integration with the AssetMaster. Your suggestion seems sensible though.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,744 @@
...
+from zipline.errors import (
+    ConsumeAssetMetaDataError,
+    InvalidAssetType,
+    SidAssignmentError,
+)
+from zipline.assets import (
+    Asset, Equity, Future
+)
+
+ASSET_FIELDS = [
```
llllllllll: Does the order matter here, if not, these should be `frozenset`.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,744 @@
...
+        allow_sid_assignment: boolean
+            If True then the class can assign sids where necessary.
+        constraints: boolean
+            If True, create SQL ForeignKey and Index constraints.
+
+        """
+
+        self.allow_sid_assignment = allow_sid_assignment
+        if allow_sid_assignment:
+            ts = normalize_date(pd.Timestamp('now', tz='UTC'))
```
llllllllll: We can cut pandas out of this and just do:

```python
self.end_date_to_assign = int(datetime.combine(
     date.today(),
     time(0, tzinfo=pytz.utc),
).timestamp())
```

This is faster and a bit more clear imho.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,744 @@
...
+        self.init_db(db_conn, constraints)
+
+        # Get the data to add to SQL
+        equities, futures, exchanges, root_symbols = self.gen_data()
+
+        c = db_conn.cursor()
+        # Write to the SQL tables, using the raw SQL driver instead
+        # of the pandas.DataFrame.to_sql method, as the former
+        # allows us to create an SQL transaction
+        c.execute('BEGIN')
```
llllllllll: It looks like we don't need to manually execute begins, just `commit`s. Is there a particular reason we are manually executing this?

As a more high level comment: I would really really like to see this using sqlalchemy or something. This would let us support any sql database with a dbapi and would make the code easier to maintain

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,744 @@
...
+        self._write_root_symbols(root_symbols, db_conn)
+        self._write_futures(futures, db_conn)
+        self._write_equities(equities, db_conn)
+        db_conn.commit()
+
+    def _write_exchanges(self, exchanges, db_conn):
+
+        data = [tuple(x) for x in exchanges.to_records()]
+        c = db_conn.cursor()
+        # The OR IGNORE syntax means we do not insert data
```
llllllllll: Should we warn the user or raise if a constraint is violated? Why are we dropping them implicitly?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,744 @@
...
+                      'ON futures_root_symbols(root_symbol_id)')
+            c.execute('CREATE UNIQUE INDEX IF NOT EXISTS ix_asset_router_sid  '
+                      'ON asset_router(sid)')
+
+        # Note: Also need a max_date table.
+
+        db_conn.commit()
+
+    @abstractmethod
+    def gen_data(self, equities, futures, exchanges, root_symbols):
```
llllllllll: The function signature here is incorrect, no?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,744 @@
...
+        """
+        Subclasses should implement this method to
+        return data in a standard format.
+        """
+
+        return NotImplementedError()
```
llllllllll: this want to be a `raise` not `return`. Also, `NotImplementedError` should be passed the name of the method that is not implemented normally like: `raise NotImplementedError('gen_data')`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,744 @@
...
+    input datatypes by implementing the gen_data method.
+
+    Methods
+    -------
+    write_all(db_conn, fuzzy_char=None, allow_sid_assignment=True,
+              constraints=False)
+        Write the data supplied at initialization to the database.
+    init_db(db_conn, constraints=False)
+        Create the SQLite tables (called by write_all).
+    gen_data(self, equities, futures, exchanges, root_symbols)
```
llllllllll: I don't think this is still the signature, we call it with no args in the body of `write_all`

zipline/assets/asset_writer.py
ssanderson: This should just be using a the cursor as a context manager:

```
with c:
    # do stuff
```
This will commit if everything in the context succeeds, and will rollback if an exception is raised.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,744 @@
...
+
+        self.allow_sid_assignment = allow_sid_assignment
+        if allow_sid_assignment:
+            ts = normalize_date(pd.Timestamp('now', tz='UTC'))
+            # Store as seconds since UNIX Epoch for compatibility
+            # with SQL.
+            self.end_date_to_assign = (ts.value // 10 ** 9)
+
+        # Create SQL tables
+        self.init_db(db_conn, constraints)
```
llllllllll: I am not sure I like the pattern of `write_all` calling `init_db` here. Shouldn't those just be separate. This seems to make it harder to extend through subclassing

zipline/assets/asset_writer.py
ssanderson: @llllllllll I don't think your example works on PY2.  Also, I strongly disagree that your version is more clear.

ssanderson: I would write this as `pd.Timestamp('now').normalize()`.

zipline/assets/asset_writer.py
StewartDouglas: You're right - the signature is out of date. Making changes as per your suggestions.

zipline/assets/asset_writer.py
StewartDouglas: Order shouldn't matter, but can frozensets be combined in the same manner as I do here?

zipline/assets/assets.py
llllllllll: This can be `from .asset_writer import (...`

zipline/assets/assets.py
```diff
@@ -26,112 +25,89 @@
...
-    SidAssignmentError,
     SidNotFound,
     SymbolNotFound,
     MapAssetIdentifierIndexError,
 )
-from zipline.assets._assets import (
+from zipline.assets import (
     Asset, Equity, Future
 )
+from abc import ABCMeta
```
llllllllll: Can we move this to the top of the imports?

imports should be in 3 groups separated by 1 line and sorted alphabetically internally:

1. stdlib
1. third party packages
1. packages defined in this project

StewartDouglas: :+1:

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,755 @@
...
+            A connection to a SQLite database.
+        constraints: boolean
+            If True, create SQL ForeignKey and Index constraints.
+        """
+
+        c = db_conn.cursor()
+
+        # The AssetDBWriter holds a nested-dict of all metadata for
+        # reference when building Assets
+        self.metadata_cache = {}
```
llllllllll: this should probably be created in the `__init__`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+
+
+class NullAssetDBWriterLegacy(AssetDBWriterLegacy):
+    """
+    An implementation of AssetDBWriterLegacy for use
+    when no data is initially specified.
+    """
+
+    def load_data(self, __):
+        for i in iter(()):
```
llllllllll: This is just `return iter(())`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            yield
+
+
+class AssetDBWriterLegacyFromList(AssetDBWriterLegacy):
+    """
+    Returns a generator yielding entries from sid_list.
+    """
+    def load_data(self, sid_list):
+
+        for i in sid_list:
```
llllllllll: `return iter(sid_list)`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        with the following format:
+
+        {id_0: {start_date : ...}, id_1: {start_data: ...}, ...}
+    """
+
+    def load_data(self, dict):
+        """
+        Returns a generator yielding pairs of (identifier, metadata)
+        """
+        for identifier, metadata in dict.items():
```
llllllllll: Best practice says to not shadow the builtin name `dict` by naming this argument something else. A common practice is to resolve this shadowing issue with a suffixing `_` like: `dict_`

`return six.iteritems(dict_)`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        """
+        for row in readable.read():
+            id_metadata = {}
+            for field in ASSET_FIELDS:
+                try:
+                    row_value = row[field]
+                    # Avoid passing placeholder strings
+                    if row_value and (row_value != 'None'):
+                        id_metadata[field] = row[field]
+                except KeyError:
```
llllllllll:
```python
except (KeyError, IndexError):
    continue
```

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+                    continue
+                except IndexError:
+                    continue
+            if 'sid' in id_metadata:
+                identifier = id_metadata['sid']
```
llllllllll: this should just be `identifier = id_metadata.pop('sid')` same in the elif

zipline/assets/assets.py
```diff
@@ -26,112 +26,87 @@
...
 # repeatedly.
-FUTURE_BY_SID_QUERY = 'select {0} from futures where sid=?'.format(
+FUTURE_BY_SID_QUERY = 'select {0} from futures_contracts where sid=?'.format(
     ", ".join(FUTURE_TABLE_FIELDS))
 
 EQUITY_BY_SID_QUERY = 'select {0} from equities where sid=?'.format(
     ", ".join(EQUITY_TABLE_FIELDS))
 
 
+def create_relevant_writer(metadata):
```
llllllllll: I feel like this is a really good place for multipledispatch (or single dispatch). This works just fine though

zipline/assets/assets.py
```diff
@@ -151,54 +126,24 @@ def __init__(self,
...
+        self.conn.close()
+        # Create new database connection in memory.
+        self.conn = sqlite3.connect(':memory:')
+        # Initialize the database tables using the same connection
+        # as used by the AssetFinder.
+        _asset_writer = NullAssetDBWriterLegacy({})
+        _asset_writer.init_db(self.conn)
 
     def asset_type_by_sid(self, sid):
+        """
```
llllllllll: is this a public api method? If so, can we use the docstring conventions

zipline/assets/assets.py
```diff
@@ -631,274 +586,49 @@ def map_identifier_index_to_sids(self, index, as_of_date):
...
-            # For example, in the creation of a DataFrameSource, the source's
-            # 'sid' args may be full Assets
-            if isinstance(identifier, Asset):
-                sid = identifier.sid
-                metadata = identifier.to_dict()
-                metadata['asset_type'] = identifier.__class__.__name__
-                self.insert_metadata(identifier=sid, **metadata)
-            else:
-                self.insert_metadata(identifier)
+        _asset_writer = AssetDBWriterLegacyFromList(identifiers)
```
llllllllll: If this is using legacy code, should we raise a `DeprecationWarning` here?

zipline/assets/assets.py
```diff
@@ -151,54 +126,24 @@ def __init__(self,
...
-        asset_type text)
-        """)
-
-        c.execute('CREATE INDEX asset_router_sid on asset_router(sid)')
-
-        self.conn.commit()
+    def clear_metadata(self):
+        """
+        Used for testing.
+        Will be deprecated in future versions of zipline.
```
llllllllll: Let's raise a `PendingDeprecationWarning` to put this into code.

They are ignored by default

zipline/assets/assets.py
```diff
@@ -26,112 +26,87 @@
...
     ", ".join(FUTURE_TABLE_FIELDS))
 
 EQUITY_BY_SID_QUERY = 'select {0} from equities where sid=?'.format(
     ", ".join(EQUITY_TABLE_FIELDS))
 
 
+def create_relevant_writer(metadata):
+    """ Create an instance of AssetDBWriter relevant to
+        processing metadata.
+        Will be deprecated in future versions of zipline.
```
llllllllll: Is this not deprecated already because we have a replacement?

zipline/assets/asset_writer.py
llllllllll: Yeah, just with `|` (union) instead of `+` (concat)

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+
+        # Note: Also need a max_date table.
+
+        db_conn.commit()
+
+    @abstractmethod
+    def load_data(self):
+        """
+        Subclasses should implement this method to return data in a standard
+        format: a pandas.DataFrame for each of the following tables:
```
llllllllll: What is the dshape of each of the dataframes?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+    input datatypes by implementing the load_data method.
+
+    Methods
+    -------
+    write_all(db_conn, fuzzy_char=None, allow_sid_assignment=True,
+              constraints=False)
+        Write the data supplied at initialization to the database.
+    init_db(db_conn, constraints=False)
+        Create the SQLite tables (called by write_all).
+    load_data(self)
```
llllllllll: omit the `self`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        # an exception when building assets.
+        if allow_sid_assignment:
+            self.end_date_to_assign = normalize_date(
+                pd.Timestamp('now', tz='UTC'))
+
+        # Create SQL tables.
+        self.init_db(self.conn, constraints)
+
+        # Write to SQL tables.
+        for sid, metadata in self.load_data(self._data):
```
llllllllll: Subclasses should not change the call signature of methods in an incompatible way. imho, the legacy writers should have a seperate class heirachy and raise deprecation warnings when instantiated.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            self.write_block(sid, **metadata)
+
+    def write_block(self, identifier, **kwargs):
+        """
+        Inserts the given metadata kwargs to the entry for the given
+        sid. Matching fields in the existing entry will be overwritten.
+        Will be deprecated in future versions of zipline.
```
llllllllll: `PendingDeprecationWarning`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            else:
+                if self.allow_sid_assignment:
+                    # Assign the sid the value of its insertion order.
+                    # This assumes that we are assigning values to all assets.
+                    entry['sid'] = len(self.metadata_cache)
+                else:
+                    raise SidAssignmentError(identifier=identifier)
+
+        # If the file_name is in the kwargs, it will be used as the symbol
+        try:
```
llllllllll: let's refactor all of these into something like `remap_key`
```python
def remap_key(mapping, from_, to):
    try:
        mapping[to_] = mapping.pop(from_)
    except KeyError:
        pass
```
This way future readers can better understand the intent here.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            pass
+        except KeyError:
+            if isinstance(identifier, string_types):
+                entry['symbol'] = identifier
+
+        # If the company_name is in the kwargs, it may be the asset_name
+        try:
+            company_name = entry.pop('company_name')
+            try:
+                entry['asset_name']
```
llllllllll: entry.setdefault('asset_name', company_name)

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        except KeyError:
+            pass
+        try:
+            entry['expiration_date'] = entry.pop('expiration_date_nano')
+        except KeyError:
+            pass
+
+        # Process dates to Timestamps
+        try:
+            entry['start_date'] = pd.Timestamp(entry['start_date'], tz='UTC')
```
llllllllll: We are duplicating this logic, why not refactor into something like:

```python
from toolz.utils import no_default
def maybe_apply_to_value(mapping, key, func, default=no_default):
    if key in mapping:
        mapping[key] = func(mapping[key])
    elif default is not no_default:
        mapping[key] = default
```

Then we can replace this with 
```python
maybe_apply_value(
    entry,
    'start_date',
     partial(pd.Timestamp, tz='utc'),
     pd.Timestamp(0, tz='utc'),
)
```
We can do simular things with `end_date`, `notice_date` and others.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            return
+
+        entry = {}
+
+        for key, value in kwargs.items():
+            # Do not accept invalid fields
+            if key not in ASSET_FIELDS:
+                continue
+            # Do not accept Nones
+            if value is None:
```
llllllllll: Should these fail silently?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        sid. Matching fields in the existing entry will be overwritten.
+        Will be deprecated in future versions of zipline.
+        """
+
+        if identifier in self.metadata_cache:
+            # Multiple pass insertion no longer supported.
+            # This could and probably should raise an Exception, but is
+            # currently just a short-circuit for compatibility with existing
+            # testing structure in the test_algorithm module which creates
+            # multiple sources which all insert redundant metadata.
```
llllllllll: Does this mean the docstring is no longer correct? It says It overrides old entries. Also, if the only reason we are supporting this is the tests then we just need to update the tests.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            if isinstance(value, float) and np.isnan(value):
+                continue
+            entry[key] = value
+
+        # Check if the sid is declared
+        try:
+            entry['sid']
+        except KeyError:
+            # If the sid is not a sid, assign one
+            if hasattr(identifier, '__int__'):
```
llllllllll:
```python
try:
    enty['sid'] = int(identifier)
except TypeError:
    pass
```

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            if value == '':
+                continue
+            # Do not accept NaNs from dataframes
+            if isinstance(value, float) and np.isnan(value):
+                continue
+            entry[key] = value
+
+        # Check if the sid is declared
+        try:
+            entry['sid']
```
llllllllll: Because we are not trying to use the `sid` here, this should just be an `if 'sid' not in entry:`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        # If the file_name is in the kwargs, it will be used as the symbol
+        try:
+            entry['symbol'] = entry.pop('file_name')
+        except KeyError:
+            pass
+
+        # If the identifier coming in was a string and there is no defined
+        # symbol yet, set the symbol to the incoming identifier
+        try:
+            entry['symbol']
```
llllllllll:
```python
if isinstance(identifier, string_types):
    entry.setdefault('symbol', identifier)
```

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            pass
+        try:
+            entry['expiration_date'] = pd.Timestamp(entry['expiration_date'],
+                                                    tz='UTC')
+        except KeyError:
+            pass
+
+        # Build an Asset of the appropriate type, default to Equity
+        asset_type = entry.pop('asset_type', 'equity')
+        if asset_type.lower() == 'equity':
```
llllllllll: I would break the writing for equities and futures into seperate methods and then dispatch to them here.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            # currently just a short-circuit for compatibility with existing
+            # testing structure in the test_algorithm module which creates
+            # multiple sources which all insert redundant metadata.
+            return
+
+        entry = {}
+
+        for key, value in kwargs.items():
+            # Do not accept invalid fields
+            if key not in ASSET_FIELDS:
```
llllllllll: This should certianly raise a `TypeError` Does this function need to accept **kwargs here?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            c.execute('CREATE UNIQUE INDEX IF NOT EXISTS ix_futures_exchanges_en '   # noqa
+                      'ON futures_exchanges(exchange_id)')
+            c.execute('CREATE UNIQUE INDEX IF NOT EXISTS ix_futures_contracts_sid '  # noqa
+                      'ON futures_contracts(sid)')
+            c.execute('CREATE UNIQUE INDEX IF NOT EXISTS ix_futures_root_symbols_id '  # noqa
+                      'ON futures_root_symbols(root_symbol_id)')
+            c.execute('CREATE UNIQUE INDEX IF NOT EXISTS ix_asset_router_sid  '
+                      'ON asset_router(sid)')
+
+        # Note: Also need a max_date table.
```
llllllllll: Do we still need this?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        """
+
+        raise NotImplementedError('load_data')
+
+
+class AssetDBWriterFromDictionary(AssetDBWriter):
+    """
+    Class used to write dictionary data to SQLite database.
+
+    Expects a dictionary to be passed to load_data
```
llllllllll: This is no longer true

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            A connection to a SQLite database.
+        fuzzy_char: string
+            A string for use in fuzzy matching.
+        allow_sid_assignment: boolean
+            If True then the class can assign sids where necessary.
+        constraints: boolean
+            If True, create SQL ForeignKey and Index constraints.
+
+        """
+        self.fuzzy_char = fuzzy_char
```
llllllllll: These all feel like `__init__` setup code. We are adding new attributes to this object in non `__init__` methods.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        self.allow_sid_assignment = allow_sid_assignment
+        if allow_sid_assignment:
+            ts = normalize_date(pd.Timestamp('now', tz='UTC'))
+            # Store as seconds since UNIX Epoch for compatibility
+            # with SQL.
+            self.end_date_to_assign = (ts.value // 10 ** 9)
+
+        # Store a nested-dict of all metadata for
+        # reference when building Assets
+        self.metadata_cache = {}
```
llllllllll: where is this used?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+        """
+        Returns a generator yielding pairs of (identifier, metadata)
+        """
+        for identifier, row in dataframe.iterrows():
+            yield identifier, row.to_dict()
+
+
+class AssetDBWriterLegacyFromReadable(AssetDBWriterLegacy):
+    """ An implementation of AssetDBWriter for use
+        with objects with a 'read' property.
```
llllllllll: `read` method, not property

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+
+        The object's read method must return rows
```
llllllllll: we should rephrase this to `must return an iterable of mapping objects where the mappings have at least one of 'sid' or 'symbol' along with ...`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,779 @@
...
+            if isinstance(identifier, Asset):
+                sid = identifier.sid
+                metadata = identifier.to_dict()
+                metadata['asset_type'] = identifier.__class__.__name__
+                self.write_block(sid, **metadata)
+            else:
+                self.write_block(identifier)
+
+
+class NullAssetDBWriterLegacy(AssetDBWriterLegacy):
```
llllllllll: This name breaks the convention set by the other legacy writers.

llllllllll:
I know this is a WIP, but we will need docs on this feature.

zipline/assets/asset_writer.py
StewartDouglas: Since we will either deprecate this code in the future or remove it entirely, I would suggest leaving it as is.

llllllllll: We haven't deprecated it and we haven't removed it so we still need to support this code. The current state masks the intent makeing it harder to update and maintain later.

StewartDouglas:
Closing as this PR has been superseded by: https://github.com/quantopian/zipline/pull/674, although I'll still reference it for Joe's feedback.

PR #665, opened 2015-07-30
ENH: Adds auto-closing feature and implements for Futures

jfkirk:
@StewartDouglas could you give this a read-through?

@ehebert this is the feature I spoke to you about this morning.

### DIFF
--- a/zipline/finance/performance/position_tracker.py
+++ b/zipline/finance/performance/position_tracker.py
@@ -11,6 +11,7 @@ except ImportError:
     from collections import OrderedDict
 from six import iteritems, itervalues
 
+from zipline.protocol import Event, DATASOURCE_TYPE
 from zipline.finance.slippage import Transaction
 from zipline.utils.serialization_utils import (
     VERSION_LABEL
@@ -42,11 +43,15 @@ class PositionTracker(object):
         )
         self._positions_store = zp.Positions()
 
+        # Dict, keyed on dates, that contains lists of close position events
+        # for any Assets in this tracker's positions
+        self._auto_close_position_sids = {}
+
     @with_environment()
     def _retrieve_asset(self, sid, env=None):
         return env.asset_finder.retrieve_asset(sid)
 
-    def _update_multipliers(self, sid):
+    def _update_asset(self, sid):
         try:
             self._position_value_multipliers[sid]
             self._position_exposure_multipliers[sid]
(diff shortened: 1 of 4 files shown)


zipline/finance/performance/tracker.py
```diff
@@ -413,6 +408,22 @@ def check_upcoming_dividends(self, completed_date):
             # notify periods to update their stats
             period.handle_dividends_paid(net_cash_payment)
 
+    def check_asset_auto_closes(self, next_trading_day):
+        """
+        Check if we currently own any futures whose notification_date is
```
StewartDouglas: Sometimes we use the term notice date, and sometimes notification date. Let's stick to notice date, since it's shorter and is used on CME's website.

jfkirk: Good catch, thanks!

zipline/finance/performance/position_tracker.py
```diff
@@ -64,6 +69,71 @@ def _update_multipliers(self, sid):
...
+            The SID of the Asset to be auto-closed
+        """
+        date_sids = self.auto_close_position_dates.setdefault(dt, [])
+        if sid not in date_sids:
+            date_sids.append(sid)
+
+    def auto_close_position_events(self, next_trading_day):
+        """
+        Generates CLOSE_POSITION events for any SIDs whose auto-close date is
+        before the given date.
```
StewartDouglas: before *or equal* to the given date?

StewartDouglas:
Looks good to me.

zipline/finance/performance/position_tracker.py
```diff
@@ -64,6 +69,71 @@ def _update_multipliers(self, sid):
...
+        Yields
+        ------
+        Event
+            A close position event for any sids that should be closed before
+            the next_trading_day parameter
+        """
+        past_asset_end_dates = set()
+
+        # Check the auto_close_position_dates dict for SIDs to close
+        for date, sids in self.auto_close_position_dates.items():
```
richafrank: Does the order of iteration here matter?

jfkirk: No, but one potential optimization is to pre-sort the auto-close dates on insertion, such that we can bail out of the generator as soon as we hit the first date after next_trading_day. What do you think?

zipline/finance/performance/position_tracker.py
```diff
@@ -64,6 +69,71 @@ def _update_multipliers(self, sid):
...
+
+        Parameters
+        ----------
+        dt : pandas.Timestamp
+            The date before-which the given SID will be auto-closed
+        sid : int
+            The SID of the Asset to be auto-closed
+        """
+        date_sids = self.auto_close_position_dates.setdefault(dt, [])
+        if sid not in date_sids:
```
richafrank: Would it make more sense to have date_sids be a ```set``` and manage duplicates itself?

zipline/finance/performance/position_tracker.py
```diff
@@ -64,6 +69,71 @@ def _update_multipliers(self, sid):
...
+                # Yield a CLOSE_POSITION event
+                event = Event({
+                    'dt': date,
+                    'type': DATASOURCE_TYPE.CLOSE_POSITION,
+                    'sid': sid,
+                })
+                yield event
+
+        # Clear out past dates
+        while len(past_asset_end_dates) > 0:
```
richafrank: This could also be ```while past_asset_end_dates:```, since a set's truthiness is whether is has elements or not.

zipline/finance/performance/position_tracker.py
jfkirk: Cool idea - modifying now

zipline/finance/performance/position_tracker.py
richafrank: That definitely makes sense to me, though I don't know enough about the number of items here or the time it takes to pre-sort versus the time saved by bailing out.

jfkirk: I can't think of a good way to implement the pre-sort that wouldn't be just as expensive, if not more expensive, than this implementation. Do you know a good way that I could implement a sort-on-insertion?

zipline/finance/performance/position_tracker.py
```diff
@@ -64,6 +71,72 @@ def _update_multipliers(self, sid):
...
+        Parameters
+        ----------
+        dt : pandas.Timestamp
+            The date before-which the given SID will be auto-closed
+        sid : int
+            The SID of the Asset to be auto-closed
+        """
+        if dt not in self._auto_close_position_sids:
+            bisect.insort(self._auto_close_position_dates, dt)
+        self._auto_close_position_sids.setdefault(dt, set()).add(sid)
```
jfkirk: @richafrank I've implemented an sort-on-insert list of keys that can then be iterated through later, allowing us to break out quickly. This should be more performant. What do you think?

zipline/finance/performance/position_tracker.py
```diff
@@ -64,6 +71,72 @@ def _update_multipliers(self, sid):
...
+        Event
+            A close position event for any sids that should be closed before
+            the next_trading_day parameter
+        """
+        past_asset_end_dates = set()
+
+        # Check the auto_close_position_dates dict for SIDs to close
+        for date in self._auto_close_position_dates:
+            if date > next_trading_day:
+                break
```
jfkirk: @richafrank This is the iteration-with-breakout

zipline/finance/performance/position_tracker.py
richafrank: Implementation-wise, probably an ```OrderedDict``` would be clearer.  I can't speak to the performance impact here without any stats, but if you say it's more performant :+1: !

richafrank: OrderedDict uses insertion order - not sure if that's the case here.

jfkirk: The catch with an OrderedDict is that it maintains order of insertion, and the only way to re-order is to re-build the dict entirely. I figured that was too expensive for our purposes, so keeping a separate, ordered, insertable list seemed the most elegant way to keep the iteration ordered.

Is there a more elegant way to re-tool this, or do you think this one is ready to go?

richafrank: I don't have any sense of the performance impact here, so if I'm judging just on elegance, I'd say go back to not breaking out early...

jfkirk:
@richafrank I reverted to the simpler logic.

richafrank:
> @richafrank I reverted to the simpler logic.

Cool - sounds fine.

jfkirk:
Perf checks out - merging.

PR #668, opened 2015-08-04
BUG: Improves lookup_future_chain to handle NaT date args

yankees714:
If lookup_future_chain was provided with an as_of_date or knowledge date that was pandas.NaT, the query we were forming wasn't what we want. Instead, as_of_date, if not NaT, is used for knowledge_date, and if both are NaT, no date filtering is done in the query.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -587,6 +587,10 @@ class AssetFinderTestCase(TestCase):
         self.assertEqual(ad_contracts[0].sid, 1)
         self.assertEqual(ad_contracts[1].sid, 0)
 
+        # Check that pd.NaT for knowledge_date uses the value of as_of_date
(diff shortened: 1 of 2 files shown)


ehebert:
The logic looks sound to me, two pieces may make this complete.

1) Update the doc string the with the combinations of parameters and the resulting query parameters, i.e. "both `as_of_date` and `knowledge_date` are not NaT, returns values where the notice_date is greater than the as_of_date and the start_date is less than the knowledge date", "both `as_of_date` and `knowledge_date` are NaT, the query is unbounded."... etc.
2) Make sure the unit tests (if they don't already) cover each of the defined cases.

yankees714:
@ehebert I changed the logic a bit, and added the updated tests.

ehebert:
Looks good to me. :+1: on removing the extra check of `knowledge_date is pd.NaT`

PR #669, opened 2015-08-05
MAINT: Read requirements from setup.py

ssanderson:
@llllllllll for you :heart:.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@
 
 import re
 import sys
-from operator import lt, gt, eq
+from operator import lt, gt, eq, le, ge
 from os.path import (
     abspath,
     dirname,
@@ -70,10 +70,17 @@ def _filter_requirements(lines_iter):
             requirement, version_spec = line.split(';')
             try:
                 groups = re.match(
-                    "(python_version)([<>=]+)(')([0-9\.]+)(')(.*)",
+                    "(python_version)([<>=]{1,2})(')([0-9\.]+)(')(.*)",
                     version_spec,
                 ).groups()
-                comp = {'<': lt, '=': eq, '>': gt}[groups[1]]
+                comp = {
+                    '<': lt,
+                    '<=': le,
+                    '=': eq,
+                    '==': eq,
+                    '>': gt,
+                    '>=': ge,
+                }[groups[1]]
                 version_spec = StrictVersion(groups[3])
             except Exception as e:
                 # My kingdom for a 'raise from'!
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -14,32 +14,101 @@
...
+            continue
+
+        # pip install -r understands line with ;python_version<'3.0', but
+        # whatever happens inside extras_requires doesn't.  Parse the line
+        # manually and conditionally add it if needed.
+        if ';' in line:
+            requirement, version_spec = line.split(';')
+            try:
+                groups = re.match(
+                    "(python_version)([<>=]+)(')([0-9\.]+)(')(.*)",
```
llllllllll: `\2` should be `[<>=]{1, 2}` or we get things like js super equals `===`

setup.py
```diff
@@ -14,32 +14,101 @@
...
+                    version_spec,
+                ).groups()
+                comp = {'<': lt, '=': eq, '>': gt}[groups[1]]
+                version_spec = StrictVersion(groups[3])
+            except Exception as e:
+                # My kingdom for a 'raise from'!
```
llllllllll: you can emulate this in python 2 by raising a tuple of `(value, None, tb)`

setup.py
```diff
@@ -14,32 +14,101 @@
...
+def _filter_requirements(lines_iter):
+    for line in lines_iter:
+        line = line.strip()
+        if not line or line.startswith('#'):
+            continue
+
+        # pip install -r understands line with ;python_version<'3.0', but
+        # whatever happens inside extras_requires doesn't.  Parse the line
+        # manually and conditionally add it if needed.
+        if ';' in line:
```
llllllllll: does it need to start with this?

setup.py
```diff
@@ -14,32 +14,101 @@
...
+            requirement, version_spec = line.split(';')
+            try:
+                groups = re.match(
+                    "(python_version)([<>=]+)(')([0-9\.]+)(')(.*)",
+                    version_spec,
+                ).groups()
+                comp = {'<': lt, '=': eq, '>': gt}[groups[1]]
```
llllllllll: what about `>=, `==', `<=`, ...

setup.py
```diff
@@ -14,32 +14,101 @@
...
+                version_spec = StrictVersion(groups[3])
+            except Exception as e:
+                # My kingdom for a 'raise from'!
+                raise ValueError(
+                    "Couldn't parse requirement line; '%s'\n"
+                    "Error was:\n"
+                    "%r" % (line, e)
+                )
+
+            sys_version = '.'.join(list(map(str, sys.version_info[:3])))
```
llllllllll: join accepts an iterable

setup.py
```diff
@@ -14,32 +14,101 @@
...
+    abspath,
+    dirname,
+    join,
+)
+from distutils.version import StrictVersion
 from setuptools import setup, Extension
-from Cython.Build import cythonize
-import numpy as np
+try:
+    from Cython.Build import cythonize
```
llllllllll: How are we putting this on pypi now?

llllllllll:
This is a good change, just had a few comments / questions.

setup.py
ssanderson: The line that this is for is `backports.ssl-match-hostname==3.4.0.2;python_version<'3.0'`.

llllllllll: kk

PR #670, opened 2015-08-05
ENH: sqlalchemy

llllllllll:


zipline/assets/assets.py
```diff
@@ -526,33 +538,38 @@ def map_identifier_index_to_sids(self, index, as_of_date):
...
+        buf = np.array(
+            tuple(map(
+                float,
+                sa.select((
+                    equities_cols.sid,
+                    equities_cols.start_date,
+                    equities_cols.end_date,
+                )).execute(),
+            )),
+            dype='<f8',  # use doubles so we get NaNs
```
ssanderson: Can we just read integers with default values instead of copying this whole array twice?

llllllllll: when you say "read with deafault values", you mean from the db?

ssanderson: Yes.  Or better yet, have those column be non-nullable and use defaults on insertion.

zipline/assets/asset_writer.py
```diff
@@ -86,232 +88,175 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
...
-
-        db_conn.commit()
+        self.sql_metadata = metadata = sa.MetaData(bind=engine)
+        self.equities = sa.Table(
+            'equities',
+            metadata,
+            sa.Column(
+                'sid',
+                sa.Integer,
+                nullable=False,  # in case constraints is False
```
StewartDouglas: Can we remove this line? If I call ```init_db``` with ```constraints=False``` I don't expect to be subject to any constraints - but this enforcing one.

zipline/assets/asset_writer.py
```diff
@@ -86,232 +88,175 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
...
+            sa.Column('end_date', sa.Integer),
+            sa.Column('first_traded', sa.Integer),
+            sa.Column('exchange', sa.Text),
+            sa.Column('fuzzy', sa.Text),
+        )
+        self.futures_exchanges = sa.Table(
+            'futures_exchanges',
+            metadata,
+            sa.Column(
+                'echange_id',
```
StewartDouglas: typo

zipline/assets/asset_writer.py
llllllllll: I was trying to match the old behavior. You create the sid column with `NOT NULL` so this just mimics that

StewartDouglas: Ok that's my bad. I'll remove these lines. The ```NOT NULL``` constraint will be enforced when we set ```constraints=True``` through our ```INTEGER PRIMARY KEY```.

richafrank:
Do we need to add sqlalchemy to the requirements?

zipline/assets/asset_writer.py
```diff
@@ -81,237 +83,176 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
...
-                      'ON asset_router(sid)')
-
-        # Note: Also need a max_date table.
-
-        db_conn.commit()
+        self.sql_metadata = metadata = sa.MetaData(bind=engine)
+        self.equities = sa.Table(
+            'equities',
+            metadata,
+            sa.Column(
```
ssanderson: Does this enforce a uniqueness contraint by default in sqlite?  If not, can we add a unique constraint on this column (and the equivalent ones on other tables)?  FFC will always crash if it gets duplicated asset ids, and the whole point of this column is that it's a unique ID.

StewartDouglas: This will be unique if ```primary_key=True```, which in turns will depend on ```constraints=True```. We should change the default of ```constraints``` to be ```True``` so that it will be compatible with FFC.

ssanderson: I'd say that regardless of whether `constraints` defaults to something, there should always be a unique constraint on this column.  Having multiple entries in the database for a given sid is always a bug.

StewartDouglas: Ok that's a fair point. In that case I also suggest we include a ```NOT NULL``` constraint on the sid column (undoing my earlier commit) as even with the ```UNIQUE``` constraint we could still have multiple ```NULLs``` in the columns, which may cause FFC issues and at the very least seems like a bad thing to do:

```sql
CREATE TABLE example (sid INTEGER UNIQUE);
INSERT INTO example VALUES ((NULL),(NULL),(NULL));
-- example now contains three NULLs --

ssanderson: :+1:

StewartDouglas:
Joe added sqlalchemy to the requirements.txt. Merging.

PR #672, opened 2015-08-07
ENH: adds auto_close as a simulation parameter

warren-oneill:
Allows users to turn off automatic closing of futures by setting an ```auto_close``` flag to ```False``` in the simulation parameters.
cc @jfkirk @StewartDouglas @yankees714

### DIFF
--- a/tests/serialization_cases.py
+++ b/tests/serialization_cases.py
@@ -67,7 +67,7 @@ def object_serialization_cases(skip_daily=False):
         (PerformancePeriod,
             (10000,), {'position_tracker': PositionTracker()}, 'to_dict'),
         (Position, (8554,), {}, 'dict'),
-        (PositionTracker, (), {}, 'dict'),
+        (PositionTracker, (), {'auto_close': True}, 'dict'),
         (PerformanceTracker, (sim_params_minute,), {}, 'to_dict'),
         (RiskMetricsCumulative, (sim_params_minute,), {}, 'to_dict'),
         (RiskMetricsPeriod,
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -31,6 +31,7 @@ from zipline.utils.test_utils import (
     teardown_logger
 )
 import zipline.utils.factory as factory
+from zipline.utils.factory import create_simulation_parameters
 import zipline.utils.simfactory as simfactory
 
 from zipline.errors import (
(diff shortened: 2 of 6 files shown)


tests/test_pickle_serialization.py
```diff
@@ -18,7 +18,7 @@
 from nose_parameterized import parameterized
 from unittest import TestCase
 
-from .serialization_cases import (
+from tests.serialization_cases import (
```
warren-oneill: I was having problems in PyCharm with relative imports but I can revert this if needs be.

jfkirk:
@ssanderson What do you think? It makes sense to me to make auto-closing configurable, but I know that we are trying to wean ourselves off of the SimulationParameters.

warren-oneill:
@jfkirk @ssanderson another option would be to add the flag to the asset metadata. In this way we can have algos where some assets are auto closed and others not.

ssanderson:
Changes that modify global simulation behavior like this make me nervous; each individual such change is innocuous, but over time they create exponentially-many possible configurations in which Zipline could be running, some of which might be mutually-exclusive, and almost all of which will be untested, because we write roughly linearly-many tests, but produce exponentially-many configurations.

@warren-oneill can you talk about about why a user would or wouldn't want to auto-close positions on a contract?  I'm wondering if, for example, the right way to do this would be to add `close_date` as an optional parameter to `order`, which would default to the requested asset's expiration.

warren-oneill:
@ssanderson for us they are a couple of cases where we would want to leave the positions open after expiration day:
* as an energy producer we can use the futures to hedge against our production
* we trade also in financial futures where there is no danger of going into delivery but instead after expiration the open position are closed using a settlement price and we may want to bet on this price
* market outages are very common in our field and there may be situations where we are unable to close our position. This is something which we would like to have in our backtests.
* some products in our markets are not as liquid as stocks and it can be that one is unable to find a buyer/seller.

ssanderson:
@warren-oneill thoughts on an alternative proposed at https://github.com/quantopian/zipline/pull/676/files#r36980067 ?  I think baking this logic into the data model for Futures would make this logic much simpler rather than having to have `if auto_close_futures` branching everywhere, we just have `if now > future.auto_close_date`, which means we'd be exercising the same code paths at all times, which means fewer opportunities for bugs to hide.

The downside of baking this into the data model is that it'd be harder to toggle back and forth between auto_close=True and auto_close=False.  The upside would be that you'd have per-asset granularity on whether you want auto-closing behavior.  Thoughts on these tradeoffs?

warren-oneill:
closed in favour of #676

PR #673, opened 2015-08-10
WIP, ENH: allows cascading of assets into children products

warren-oneill:
In European gas and energy futures larger products (years, quarters) can be cascade into smaller "children" products (months, days). 
67304_ill_cascading
Examples can be found [here](http://kwa-analytics.com/access-level/public-access/european-gas-futures-cascade-rules-overview-1/) and [here](http://www.nasdaqomx.com/transactions/markets/commodities/markets/power/power-ds-futures)
This is very important functionality for us but I was unable to find any U.S. examples. Is cascading possible in North American markets? If not another usage of this functionality would be the opening of arbitrage possibility between the larger and smaller products.

To-dos:
- add algo unittest
- combine with auto-close (i.e. the first quarter of a year product is closed and the remaining 3 quarters are opened)

cc @jfkirk @StewartDouglas @yankees714 @dhexus

### DIFF
--- a/tests/test_perf_tracking.py
+++ b/tests/test_perf_tracking.py
@@ -2020,6 +2020,51 @@ class TestPerformanceTracker(unittest.TestCase):
                 # Test not-owned SID
                 self.assertIsNone(txn)
 
+    @with_environment()
+    def test_cascade(self, env=None):
+        metadata = {4: {'asset_type': 'future', 'children': [5]},
+                    5: {'asset_type': 'future', 'children': [4, 6]},
+                    6: {'asset_type': 'future'}}
+        env.update_asset_finder(asset_metadata=metadata)
+        pt = perf.PositionTracker()
+        dt = pd.Timestamp("1984/03/06 3:00PM")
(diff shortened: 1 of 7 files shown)


warren-oneill:
If this PR won't be accepted would it at least possible to make it easier to add new variables to the metadata. I'm struggling a bit to reconcile my additions with the latest version of master. 
cc @ssanderson

StewartDouglas:
FWIW I've not seen any mention of cascading in North American markets, so won't be a priority for us in the short term.

warren-oneill:
It could also be used as an arbitrage function between, eg, weeks and months.

warren-oneill:
Regarding the ease of use of the metadata. Would it not be possible to define all metadata variables in one array where one gives name and type? Currently assets.py, asset_writer.py and _assets.pyx have to be edited to add a new entry.

warren-oneill:
It would also be great to have an easier way to add new event types.

freddiev4:
Closing this due to inactivity / many conflicts.

PR #674, opened 2015-08-11
Create AssetDBWriter, update TradingEnvironment and tests

StewartDouglas:
DO NOT MERGE

@jfkirk @richafrank @ssanderson 

TL;DR: I’m opening up this PR  as a means of soliciting feedback on the new ```write_data``` method of the TradingAlgorithm and the associated changes to the test suite.

Going forward the AssetFinder will exclusively be a reader class. Therefore we should deprecate the TradingEnvironment’s ```update_asset_finder``` method. In it’s place I have created a new method, ```write_data```, which will write data to the underlying database using the AssetDBWriter class (using either the TradingEnvironment’s existing database engine, or the engine you specify in the ```write_data``` method).

This change in turn necessitates various changes across our test suite. Currently we write data to our underlying storage in a number of different places (which quickly becomes an issue with the enforcement of SQL ```UNIQUE``` constraints.)

**Test suite changes**

The creation of the AssetDBWriter creates an opportunity for some important changes to the test suite. 

Remove unnecessary writing functionality from:

- ```SpecificEquityTrades``` in zipline/sources/test_source.py
- ```create_trade_history()``` in zipline/utils/factory.py
- ```RandomWalkSource``` in zipline/sources/simulated.py

If a test wants to write data to the db, they must either 1) use the TradingEnvironment’s write_data method directly via:

```
trading.environment.write_data(equities_data = ..., futures_data = ...)
```

or 2) supply it in an instance of the TradingAlgorithm (however currently this only handles equity data)

```
TradingAlgorithm(asset_metadata=data)
```

If a test wishes to reuse sids (popular choices for testing are 0, 1 or 133) it it should instantiate a new trading environment by calling:

```trading.environment = trading.TradingEnvironment()```

This will initialize a new TradingEnvironment and with it a new empty in-memory database. This in particular is a point I would like feedback on. An alternative would be to add a method ‘update_engine’ to the TradingEnvironment, which would be less costly.

As an example of these changes taking effect I have updated **tests/test_algorithm.py to be consistent with the updates - I would particularly appreciate feedback on the changes I've made to this file.**

**To do**:
- [x] Update outstanding tests
- [x] Refactor TradingAlgorithm to accept other data (e.g. futures), not just equities.

### CODE REVIEW

File: zipline/finance/trading.py [hunk 1]
```diff
@@ -149,44 +157,67 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         # stack.
         return False
 
-    def update_asset_finder(self,
-                            clear_metadata=False,
-                            asset_finder=None,
-                            asset_metadata=None,
-                            identifiers=None):
+    def write_data(self,
```
jfkirk: Should this method be named something like 'write_asset_metadata'? TradingEnvironment.write_data seems really vague to me - from the outside, I wouldn't know what data was being written.

zipline/algorithm.py
```diff
@@ -204,11 +204,11 @@ def __init__(self, *args, **kwargs):
         # Update the TradingEnvironment with the provided asset metadata
         self.trading_environment = kwargs.pop('env',
                                               TradingEnvironment.instance())
-        self.trading_environment.update_asset_finder(
-            asset_finder=kwargs.pop('asset_finder', None),
-            asset_metadata=kwargs.pop('asset_metadata', None),
-            identifiers=kwargs.pop('identifiers', None)
+        self.trading_environment.write_data(
```
jfkirk: If the desired pattern now is to write the metadata directly to the environment, should we just go ahead and scrap the ability to write metadata here?

I still do see a use case for providing identifiers -- its the simplest way to spin up an algorithm. However, if the user wants to provide more metadata about their assets, we might now be wanting them to do it directly.

Also, here we are assuming that all metadata coming in is equity data?

jfkirk:
Self-assigning, as I am taking over this branch

tests/test_algorithm.py
```diff
@@ -360,7 +364,7 @@ def test_asset_lookup(self):
     def test_future_chain(self):
         """ Tests the future_chain API function.
         """
-
+        trading.environment = trading.TradingEnvironment()
```
llllllllll: This pattern can potentially be cleaned up with the `with_env` decorator.

zipline/algorithm.py
llllllllll: Why is the pattern that we write data to the environment? I am not sure why we need to store this as global state.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,679 @@
...
+        """
+        Returns a standard set of pandas.DataFrames:
+        equities, futures, exchanges, root_symbols
+        """
+
+        data = self._load_data()
+
+        ###############################
+        # Generate equities DataFrame #
+        ###############################
```
llllllllll: We don't really need these fancy boxes, a colon at the end would suffice.

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,679 @@
...
+                         exchanges=exchanges_output,
+                         root_symbols=root_symbols_output)
+
+    def convert_datetime(self, dt):
+        """Convert a datetime variable to integer of nanoseconds
+           since UNIX Epoch.
+
+        Parameters
+        ----------
+        dt
```
llllllllll: missing a type here, `: datetime-coercable` maybe?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,679 @@
...
+            A string, int or pd.Timestamp instance representing a datetime.
+
+        Returns
+        -------
+        float
```
llllllllll: `ns : float`

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,679 @@
...
+            nanoseconds since UNIX Epoch.
+
+        """
+
+        if pd.isnull(dt):
+            return None
```
llllllllll: The docstring above does not mention that this is the case. Should this not just raise some error? Do we need to explicitly guard against this?

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,679 @@
...
+        except TypeError:
+            dt = pd.Timestamp(dt).tz_convert('UTC')
+
+        # Get seconds from UNIX Epoch
+        total_seconds_from_epoch = self._seconds_from_unix_time(dt)
+
+        # Return nanoseconds since UNIX Epoch
+        return int(total_seconds_from_epoch * 1000000000)
+
+    def _seconds_from_unix_time(self, dt):
```
llllllllll: https://github.com/quantopian/qcommons/blob/master/qcommons/date_utils.py#L11

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,679 @@
...
+        float
+            Seconds between dt and UNIX Epoch.
+
+        """
+        epoch = pd.to_datetime(0, utc=True)
+        delta = dt - epoch
+        return delta.total_seconds()
+
+    @staticmethod
+    def dict_subset(dict_, subset):
```
llllllllll: This wants top be a free function

zipline/assets/asset_writer.py
```diff
@@ -0,0 +1,679 @@
...
+
+EQUITY_TABLE_FIELDS = ASSET_TABLE_FIELDS
+
+EXCHANGE_TABLE_FIELDS = frozenset({
+    'exchange_id',
+    'exchange',
+    'timezone'
+})
+
+ROOT_SYMBOL_TABLE_FIELDS = ({
```
llllllllll: I think this should be `frozenset({`

zipline/assets/asset_writer.py
jfkirk: qcommons aren't available in Zipline - what do you recommend?

llllllllll: oh, sorry about that

zipline/assets/_assets.pyx
```diff
@@ -232,6 +232,7 @@ cdef class Future(Asset):
     def __cinit__(self,
                   int sid, # sid is required
                   object symbol="",
+                  object root_symbol_id = "",
```
jfkirk: This should be cut.

tests/modelling/test_modelling_algo.py
```diff
@@ -200,6 +203,11 @@ def handle_data(context, data):
         # Do the same checks in before_trading_start
         before_trading_start = handle_data
 
+        # Create fresh trading environment as the algo.run()
+        # method will attempt to write data to disk, and could
+        # violate SQL constraints.
+        trading.environment = trading.TradingEnvironment()
```
jfkirk: This assignment is impotent - remove.

tests/test_algorithm_gen.py
```diff
@@ -196,14 +208,19 @@ def test_progress(self):
         Ensure the pipeline of generators are in sync, at least as far as
         their current dates.
         """
+        # Ensure we are pointing to the TradingEnvironment for this class
+        trading.environment = AlgorithmGeneratorTestCase.env
```
jfkirk: Impotent - remove

tests/modelling/base.py
```diff
@@ -48,15 +48,16 @@ class BaseFFCTestCase(TestCase):
...
+            equities_df=make_simple_asset_info(
                 assets,
                 self.__calendar[0],
                 self.__calendar[-1],
-            ),
-            db_path=':memory:',
-            create_table=True,
-        )
+            ))
+        self.__finder = self.__env.asset_finder
```
jfkirk: @ssanderson I added this for compatibility when rebasing

zipline/finance/trading.py
```diff
@@ -132,61 +113,95 @@ def __init__(
...
-        :param asset_metadata: A dict, DataFrame, or readable object
-        :param identifiers: A list of identifiers to be inserted
-        :return:
+        self.engine = engine = create_engine('sqlite:///%s' % asset_db_path)
+        AssetDBWriterFromDictionary().init_db(engine)
+        self.asset_finder = AssetFinder(engine)
+
+    def write_data(self,
+                   engine=None,
+                   equities_data={},
```
ssanderson: [Please don't use mutable values as default arguments.](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments)

jfkirk: Yep - Andrew called that out, too. Fixing now.

tests/modelling/base.py
ssanderson: We shouldn't be storing `__env` on self if nothing else uses it.

Also :-1: on the fact that I now need to construct an entire `TradingEnvironment` in order to make an `AssetFinder`.

.travis.yml
```diff
@@ -21,6 +21,26 @@ install:
   - conda install --yes -c https://conda.binstar.org/Quantopian numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy==$SCIPY_VERSION matplotlib Cython patsy statsmodels tornado pyparsing xlrd mock pytz requests six dateutil ta-lib logbook
   - pip install --upgrade pip nose-timer coverage
   - pip install -e .[dev]
+  - grep bottleneck== etc/requirements.txt | xargs pip install
```
ssanderson: I think this inadvertently reverted recent changes to this file?

zipline/finance/trading.py
jikamens: Ooh, feel the Python-fu :fire:

.travis.yml
jfkirk: Oh, interesting -- good catch, thanks.

zipline/assets/asset_writer.py
jfkirk: I slimmed this method down as suggested.

zipline/utils/serialization_utils.py
```diff
@@ -13,6 +13,66 @@
...
+
+
+def _persistent_load(persid, env):
+    if persid == AssetFinder.PERSISTENT_TOKEN:
+        return env.asset_finder
+    if persid == TradingEnvironment.PERSISTENT_TOKEN:
+        return env
+
+
+def dump_with_persistent_ids(obj, protocol=None):
```
jfkirk: @richafrank This is the method we discussed that can be used in place of pickle.dumps and performs the tokenization of the AssetFinder and TradingEnvironment.

There is a corresponding method below to be used in place of pick.load.

jfkirk:
@ssanderson @richafrank @ehebert This PR now contains a commit to remove the global TradingEnvironment and, by extension, the global AssetFinder. To compensate for this, the classes and methods that used the global env now accept the env/finder that it needs.

A TradingAlgorithm accepts an env as an arg. If an env is not provided, a default TradingEnvironment is build by the TradingAlgorithm, and all subclasses will use that TradingEnvironment.

zipline/utils/serialization_utils.py
richafrank: Looks right to me!  Naming-wise, if we want to match the schema that pickle uses, these should probably be called ```dumps_with...``` and ```loads_with...``` (with an 's'), since they return/take bytes as opposed to files.

jfkirk: Ah, hadn't noticed the distinction. Updating now - thanks!

jfkirk: @richafrank These methods are using dump() and load() internally - does it still make sense to refactor the methods to loads_with_persistent_ids() and dumps_with_persistent_ids()

jfkirk:
1. Add class tear-downs the del the env
2. TradingEnvironment init kwargs for futures metadata
3. PositionTracker pickling unnecessary OrderedDicts
4. Cumulative risk metrics treasury curve serialization

PR #676, opened 2015-08-13
ENH: Adds auto_close_date field to Future objects

jfkirk:
This commit adds the auto_close_date property to Future objects, and it is calculated as the earliest of either the notice_date or expiration_date, or None if both of those dates are None.

@yankees714 This PR does not impact the AssetFinder's future chain logic at all. It may be appropriate for us to put in those changes in this PR once we figure out how we want it to look.

### DIFF
--- a/zipline/assets/_assets.pyx
+++ b/zipline/assets/_assets.pyx
@@ -227,6 +227,7 @@ cdef class Future(Asset):
     cdef readonly object root_symbol
     cdef readonly object notice_date
     cdef readonly object expiration_date
+    cdef readonly object auto_close_date
     cdef readonly float contract_multiplier
 
     def __cinit__(self,
@@ -238,6 +239,7 @@ cdef class Future(Asset):
                   object end_date=None,
                   object notice_date=None,
                   object expiration_date=None,
+                  object auto_close_date=None,
                   object first_traded=None,
                   object exchange="",
                   float contract_multiplier=1):
@@ -245,6 +247,7 @@ cdef class Future(Asset):
         self.root_symbol         = root_symbol
         self.notice_date         = notice_date
         self.expiration_date     = expiration_date
+        self.auto_close_date     = auto_close_date
         self.contract_multiplier = contract_multiplier
 
     def __str__(self):
@@ -256,7 +259,7 @@ cdef class Future(Asset):
     def __repr__(self):
         attrs = ('symbol', 'root_symbol', 'asset_name', 'exchange',
                  'start_date', 'end_date', 'first_traded', 'notice_date',
-                 'expiration_date', 'contract_multiplier')
+                 'expiration_date', 'auto_close_date', 'contract_multiplier')
         tuples = ((attr, repr(getattr(self, attr, None)))
                   for attr in attrs)
         strings = ('%s=%s' % (t[0], t[1]) for t in tuples)
@@ -278,6 +281,7 @@ cdef class Future(Asset):
                                  self.end_date,
                                  self.notice_date,
                                  self.expiration_date,
+                                 self.auto_close_date,
                                  self.first_traded,
                                  self.exchange,
                                  self.contract_multiplier,))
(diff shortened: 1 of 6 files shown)


zipline/assets/_assets.pyx
```diff
@@ -251,6 +251,21 @@ cdef class Future(Asset):
         if self.end_date is None:
             self.end_date = expiration_date
 
+    property auto_close_date:
```
ssanderson: @jfkirk is there a reason users can't specify this date explicitly?  I wonder if that would be a cleaner solution to @warren-oneill's need to toggle this behavior (https://github.com/quantopian/zipline/pull/672).  Then this just becomes a line in the constructor that looks like:

```
def __cinit__(self, expiration_date, notice_date, auto_close_date, ...):
    if auto_close_date is None:
        auto_close_date = min(self.notice_date, self.expiration_date)
    self.auto_close_date = auto_close_date
```
You could then effectively disable auto_closing by passing some far-future date as your "auto-close-date".  (or we could have a sentinel value that expliciltly means "never auto close".

jfkirk: Yep - Andrew and Stewart suggested the same thing, and I'm leaning toward it, too. It has a bigger impact on look-up in the AssetFinder, but that is a surmountable issue.

jfkirk:
@ssanderson I changed it to be a field of the Future, rather than a property.

I'm going to hold off on this PR until Stewart's writer/finder PRs are all in, then I'll rebase. That also buys us time to figure out if/how we want to change the finder for this.

warren-oneill:
@jfkirk @ssanderson looks good. Just have one question: is it possible to have an asset with an expiration date but without closing? It seems like this is the case but it wasn't so in an earlier version of this branch. Just want to make sure.

jfkirk:
@warren-oneill That is correct - if you do not set an auto_close_date on your Future objects, they will not be closed unless it is by your algorithm or by a CLOSE_POSITION event from your sources.

warren-oneill:
hey @jfkirk @ssanderson could you guys give me an estimate of when this could be in master? Want to know if I should start implementing on our own branch or hold off till this is accepted.

ssanderson:
ping @jfkirk

jfkirk:
Assigning Stewart to see this through merge - Thanks, Stewart!

jfkirk:
This PR is going to be dependent on the asset-db-writer-update branch, which should be coming in this week.

jfkirk:
@StewartDouglas This PR is unblocked and re-based. Could you review?

StewartDouglas:
ACK

zipline/assets/assets.py
```diff
@@ -259,6 +259,10 @@ def _retrieve_futures_contract(self, sid):
                 data['expiration_date'] = pd.Timestamp(
                     data['expiration_date'], tz='UTC')
 
+            if data['auto_close_date']:
+                data['auto_close_date'] = pd.Timestamp(
+                    data['auto_close_date'], tz='UTC')
+
```
StewartDouglas: maybe doesn't belong in this PR, but we could collapse the above logic to the following:
```python
 date_strs = ['start_date', 'end_date', 'first_traded',
             'notice_date', 'expiration_date', 'auto_close_date']
for param in date_strs:
    if param in data:
        data[param] = pd.Timestamp(data[param], tz='UTC')
```

jfkirk: I like that, and it fits well parallel to the _convert_asset_str that we already do. Adding now.

jfkirk:
@yankees714 I don't see the None-check - can you link it for me?

yankees714:
Just that we were doing `if data['start_date']` before, not `if 'start_date' in data`.

https://github.com/quantopian/zipline/commit/4b2e358e408595d2727c6a0ca09de21d6497fd0b#diff-e38a95b451336da451f69f2b60840da6L244

jfkirk:
Ahh - makes sense. As-is, it would try to make timestamps out of Nones if Nones came in the dict. I will adjust accordingly. Thanks!

StewartDouglas:
LGTM :+1:

PR #679, opened 2015-08-17
BUG: Interface of load_adjusted_array is to return a list of arrays

richafrank:
Fixes MultiColumnLoader to return a list of adjusted arrays, not a list of lists of adjusted arrays.

### DIFF
--- a/tests/modelling/test_engine.py
+++ b/tests/modelling/test_engine.py
@@ -16,6 +16,7 @@ from pandas import (
     Int64Index,
     rolling_mean,
     Timestamp,
+    Series,
 )
 from pandas.util.testing import assert_frame_equal
 from testfixtures import TempDirectory
@@ -48,6 +49,7 @@ from zipline.utils.test_utils import (
     make_rotating_asset_info,
     make_simple_asset_info,
     product_upper_triangle,
+    check_arrays,
 )
 
 
@@ -464,9 +466,8 @@ class MultiColumnLoaderTestCase(TestCase):
                                       self.dates[2],
                                       self.dates[-1])
         self.assertIsNotNone(result)
-        self.assertIn('f', result)
-        self.assertFalse(result['f'].empty)
+        self.assertEqual({'f'}, set(result.columns))
         # (close - open) * window = (1 - 2) * 3 = -3
         # skipped 2 from the start, so that the window is full
-        self.assertEqual(list(result['f']),
-                         [-3] * len(self.assets) * (len(self.dates) - 2))
+        check_arrays(result['f'],
+                     Series([-3] * len(self.assets) * (len(self.dates) - 2)))
--- a/tests/modelling/test_engine.py
+++ b/tests/modelling/test_engine.py
@@ -430,3 +430,43 @@ class SyntheticBcolzTestCase(TestCase):
                 columns=assets,
             ),
         )
+
+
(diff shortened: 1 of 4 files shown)


tests/modelling/test_engine.py
```diff
@@ -430,3 +430,43 @@ def test_drawdown(self):
...
+        })
+
+        engine = SimpleFFCEngine(loader, self.dates, self.asset_finder)
+
+        factor = RollingSumDifference()
+
+        result = engine.factor_matrix({'f': factor},
+                                      self.dates[2],
+                                      self.dates[-1])
+        self.assertIsNotNone(result)
```
ssanderson: This check is implied by the subsequent checks.  Is there a specific reason you prefer to have this?

tests/modelling/test_engine.py
```diff
@@ -430,3 +430,43 @@ def test_drawdown(self):
...
+        self.assertIn('f', result)
```
ssanderson: Can we assert that `f` is the only key in the returned frame (e.g. `self.assertEqual({'f'}, set(result.columns)`)?

tests/modelling/test_engine.py
```diff
@@ -430,3 +430,43 @@ def test_drawdown(self):
...
+        self.assertFalse(result['f'].empty)
+        # (close - open) * window = (1 - 2) * 3 = -3
+        # skipped 2 from the start, so that the window is full
+        self.assertEqual(list(result['f']),
```
ssanderson: Rather than casting both of these to lists, can we use `zipline.utils.test_utils.check_arrays` to actually check the numpy arrays here?  I've had bugs in this code where I was expecting an array in one place and returning a Series or vice versa, and just casting both inputs to lists will mask that error.

tests/modelling/test_engine.py
richafrank: My thought was to be explicit in each check such that they fail, as opposed to producing errors from unexpected input.  If ```result``` is ```None```, I thought it would be nice to have the test fail with "unexpectedly None" as opposed to "TypeError: argument of type 'NoneType' is not iterable".

tests/modelling/test_engine.py
richafrank: Sounds good to me.

tests/modelling/test_engine.py
richafrank: Sounds great - will take a look.

tests/modelling/test_engine.py
ssanderson: > I thought it would be nice to have the test fail with "unexpectedly None" as opposed to "TypeError: argument of type 'NoneType' is not iterable".

The sentiment here makes sense to me, but on the other hand, it seems like this implies that we should put `self.assertIsNotNone` as the first check before every function we test anywhere, which seems incorrect to me.  I don't care super strongly either way though.

richafrank: I do tend to put extra asserts where I want to check something more basic before I check something more complex, so I guess it doesn't seem incorrect to me.  I'm going to leave it, squash, and merge, unless there's anything else!

ssanderson:
LGTM, :dragon:

richafrank:
Merged with https://github.com/quantopian/zipline/commit/30847a10a7ef2671f6ed770affd417d571236a0d

PR #681, opened 2015-08-19
Sanest setup

ssanderson:


ssanderson:
@jikamens @llllllllll I think this is a worthy successor to the previous attempts.

Changes:

- Use @jikamens LazyCythonizingList for ext_modules.
- Parse our requirements.txt to keep `install_requires` up to date.
- Create `extras` builds for `talib` and `dev`.
- Use `pip install -e .[dev]` on Travis to install Zipline before testing.

PR #684, opened 2015-08-21
BUG: Use NAN from numpy.

ssanderson:
MSVC doesn't define NAN in math.h because they only implement C89.

See http://tdistler.com/2011/03/24/how-to-define-nan-not-a-number-on-windows.


llllllllll:
lgtm

Issue #634, opened 2015-07-06
TradingAlgorithm.initialize accepts and silently ignores *args and **kwargs.


ssanderson:
c.f. https://github.com/quantopian/zipline/blob/master/zipline/algorithm.py#L285

These either shouldn't be accepted or should be forwarded to `_initialize`.

cc @jfkirk

jfkirk:
@ssanderson My understanding was that the *args and **kwargs are accepted because extensions of TradingAlgorithm that override TradingAlgorithm.initialize may accept other args. Is this not the case?

ssanderson:
@jfkirk if a subclass overrides `TradingAlgorithm.initialize`, then it doesn't matter if the base class accepts additional args, because the base class function isn't being being called.

jfkirk:
![](http://st.depositphotos.com/2602289/4400/i/950/depositphotos_44001039-Man-slapping-hand-on-head-to-say-duh.jpg)

Makes sense. Thanks!

richafrank:
Is this all set with #687 ?

twiecki:
Yes.

PR #687, opened 2015-08-25
BUG Forward initialize args and kwargs to user-defined function.

twiecki:


jfkirk:
LGTM - have you checked qexec compatibility? I don't know of anything explicit that would conflict, but I'm not sure.

ssanderson:
c.f. https://github.com/quantopian/zipline/issues/634

twiecki:
@jfkirk ping

StewartDouglas:
I don’t think we should be passing these args and kwargs to `self._initialize`.

`self._initiliaze` is set to whatever `initialize(context)` is in the algo code, which does not expect any args or kwargs. Consequently it will crash when we try to supply it with non-empty args and kwargs (which can happen, for example, when running algos with scripts/run_algo.py).

I’m looking at removing args and kwargs from the definition of the initialize method.

twiecki:
I think the use-case is that you would define your own `initialize(context, my_kwarg=5)` so that you can pass in parameters externally. 

I don't see when we would pass *args or **kwargs when calling via `run_algo.py`?

StewartDouglas:
`PYTHONPATH=. python scripts/run_algo.py -f zipline/examples/buyapple.py --start 2011-1-1 --end 2012-1-1`

If you set a break point in `def initialize` you'll see that kwargs has the key 'namespace'. This could be resolved however by simply 'poping' namespace, rather than our current logic: 

`self.namespace = kwargs.get('namespace', {})`

StewartDouglas:
Ok @twiecki I think we can incorporate your original suggestion, subject to me changing:

`self.namespace = kwargs.get('namespace', {})`
to
`self.namespace = kwargs.pop('namespace', {})`

in algorithm.py, and undoing a change I made to test_modelling_algo.py, which has since become unecessary after yesterday's big asset_writer merge (it passed an asset_finder to the TradingAlgorithm, which now does nothing other than cause self._initialize to barf).

Would you be ok if I make these changes and squash them into your original commit (to reduce commit noise and ensure that the tests suite passes for this commit.)

twiecki:
That sounds great.

To give a bit of context. This is really critical for parameter optimization and we have some great research that uses this feature.

StewartDouglas:
@ssanderson mind taking a quick look over this, as it touches one of the tests your wrote (although it simply undoes some changes that I had previously made)

StewartDouglas:
Qexec unit tests passing but 3 integration tests failing (investigating, but I think these are unrelated failures)

ssanderson:
Assuming this is still passing downstream, LGTM.

StewartDouglas:
My Qexec tests failures were due to local configuration issues, and are now passing. Just rebased. @twiecki can you merge if Jenkins passes.

twiecki:
@StewartDouglas Sure, can you let me know once it does?

StewartDouglas:
@twiecki sure

StewartDouglas:
Merged.

twiecki:
:+1:

PR #690, opened 2015-08-26
TEST: Run nosetests with timings.

ssanderson:
We already did this on travis.  Make it part of the normal test workflow
instead of a special case for travis.


ssanderson:
@ehebert putting this on your plate since it's relevant for keeping people cognizant of test timings.  Also ping @jikamens in case you have any reason to think this interacts strangely with nose-gevented-multiprocess.

PR #691, opened 2015-08-28
Add SidNotFound exception case

dmichalowicz:


zipline/assets/assets.py
```diff
@@ -529,33 +529,34 @@ def _lookup_generic_scalar(self,
...
-                matches.append(result)
+            except SidNotFound:
+                missing.append(asset_convertible)
+                return None
+            matches.append(result)
 
-            elif isinstance(asset_convertible, string_types):
-                # Throws SymbolNotFound on failure to match.
+        elif isinstance(asset_convertible, string_types):
+            # Throws SymbolNotFound on failure to match.
```
ssanderson: This comment probably isn't necessary anymore given that it's in a try/catch block catching `SymbolNotFound`.

Issue #663, opened 2015-07-30
most examples don't work outside quantopian env


fraka6:
Hi guys,

Please update http://www.zipline.io/ because dual_moving_avg.py doesn't exist, its dual_moving_average.py. This strategy work perfectly in quantopian environment but not with zipline CLI.
I suspect the ichart.finance.yahoo path to be broken (see details at the end).

The only example that works or don't crash is  buyapple.  
dual_ema_talib.py -> ImportError: No module named talib
pairtrade.py-> ValueError: You must define a handle_data function.
olmar.py-> AttributeError: 'TradingAlgorithm' object has no attribute 'symbol'
buyapple_analyze.py-> ValueError: You must define a handle_data function.

Thanks, 
Fraka6
========
zipline.__version__ 
Out[39]: '0.7.0'

python scripts/run_algo.py -f zipline/examples/dual_moving_average.py --symbols AAPL --start 2011-1-1 --end 2012-1-1 -o dma.pickle
...
[2015-07-30 04:14:04.415502] WARNING: Loader: No benchmark data found for date range.
start_date=2015-07-30 00:00:00+00:00, end_date=2015-07-30 04:14:04.295081, url=http://ichart.finance.yahoo.com/table.csv?a=6&c=2015&b=30&e=30&d=6&g=d&f=2015&s=%5EGSPC
[2015-07-30 04:14:24.025207] INFO: Performance: Simulated 252 trading days out of 252.
[2015-07-30 04:14:24.025510] INFO: Performance: first open: 2011-01-03 14:31:00+00:00
[2015-07-30 04:14:24.025611] INFO: Performance: last close: 2011-12-30 21:00:00+00:00
Traceback (most recent call last):
...
  File "scripts/run_algo.py", line 25, in <module>
File "pandas/hashtable.pyx", line 694, in pandas.hashtable.PyObjectHashTable.get_item (pandas/hashtable.c:12231)
KeyError: 'AAPL'

dgreenberg1:
I'm having the same problem, getting the same error when trying to run dual_moving_average.py.

Has this issue been fixed?

twiecki:
Can you try with a more recent version?
`pip install -U git+https://github.com/quantopian/zipline`

fraka6:
I did tried it at the time with the latest version and just retried after a pull and I get the same error.

jfkirk:
Self-assigning to investigate.

twiecki:
[...]

PR #697, opened 2015-09-03
Fix examples

StewartDouglas:
Addresses the issues flagged here: https://github.com/quantopian/zipline/issues/663

- dual_moving_average_analyze.py now checks for data in the perf df prior to plotting. 
- *_analyze.py files have been moved to a utils folder to avoid confusion (and the cli.py file amended accordingly).
- Additionally I've made some cosmetic improvements to the graphical output of the algorithms.


twiecki:
How about removing the _analyze.py files all together and just putting it in the algorithm file itself? I'm aware that I was the one to come up with that weird construct in the first place. The motivation back then was to have that file run with minimal modifications on Quantopian, but maybe it's not worth it.

StewartDouglas:
I like the way that this construct allows the code to be run with minimal modifications on Quantopian. The downside I see is that it's quite opaque to the user where the analysis gets done, and if they write their own new algo and run it they might understandably assume that the analysis will be generated automatically, which is not the case. On balance I suggest either 1) sticking with the status quo, or; 2) moving the _analyze.py logic into the example files, but clearly commenting that this logic is unecessary if the algorithm is run on Quantopian.com. 

@jfkirk @ssanderson - thoughts?

StewartDouglas:
Alternatively, how about we scrap the logic in cli.py* which appends the analyze methods and instead include a line in each example file which imports the analyze functionality? In buyapple.py we would have the following line:
```python
from zipline.examples.utils.buyapple_analyze import analyze  # noqa
```
This has the benefit of minimzing the amount of code in the example files, aligning them as closely as possible with what a user would execute on Quantopian.com, while also making it clear where the analyze functionality actually comes from. Additionally, we could document the fact that users can themselves create analysis output for their own algorithm by adding a _analyze.py file to the examples/utils directory.

*https://github.com/quantopian/zipline/blob/master/zipline/utils/cli.py#L232-L236

twiecki:
@StewartDouglas I like that idea.

StewartDouglas:
Implemented the suggestion I made. Will rebase once https://github.com/quantopian/zipline/pull/698 is merged into master, then merge this PR.

tests/test_examples.py
```diff
@@ -43,3 +45,11 @@ class ExamplesTests(TestCase):
...
     def test_example(self, name, example):
         imp.load_source('__main__', os.path.basename(example), open(example))
+        # The algorithms buyapple.py and dual_moving_average.py have no
+        # if __name__ == '__main__' code block, and consequently are not
+        # executed by the imp.load_source() call. Instead, call these
+        # example algorithms directly as if using scripts/run_algo.py
+        if name == 'buyapple_py' or name == 'dual_moving_average_py':
+            confs = ['-f', example, '--start', '2011-1-1', '--end', '2012-1-1']
+            parsed_args = parse_args(confs)
+            run_pipeline(**parsed_args)
```
StewartDouglas: @twiecki  I realized that to date we have not actually been calling the buyapple.py or dual_moving_average.py examples algorithms, as they have no `if __name__ == ‘__main__’` code block. The above if-statement addresses this. The alternative was to include an `if __name__ == ‘__main__’` block in each of these algorithms, but I wanted to avoid this as it would introduce yet more differences between pure zipline code and that run on quantopian.com. This update will ensure we have test coverage for these two algorithms, and their corresponding analyze functions. The absence of this coverage resulted in the original user-reported error https://github.com/quantopian/zipline/issues/663.

What do you think about going one step further and removing the `if __name__ == ‘__main__’` from the other example algorithms? The advantage of this would be that there would only be only way of executing an algorithm from the command line, using `python scripts/run_algo.py`, which would clear up some user confusion, and users could more easily copy-paste into quantopian.com.

StewartDouglas: @jfkirk chatted about this. James suggested adding `if __name__ == '__main__'` blocks into the two examples which are currently missing them. This has the advantage of being consistent across all algos and it allows the user to execute the algorithm directly using `python zipline/examples/<algo>.py` without the need for lots of command line flags. There are pros and cons to both idea. @twiecki have any thoughts?

twiecki: Yeah, I think adding the if-statement makes sense. That way all examples are directly runnable but they are still examples of using `run_algo.py`.

StewartDouglas:
Ok, `if __name__ == '__main__'` blocks have been added. The test_examples.py file needed to be updated so that we cover both the code in these if-blocks as well as in the separate _analyze.py files. @twiecki @jfkirk mind taking another look? If you're happy I'll merge.

jfkirk:
ACK - reviewing

twiecki:
LGTM :+1:

StewartDouglas:
Going to rebase to consolidate some of the commits.

StewartDouglas:
Merged. Closing.

PR #698, opened 2015-09-03
Allow user to set the symbol lookup date

StewartDouglas:


zipline/algorithm.py
```diff
@@ -693,9 +697,14 @@ def symbol(self, symbol_str):
         Default symbol lookup for any source that directly maps the
         symbol to the Asset (e.g. yahoo finance).
         """
+        # If the user has not set the symbol lookup date,
+        # use the period_end as the date for sybmol->sid resolution.
+        _lookup_date = self.symbol_lookup_date if self.symbol_lookup_date \
```
jfkirk: What is the truthy-value of a datetime?-

zipline/algorithm.py
```diff
@@ -986,6 +995,15 @@ def set_commission(self, commission):
...
         self.commission = commission
 
+    @api_method
+    def set_symbol_lookup_date(self, dt):
+        """
+        Set the date for which symbols will be resolved to their sids
+        (symbols may map to different firms or underlying assets at
+        different times)
+        """
+        self.symbol_lookup_date = dt
```
jfkirk: Should we check and ensure the type of the dt passed in?

yankees714: You could just say `pd.Timestamp(dt)`, which will accept datetimes and Timestamps, plus strings, and take care of the check for you.

StewartDouglas: I thought about this. My initial reaction was to the variable flow through to here: https://github.com/quantopian/zipline/blob/master/zipline/assets/assets.py#L324-L325, but a more explicit try-except would probably be more useful. Will update.

zipline/algorithm.py
StewartDouglas: datetimes and pd.Timestamps are True, but confusingly the time component of them is False at UTC midnight:

```python
>>> dt = Timestamp('1970-01-01 00:00:00+0000', tz='UTC')
>>> bool(dt)
True
>>> midnight = Timestamp('1970-01-01 00:00:00+0000', tz='UTC').time()
>>> bool(midnight)
False
```

However, just to be extra-specially sure I'll add a `is not None` check.

zipline/algorithm.py
jfkirk: +1 for try-except raising a value error if its not Datetime-coerceable

zipline/algorithm.py
```diff
@@ -986,6 +995,15 @@ def set_commission(self, commission):
...
         self.commission = commission
 
+    @api_method
+    def set_symbol_lookup_date(self, dt):
+        """
+        Set the date for which symbols will be resolved to their sids
+        (symbols may map to different firms or underlying assets at
+        different times)
+        """
+        self.symbol_lookup_date = pd.Timestamp(dt)
```
StewartDouglas: What's the difference between the above and just calling:
```python
return Timestamp(obj, tz=tz)
```

ssanderson: Looks like there isn't any.

```In [12]: wtz = Timestamp('2014-01-01', tz='US/Eastern')

In [13]: wtz.tz_convert("UTC") == Timestamp(wtz, tz='UTC')
Out[13]: True
```
I'm not sure why I wrote it that way originally.

StewartDouglas:
Tests added

tests/test_algorithm.py
```diff
@@ -428,6 +429,55 @@ def test_future_chain(self):
...
+    def test_set_symbol_lookup_date(self):
+        """
+        Test the set_symbol_lookup_date API method.
+        """
+        # Note we start sid enumeration at i+3 so as not to
+        # collide with sids [1, 2] added in the setUp() method.
+        dates = pd.date_range('2013-01-01', freq='2D', periods=2, tz='UTC')
+        # Create two assets with the same symbol but different
+        # non-overlapping date ranges.
+        metadata = pd.DataFrame.from_records(
```
StewartDouglas: @jfkirk As mentioned, this will require refactoring when we merge the asset-db-writer-update branch, but should only involve moving the writing of this metadata into the `setUpClass` method and modifying the call to `TradingAlgorithm`

zipline/algorithm.py
```diff
@@ -225,6 +226,10 @@ def __init__(self, *args, **kwargs):
         # Set the dt initally to the period start by forcing it to change
         self.on_dt_changed(self.sim_params.period_start)
 
+        # The symbol lookup date specifies the date to use when resolving
+        # symbols to sids, and can be set using set_symbol_lookup_date()
+        self.symbol_lookup_date = None
```
jfkirk: Could you rename this to self._symbol_lookup_date just to be clear that the field shouldn't be externally referenced?

jfkirk:
One note about a variable name, but other than that this LGTM! :+1:

zipline/algorithm.py
StewartDouglas: good call

PR #701, opened 2015-09-08
PERF: ~40% on import of tradingcalendar.py.

ssanderson:


jfkirk:
Makes sense to me.

ssanderson:
I think this is going to be rendered obsolete by @jfkirk's latest.  Closing.

PR #702, opened 2015-09-14
MAINT: Removes unused __init__ from AssetDBWriter

jfkirk:
@StewartDouglas Mind checking this?


StewartDouglas:
Seems sensible. AssetDBWriter is an abstract class an so shouldn't be initialized anyway, and all its subclasses implement their own `__init__`. @yankees714 what do you think?

yankees714:
Sounds good to me :+1:

PR #704, opened 2015-09-14
MAINT: Moves market_minutes_for_day into tradingcalendar

yankees714:
This makes the function accessible directly from tradingcalendar. It previously was a method of TradingEnvironment, which meant you needed an env to use it.

### DIFF
--- a/zipline/utils/tradingcalendar.py
+++ b/zipline/utils/tradingcalendar.py
@@ -403,3 +403,9 @@ def get_open_and_closes(trading_days, early_closes, get_open_and_close):
 
 open_and_closes = get_open_and_closes(trading_days, early_closes,
                                       get_open_and_close)
+
+
+def market_minutes_for_day(stamp):
+    index = open_and_closes.index.get_loc(stamp.date())
+    todays_minutes = open_and_closes.values[index]
+    return pd.date_range(todays_minutes[0], todays_minutes[1], freq='T')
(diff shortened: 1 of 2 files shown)


zipline/utils/tradingcalendar.py
```diff
@@ -403,3 +403,9 @@ def get_open_and_closes(trading_days, early_closes, get_open_and_close):
 
 open_and_closes = get_open_and_closes(trading_days, early_closes,
                                       get_open_and_close)
+
+
+def market_minutes_for_day(stamp):
+    index = open_and_closes.index.get_loc(stamp.date())
```
jfkirk: Since this body is basically a dupe of TradingEnvironment.get_open_and_close, should we just move that method over to the calendar as well?

jfkirk:
How are the non-NYSE calendars impacted by this?

zipline/utils/tradingcalendar.py
yankees714: There's a different `get_open_and_close` in `tradingcalendar` already, which is what actually figures out the opens and closes, as opposed to just reading them. Maybe we could rename one of them?

ssanderson:
This looks like it's moving in the direction of https://github.com/quantopian/zipline/pull/556, where I built a proper object-oriented `TradingCalendar` class.  As of when I last worked on it, the only thing that was missing was implementations of the `lse` and `bmf` calendars.  Maybe we should just finish that work instead of migrating everything from `TradingEnvironment` to `TradingCalendar`?

yankees714:
Agreed that finishing #556 is the way to go here.

PR #705, opened 2015-09-15
MAINT: AssetFinder now takes fuzzy characters on look-up, not init

jfkirk:
Its more graceful for many applications to provide the Finder with a character on look-up, rather than provide a single fixed character on init.


jdricklefs:
Looks right to me!

PR #708, opened 2015-09-15
ENH: Simplifies AssetFinder symbol lookup by making fuzzy lookup the …

jfkirk:
…default

This branch adds two columns to the equities table: 'company_symbol' and 'share_class_symbol'.

Whenever a symbol is looked up in the equities table, the symbol is split on the first-occurring share class delimiter, and the search is performed using the split-up company and share-class symbols.

The symbols column is preserved, and its value is unchanged. This means that the Equity object returned has the 'symbol' columns value, meaning that the delimiter in the returned Equity is whatever delimiter came in on the original insertion.

@jdricklefs This is the change we discussed.
@kglowinski Thank you for your help!

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -292,12 +292,11 @@ class AssetFinderTestCase(TestCase):
             [
                 {
                     'sid': i,
-                    'symbol':  'TEST@%d' % i,
+                    'symbol':  'TEST.%d' % i,
                     'company_name': "company%d" % i,
                     'start_date': as_of.value,
                     'end_date': as_of.value,
-                    'exchange': uuid.uuid4().hex,
-                    'fuzzy': 'TEST%d' % i
+                    'exchange': uuid.uuid4().hex
                 }
                 for i in range(3)
             ]
@@ -308,25 +307,26 @@ class AssetFinderTestCase(TestCase):
             finder.retrieve_asset(i) for i in range(3)
         )
 
-        for i in range(2):  # we do it twice to test for caching bugs
+        # we do it twice to catch caching bugs
+        for i in range(2):
+            # Shouldn't find this with no fuzzy_str passed.
             self.assertIsNone(finder.lookup_symbol('test', as_of))
+            self.assertIsNone(finder.lookup_symbol('test1', as_of))
             self.assertEqual(
                 asset_1,
-                finder.lookup_symbol('test@1', as_of)
+                finder.lookup_symbol('test.1', as_of)
             )
 
             # Adding an unnecessary fuzzy shouldn't matter.
             self.assertEqual(
                 asset_1,
-                finder.lookup_symbol('test@1', as_of, fuzzy=True)
+                finder.lookup_symbol('test/1', as_of)
             )
 
-            # Shouldn't find this with no fuzzy_str passed.
-            self.assertIsNone(finder.lookup_symbol('test1', as_of))
             # Should find exact match.
             self.assertEqual(
                 asset_1,
-                finder.lookup_symbol('test1', as_of, fuzzy=True),
+                finder.lookup_symbol('test-1', as_of),
             )
 
     def test_lookup_symbol_resolve_multiple(self):
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -286,7 +286,7 @@ class AssetFinderTestCase(TestCase):
     def setUp(self):
         self.env = TradingEnvironment()
 
-    def test_lookup_symbol_fuzzy(self):
+    def test_lookup_symbol_delimited(self):
         as_of = pd.Timestamp('2013-01-01', tz='UTC')
         frame = pd.DataFrame.from_records(
             [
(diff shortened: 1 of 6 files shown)


jfkirk:
Note to self: Fix comment on AssetFinder.lookup_symbol

Also, we may want to deprecate either lookup_symbol or lookup_symbol_resolve_multiple as they are practically the same at this point, just with one difference on None-handling which could be done with an arg.

jfkirk:
@kglowinski This is the PR we discussed.

tests/test_assets.py
```diff
@@ -308,28 +307,20 @@ def test_lookup_symbol_fuzzy(self):
             finder.retrieve_asset(i) for i in range(3)
         )
 
-        for i in range(2):  # we do it twice to test for caching bugs
+        # we do it twice to catch caching bugs
+        for i in range(2):
+            # Shouldn't find this with no fuzzy_str passed.
```
kglowinski: I'm confused what this comment means- lookup symbol doesn't take a fuzzy str. Is what you meant rather that we shouldn't find either of the first two cases since the symbol string doesn't contain a fuzzy character?

tests/test_assets.py
```diff
@@ -308,28 +307,20 @@ def test_lookup_symbol_fuzzy(self):
...
+            self.assertEqual(asset_1, finder.lookup_symbol('test/1', as_of))
 
-            # Shouldn't find this with no fuzzy_str passed.
-            self.assertIsNone(finder.lookup_symbol('test1', as_of))
             # Should find exact match.
-            self.assertEqual(
-                asset_1,
-                finder.lookup_symbol('test1', as_of, fuzzy=True),
-            )
+            self.assertEqual(asset_1, finder.lookup_symbol('test-1', as_of))
```
kglowinski: Are this and the above case different?

tests/test_assets.py
jfkirk: Cruft comment from old version - will delete

tests/test_assets.py
jfkirk: It's using a different fuzzy character so, ideally, the result is the exact same asset.

kglowinski: I don't know that we need multiple tests, or if we do, maybe we should run each of the possible fuzzy characters in a nice comprehension.

Additionally, we should have a test for a fuzzy symbol that doesn't exist within the list and test that nothing is returned. As well as having one with two delimiters, to check that it's doing only one mapping.

zipline/algorithm.py
```diff
@@ -749,9 +749,10 @@ def symbol(self, symbol_str):
         _lookup_date = self._symbol_lookup_date if self._symbol_lookup_date is not None \
             else self.sim_params.period_end
 
-        return self.asset_finder.lookup_symbol_resolve_multiple(
+        return self.asset_finder.lookup_symbol(
             symbol_str,
-            as_of_date=_lookup_date
+            as_of_date=_lookup_date,
+            default_None=False,
```
kglowinski: Remind me- is the difference between default_None= True/False that None is returned vs empty string? Out of curiosity what's the use case for wanting it to return one vs the other?

jfkirk: By default, a call to lookup_symbol will return either an Equity or a None if no equity could be mapped. If you set default_None=False, a None will never be returned but, instead, a SymbolNotFound error will be raised.

ssanderson: > Out of curiosity what's the use case for wanting it to return one vs the other?

Legacy code support I imagine.  This is not behavior (I hope) that anyone wants or is excited about.

jfkirk: Correct - I am comfortable cutting that out and making SymbolNotFound catches explicit if you think that is wise.

jfkirk: You know what? I'm going to cut it. :yellow_heart:

kglowinski: Does that mean that every time we don't symbol map correctly, it's going to raise an error? tl;dr we need to try/except SymbolNotFound for our calls?

jfkirk: Yes, which makes sense to me. Right now we're split 50/50 between places that have default_none=True and do none-checking, and the other half are places that do try/catch. Now they would all just do try/catch.

tests/test_assets.py
```diff
@@ -286,50 +286,83 @@ class AssetFinderTestCase(TestCase):
...
-            )
-
-            # Shouldn't find this with no fuzzy_str passed.
-            self.assertIsNone(finder.lookup_symbol('test1', as_of))
-            # Should find exact match.
-            self.assertEqual(
-                asset_1,
-                finder.lookup_symbol('test1', as_of, fuzzy=True),
-            )
+            for fuzzy_char in ['-', '/', '_', '.']:
```
kglowinski: `for fuzz_char in set('-_/.'):`

zipline/algorithm.py
```diff
@@ -749,9 +749,9 @@ def symbol(self, symbol_str):
         _lookup_date = self._symbol_lookup_date if self._symbol_lookup_date is not None \
             else self.sim_params.period_end
 
-        return self.asset_finder.lookup_symbol_resolve_multiple(
+        return self.asset_finder.lookup_symbol(
             symbol_str,
-            as_of_date=_lookup_date
+            as_of_date=_lookup_date,
```
kglowinski: cruft.

ssanderson: Is this comment about the trailing comma here?  IMO that's correct.

zipline/assets/assets.py
```diff
@@ -261,28 +261,50 @@ def _retrieve_futures_contract(self, sid):
...
-        If no Asset was active at as_of_date raises SymbolNotFound.
+        If no Equity was active at as_of_date raises SymbolNotFound.
         """
+
+        # Format inputs
+        symbol = symbol.upper()
         if as_of_date is not None:
             as_of_date = pd.Timestamp(normalize_date(as_of_date))
 
+        company_symbol, share_class_symbol, fuzzy_symbol = \
```
kglowinski: Don't we really only want to do this if either we have fuzzy or if we have the first failure case? Because in the "normal" case, we're assuming the symbol is just a single complete, and then if multiple THAT'S when we try to split on delimiter.

kglowinski: No, I lied. We search fuzzy first. Carry on.

PR #710, opened 2015-09-17
more progress on tests

jbredeche:
Dataportal now supports daily mode for spot prices.

I think all the remaining test failures (and there are a bunch of them) in TestTransformAlgorithm are legitimate order-related failures.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -524,51 +524,43 @@ class TestTransformAlgorithm(TestCase):
 
     @classmethod
     def setUpClass(cls):
+        setup_logger(cls)
         futures_metadata = {3: {'asset_type': 'future',
                                 'contract_multiplier': 10}}
         cls.env = TradingEnvironment()
-        cls.env.write_data(equities_identifiers=[0, 1, 133],
+        cls.sim_params = factory.create_simulation_parameters(num_days=4,
+                                                              env=cls.env)
+        equities_metadata = {
+            0: {
+                'start_date': cls.sim_params.period_start,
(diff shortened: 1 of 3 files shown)


tests/test_algorithm.py
```diff
@@ -578,35 +571,35 @@ def test_invalid_order_parameters(self):
...
-        assert isinstance(algo.sources[0], DataFrameSource)
-
-    def test_panel_as_input(self):
-        algo = TestRegisterTransformAlgorithm(
-            sim_params=self.sim_params,
-            env=self.env,
-            sids=[0, 1])
-        algo.run(self.panel)
-        assert isinstance(algo.sources[0], DataPanelSource)
+    # def test_multi_source_as_input(self):
```
ehebert: I think we can just delete these as they won't be supported?

zipline/data/data_portal.py
```diff
@@ -94,6 +94,7 @@ def __init__(self,
         self.sources_map = {}
 
         self.sim_params = sim_params
+        self.daily_mode = (self.sim_params.data_frequency == "daily")
```
ehebert: Could we use `data_frequency` here instead? Mainly for consistency, since we do that elsewhere.

Also, I was a little surprised by this result:

```
In [9]:
daily_bar = True

In [10]:
%%timeit
daily_bar == True
10000000 loops, best of 3: 52 ns per loop


In [4]:
data_frequency = 'daily'

In [7]:
%%timeit
data_frequency == 'daily'
10000000 loops, best of 3: 38.8 ns per loop
```

PR #711, opened 2015-09-17
Normalize dates in AssetDateBounds control checks

jdricklefs:
@ssanderson @jfkirk As discussed. Seem ok to you guys?

Commit message:
```
Assumes that if a given asset's end_date is
e.g. 9/17/2015 00:00:00 UTC that it means the
asset is still tradeable on 9/17/2015 during
the market day.
```

### DIFF
--- a/zipline/finance/controls.py
+++ b/zipline/finance/controls.py
@@ -297,6 +297,10 @@ class AssetDateBounds(TradingControl):
         Fail if the algo has passed this Asset's end_date, or before the
         Asset's start date.
         """
+        # If the order is for 0 shares, then silently pass through.
+        if amount == 0:
+            return
+
         normalized_algo_dt = pd.Timestamp(algo_datetime).normalize()
 
         # Fail if the algo is before this Asset's start_date
--- a/zipline/finance/controls.py
+++ b/zipline/finance/controls.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 import abc
 
+import pandas as pd
+
 from six import with_metaclass
 
 from zipline.errors import (
@@ -55,14 +57,23 @@ class TradingControl(with_metaclass(abc.ABCMeta)):
         """
         raise NotImplementedError
 
-    def fail(self, asset, amount, datetime):
+    def fail(self, asset, amount, datetime, metadata=None):
         """
         Raise a TradingControlViolation with information about the failure.
+
+        If dynamic information should be displayed as well, pass it in via
+        `metadata`.
         """
+        constraint = repr(self)
+        if metadata:
+            constraint = "{constraint} (Metadata: {metadata})".format(
+                constraint=constraint,
+                metadata=metadata
+            )
         raise TradingControlViolation(asset=asset,
                                       amount=amount,
                                       datetime=datetime,
-                                      constraint=repr(self))
+                                      constraint=constraint)
 
     def __repr__(self):
         return "{name}({attrs})".format(name=self.__class__.__name__,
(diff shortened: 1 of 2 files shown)


ssanderson:
So are we not adding these to algos by default?

jdricklefs:
No, not yet anyway.

zipline/finance/controls.py
```diff
@@ -286,12 +288,34 @@ def validate(self,
...
+                self._fail_name = "asset_end_date"
+                self._fail_date = normalized_end
+                self.fail(asset, amount, algo_datetime)
+
+    def __repr__(self):
+        # More descriptive repr, since AssetDateBounds depend on a dynamic
+        # comparison
+        return "{name}(\"{fail_name}\":\"{fail_date})\"".format(
+            name=self.__class__.__name__,
+            fail_name=self._fail_name,
```
richafrank: It's great that this is more descriptive, but it's also confusing to me that I can't instantiate an AssetDateBounds and get its repr without having called validate first.  Also, if I call validate again, I'll see a different repr.  I think that a repr usually represents how you might instantiate this object, which this also fails on (no pun intended).

When are we looking at this repr?  Should whatever ```fail``` returns or raises have a better repr instead?

zipline/finance/controls.py
```diff
@@ -286,12 +288,34 @@ def validate(self,
...
+            normalized_end = pd.Timestamp(asset.end_date).normalize()
+            if normalized_algo_dt > normalized_end:
+                self._fail_name = "asset_end_date"
+                self._fail_date = normalized_end
+                self.fail(asset, amount, algo_datetime)
+
+    def __repr__(self):
+        # More descriptive repr, since AssetDateBounds depend on a dynamic
+        # comparison
+        return "{name}(\"{fail_name}\":\"{fail_date})\"".format(
```
richafrank: Probably easier to read if the outer quotes were single quotes.

zipline/finance/controls.py
```diff
@@ -286,12 +288,34 @@ def validate(self,
         Fail if the algo has passed this Asset's end_date, or before the
         Asset's start date.
         """
+        normalized_algo_dt = pd.Timestamp(algo_datetime).normalize()
+
+        # If the order is for 0 shares, then silently pass through.
+        if amount == 0:
```
richafrank: Could do this first, since the normalization is unnecessary in this case.

zipline/finance/controls.py
jdricklefs: Hm, good point about the errors if `validate` wasn't invoked. I'd tried to just throw some more data in self.__fail_kwargs during the failure, but that `self` wasn't the same `self` as in the base class repr (something to do with the metaclasses I think). 

My concern is the UX of failing this check being utterly unhelpful without the context of the asset's actual `start_date`/`end_date`.

Suppose I could add another kwarg to fail() to an additional dict to be passed?

richafrank: > Suppose I could add another kwarg to fail() to an additional dict to be passed?

Yea, that's what I was thinking too.  It might require updating or subclassing ```TradingControlViolation```, since it looks like that only expects the current kwargs and no others.

jdricklefs: Implemented this via optional pre-processing in `fail()`. 

```
We've detected that your algorithm would have violated the following trading
guard if it had been run from 2013-05-16 00:00:00+00:00 to 2013-05-16 13:31:00+00:00:

Order for 1 shares of Equity(14882 [BPLS]) at 2013-05-16 13:31:00+00:00 violates trading constraint
AssetDateBounds({}) (Metadata: {'asset_end_date': Timestamp('2002-01-01 00:00:00+0000', tz='UTC')}).

To debug this issue, run a backtest that includes the above dates.

```

zipline/finance/controls.py
```diff
@@ -286,12 +297,28 @@ def validate(self,
...
+                metadata = {
+                    'asset_start_date': normalized_start
+                }
+                self.fail(asset, amount, algo_datetime, metadata=metadata)
         # Fail if the algo has passed this Asset's end_date
-        if asset.end_date and (algo_datetime >= asset.end_date):
-            self.fail(asset, amount, algo_datetime)
+        if asset.end_date:
+            normalized_end = pd.Timestamp(asset.end_date).normalize()
+            if normalized_algo_dt > normalized_end:
```
richafrank: Just to confirm, the change from ```>=``` to ```>``` is intentional?

jdricklefs: Yes (and the test case I added is to confirm it). The normalized end_date of "9/17/2015 00:00" should mean "This asset is still trading on the day of 9/17/2015". So an algo datetime of mid-day 9/17/2015 should be able to buy this.

richafrank: Ok great.

richafrank:
Looks good to me!

jdricklefs:
:heart: Thanks Rich!

jdricklefs:
Merged via 38ff4cc9137a97d631ffe79782122379ad264788 + fd1bee9bf6f739f454b8fcf07b010b655e434b73

PR #712, opened 2015-09-18
Wire data portal into tracker

ehebert:
Prepare for data_portal powered last_sale_price by passing data_portal down to the tracker.
Also remove benchmark via DataPortal, since benchmark iter was using PerformanceTracker which caused a dependency cycle.

This temporarily hardcodes benchmark returns in tradesimulation, to be implemented after last_sale_price, the new benchmark implementation will probably use access of daily data from the dataportal from the perf tracker.

Depends on https://github.com/quantopian/qexec/pull/7672


ehebert:
Closing, but merging into lazy-mainline, since this was accidentally opened against master.

PR #715, opened 2015-09-18
MAINT: Removes batch transform's environment dependency

jfkirk:
@richafrank This should resolve the issue you saw

### DIFF
--- a/zipline/transforms/batch_transform.py
+++ b/zipline/transforms/batch_transform.py
@@ -32,14 +32,9 @@ from six import (
     iteritems
 )
 
+from zipline.utils.algo_instance import get_algo_instance
 from zipline.utils.data import MutableIndexRollingPanel
 from zipline.protocol import Event
-from zipline.finance.trading import TradingEnvironment
-
-# HACK the BatchTransform module stores a trading environment to be used by
-# the transforms
-# TODO remove this hack, if not this whole module
-_batch_transform_env = TradingEnvironment()
 
 log = logbook.Logger('BatchTransform')
 func_map = {'open_price': 'first',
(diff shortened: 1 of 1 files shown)


zipline/transforms/batch_transform.py
```diff
@@ -34,12 +34,7 @@
 
 from zipline.utils.data import MutableIndexRollingPanel
 from zipline.protocol import Event
-from zipline.finance.trading import TradingEnvironment
-
-# HACK the BatchTransform module stores a trading environment to be used by
-# the transforms
-# TODO remove this hack, if not this whole module
-_batch_transform_env = TradingEnvironment()
```
richafrank: We chatted about possibly using ```get_algo_instance().trading_environment``` instead of duplicating the code from ```TradingEnvironment``` here.  And if that's too difficult for now, we should add a new comment about why we're duplicating it.

jfkirk:
@richafrank Just pushed again to use get_algo_instance

zipline/transforms/batch_transform.py
```diff
@@ -33,13 +33,7 @@
...
 
 from zipline.utils.data import MutableIndexRollingPanel
-from zipline.protocol import Event
-from zipline.finance.trading import TradingEnvironment
-
-# HACK the BatchTransform module stores a trading environment to be used by
-# the transforms
-# TODO remove this hack, if not this whole module
-_batch_transform_env = TradingEnvironment()
+from zipline.protocol import Event, get_algo_instance
```
richafrank: Surprised to see get_algo_instance imported from protocol instead of zipline.utils.algo_instance, but not a big deal.

richafrank:
LGTM!

Issue #119, opened 2013-04-02
Add an option to turn off benchmark/treasury updating


ehebert:
So that Zipline can be run offline, i.e. so that a downloaded dataset can be used when there is no internet connection.

Without the option to disable benchmarks Zipline will currently crash on a socket exception while calling `data.loader.update_benchmarks`

cosnis:
dose it worked for now?

florentchandelier:
is urllib2 appropriate for zipline ? I believe the best way is not to provide an option but simply check for a connection status. Yet I would understand why urllib2 cannot be accessed from within zipline (specifically Quantopian), at which point an option is necessary ; 

At least another contributor can pass a variable from the TradingAlgorithm class down to the same location in the code in lieu of  'connection_status'. This was a quick suggestion so that I can work on few things over the next couple days without internet.

ssanderson:
Closing this in favor of https://github.com/quantopian/zipline/issues/119.

PR #718, opened 2015-09-19
fix issue #119 :Add an option to turn off benchmark/treasury updating

florentchandelier:
Issue: So that Zipline can be run offline, i.e. so that a downloaded dataset can be used when there is no internet connection. Without the option to disable benchmarks Zipline will currently crash on a socket exception while calling data.loader.update_benchmarks

Proposed Solution: Simply check for a connection status, and if no connection bypassing the update_benchmark() in loader.py, and use the data previously cached in ./zipline/data/\* for benchmark and treasuries, and ./zipline/cache/\* for stocks ... thus as if the connection was live.

### DIFF
--- a/zipline/data/loader.py
+++ b/zipline/data/loader.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import urllib2
 
 import importlib
 import os
@@ -49,7 +50,14 @@ INDEX_MAPPING = {
     ('treasuries', 'treasury_curves.csv', 'data.treasury.gov'),
 }
 
-
+def get_connection_status():
+    try:
+        res = urllib2.urlopen('http://www.google.com', timeout=1)
+        return True
+    except urllib2.URLError as err: 
+        pass
+    return False
+    
 def get_data_filepath(name):
     """
     Returns a handle to data file.
@@ -146,6 +154,11 @@ def get_benchmark_filename(symbol):
 
 def load_market_data(trading_day=trading_day_nyse,
                      trading_days=trading_days_nyse, bm_symbol='^GSPC'):
+    #
+    # bypass update_benchmark if zipline is used offline
+    #
+    connection_status = get_connection_status()
+    
     bm_filepath = get_data_filepath(get_benchmark_filename(bm_symbol))
     try:
         saved_benchmarks = pd.Series.from_csv(bm_filepath)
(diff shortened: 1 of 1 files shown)


ssanderson:
@florentchandelier I'll take a look at this soon.  Sorry for the radio silence.

ssanderson:
@florentchandelier I'm not sure how robust checking against google will be.  How would you feel about just adding a `nodownload` flag to `load_market_data` so that you could explicitly say not to query remote data?  Another variation on this would be to check for an environment variable like `ZIPLINE_DISABLE_NETWORK` or somesuch and check that before attempting to download.

florentchandelier:
Considering this relates only to beeing able to dl data, and more specifically to have an internet connection, I fell this must not have any impact on a written code, and thus I would advise against using a flag.

I agree checking against google is not elegant. Maybe a cleaner way would be an appropriate try/error catch ? This would be transparent for the end programmer when backtesting, and I believe this should be the objective here.

Thus, I would advocate for a clean implementation of an environment variable, as you suggested, yet performed automatically by the system.

If you could point me to where this should sit in the code, I could try to code such approach.

ssanderson:
@florentchandelier there are now two places in `ensure_benchmark_data` and `ensure_treasury_data` where the logic looks like:

```
        data = pd.Series.from_csv(path).tz_localize('UTC')
        if has_data_for_dates(data, first_date, last_date):
            return data
```

If you wanted to allow and environment-variable-based short-circuit, that logic should probably become something like:

```
        data = pd.Series.from_csv(path).tz_localize('UTC')
        if os.environ.get("ZIPLINE_NO_REMOTE_DATA"):
            logger.info(
                "Using cached data because ZIPLINE_NO_REMOTE_DATA was set."
            )
            return data
        if has_data_for_dates(data, first_date, last_date):
            return data
```

ssanderson:
FYI, I recently rewrote the benchmark/treasury loader pretty thoroughly.  It should be on the order of 5 to 10x faster now.

llllllllll:
@florentchandelier still working on this or should we close this for now?

florentchandelier:
Thx for popping this out again. Was definitely not on my agenda anymore, kindly let me have a look at it over the next few weeks, until late february. Thanks.

florentchandelier:
I'm finalizing some simple code that consolidates all .csv file in .zipline/cache directory and further pass properly formatted data to algo.run(). I have to complete same for teasuries/benchmark.

With the previous, I have to check if that all makes sense considering the original issue (mostly around benchmark).

That should be finished within the next few days/weeks.

florentchandelier:
Hi there, 
I think this issue could be closed but marked as unresolved. I have indeed decide to move forward with a more traditional: unlike my original objective that was for zipline to automatically detect whether or not it is online, the burden is on the user side to either load locally files, or use internet to fetch data if an internet connection is available..

I will thus keep working on a data loader and consolidator that eventually check if data to be used for backtesting could be harvested from the zipline/cache directory by default.

I have currently commited this but more is to come: https://github.com/florentchandelier/zipline2quantopian/blob/f11360241dbabf333e114bd719e46cbf95008079/skeleton/generic_modules/custom_data.py

freddiev4:
Going to close this considering the most recent comment, as well as that there are some other PRs that address this in different ways.

PR #720, opened 2015-09-21
Enable test algorithm

ehebert:
@jbredeche these patches enable more of test_algorithm, by using the old trade data to create DataPortal's, up to the point of discovering a bug in when orders are filled in test_volshare_slippage

### DIFF
--- a/tests/utils/test_utils.py
+++ b/tests/utils/test_utils.py
@@ -64,12 +64,11 @@ def create_data_portal(env, tempdir, sim_params, sids):
         )
 
 
-def create_data_portal_from_trade_history(
-        tempdir, trades_by_sid, sim_params, env):
+def create_data_portal_from_trade_history(env, tempdir, sim_params,
+                                          trades_by_sid):
     if sim_params.data_frequency == "daily":
         path = os.path.join(tempdir.path, "testdaily.bcolz")
         assets = {}
-        length = sim_params.days_in_period
         for sidint, trades in trades_by_sid.iteritems():
             opens = []
             highs = []
--- a/tests/utils/test_utils.py
+++ b/tests/utils/test_utils.py
@@ -65,30 +65,31 @@ def create_data_portal(env, tempdir, sim_params, sids):
 
 
 def create_data_portal_from_trade_history(
-        tempdir, sidint, trades, sim_params, env):
+        tempdir, trades_by_sid, sim_params, env):
     if sim_params.data_frequency == "daily":
         path = os.path.join(tempdir.path, "testdaily.bcolz")
         assets = {}
         length = sim_params.days_in_period
-        opens = []
-        highs = []
-        lows = []
-        closes = []
-        volumes = []
-        for trade in trades:
-            opens.append(trade.open_price)
-            highs.append(trade.high)
-            lows.append(trade.low)
-            closes.append(trade.close_price)
-            volumes.append(trade.volume)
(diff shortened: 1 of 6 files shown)


tests/utils/test_utils.py
```diff
@@ -62,3 +62,72 @@ def create_data_portal(env, tempdir, sim_params, sids):
             sim_params=sim_params,
             asset_finder=env.asset_finder
         )
+
+
+def create_data_portal_from_trade_history(
```
jbredeche: can this method have the params in the same order as ```create_data_portal```?

ie:
```
def create_data_portal_from_trade_history(env, tempdir, sim_params, trades_by_sid)
```

zipline/finance/slippage.py
```diff
@@ -314,10 +314,13 @@ def __init__(self, spread=0.0):
         self.spread = spread
 
     def process_order(self, event, order, dt):
```
ssanderson: It looks like this no longer needs to be passed the event, which raises the question: **anything** should ever get passed an `Event` object anymore?  It seems like the answer should be "No"?

zipline/finance/slippage.py
```diff
@@ -314,10 +314,13 @@ def __init__(self, spread=0.0):
...
+
+        price = self.data_portal.get_current_price_data(order.sid, 'close')
```
jbredeche: where is ```self.data_portal``` assigned?

zipline/gens/tradesimulation.py
```diff
@@ -188,6 +188,10 @@ def inner_loop(dt_to_use):
                     perf_tracker_benchmark_returns[trading_day] = 0.001
                     # use the last dt as market close
                     yield self.get_message(trading_day)
+
+                    self.algo.portfolio_needs_update = True
```
jbredeche: why the change?

zipline/algorithm.py
```diff
@@ -445,6 +440,10 @@ def _create_generator(self, sim_params, source_filter=None):
                 data_portal=self.data_portal
             )
 
+        if not self.initialized:
```
jbredeche: why push back initialization to here?

zipline/finance/slippage.py
ehebert: I think when the dust settles, that will be the case.

zipline/gens/tradesimulation.py
ehebert: IIRC, found this when a unit test tried to access account, and the `needs_update` flag was initially False... the `_needs_update` fields may subject to removal, since some of the bottlenecks they were addressing may go away.

zipline/finance/slippage.py
ehebert: It is 'hotwired' in in tradesimulation, which was pre-existing to this patch. (VolumeShare also uses the data_portal.)

We should pass it explicitly.

zipline/algorithm.py
ehebert: I can rewind the history a bit to find the exact algo, but IIRC it was because an algorithm needed access to a perf_tracker during initialize.

Note that between the lazy-mainline branch and master we made this change:

```
-        # Build a perf_tracker
-        self.perf_tracker = PerformanceTracker(sim_params=self.sim_params,
-                                               env=self.trading_environment)
-
+        self.perf_tracker = None
         # Pull in the environment's new AssetFinder for quick reference
```

PR #722, opened 2015-09-21
ENH: Allow passing an existing engine to TradingEnvironment

jdricklefs:
Specifically to allow the use case of creating an in-memory SQLite database and populating it with assets before creating the trading environment.


ehebert:
Looks good to me.

To be completely thorough we could add a test which instantiates the class using both a path and a connection, but could follow up with that later.

jdricklefs:
Thanks, @ehebert!

PR #724, opened 2015-09-23
Begin unlocking test_perf_tracking tests.

ehebert:
- Add a 'dt' arg to get_current_price, so that the `current_day` member
  does not need to be mutated when using just the perf period and
  tracker.

- Change the frequency of the test data to daily from one second(?!)

- Create a data_portal in the test using sim_params which cover just the
  range of the test data.

This makes it possible to at least run:
$ nosetests -x tests/test_perf_tracking.py:TestPositionPerformance

Albeit, with a failure in the pnl result.

TODO: Figure out why transactions are generated past the end of the
sim_params.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -170,8 +170,8 @@ class DataPortal(object):
         if asset in self.sources_map:
             # go find this asset in our custom sources
             try:
-                return self.sources_map[asset].loc[self.current_day].\
-                    loc[column]
+                # TODO: Change to index both dt and column at once.
+                return self.sources_map[asset].loc[dt].loc[column]
             except:
                 log.error(
                     "Could not find price for asset={0}, current_day={1},"
@@ -209,7 +209,7 @@ class DataPortal(object):
                     "because it only started trading on {2}!".
                     format(
                         str(asset),
-                        str(self.current_day),
+                        str(dt),
                         str(asset_data_start_date)
                     )
                 )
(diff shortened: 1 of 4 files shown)


zipline/data/data_portal.py
```diff
@@ -163,7 +163,10 @@ def _open_minute_file(self, field, sid):
 
         return carray
 
-    def get_current_price_data(self, asset, column):
+    def get_current_price_data(self, asset, column, dt=None):
+        if dt is None:
+            dt = self.current_day
```
jbredeche: you don't use dt anywhere below.

zipline/data/data_portal.py
```diff
@@ -193,17 +196,21 @@ def get_current_price_data(self, asset, column):
...
-
-            if self.current_day < asset_start_trading_date:
+            # TODO: only access this info once.
+            calendar = daily_attrs['calendar']
+            asset_data_start_date = \
+                pd.Timestamp(
+                    calendar[daily_attrs['calendar_offset'][str(asset_int)]],
+                    tz='UTC')
+
+            if dt < asset_data_start_date:
```
ehebert: @jbredeche re: https://github.com/quantopian/zipline/pull/724/files#r40274701 here is where dt is used.

jbredeche: got it.

I think you need to use it on line 221 as well - that's where we do the offset calc right?  otherwise the daily data returned is still using self.current_day

PR #726, opened 2015-09-24
MAINT: Only calc position values once per packet.

ehebert:
Instead of calculating the position values for each stat result, e.g.
gross_exposure, net_liquidity etc.; get the positions upfront and then
calculate the period and position stats in order, passing each value
explicitly to the ones that follow it in the dependency chain.

e.g. the gross_value depends on the long_value and the short_value,
which called the position_values property for calculating both the
long_value and the short_value.

Removing the repeated calls to position_values (and
position_exposures) removes the need for the caching the last sale
prices and position amounts in separate vectors, since it is inexpensive
enough to read those values off of the positions dictionary held in the
position tracker.

This patch gives a small gain to ~500 sized portfolios, but the main
intent is to clear the path to not storing last_sale_prices on the
position objects at all. Removing all of the caching layer in this class
makes that change easier to apply. Removing the extra calls to
position_values also made this class easier to step through/reason about
when splicing in the new last sale price access, as well.

### DIFF
--- a/tests/test_perf_tracking.py
+++ b/tests/test_perf_tracking.py
@@ -35,6 +35,7 @@ from six.moves import range, zip
 
 import zipline.utils.factory as factory
 import zipline.finance.performance as perf
+from zipline.finance.performance import position_tracker
 from zipline.finance.slippage import Transaction, create_transaction
 import zipline.utils.math_utils as zp_math
 
@@ -2181,22 +2182,22 @@ class TestPositionTracker(unittest.TestCase):
         np.bool_(False)
         """
         pt = perf.PositionTracker(self.env.asset_finder)
+        pos_stats = position_tracker.calc_position_stats(pt)
 
         stats = [
-            'calculate_positions_value',
-            '_net_exposure',
-            '_gross_value',
-            '_gross_exposure',
-            '_short_value',
-            '_short_exposure',
-            '_shorts_count',
-            '_long_value',
-            '_long_exposure',
-            '_longs_count',
+            'net_value',
+            'net_exposure',
+            'gross_value',
+            'gross_exposure',
+            'short_value',
+            'short_exposure',
+            'shorts_count',
+            'long_value',
+            'long_exposure',
+            'longs_count',
         ]
         for name in stats:
-            meth = getattr(pt, name)
-            val = meth()
+            val = getattr(pos_stats, name)
             self.assertEquals(val, 0)
             self.assertNotIsInstance(val, (bool, np.bool_))
 
@@ -2234,20 +2235,22 @@ class TestPositionTracker(unittest.TestCase):
         pt.update_positions({1: pos1, 2: pos2, 3: pos3, 4: pos4})
 
         # Test long-only methods
-        self.assertEqual(100, pt._long_value())
-        self.assertEqual(100 + 300000, pt._long_exposure())
+
+        pos_stats = position_tracker.calc_position_stats(pt)
+        self.assertEqual(100, pos_stats.long_value)
+        self.assertEqual(100 + 300000, pos_stats.long_exposure)
 
         # Test short-only methods
-        self.assertEqual(-200, pt._short_value())
-        self.assertEqual(-200 - 400000, pt._short_exposure())
+        self.assertEqual(-200, pos_stats.short_value)
+        self.assertEqual(-200 - 400000, pos_stats.short_exposure)
 
         # Test gross and net values
-        self.assertEqual(100 + 200, pt._gross_value())
-        self.assertEqual(100 - 200, pt._net_value())
+        self.assertEqual(100 + 200, pos_stats.gross_value)
+        self.assertEqual(100 - 200, pos_stats.net_value)
 
         # Test gross and net exposures
-        self.assertEqual(100 + 200 + 300000 + 400000, pt._gross_exposure())
-        self.assertEqual(100 - 200 + 300000 - 400000, pt._net_exposure())
+        self.assertEqual(100 + 200 + 300000 + 400000, pos_stats.gross_exposure)
+        self.assertEqual(100 - 200 + 300000 - 400000, pos_stats.net_exposure)
 
     def test_serialization(self):
         pt = perf.PositionTracker(self.env.asset_finder)
@@ -2260,9 +2263,6 @@ class TestPositionTracker(unittest.TestCase):
         pt.update_positions({1: pos1, 3: pos3})
         p_string = dumps_with_persistent_ids(pt)
         test = loads_with_persistent_ids(p_string, env=self.env)
-        nt.assert_dict_equal(test._position_amounts, pt._position_amounts)
-        nt.assert_dict_equal(test._position_last_sale_prices,
-                             pt._position_last_sale_prices)
         nt.assert_count_equal(test.positions.keys(), pt.positions.keys())
         for sid in pt.positions:
             nt.assert_dict_equal(test.positions[sid].__dict__,
--- a/tests/test_perf_tracking.py
+++ b/tests/test_perf_tracking.py
@@ -2239,10 +2239,12 @@ class TestPositionTracker(unittest.TestCase):
         pos_stats = position_tracker.calc_position_stats(pt)
         self.assertEqual(100, pos_stats.long_value)
         self.assertEqual(100 + 300000, pos_stats.long_exposure)
+        self.assertEqual(2, pos_stats.longs_count)
 
         # Test short-only methods
         self.assertEqual(-200, pos_stats.short_value)
         self.assertEqual(-200 - 400000, pos_stats.short_exposure)
+        self.assertEqual(2, pos_stats.shorts_count)
 
         # Test gross and net values
         self.assertEqual(100 + 200, pos_stats.gross_value)
--- a/zipline/finance/performance/period.py
+++ b/zipline/finance/performance/period.py
@@ -75,6 +75,7 @@ import logbook
 
 import numpy as np
 
+from collections import namedtuple
 from zipline.assets import Future
 
 try:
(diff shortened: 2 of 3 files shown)


ehebert:
@ssanderson this is the PR I mentioned to you earlier today.

@jbredeche you may be a good candidate for reviewing this, as well, since it is based on some work contained in the lazy-mainline branch.

ehebert:
Ran this on our internal performance harness, and the change had some slight wins, but not really outside of variance (though was seeing some >3-4% gains for universes that held positions for universes sized ~500). Having no perf regression would be the perf goal here, as the main impetus behind this patch is reducing the number of calculations for clarity and ease of changing how the data is sourced for the position values.

tests/test_perf_tracking.py
```diff
@@ -2234,20 +2235,22 @@ def test_position_values_and_exposures(self):
...
-        self.assertEqual(100 + 200, pt._gross_value())
-        self.assertEqual(100 - 200, pt._net_value())
+        self.assertEqual(100 + 200, pos_stats.gross_value)
+        self.assertEqual(100 - 200, pos_stats.net_value)
 
         # Test gross and net exposures
-        self.assertEqual(100 + 200 + 300000 + 400000, pt._gross_exposure())
-        self.assertEqual(100 - 200 + 300000 - 400000, pt._net_exposure())
+        self.assertEqual(100 + 200 + 300000 + 400000, pos_stats.gross_exposure)
+        self.assertEqual(100 - 200 + 300000 - 400000, pos_stats.net_exposure)
```
ssanderson: Looks like there are a couple entries that we're testing in `test_empty_positions` that we're not testing here.  Should we be doing those as well?

ssanderson: We're testing:
```
short_value
short_exposure
gross_value
net_value
gross_exposure
net_exposure
```
but we compute
```
net_value
net_exposure
gross_value
gross_exposure
short_value
short_exposure
shorts_count
long_value
long_exposure
longs_count
```

```
In [8]: x
Out[8]: 
{'gross_exposure',
 'gross_value',
 'long_exposure',
 'long_value',
 'longs_count',
 'net_exposure',
 'net_value',
 'short_exposure',
 'short_value',
 'shorts_count'}

In [9]: y
Out[9]: 
{'gross_exposure',
 'gross_value',
 'net_exposure',
 'net_value',
 'short_exposure',
 'short_value'}

In [10]: x - y
Out[10]: {'long_exposure', 'long_value', 'longs_count', 'shorts_count'}
```

zipline/finance/performance/period.py
```diff
@@ -402,11 +429,12 @@ def as_account(self):
                     self.ending_cash / (self.ending_cash + self.ending_value))
         account.day_trades_remaining = \
             getattr(self, 'day_trades_remaining', float('inf'))
-        account.leverage = \
-            getattr(self, 'leverage', self._gross_leverage())
-        account.net_leverage = self._net_leverage()
-        account.net_liquidation = \
-            getattr(self, 'net_liquidation', self._net_liquidation_value)
+        account.leverage = getattr(self, 'leverage',
```
ssanderson: I know you didn't write this, but can we not use `getattr` everywhere here?

ssanderson: (That change might actually have a nontrivial perf impact).

zipline/finance/performance/position_tracker.py
```diff
@@ -27,6 +28,140 @@
...
+                            'long_value',
+                            'long_exposure',
+                            'longs_count',
+                            'net_value'])
+
+
+def calc_position_values(amounts,
+                         last_sale_prices,
+                         value_multipliers):
+    iter_amount_price_multiplier = zip(
```
ssanderson: You could also write this as a map over the three iterators:
```
def product3(x, y, z):
    return x * y * z

def calc_position_values(amounts, last_sale_prices, value_multpliers):
    return map(product3, amounts, last_sale_prices, value_multipliers)
```

zipline/finance/performance/position_tracker.py
```diff
@@ -27,6 +28,140 @@
...
+        last_sale_prices,
+        itervalues(value_multipliers),
+    )
+    return [
+        price * amount * multiplier for
+        price, amount, multiplier in iter_amount_price_multiplier
+    ]
+
+
+def calc_net_value(position_values):
```
ssanderson: Can we add docstrings to all of these methods?  They're fairly important, and it's not always obvious what the difference is between, e.g. "net" vs "gross" versions of various metrics.

zipline/finance/performance/position_tracker.py
```diff
@@ -27,6 +28,140 @@
...
+    if len(position_values) == 0:
```
ssanderson: Is this check to prevent us from getting an int back in the case of an empty list?  If so you can write this as `return sum(position_values, 0.0)`.  (or `np.float64()` if you prefer).

zipline/finance/performance/position_tracker.py
```diff
@@ -345,15 +398,20 @@ def pay_dividends(self, dividend_frame):
         return net_cash_payment
 
     def maybe_create_close_position_transaction(self, event):
-        if not self._position_amounts.get(event.sid):
+        try:
+            pos = self.positions[event.sid]
+            amount = pos.amount
+            if amount == 0:
+                return None
+        except KeyError:
```
ssanderson: Can we move this except block up to wrap just the key access here?

zipline/finance/performance/period.py
ehebert: In this case we are providing a default of `period_stats.gross_leverage` when a `self.leverage` is not an attribute.

(I'd lean towards checking for a `None` value instead, or some other way of determining whether the default should be used, but not in scope of this change.)

zipline/finance/performance/period.py
```diff
@@ -90,11 +91,49 @@
...
 TRADE_TYPE = zp.DATASOURCE_TYPE.TRADE
 
 
+PeriodStats = namedtuple('PeriodStats',
+                         ['net_liquidation',
+                          'gross_leverage',
+                          'net_leverage'])
+
+
+def calc_net_liquidation(ending_cash, long_value, short_value):
```
ssanderson: Can we add docstrings for these?

zipline/finance/performance/period.py
```diff
@@ -90,11 +91,49 @@
...
+    return ending_cash + long_value + short_value
+
+
+def calc_gross_leverage(gross_exposure, net_liq):
+    if net_liq != 0:
```
ssanderson: I think `numpy.divide` does what you want here.

zipline/finance/performance/period.py
```diff
@@ -90,11 +91,49 @@
...
+        return gross_exposure / net_liq
+
+    return np.inf
+
+
+def calc_net_leverage(net_exposure, net_liq):
```
ssanderson: Can we just call this function `calc_leverage` and call it twice with different arguments.

zipline/finance/performance/position_tracker.py
ehebert: Noted, though my mindset for this patch is as a `mv` of existing implementations/logic.

zipline/finance/performance/position_tracker.py
ehebert: I would like to, can we compromise on opening an issue on it? The documentation was not pre-existing, so would have to do some research to get the terminology right, etc.

zipline/finance/performance/position_tracker.py
ehebert: I think I would prefer the `np.float64`, and can follow up with it. I feel strongly that this PR should use the same logic in all functions to minimize change while the order of calling is overhauled.

tests/test_perf_tracking.py
ehebert: We probably should, but had not been. Will open an issue for this as well.

zipline/finance/performance/period.py
ehebert: Yes, I agree that would be preferable, will change.

ehebert: https://github.com/quantopian/zipline/commit/31d214d8ff81d6d01d0adab826d324135f63cd9e

tests/test_perf_tracking.py
ssanderson: Given that we moved around the behavior that calculates these values, I'd argue that we should add those tests here.  I think they're pretty trivial to write?

zipline/finance/performance/period.py
ssanderson: Yeah, this is actually compat for IB stuff on Quantopian...not how I would have written this, but agreed not in the scope of your change.

zipline/finance/performance/position_tracker.py
ssanderson: I agree that they should be in different commits.I'm not sure I buy the benefit of separating out a change like this into a separate PR.  Follow-ups have a tendency to either not be followed-up on, or else they sit around unreviewed forever.  I'm increasingly of the opinion that the only way our codebase ever improves is if we actively refactor it while we're working on features.

tests/test_perf_tracking.py
ehebert: In adding `longs_count` and `shorts_count`, found that we were already covering the long value and exposure.

```
        self.assertEqual(100, pos_stats.long_value)
        self.assertEqual(100 + 300000, pos_stats.long_exposure)
```

zipline/finance/performance/position_tracker.py
ehebert: https://github.com/quantopian/zipline/commit/20c64b591f3050ff5eb1c65112f55368e9db5bfe

tests/test_perf_tracking.py
ehebert: https://github.com/quantopian/zipline/commit/c88412e3f1ac28eb6affb8ffc4caa5b9f8d710e7

PR #728, opened 2015-09-24
Run panel of assets

richafrank:
Running an algo with a df/panel of Assets was raising SidNotFound

```retrieve_asset``` returns its argument if it's an Asset.  In ```run```, we want to ensure that Asset is in the db first, so pass the asset's sid attribute to ```retrieve_asset``` to get a real answer about whether it's already in the db.


richafrank:
cc @jfkirk

adzuci:
Yay =).

KarenRubin:
@richafrank do you need help testing this?

richafrank:
@KarenRubin I should be good.

jfkirk:
This is a lot cleaner than what I had -- thanks, Rich!

PR #730, opened 2015-09-25
fix some more test failures

jbredeche:
- ensure dt is midnight utc inside dataportal when doing daily lookup

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -424,6 +424,9 @@ class TradingAlgorithm(object):
         processed by the zipline, and False for those that should be
         skipped.
         """
+        if sim_params is not None:
+            self.sim_params = sim_params
+
         if self.perf_tracker is None:
             # Build a perf_tracker
             self.perf_tracker = PerformanceTracker(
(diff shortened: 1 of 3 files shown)


zipline/algorithm.py
```diff
@@ -424,6 +424,9 @@ def _create_generator(self, sim_params, source_filter=None):
         processed by the zipline, and False for those that should be
         skipped.
         """
+        if sim_params is not None:
```
jbredeche: this wasn't necessary for any tests, but seemed odd not to have.  we pass in sim_params into this method and they weren't being used.

zipline/data/data_portal.py
```diff
@@ -190,6 +190,8 @@ def get_current_price_data(self, asset, column, dt=None):
         column_to_use = self.column_lookup[column]
 
         if self.data_frequency == "daily":
+            dt = pd.Timestamp(dt.date(), tz='utc')
```
ehebert: I'm still hesitant about fixing the input here, as opposed to having the caller (period/tracker) pass `datetime.date` for which it requesting the data, but it is definitely ok while everything is coming together and perhaps more long term than that.

zipline/algorithm.py
ehebert: It looks like they are being used here, https://github.com/quantopian/zipline/blob/jean-zipline-tests/zipline/algorithm.py#L454

The mixture usage of `self.sim_params` vs. `sim_params` is confusing in this method is confusing.

PR #731, opened 2015-09-28
TST: Remove instant fill from close position test.

ehebert:
'instant fill' execution may be on the path to deprecation, (the removal
is currently proposed by the branch which removes events in favor of
operating over dts.), so remove instant_fill from the close position
tests so that the input and output is the same between the proposed
branch and here.

(This was one of the few places where instant fill was used that
was not testing instant fill behavior explicitly, so may be better to
align this test with the rest of the suite.)

Also, change test methods to make it more clear on which day the values
the expected and actual results differed.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -1681,21 +1681,22 @@ class TestClosePosAlgo(TestCase):
 
     def setUp(self):
         self.env = TradingEnvironment()
-        self.days = self.env.trading_days
-        self.index = [self.days[0], self.days[1], self.days[2]]
+        self.days = self.env.trading_days[:4]
         self.panel = pd.Panel({1: pd.DataFrame({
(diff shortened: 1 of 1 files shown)


tests/test_algorithm.py
```diff
@@ -1735,25 +1736,24 @@ def test_close_position_future(self):
...
         data = DataPanelSource(self.no_close_panel)
 
         # Check results
         results = algo.run(data)
 
-        expected_pnl = [0, 1, 2]
-        self.check_algo_pnl(results, expected_pnl)
-
-        expected_positions = [1, 1, 0]
+        expected_positions = [0, 1, 1, 0]
```
ehebert: Flipped the order of checking pnl vs. positions here since pnl is dependent on position values, so needed to get those passing first.

jfkirk:
LGTM!

ehebert:
Thanks for the quick review!

PR #733, opened 2015-09-28
(WIP) Pipeline API

ssanderson:
Exchanges the implicit global state of the previous Modeling API (using `add_factor` and `add_filter`) for an object-oriented API built in a new `Pipeline` class.

A `Pipeline` contains two pieces of state:

1. `columns`, a dictionary mapping column names to instances of `Filter`, `Factor`, or `Classifier`.
2. `screen`, an instance of `Filter`, which controls what rows, if any, to throw out before providing results back to the user.

The top-level entrypoint into Modelling API calculations is now `FFCEngine.run_pipeline`.


ssanderson:
Still missing on this branch:
- [ ] Use `pipeline.screen` instead of implicitly `&`ing all Filters.
- [ ] Add tests for `screen` usage.
- Fix internal breakage.  (Should be minimal.)

PR #734, opened 2015-09-29
move order handling back into blotter

jbredeche:
- moved a lot of the order-related logic in tradesim's inner loop back into blotter
- added ```blotter.process_open_orders``` and removed ```blotter.process_trade``` and ```blotter.process_transactions```
- Blotter now takes slippage and commission in the constructor
- got blotter tests passing
- got rid of transact_partial
- ```algorithm```no longer stores slippage and commission, it passes them through (in ```set_slippage``` and ```set_commission```) to blotter

### DIFF
--- a/zipline/finance/blotter.py
+++ b/zipline/finance/blotter.py
@@ -46,7 +46,6 @@ class Blotter(object):
         # holding orders that have come in since the last event.
         self.new_orders = []
 
-        self.current_dt = None
         self.max_shares = int(1e+11)
 
         self.slippage_func = slippage_func or VolumeShareSlippage()
@@ -59,20 +58,15 @@ class Blotter(object):
     commission={commission},
     open_orders={open_orders},
     orders={orders},
-    new_orders={new_orders},
-    current_dt={current_dt})
+    new_orders={new_orders})
 """.strip().format(class_name=self.__class__.__name__,
                    slippage_func=self.slippage_func,
                    commission=self.commission,
                    open_orders=self.open_orders,
                    orders=self.orders,
-                   new_orders=self.new_orders,
-                   current_dt=self.current_dt)
+                   new_orders=self.new_orders)
 
-    def set_date(self, dt):
-        self.current_dt = dt
-
-    def order(self, sid, amount, style, order_id=None):
+    def order(self, sid, amount, style, order_id=None, dt=None):
         # something could be done with amount to further divide
         # between buy by share count OR buy shares up to a dollar amount
         # numeric == share count  AND  "$dollar.cents" == cost amount
@@ -86,6 +80,9 @@ class Blotter(object):
         StopLimit order: order(sid, amount, style=StopLimitOrder(limit_price,
                                stop_price))
         """
+        if dt is None:
+            raise ValueError("dt cannot be None!")
+
         if amount == 0:
             # Don't bother placing orders for 0 shares.
             return
@@ -97,7 +94,7 @@ class Blotter(object):
 
         is_buy = (amount > 0)
         order = Order(
-            dt=self.current_dt,
+            dt=dt,
             sid=sid,
             amount=amount,
             stop=style.get_stop_price(is_buy),
@@ -111,7 +108,7 @@ class Blotter(object):
 
         return order.id
 
-    def cancel(self, order_id):
+    def cancel(self, order_id, dt):
         if order_id not in self.orders:
             return
 
@@ -125,12 +122,12 @@ class Blotter(object):
             if cur_order in self.new_orders:
                 self.new_orders.remove(cur_order)
             cur_order.cancel()
-            cur_order.dt = self.current_dt
+            cur_order.dt = dt
             # we want this order's new status to be relayed out
             # along with newly placed orders.
             self.new_orders.append(cur_order)
 
-    def reject(self, order_id, reason=''):
+    def reject(self, order_id, dt, reason=''):
         """
         Mark the given order as 'rejected', which is functionally similar to
         cancelled. The distinction is that rejections are involuntary (and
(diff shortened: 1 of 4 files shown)


zipline/finance/blotter.py
```diff
@@ -33,30 +34,36 @@
 
 log = Logger('Blotter')
 
-class Blotter(object):
 
-    def __init__(self):
-        self.transact = transact_partial(VolumeShareSlippage(), PerShare())
```
ehebert: Why the removal here, besides clarity?
(In hindsight the `transact_partial` is a little too indirect.)

jbredeche: before moving things back, it wasn't being used.

after I added things back, I thought it was too indirect, and would rather explicitly set slippage and commission functions.

zipline/finance/blotter.py
```diff
@@ -179,64 +185,62 @@ def process_benchmark(self, benchmark_event):
         return
         yield
 
-    def process_trade(self, trade_event):
+    def process_open_orders(self):
```
ssanderson: This feels to me like it should be passed the current simulation time, rather than having it stored somewhere.

zipline/gens/tradesimulation.py
```diff
@@ -121,14 +104,10 @@ def transform(self, todo_remove):
         day_engine = DayEngine(market_opens, market_closes)
 
         blotter = self.algo.blotter
+        blotter.slippage_func.data_portal = data_portal
```
jbredeche: this is a bit ugly.

zipline/finance/blotter.py
jbredeche: That's actually how I did it on my first pass, but then realized that blotter has current_dt on it, so ended up using current_dt.  Thought about removing current_dt, but the various order methods use it to set the date on the new orders.

I think we should stay consistent - either blotter knows about current simulation time, or it doesn't.  Since it does right now, I went with it.

zipline/finance/blotter.py
```diff
@@ -179,64 +185,62 @@ def process_benchmark(self, benchmark_event):
...
+                if txn.commission is not None:
+                    order.commission = (order.commission or 0.0) + \
+                                       txn.commission
 
-        if trade_event.volume < 1:
-            # there are zero volume trade_events bc some stocks trade
-            # less frequently than once per minute.
-            return
+                txn.dt = pd.Timestamp(txn.dt, tz='UTC')
+                order.dt = txn.dt
```
ehebert: Do we currently set the `dt` of the order to the txn `dt`. My understanding is that those are two distinct moments.

jbredeche: good call, not sure.  This line replaces https://github.com/quantopian/zipline/pull/734/files#diff-1f85e36f1c099fa9d61d0b1651065ddbL166, which was not doing much :)

I'll investigate

jbredeche: Looks like we do set the order's date: https://github.com/quantopian/zipline/blob/master/zipline/finance/blotter.py#L252

zipline/finance/blotter.py
ehebert: I concur.

jbredeche: I guess we could make everything in blotter that needs to know the current simulation time takes that as a parameter.  I'll look into that.

ssanderson: My strong preference would be for the blotter to not know anything about the current simulation time.  If it's just the order methods that need the `dt`, that's easily solved by passing `current_dt` to those as well.

jbredeche: Already making it so.  Sounds good.

zipline/finance/blotter.py
```diff
@@ -179,64 +185,62 @@ def process_benchmark(self, benchmark_event):
...
-                if txn.amount == 0:
-                    raise zipline.errors.TransactionWithNoAmount(txn=txn)
-                if math.copysign(1, txn.amount) != order.direction:
-                    raise zipline.errors.TransactionWithWrongDirection(
-                        txn=txn, order=order)
-                if abs(txn.amount) > abs(self.orders[txn.order_id].amount):
-                    raise zipline.errors.TransactionVolumeExceedsOrder(
-                        txn=txn, order=order)
+            sid = order.sid
+            if sid not in self.open_orders:
```
ehebert: To reduce the number of calls this could be written as:

```
try:
     sid_orders = self.open_orders[sid]
     sid_orders.remove(order)
except KeyError:
      continue
```

Is self.open_orders[sid] a list? Something that we want to do targeted removal on may work better as a `dict` keyed by the order.id

zipline/finance/blotter.py
```diff
@@ -179,64 +182,63 @@ def process_benchmark(self, benchmark_event):
...
-                                        txn.commission)
+            if order in self.open_orders[sid]:
+                self.open_orders[sid].remove(order)
 
-            # mark the date of the order to match the transaction
-            # that is filling it.
-            order.dt = txn.dt
+            remaining_open_orders = self.open_orders[sid]
+            if len(remaining_open_orders) == 0 or \
+               all(not order.open for order in remaining_open_orders):
```
ehebert: Why isn't the removal of the closed orders sufficient here?

zipline/gens/tradesimulation.py
ehebert: Can you help point me to where this is used?

ehebert: Nm, found it it in `process_order`.

jbredeche:
@ehebert @ssanderson thanks for reviewing, should be all set now.  merging.

ssanderson:
@jbredeche I think `process_open_orders()` is still being called with the wrong signature in `test_blotter`.  See, for example https://github.com/quantopian/zipline/pull/734/files#diff-9dc7cb37547c5feca2c0771447ce1a63R162.  Are those tests passing for you?

jbredeche:
@ssanderson I totally spaced out and didnt update tests.  Updated them on the lazy-mainline branch (https://github.com/quantopian/zipline/commit/4605506c629635b40e0f934b656b8c315f70d0b0)

PR #735, opened 2015-09-29
MAINT: Remove unused minute performance period.

ehebert:
Minute performance period is no longer used. The minute risk containers,
which had been removed, were the last usage of the minute performance
period.


ehebert:
@jdricklefs rightly asked in person/chat that if `minute_performance` isn't used for the 'live' mode i.e. frequency of 'minute' and emission of 'minute', than what is?

The answer is that `dataloader.tracker.PerformanceTracker.to_dict` uses the following logic.
```
        if emission_type == 'daily':
            _dict['daily_perf'] = self.todays_performance.to_dict()
                self.saved_dt)
```

Following refactorings may make that more clear.

PR #736, opened 2015-09-29
Calculate futures payouts

ehebert:
This may only work in daily mode, but it is a start in calculating the
payouts during the period performance calculaiton.

Should add a minute test case to master to ensure that we do not break
it.

### DIFF
--- a/zipline/finance/performance/period.py
+++ b/zipline/finance/performance/period.py
@@ -133,6 +133,7 @@ class PerformancePeriod(object):
             self,
             starting_cash,
             asset_finder,
+            data_portal,
             period_open=None,
             period_close=None,
             keep_transactions=True,
@@ -140,6 +141,7 @@ class PerformancePeriod(object):
             serialize_positions=True):
 
         self.asset_finder = asset_finder
+        self.data_portal = data_portal
 
         self.period_open = period_open
         self.period_close = period_close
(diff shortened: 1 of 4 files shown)


zipline/finance/performance/period.py
```diff
@@ -215,7 +217,34 @@ def calculate_performance(self):
...
+            if isinstance(asset, Future):
+                old_price_dt = max(pos.last_sale_date,
+                                   self.period_open)
+                if old_price_dt == pos.last_sale_date:
+                    continue
+                old_price = self.data_portal.get_previous_price(
+                    sid,
+                    'close',
+                    dt=old_price_dt)
+                price = self.data_portal.get_current_price_data(
```
jbredeche: do you mean ```get_spot_price```?

zipline/finance/performance/tracker.py
```diff
@@ -162,12 +167,13 @@ def __init__(self, sim_params, env, data_portal):
...
             serialize_positions=False,
             asset_finder=self.env.asset_finder,
+            data_portal=data_portal,
         )
         self.cumulative_performance.position_tracker = self.position_tracker
         self.perf_periods.append(self.cumulative_performance)
 
         # this performance period will span just the current market day
-        self.todays_performance = PerformancePeriod(
+        self.todays_performance = TodaysPerformance(
```
jbredeche: :+1: :+1: :+1: :+1: :+1:

zipline/finance/performance/period.py
ehebert: Yes. Did not update after resolving merge conflicts.

PR #738, opened 2015-10-01
BUG: missing comma in _asset_timestamp_fields

warren-oneill:
surprised this wasn't caught by any of the tests. Should we add an extra one?

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -56,7 +56,7 @@ _asset_str_fields = frozenset({
 _asset_timestamp_fields = frozenset({
     'start_date',
     'end_date',
-    'first_traded'
+    'first_traded',
     'notice_date',
     'expiration_date',
     'auto_close_date',


ssanderson:
Oof.  Thanks for the catch.  We're currently doing:

```
def _convert_asset_timestamp_fields(dict):
    """
    Takes in a dict of Asset init args and converts dates to pd.Timestamps
    """
    for key, value in dict.items():
        if (key in _asset_timestamp_fields) and (value is not None):
            dict[key] = pd.Timestamp(value, tz='UTC')
```
when we pull an entry out of the database.  Both for efficiency and correctness, this loop should probably be:

```
def _convert_asset_timestamp_fields(dict):
    """
    Takes in a dict of Asset init args and converts dates to pd.Timestamps
    """
    for key in _asset_timestamp_fields:
        value = dict[key]

        if (key in _asset_timestamp_fields) and (value is not None):
            dict[key] = pd.Timestamp(value, tz='UTC')
```

ssanderson:
@jfkirk @StewartDouglas `_convert_asset_str_fields` is also almost certainly broken.

```
def _convert_asset_str_fields(dict):
    """
    Takes in a dict of Asset init args and converts from unicode to string
    where applicable
    """
    for key, value in dict.items():
        if key in _asset_str_fields:
            dict[key] = str(value)
```

This will return `bytes` in Python 2 and `unicode` in Python 3.  The statement "converts from unicode to string" is not meaningful in a Python 2/3 compat library.

jfkirk:
Good catches all around - apologies for not factoring in 2/3 compatibility when writing that.

This PR looks good - I'm merging it then opening an issue to tackle Scott's catches.

jfkirk:
Issue open for other fixes:
https://github.com/quantopian/zipline/issues/740

Merging this PR now.

PR #741, opened 2015-10-01
more progress

jbredeche:
- handle splits and dividends in the main tradesim loop (adjust open orders and positions)
- finance tests passing!
- serialization tests passing


ehebert:
SYN on first round of comments.

jbredeche:
ACK

Issue #740, opened 2015-10-01
Converting asset fields may be broken


jfkirk:
From @ssanderson : 

Oof.  Thanks for the catch.  We're currently doing:

```
def _convert_asset_timestamp_fields(dict):
    """
    Takes in a dict of Asset init args and converts dates to pd.Timestamps
    """
    for key, value in dict.items():
        if (key in _asset_timestamp_fields) and (value is not None):
            dict[key] = pd.Timestamp(value, tz='UTC')
```
when we pull an entry out of the database.  Both for efficiency and correctness, this loop should probably be:

```
def _convert_asset_timestamp_fields(dict):
    """
    Takes in a dict of Asset init args and converts dates to pd.Timestamps
    """
    for key in _asset_timestamp_fields:
        value = dict[key]
        if value is not None:
            dict[key] = pd.Timestamp(value, tz='UTC')
```

 `_convert_asset_str_fields` is also almost certainly broken.

```
# A set of fields that need to be converted to strings before building an
# Asset to avoid unicode fields
_asset_str_fields = frozenset({
    'symbol',
    'asset_name',
    'exchange',
})

def _convert_asset_str_fields(dict):
    """
    Takes in a dict of Asset init args and converts from unicode to string
    where applicable
    """
    for key, value in dict.items():
        if key in _asset_str_fields:
            dict[key] = str(value)
```

This will return `bytes` in Python 2 and `unicode` in Python 3.  The statement "converts from unicode to string" is not meaningful in a Python 2/3 compat library.

jfkirk:
@StewartDouglas Could you tackle these fixes at your earliest opportunity?

StewartDouglas:
ACK looking now

ssanderson:
Adding to AssetFinder bugs noticed while looking at this: **we're no longer correctly testing for duplicate symbol resolution in `test_lookup_generic`**.  The code that builds the test cases for that method currently uses these two assets:

```
            {
                'sid': 0,
                'symbol': 'duplicated_0',
                'start_date': dupe_0_start.value,
                'end_date': dupe_0_end.value,
                'exchange': '',
            },
            {
                'sid': 1,
                'symbol': 'duplicated_1',
                'start_date': dupe_1_start.value,
                'end_date': dupe_1_end.value,
                'exchange': '',
            }
```
We then query for `'duplicated_0'` and `'duplicated_1'`, but
[...]

PR #742, opened 2015-10-01
Correctly tests for duplicate symbols and remove _convert_asset_str_fields

StewartDouglas:
PR to address two issues:
1) In the process of an earlier refactoring, the symbol names in `build_lookup_generic_cases()` were changed to be different, meaning we were no longer testing the relevant functionality of `lookup_generic`. The symbol names have been updated to once again be the same.
2) `_convert_asset_str_fields()` is no longer required. `_convert_asset_str_fields()` was added to address the fact that the unicode returned by SQLAlchemy was causing test failures, in particular `test_repr` and `test_root_symbols` failed because of the leading 'u' character in the unicode of python-2.x. Later it was pointed out that this method would behave differently in python 2 and 3, in particular the `str()` method would return bytes in python 2 and unicode in python 3. After investigating it was found that the `_convert_asset_str_fields()` was no longer  necessary. It is unclear why the original tests were failing. It is not due to SQLAlchemy versions, as tests pass for version >1 and <1.

Addresses issued raised in: https://github.com/quantopian/zipline/issues/740

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -65,14 +65,14 @@ def build_lookup_generic_cases():
         [
             {
                 'sid': 0,
-                'symbol': 'duplicated_0',
+                'symbol': 'duplicated',
                 'start_date': dupe_0_start.value,
                 'end_date': dupe_0_end.value,
                 'exchange': '',
             },
             {
                 'sid': 1,
-                'symbol': 'duplicated_1',
+                'symbol': 'duplicated',
                 'start_date': dupe_1_start.value,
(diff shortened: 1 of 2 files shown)


richafrank:
Makes sense to me on both accounts.

1. Sounds like the ```lookup_generic``` functionality was still working, but the test wasn't actually testing it.  Is that right?
1. Glad to hear we don't need to do any str/unicode conversions.

StewartDouglas:
@richafrank Yup, we just were testing `lookup_generic` properly.

Will merge shortly.

PR #743, opened 2015-10-01
Add lookup_future_symbol method and future_symbol API method

StewartDouglas:
cc @jfkirk @warren-oneill @yankees714 

Opening up this PR to discuss the addition of a `future_symbol()` API method and an internal `lookup_future_symbol()` method. 

I have added an index to the symbol column in the futures SQL table to speed up lookup times, and while I was at it I felt it that an SQL UNIQUE constraint would also be relevant since there should never be a collision of futures symbols. Note that this change initially caused one of the existing tests (which duplicates an entry in this column) to fail, so the test has been changed.
Update: the root symbols coming from our data provider are unique, and hence the symbols of individual contracts are unique. Hence it makes sense to impose the UNQIUE constraint.

Outstanding:

- [x] Refactor failing test or remove UNIQUE constraint.
- [x] Add future_symbol API method.
- [x] Take into account the changes in https://github.com/quantopian/zipline/pull/746
- [x] Add more tests. 
- [x] Move lookup_future_symbol tests to their own class, or more appropriate place.

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -175,6 +175,10 @@ class AssetFinder(object):
     def _select_asset_by_sid(asset_tbl, sid):
         return sa.select([asset_tbl]).where(asset_tbl.c.sid == int(sid))
 
+    @staticmethod
+    def _select_asset_by_symbol(asset_tbl, symbol):
+        return sa.select([asset_tbl]).where(asset_tbl.c.symbol == symbol)
+
     def _retrieve_asset(self, sid, cache, asset_tbl, asset_type):
         try:
             return cache[sid]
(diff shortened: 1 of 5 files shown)


warren-oneill:
this looks great :smile: exactly what I need

jfkirk:
ACK - reviewing

zipline/assets/assets.py
```diff
@@ -175,6 +175,10 @@ def _retrieve_futures_contract(self, sid):
     def _select_asset_by_sid(asset_tbl, sid):
         return sa.select([asset_tbl]).where(asset_tbl.c.sid == int(sid))
 
+    @staticmethod
+    def _select_asset_by_symbol(asset_tbl, symbol):
+        return sa.select([asset_tbl]).where(asset_tbl.c.symbol == symbol)
```
jfkirk: How would this behave on a table where the symbols are not unique?

StewartDouglas: It will result in multiple rows. This won't happen internally, as our futures symbols are unique. Additionally, we enforce a `UNIQUE` constraint on symbols for futures, so if this would only ever return multiple rows if 1) a user had removed this constraint and 2) the user's data contained duplicate symbols. In such a case it would probably make more sense for the user to change the root symbol of one of the duplicates.

jfkirk: OK, works for me

jfkirk:
LGTM

warren-oneill:
@StewartDouglas @jfkirk tnx guys! you're the best :wink:

PR #744, opened 2015-10-01
Perf tracker refactor preparing for lazy.

ehebert:
** Main Commit

    Move the calculation of position stats and perfomance period
    calculations up the stack so that the performance tracker controls the
    timing (and frequency) of snapshotting the position and performance
    periods' data.

    This intent of this change is to prepare the tracker and related classes
    so that a dt can be passed explicitly during calculations when attached
    to the incoming data portal changes.

    Should have some slight performance benefits since required calculations
    are done closer to once a bar.

    This patch also includes changes to how the algorithm object caches the
    portfolio and account objects. They are changed here instead of
    separately because the `update_portfolio` and related flags were
    removed.

Also, includes some small refactorings/changes that clear the way for the change in the timing of when perf period and supporting position stats are calculated and passed around.

This branch prepares for picking https://github.com/quantopian/zipline/tree/test-perf-with-dt back up where it stalled out around getting the right dt (the dt will be added as a param to 'perf_period.stats') from the data portal during perf period calculations, especially during 'rollovers'.

What I have in mind is to build on top of this branch with changing:

```
def handle_minute_close(self, dt):
    ...
   pos_stats = self.position_tracker.stats()
   cumulative_stats = self.cumulative_performance.stats(pos_stats)
   todays_stats = self.todays_performance.stats(pos_stats)
```

to

```
def handle_minute_close(self, dt):
    ...
    data_portal = self.data_portal
    pos_stats = self.position_tracker.stats(data_portal, self.dt)
    cumulative_stats = self.cumulative_performance.stats(pos_stats, data_portal, self.dt)
    todays_stats = self.todays_performance.stats(pos_stats, data_portal, self.dt)    
```

The performance stats require data_portal and dt for lookups of futures data and cash adjustments.

This branch is PRed against master because these wholesale changes were easy to develop against that known base, but this branch could merged to lazy-mainline and/or master.
(There is a internal branch which supports these changes if we do merge to master, but understand if we want to be cautious and just merge to lazy-mainline for now.)

Once this is merged to either location, then test-perf-with-dt, which has about 90% of the tests passing will be rebased on to the resulting lazy-mainline branch.


ehebert:
On lazy via 5c36a5cf2badb5b7cb789f99686f25743cf3342e

PR #745, opened 2015-10-01
Perf tracker refactor on lazy

ehebert:
This is https://github.com/quantopian/zipline/pull/744 with lazy-mainline merged and the conflicts resolved with addition of passing data_portal and dt to positions stats, etc.

Probably could use a better commit message on the Merge commit, but this should help show how the refactoring works in the lazy-mainline scope (i.e. providing a clear way for tracker/tradesim/algorithm to pass the dt when accessing positions tracker and perf period backed data.)


zipline/finance/performance/position_tracker.py
```diff
@@ -106,18 +106,21 @@ def calc_gross_value(long_value, short_value):
...
+                        position_value_multipliers,
+                        position_exposure_multipliers,
+                        data_portal,
+                        dt):
     sids = []
     amounts = []
     last_sale_prices = []
-    position_value_multipliers = pt._position_value_multipliers
-    position_exposure_multipliers = pt._position_exposure_multipliers
+    position_value_multipliers = position_value_multipliers
```
jbredeche: cruft?

zipline/finance/performance/tracker.py
```diff
@@ -243,51 +238,75 @@ def handle_sid_removed_from_universe(self, sid):
...
+                             todays_stats,
+                             self.emission_rate)
+
+    def _to_dict(self, pos_stats, cumulative_stats, todays_stats,
+                 emission_type=None):
         """
         Creates a dictionary representing the state of this tracker.
         Returns a dict object of the form described in header comments.
+
+        Use this method internally, when stats are availble.
```
jbredeche: typo "availble"

zipline/finance/performance/tracker.py
```diff
@@ -298,14 +317,14 @@ def process_trade(self, event):
         # update last sale, and pay out a cash adjustment
         cash_adjustment = self.position_tracker.update_last_sale(event)
         if cash_adjustment != 0:
-            for perf_period in self.perf_periods:
-                perf_period.handle_cash_payment(cash_adjustment)
+            self.cumulative_performance.handle_cash_payment(cash_adjustment)
```
jbredeche: :+1: on calling out (and naming) specific perf periods

zipline/finance/performance/tracker.py
```diff
@@ -314,18 +333,17 @@ def process_dividend(self, dividend):
...
         if leftover_cash > 0:
-            for perf_period in self.perf_periods:
-                perf_period.handle_cash_payment(leftover_cash)
+            self.cumulative_performance.handle_cash_payment(leftover_cash)
+            self.todays_performance.handle_cash_payment(leftover_cash)
 
     def process_order(self, event):
-        for perf_period in self.perf_periods:
-            perf_period.record_order(event)
+        self.cumulative_performance.record_order(event)
```
jbredeche: can we rename these parameters to something more specific , like "order" in this case?  and if you don't mind, can you put some docstrings on these methods?

zipline/gens/tradesimulation.py
```diff
@@ -208,17 +204,13 @@ def _call_before_trading_start(self, dt):
     def on_dt_changed(self, dt):
         if self.algo.datetime != dt:
             self.algo.on_dt_changed(dt)
+            self.algo._portfolio = None
```
jbredeche: seems a bit weird to be mucking with algo's internal state here, and it's not obvious (from here) what is going on.  I think it would be better to have a method on algo that cleared `_portfolio` and `_account`, and call that method from here.

jbredeche:
Nearly all the perf tracking tests fail on this branch because no data portal is being passed into PerformancePeriod: https://github.com/quantopian/zipline/blob/perf-tracker-refactor-on-lazy/tests/test_perf_tracking.py#L187

zipline/finance/performance/tracker.py
ehebert: I agree can do in a follow up patch, but not in scope with these changes.

zipline/gens/tradesimulation.py
ehebert: I think these originally lived in `algo.on_dt_changed` on can be put back there.

ehebert:
On lazy via, 5c36a5cf2badb5b7cb789f99686f25743cf3342e

PR #746, opened 2015-10-01
Write faster

richafrank:
Speeds up the writing of all the assets to the assets.db, mostly by chunking the inserts in groups of 999 rows and by using vectorized computations for the date parsing.

Went from about 20s locally to about 5s.

### DIFF
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -310,7 +310,8 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
         constraints : bool, optional
             If True, create SQL ForeignKey and PrimaryKey constraints.
         """
-        self.sql_metadata = metadata = sa.MetaData(bind=engine)
+        metadata = sa.MetaData(bind=engine)
+
         self.equities = sa.Table(
             'equities',
             metadata,
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -209,6 +209,29 @@ class AssetDBWriter(with_metaclass(ABCMeta)):
         Returns data in standard format.
 
     """
+    def __init__(self, equities=None, futures=None, exchanges=None,
+                 root_symbols=None):
+
+        if equities is None:
+            equities = self.defaultval()
+        self._equities = equities
+
+        if futures is None:
+            futures = self.defaultval()
+        self._futures = futures
+
+        if exchanges is None:
+            exchanges = self.defaultval()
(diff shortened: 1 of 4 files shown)


richafrank:
cc @StewartDouglas @yankees714 @jfkirk

jfkirk:
All makes sense to me -- thanks, Rich!

@StewartDouglas I believe this will impact the new future lookup methods you were working on

zipline/assets/asset_writer.py
```diff
@@ -348,28 +324,26 @@ def init_db(self, engine, constraints=True):
...
-                *((sa.ForeignKey(self.futures_exchanges.c.exchange),)
-                  if constraints else ())
+                sa.ForeignKey(self.futures_exchanges.c.exchange),
             ),
-            sa.Column('notice_date', sa.Integer),
-            sa.Column('expiration_date', sa.Integer),
-            sa.Column('auto_close_date', sa.Integer),
+            sa.Column('notice_date', sa.Integer, nullable=False),
+            sa.Column('expiration_date', sa.Integer, nullable=False),
+            sa.Column('auto_close_date', sa.Integer, nullable=False),
```
StewartDouglas: @richafrank I'm not sure whether these should be nullable. This information is not provided by TDI, and instead we're going to have to source it from (a yet-to-be-chosen) third party data provider. We only have a stub method for this populating these variables right now: https://github.com/quantopian/dataloader/blob/master/dataloader/asset_master.py#L575.

richafrank: Sorry I didn't explain this better in the PR itself - it's in [this commit message] (https://github.com/quantopian/zipline/commit/21366e8529f9df1f0348fd3bc43bd15e684a1085).  There's still a sentinel value for missing data, but now that we're using [vectorized operations] (https://github.com/quantopian/zipline/commit/21366e8529f9df1f0348fd3bc43bd15e684a1085#diff-33cdc2260291626609681cc61297818aR495) to serialize the dts, we're storing only ints, so the sentinel is NaT instead of NULL.

richafrank: Correspondingly, the [reads] (https://github.com/quantopian/zipline/commit/21366e8529f9df1f0348fd3bc43bd15e684a1085#diff-e38a95b451336da451f69f2b60840da6R73) use ```pd.isnull``` instead of ```is None```.

StewartDouglas: Ah nice - remember you mentioning this yesterday.

PR #749, opened 2015-10-02
call before_trading_start, apply dividends

jbredeche:
Clean up some date management, etc in the tradesim inner loops.

Once per day, get dividends for relevant sids and apply them.

Properly call `before_trading_start` now.

### DIFF
--- a/tests/test_tradesimulation.py
+++ b/tests/test_tradesimulation.py
@@ -30,8 +30,16 @@ class BeforeTradingAlgorithm(TradingAlgorithm):
     def before_trading_start(self, data):
         self.before_trading_at.append(self.datetime)
 
+    def handle_data(self, data):
+        pass
+
 
 FREQUENCIES = {'daily': 0, 'minute': 1}  # daily is less frequent than minute
+
+
+class FakeDataPortal(object):
+    def __setattr__(self, key, value):
+        pass
 
 
 class TestTradeSimulation(TestCase):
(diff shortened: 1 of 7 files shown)


zipline/finance/performance/tracker.py
```diff
@@ -364,35 +364,31 @@ def check_upcoming_dividends(self, next_trading_day):
...
+            self.position_tracker.earn_dividends(earnable_dividends)
+
+        if len(payable_dividends) > 0:
+            net_cash_payment = self.position_tracker.\
+                pay_dividends(payable_dividends)
 
-        for period in self.perf_periods:
-            # notify periods to update their stats
-            period.handle_dividends_paid(net_cash_payment)
+            for period in self.perf_periods:
```
ehebert: This will break with the recent merge.

Also, I think only cumulative needs to handle this?

tests/test_tradesimulation.py
```diff
@@ -41,7 +49,7 @@ def test_minutely_emissions_generate_performance_stats_for_last_day(self):
                                                       data_frequency='minute',
                                                       emission_rate='minute')
         algo = NoopAlgorithm(sim_params=params)
-        algo.run(source=[], overwrite_sim_params=False)
+        algo.run(data_portal=FakeDataPortal())
```
ehebert: Should we give the random data portal to this test? Or one from a trade history of 0's/NaNs?
e.g. could do.
```
        data_portal = create_data_portal(self.env,
                                         self.tempdir,
                                         params,
                                         [1])
```

jbredeche: we could, but I was just sticking with what was being tested (ie, no trades).  Can add later if needed.

PR #753, opened 2015-10-05
Blaze pipeline loader

llllllllll:
A lot of the blaze loader is making sure that the expr passed trivial. I haven't thought about how deltas play with expressions so we are just disallowing them.


ssanderson:
@llllllllll travis failed on 2.7?

ssanderson:
Nvm, just realized those are newer commits than the original PR.

llllllllll:
I have been fighting travis all day, trying to figure out what it is failing with.

ssanderson:
> I have been fighting travis all day

Godspeed

llllllllll:
You should be able to run locally though

tests/pipeline/test_blaze.py
```diff
@@ -0,0 +1,455 @@
...
+                loader=self.garbage_loader,
+                no_deltas_rule='ignore',
+            ),
+            ds,
+        )
+
+    def test_column(self):
+        exprname = 'expr'
+        expr = bz.Data(self.df, name=exprname, dshape=self.dshape)
+        value = pipeline_api_from_blaze(
```
ssanderson: I think we can just call this function `from_blaze`.  I think it's clear from the import path that it's producing Pipeline objects.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+            warnings.warn(NoDeltasWarning(expr))
+    return None
+
+
+def pipeline_api_from_blaze(expr,
+                            deltas='auto',
+                            loader=None,
+                            resources=None,
+                            no_deltas_rule=_valid_no_deltas_rules[0]):
+    """Create a pipeline api object from a blaze expression.
```
ssanderson: I've been treating Pipeline API as a proper name (and thus capitalizing) elsewhere.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+
+    Parameters
+    ----------
+    expr : Expr
```
ssanderson: Are there restrictions on the kinds of expressions available here?

llllllllll: Only if deltas are passed.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+        return str(self.expr)
+    __str__ = __repr__
+
+
+class ExprData(namedtuple('ExprData', 'expr deltas resources')):
+    """A pair of expressions and a data resources.
+
+    Parameters
+    ----------
+    epxr : Expr
```
ssanderson: typo

tests/pipeline/test_blaze.py
llllllllll: kk, I think that is clear and concise

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+    ----------
+    name : str
+        The name of the field to check.
+    measure : Record
+        The record to check the field of.
+
+    Raises
+    ------
+    TypeError
+        if the field is not a datetime inside ``measure``
```
ssanderson: Caps `If` and period after `measure`.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+        The name of the argument we are checking.
+    expr : Expr
+        The potentially bound expr.
+    resources
+        The explicitly passed resources to compute expr.
+
+    Raises
+    ------
+    ValueError
+        if the resources to not match for an expression
```
ssanderson: Caps If and period.

zipline/pipeline/loaders/blaze.py
llllllllll: This isn't a sentance though.

llllllllll: looks like the numpy examples show this the way you described. I can update it.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+
+
+def _get_deltas(expr, deltas, no_deltas_rule):
+    """Find the correct deltas for the expression.
+
+    Parameters
+    ----------
+    expr : Expr
+        The base expression.
+    deltas : Expr, 'auto', or None
```
ssanderson: Lots of these docstrings reference `deltas` as a concept, but it's not obvious what they're about just from reading.  This might be a good use-case for a module docstring explaining the expected structure of the baseline vs. deltas tables that the blaze loader expects

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+    expr : Expr
+        The blaze expression to use.
+    deltas : Expr or 'auto', optional
+        The expression to use for the point in time adjustments.
+        If the string 'auto' is passed, a deltas expr will be looked up
+        by stepping up the expression tree and looking for another field
+        with the name of ``expr`` + '_deltas'. If None is passed, no deltas
+        will be used.
+    loader : BlazeLoader, optional
+        The blaze loader to attach this pipeline dataset to. If none is passed,
```
ssanderson: Capital None

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+    no_deltas_rule : {'warn', 'raise', 'ignore'}
+        What should happen if ``deltas='auto'`` but no deltas can be found.
+        'log' says to log a message but continue.
+        'raise' says to raise an exception if no deltas can be found.
+        'ignore' says take no action and proceed with no deltas.
+
+    Returns
+    -------
+    pipeline_api_obj : DataSet or BoundColumn
+        Either a new dataset or bound column based on the shape of the expr
```
ssanderson: :+1: on this explanation.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+        passed in. If a tabular shaped expression is passed, this will return
```
ssanderson: I'd probably write `table-shaped` here.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+        a ``DataSet`` that represents the whole table. If an array-like shape
+        is passed, a ``BoundColumn`` on the dataset that would be constructed
+        from passing the parent is returned.
+    """
+    deltas = _get_deltas(expr, deltas, no_deltas_rule)
+    if deltas is not None:
+        invalid_nodes = tuple(filter(
+            complement(flip(isinstance, valid_deltas_node_types)),
+            expr._subterms(),
+        ))
```
ssanderson: I'd probably write this as `tuple(node for node in nodes if not isinstance(node, valid_deltas_node_types))`.

A compromise might be to define `is_invalid_node = complement(flip(isinstance, valid_deltas_node_types))` and write this as `invalid_nodes = tuple(filter, is_invalid_node, expr._subterms())`.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+                " found: %s" % (
+                    ', '.join(map(getname, valid_deltas_node_types)),
+                    ', '.join(map(compose(getname, type), invalid_nodes)),
+                ),
+            )
+
+    # Check if this is a single column out of a dataset.
+    single_column = None
+    if isscalar(expr.dshape.measure):
+        # This is a single column, record which column we are to return
```
ssanderson: Comma here should be a period.  I actually stumbled trying to read this as a result, otherwise I wouldn't comment.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+        # but create the entire dataset.
+        single_column = expr._name
+        col = expr
+        for expr in expr._subterms():
```
ssanderson: Is `_subterms` a safe interface for us to be consuming here?

llllllllll: The prefix underscore is to minimize collisions with field names, like namedtuple. It is just a wrapper of `blaze.expr.core.subterm(self)`.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+            if isrecord(expr.dshape.measure):
```
ssanderson: Are we guaranteed to hit this conditional?

llllllllll: no, that is why there is the else on the for.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+                break
+        else:
+            expr = bz.Data(col, name=single_column)
+
+    measure = expr.dshape.measure
+    if not isrecord(measure) or AD_FIELD_NAME not in measure.names:
+        raise TypeError(
+            "expr must be a collection of records with at least an '{ad}'"
+            " field. Fields provided: '{fields}'\nhint: maybe you need to use "
+            ' `relabel` to change your field names'.format(
```
ssanderson: Switching to single quotes here confused me for a second.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+                ad=AD_FIELD_NAME,
+                fields=measure,
+            ),
+        )
+    _check_datetime_field(AD_FIELD_NAME, measure)
```
ssanderson: Every time I end up relying on global constants like this, I end up wanting to change it later and regretting how hard it is to do so.  How hard would it be to make these just default parameters to this function.  Are they used in multiple other places?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+                'expression with deltas may only contain (%s) nodes,'
+                " found: %s" % (
+                    ', '.join(map(getname, valid_deltas_node_types)),
+                    ', '.join(map(compose(getname, type), invalid_nodes)),
+                ),
+            )
+
+    # Check if this is a single column out of a dataset.
+    single_column = None
+    if isscalar(expr.dshape.measure):
```
ssanderson: Is there a clean example of a dshape whose measure is nonscalar vs one that is?  I suspect that the check here is conceptually simple, but we're deep into blaze-specific terminology here, so I'm not sure what's actually being expressed.

ssanderson: (Checking `isscalar` makes me think that this is looking for things like `int`, but I think what's actually happening is that we're looking for things like `var * int`?

zipline/pipeline/loaders/blaze.py
llllllllll: I haven't had the same experience, I updated this in the databazaar and found it very simple to quickly rename all occurences. What do I get by making it a default arg. Also, with no kwonly in py2 I either need to take **kwargs or allow this to be passed positionally.

zipline/pipeline/loaders/blaze.py
llllllllll: You are correct, it is looking only for scalar types like `int`. We are looking for things like `var * int` by checking against the measure, not the whole shape.

```python
>> dshape('var * int').measure
ctype("int32")
>>> isscalar(dshape('var * int').measure)
True
```

zipline/pipeline/loaders/blaze.py
ssanderson: This seems to be wrong, we talked about it.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+    -------
+    new_base : pd.DataFrame
+        The new base data with novel deltas inserted.
+    """
+    get_indexes = dates.searchsorted
+    return pd.concat(
+        (base,
+         deltas.loc[
+             (get_indexes(deltas[TS_FIELD_NAME].values, 'right') -
+              get_indexes(deltas[AD_FIELD_NAME].values, 'letf')) <= 1
```
ssanderson: Pretty sure `'letf'` isn't correct here.  Is this untested, or is pandas doing like "if how.startswith('l')"?

zipline/pipeline/loaders/blaze.py
ssanderson: This probably isn't quite right.  Talked about it in person.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+        The value to overwrite with.
+
+    Returns
+    -------
+    overwrite : Float64Overwrite
+        The overwrite that will apply the new value to the data.
+    """
+    return Float64Overwrite(
+        dates.searchsorted(asof),
+        dates.get_loc(dense_dates[dense_dates.searchsorted(asof) + 1]) - 1,
```
ssanderson: The +1 and -1 set off alarms for me.  What is this trying to do?  Is it meaningfully different from `get_adjustment_locs` in adjustment.pyx?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,697 @@
...
+    return {
+        dates.get_loc(kd): tuple(
+            overwrite_from_dates(
+                deltas.loc[kd, AD_FIELD_NAME],
+                dates,
+                dense_dates,
+                n,
+                v,
+            ) for n in range(len(assets))
+        ) for kd, v in deltas[column_name].iteritems()
```
ssanderson: Can we just pass in the column as a series rather than indexing into it over and over in this loop?

llllllllll: it looks it up once and then leave the iter on the stack. I did refactor the `deltas` above to cache the `deltas.loc[: AD_FIELD_NAME]`

zipline/pipeline/loaders/blaze.py
llllllllll: I am not sure if this was written when I wrote this. I think what I have is the scalar version of what you have. I am looking to see if I can simplifiy what I have with this.

llllllllll:
closing to make a pr against rich's branch

PR #754, opened 2015-10-06
removed DGZ from leveraged_etf_list

jmccorriston:
Removed DGZ from leveraged ETF list per the discussion on this thread:
https://www.quantopian.com/posts/dgz-a-leveraged-etf#561440cf8c09fa5208000096


jmccorriston:
@jfkirk Could you take a look at this one?

llllllllll:
This list is meant to be point in time. We should not be mutating files in place. We would need to create new remove entry instead of deleting the add. @fawce, can you confirm that I have translated your intent correctly.

fawce:
@llllllllll is correct - this should be a new delete file.

jmccorriston:
wait does this mean I should be making a new date folder with today's date, having a blank add file and then a delete file with DGZ?

jmccorriston:
Need to hold off on merging this one. I get the functionality of the date files but we need to decide what we want in terms of retroactively removing it or removing it from today-onward.

jmccorriston:
@llllllllll Changed it to a point-in-time deletion today, can you check that it looks ok?

llllllllll:
Hey, I was looking this over but I think I need to talk to fawce quick to answer some questions  I have around the security lists

ssanderson:
@llllllllll @jmccorriston @fawce what's the status on this?

jmccorriston:
I think it's blocked on a dicussion between @llllllllll and @fawce

jbredeche:
@jmccorriston @llllllllll what should we do about this?

jmccorriston:
No clue, I just know that DGZ shouldn't be on the list. @llllllllll ?

jmccorriston:
Moved to https://github.com/quantopian/zipline/pull/1434.

PR #756, opened 2015-10-07
Batch load Pipeline columns

richafrank:
Pass multiple atomic terms to load_adjusted_array if they share the same loader and extra_rows.


ssanderson:
@richafrank looks like the test for this is failing on travis.  Is this ready to review?

richafrank:
Apparently that test isn't deterministic, but feel free to check out everything else.

tests/pipeline/test_engine.py
```diff
@@ -115,7 +154,8 @@ def setUp(self):
 
     def test_bad_dates(self):
         loader = self.loader
-        engine = SimplePipelineEngine(loader, self.dates, self.asset_finder)
+        engine = SimplePipelineEngine(lambda column: loader,
```
ssanderson: Style nit: can we put the args here on separate lines?  (I assume they no longer fit in 80 characters on one line.)

tests/pipeline/test_engine.py
```diff
@@ -271,6 +315,123 @@ def test_numeric_factor(self):
...
 
+    def test_rolling_and_nonrolling(self):
+        open_ = USEquityPricing.open
+        close = USEquityPricing.close
+        volume = USEquityPricing.volume
+
+        # Test for thirty days up to the second to last day that we think all
+        # the assets existed.  If we test the last day of our calendar, no
+        # assets will be in our output, because their end dates are all
+        dates_to_test = self.dates[-32:-2]
```
ssanderson: Assets should still appear on the last day of their lifetime I think?  They don't appear on the **first** day of their lifetime, because we don't have any prior trading days.

ssanderson: Also, the expression here is actually going up to the third to last entry:

```
In [1]: x = arange(10)

In [2]: x
Out[2]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

In [3]: x[:-1]
Out[3]: array([0, 1, 2, 3, 4, 5, 6, 7, 8])

In [4]: x[:-2]
Out[4]: array([0, 1, 2, 3, 4, 5, 6, 7])
```

zipline/pipeline/term.py
```diff
@@ -157,80 +125,25 @@ def static_identity(cls, inputs, mask, window_length, domain, dtype):
...
-            # This isn't user error, this is a bug in our code.
-            raise AssertionError("{term} has no mask".format(term=self))
-
-        if self.window_length:
-            for child in self.inputs:
-                if not child.atomic:
-                    raise InputTermNotAtomic(parent=self, child=child)
 
-    @lazyval
+    @property
```
ssanderson: Is there a reason we switched this from being a lazyval to a property?

ssanderson: Ah, I guess that doesn't play nicely with `NotImplementedError`?

ssanderson: Do we even need the `atomic` attribute anymore if atomicity is segregated by type?  i.e. can we just `.atomic` accesses with `isinstance` calls?

zipline/pipeline/engine.py
```diff
@@ -273,6 +274,25 @@ def _inputs_for_term(self, term, workspace, graph):
...
+        for term in graph.atomic_terms:
+            if (loader_dispatch(term) == loader
+                    and graph.extra_rows[term] == extra_rows):
+                yield term
+
+    def loader_dispatch(self, term):
+        if term is AssetExists():
+            return None
+
+        loader = self._loader_dispatch(term)
```
ssanderson: I feel like `None` should never be a valid return from whatever does the dispatching here.  I'd have the contract be that the dispatcher should raise something like `NoSuchLoader(term)`

tests/pipeline/test_engine.py
```diff
@@ -271,6 +315,123 @@ def test_numeric_factor(self):
...
+                columns={
+                    'pipe_col1': pipe_col1,
+                    'pipe_col2': pipe_col2,
+                    'pipe_col3': pipe_col3,
+                },
+            ),
+            self.dates[2],  # index is >= the largest window length - 1
+            self.dates[-1]
+        )
+        self.assertIsNotNone(result)
```
ssanderson: Should we be asserting that this actually produced the expected outputs?

zipline/pipeline/engine.py
```diff
@@ -236,11 +236,12 @@ def _mask_and_dates_for_term(self, term, workspace, graph, dates):
         """
         Load mask and mask row labels for term.
         """
-        mask = term.mask
+        mask = term.mask if not term.atomic else self._root_mask_term
         offset = graph.extra_rows[mask] - graph.extra_rows[term]
         return workspace[mask][offset:], dates[offset:]
 
-    def _inputs_for_term(self, term, workspace, graph):
+    @staticmethod
```
ssanderson: I assume `self` was already unused here?

zipline/pipeline/engine.py
```diff
@@ -273,6 +274,25 @@ def _inputs_for_term(self, term, workspace, graph):
             out.append(input_data)
         return out
 
+    def _similar_atomic_terms(self, graph, atomic_term):
```
ssanderson: Since this is really about grouping terms by dispatched loader and extra_terms, I think it might be cleaner (also slightly faster?) to do this as a groupby at the top of `compute_chunk`.

```
from toolz import groupby, juxt # maps tuple of functions over a sequence of scalars

atomic_groups = groupby(juxt(dispatcher, graph.extra_rows.__getitem__), graph.atomic_terms)
```

zipline/pipeline/graph.py
```diff
@@ -168,6 +169,10 @@ def ordered(self):
         """
         return iter(self._ordered)
 
+    @lazyval
+    def atomic_terms(self):
+        return [term for term in self if term.atomic]
```
ssanderson: If we're using a `lazyval` here, this should probably be a tuple.

zipline/pipeline/term.py
```diff
@@ -307,7 +220,128 @@ def _compute(self, windows, dates, assets, mask):
         return out
 
 
-class AssetExists(Term):
+class AtomicTerm(Term):
```
ssanderson: There are a couple places where we do `if not term.atomic: <iterate over term.inputs>`.  I wonder if those would be made cleaner if `AtomicTerm` just defined `inputs` as an empty tuple?

zipline/pipeline/engine.py
ssanderson: @llllllllll points out that this the getitem might be better written as

```
from toolz.curried.operator import getitem

atomic_groups = groupby(juxt(dispatcher, getitem(graph.extra_rows)), graph.atomic_terms)

**Comment by ssanderson** [comment]
@richafrank this all seems reasonable to me.  I had a couple inline comments on particular implementations, but nothing major.

Two thoughts:

1.  I had imagined the loader-dispatching mechanism being encapsulated in a class rather than just passing a function to `SimplePipelineEngine`.  The main scenario where I imagine it being useful for the dispatching to be stateful is if we want to memoize lazy construction of loader objects, which we'll probably want to do downstream.  I suppose there's nothing preventing us from just passing around a method of an object as the dispatch function, but that starts to feel a bit obscure to me.  Thoughts?

2.  There are a few places where we branch on `term.atomic` and then iterate over `term.inputs` and `term.mask`.  It might make sense to have all terms provide a `dependencies` property or somesuch, which is either `term.inputs + (term.mask,)` or `()`, so that the right behavior for iterating over the dependencies of a term falls out correctly no matter what.  This comment is easier to change later, so it's probably less important to do right now.

File: tests/pipeline/test_engine.py (hunk 1 above)
Comment (richafrank) [inline tests/pipeline/test_engine.py]: Sure

File: tests/pipeline/test_engine.py (hunk 2 above)
Comment (richafrank) [inline tests/pipeline/test_engine.py]: My intended change was to merge in the MultiColumnLoaderTests, since I removed the MultiColumnLoader.

I couldn't really follow [the original intent] (https://github.com/quantopian/zipline/commit/26fd6fda8b095d052374fb27812dca3a6cad4869#diff-34c73f1261aca33c4c7c106482806900R509). What do you suggest?

File: zipline/pipeline/engine.py (hunk 6 above)
Comment (richafrank) [inline zipline/pipeline/engine.py]: Yep, that's right.

File: zipline/pipeline/engine.py (hunk 7 above)
Comment (richafrank) [inline zipline/pipeline/engine.py]: I can buy that.

File: zipline/pipeline/graph.py (hunk 8 above)
Comment (richafrank) [inline zipline/pipeline/graph.py]: Agreed.

**Comment by richafrank** [comment]
@ssanderson 
Agreed about `term.atomic` - this had been my intent originally, but got delayed.  I've added `term.dependencies` now.

I disagree about passing a class to `SimplePipelineEngine` - passing around a boundmethod feels very pythonic to me.  Passing an object whose method we call isn't too bad, but it feels like bloat of that interface, since it's currently a single method.  If we continue to add more functionality to it and pass those methods explicitly as well, then let's reconsider when the interface to `SimplePipelineEngine.__init__` grows too large.

File: zipline/pipeline/data/dataset.py [hunk 10]
```diff
@@ -29,12 +30,14 @@ class BoundColumn(Term):
     """
     A Column of data that's been concretely bound to a particular dataset.
     """
+    mask = AssetExists()
+    extra_input_rows = 0
+    inputs = ()
+    dependences = ()
```
ssanderson: dependencies?

richafrank: true story

richafrank: Luckily the base class behavior works out the same value - removing this one.

zipline/pipeline/engine.py
```diff
@@ -273,6 +276,12 @@ def _inputs_for_term(self, term, workspace, graph):
             out.append(input_data)
         return out
 
+    def get_loader(self, term):
+        if term is AssetExists():
```
ssanderson: Do we ever actually hit this branch?  It doesn't look like we're handling a `None` return anywhere.

zipline/pipeline/engine.py
```diff
@@ -315,10 +329,13 @@ def compute_chunk(self, graph, dates, assets, initial_workspace):
...
                 term, workspace, graph, dates
             )
+
             if term.atomic:
-                # FUTURE OPTIMIZATION: Scan the resolution order for terms in
-                # the same dataset and load them here as well.
-                to_load = [term]
+                to_load = sorted(
+                    atomic_groups[atomic_group_key(term)],
+                    key=lambda t: t.dataset
```
ssanderson: Why do we care about the sort order here?

zipline/pipeline/engine.py
richafrank: I becomes one of the keys in the `atomic_groups`, but that key is never looked up later because that term is already in the workspace.

zipline/pipeline/factors/factor.py
```diff
@@ -450,7 +450,7 @@ class CustomFactor(RequiredWindowLengthMixin, CustomTermMixin, Factor):
         class-level attribute named `inputs`.
     window_length : int, optional
         Number of rows of rows to pass for each input.  If this
-        argument is passed to the CustomFactor constructor, we look for a
+        argument is not passed to the CustomFactor constructor, we look for a
```
ssanderson: The classic binary off-by-one...

zipline/pipeline/engine.py
ssanderson: Ah.  Can we throw in a comment here explaining why we're handling this case?

zipline/pipeline/graph.py
```diff
@@ -199,12 +197,6 @@ def _add_to_graph(self, term, parents, extra_rows):
             )
             self.add_edge(dependency, term)
 
-        # Add term's mask, which is really just a specially-enumerated input.
```
ssanderson: :+1:

zipline/pipeline/loaders/base.py
```diff
@@ -17,5 +17,5 @@ class PipelineLoader(with_metaclass(ABCMeta)):
     TODO: DOCUMENT THIS MORE!
     """
     @abstractmethod
-    def load_adjusted_array(self, columns, mask):
+    def load_adjusted_array(self, columns, dates, assets, mask):
```
ssanderson: Good catch here.  Some day I want to write a better `ABCMeta` that cares about method signature compatibility.

zipline/pipeline/engine.py
richafrank: Since the hierarchy is generally loader:dataset:column, I thought loaders would often want to group by dataset - I know this is the case for fundamentals.  Instead of re-sorting to enable a lazy groupby inside each loader, my though was to do it once.

zipline/pipeline/term.py
```diff
@@ -157,79 +128,57 @@ def static_identity(cls, inputs, mask, window_length, domain, dtype):
...
+    @abstractproperty
+    def inputs(self):
+        raise NotImplementedError()
 
-        If term.atomic is truthy, it should have dataset and dtype attributes.
-        """
-        return len(self.inputs) == 0
+    @abstractproperty
+    def mask(self):
+        raise NotImplementedError
```
ssanderson: This probably wants to be raising an instance.

zipline/pipeline/term.py
```diff
@@ -157,79 +128,57 @@ def static_identity(cls, inputs, mask, window_length, domain, dtype):
...
-            for child in self.inputs:
-                if not child.atomic:
-                    raise InputTermNotAtomic(parent=self, child=child)
 
-    @lazyval
-    def atomic(self):
-        """
-        Whether or not this term has dependencies.
+    @abstractproperty
+    def inputs(self):
```
ssanderson: Can we put docstrings on this and `mask`?

zipline/pipeline/term.py
```diff
@@ -157,79 +128,57 @@ def static_identity(cls, inputs, mask, window_length, domain, dtype):
...
+    @abstractproperty
+    def mask(self):
+        raise NotImplementedError
 
     @lazyval
-    def windowed(self):
-        """
-        Whether or not this term represents a trailing window computation.
+    def dependencies(self):
+        return tuple(chain(self.inputs, (self.mask,)))
```
ssanderson: Since `inputs` should always be a tuple, can we just use `+` here?

zipline/pipeline/engine.py
richafrank: Yep, sounds good.

zipline/pipeline/loaders/base.py
richafrank: That would be nice!

zipline/pipeline/term.py
richafrank: Luckily it doesn't have an actual effect for us, but I agree, let's raise an instance.

zipline/pipeline/term.py
richafrank: Agreed.

zipline/pipeline/term.py
richafrank: Good call.

ssanderson:
@richafrank this looks good to me modulo the couple inline comments I left.

richafrank:
Thanks @ssanderson - patching those up now.

ssanderson:
@richafrank FYI, I'm looking at the downstream changes now.

zipline/pipeline/engine.py
ssanderson: Makes sense.  Might be worth commenting that the sort order is explicitly part of the interface for consumers so that we don't break that guarantee later.

richafrank: I actually have a test for it: https://github.com/quantopian/zipline/pull/756/files#diff-9b17b1ed8b53c49c54dccce45979890fR463

PR #758, opened 2015-10-07
ENH: Coerce user input with API method decorator

StewartDouglas:
Previously we have capitalized input strings at different levels in
our code: in the user-facing API methods and in the asset finder.
This commit moves input string capitalization exclusively to the API
method to which the string was supplied. Specifically, the string is
capitalized by a preprocess API method decorator. The preprocess
decorator passes the input string to the newly defined
ensure_upper_case() method, which returns a TypeError if the argument
supplied is not of type str.

ensure_upper_case() is defined in a new file, zipline/utils/api_utils.py.
The existing expect_types() method is also moved there.

Various tests in tests/test_assets.py are modified to account for the
fact that the asset finder method lookup_symol() no longer capitalizes
its supplied argument.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -425,6 +425,23 @@ class TestMiscellaneousAPI(TestCase):
         self.assertIsInstance(algo.sid(3), Equity)
         self.assertIsInstance(algo.sid(4), Equity)
 
+        # Supplying a non-string argument to symbol()
+        # should result in a TypeError.
+        with self.assertRaises(TypeError):
+            algo.symbol(1)
+
+        with self.assertRaises(TypeError):
+            algo.symbol((1,))
+
+        with self.assertRaises(TypeError):
+            algo.symbol({1})
+
+        with self.assertRaises(TypeError):
+            algo.symbol([1])
+
+        with self.assertRaises(TypeError):
+            algo.symbol({'foo': 'bar'})
+
     def test_future_symbol(self):
         """ Tests the future_symbol API function.
         """
@@ -449,6 +466,23 @@ class TestMiscellaneousAPI(TestCase):
 
         with self.assertRaises(SymbolNotFound):
             algo.future_symbol('FOOBAR')
+
+        # Supplying a non-string argument to future_symbol()
+        # should result in a TypeError.
+        with self.assertRaises(TypeError):
+            algo.future_symbol(1)
+
+        with self.assertRaises(TypeError):
+            algo.future_symbol((1,))
+
+        with self.assertRaises(TypeError):
+            algo.future_symbol({1})
+
+        with self.assertRaises(TypeError):
+            algo.future_symbol([1])
+
+        with self.assertRaises(TypeError):
+            algo.future_symbol({'foo': 'bar'})
 
     def test_future_chain(self):
         """ Tests the future_chain API function.
@@ -492,6 +526,23 @@ class TestMiscellaneousAPI(TestCase):
 
         with self.assertRaises(UnsupportedDatetimeFormat):
             algo.future_chain('CL', '2015-09-')
+
+        # Supplying a non-string argument to future_chain()
+        # should result in a TypeError.
+        with self.assertRaises(TypeError):
+            algo.future_chain(1)
+
+        with self.assertRaises(TypeError):
+            algo.future_chain((1,))
+
+        with self.assertRaises(TypeError):
+            algo.future_chain({1})
+
+        with self.assertRaises(TypeError):
+            algo.future_chain([1])
+
+        with self.assertRaises(TypeError):
+            algo.future_chain({'foo': 'bar'})
 
     def test_set_symbol_lookup_date(self):
         """
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -109,11 +109,11 @@ def build_lookup_generic_cases():
         (finder, 1, None, assets[1]),
         (finder, 2, None, assets[2]),
         # Duplicated symbol with resolution date
-        (finder, 'duplicated', dupe_0_start, dupe_0),
-        (finder, 'duplicated', dupe_1_start, dupe_1),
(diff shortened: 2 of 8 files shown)


zipline/utils/api_utils.py
```diff
@@ -0,0 +1,124 @@
+# Copyright 2015 Quantopian, Inc.
```
richafrank: Regarding module names, when should I put a new utility in `utils/api_support.py` versus `utils/api_utils.py`?  Do api_utils work with non-`api_method`s?

StewartDouglas: I considered placing the new code in `utils/api_support.py`, but @ssanderson felt that the logic in there was quite distinct from that which is now in `api_utils.py`. The logic in `api_utils.py` relates specifically to user input, while that in `api_support.py` are for internal control and checks.

`api_utils` would work with non-api-methods, but the methods in there have been specifically designed to decorate our api methods.

Perhaps `api_validation.py` would be a better name for `api_utils.py`?

richafrank: > api_utils would work with non-api-methods, but the methods in there have been specifically designed to decorate our api methods.

I had been thinking maybe just `validation.py`, assuming they would work fine with non `api_method`s...

richafrank: (Agreed they don't belong with api_support.py - but the names being so similar I find confusing.)

StewartDouglas: yup. or maybe `input_validation.py` to be even more explicit?

richafrank: That sounds good to me.

zipline/utils/api_utils.py
```diff
@@ -0,0 +1,124 @@
...
+# limitations under the License.
+
+from six import iteritems
+from toolz import valmap
+
+from zipline.utils.preprocess import preprocess
+
+
+def ensure_upper_case(func, argname, arg):
+    if isinstance(arg, str):
```
richafrank: Does this work in py3?

StewartDouglas:
```
Python 3.3.6 (default, Aug 14 2015, 17:30:04) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> str_arg = 'foo'
>>> isinstance(str_arg, str)
True
>>> int_arg = 1
>>> isinstance(int_arg, str)
False
```

tests/test_algorithm.py
```diff
@@ -425,6 +425,23 @@ def test_asset_lookup(self):
...
+            algo.symbol(1)
+
+        with self.assertRaises(TypeError):
+            algo.symbol((1,))
+
+        with self.assertRaises(TypeError):
+            algo.symbol(set(1,))
+
+        with self.assertRaises(TypeError):
+            algo.symbol(list(1,))
```
richafrank: `set(1,)` and `list(1,)` definitely take me longer to parse than `{1}` and `[1]`.  Any reason for the former?

llllllllll: `set(1,) and list(1,)` will raise a typeerror because they are the same as `list(1)` and `set(1)` where `1` is an `int` which is not iterable.

ssanderson: To be clear, these expressions aren't even making it into the `symbol` call.  They're failing in the calls to `list` and `set`.

tests/test_algorithm.py
```diff
@@ -425,6 +425,23 @@ def test_asset_lookup(self):
         self.assertIsInstance(algo.sid(3), Equity)
         self.assertIsInstance(algo.sid(4), Equity)
 
+        # Supplying a non-string argument to future_symbol()
```
ssanderson: This comment is referring to a different method than the one being tested.

zipline/utils/api_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    ...
+    >>> foo(2, '3')
+    (2, '3')
+    >>> foo(2.0, '3')
+    Traceback (most recent call last):
+       ...
+    TypeError: foo() expected an argument of type 'int' for argument 'x', but got float instead.  # noqa
+    """
+    if _pos:
+        raise TypeError("expect_types() only takes keyword arguments.")
```
richafrank: Would defining `expect_types` as `def expect_types(**named):` have the same effect as our check here?

tests/test_algorithm.py
richafrank: Haha, good point @ssanderson

zipline/utils/api_utils.py
richafrank: And are unicode args allowed?

tests/test_algorithm.py
StewartDouglas: Thanks guys. I meant to write `set((1,))` and `list((1,))`. @rich I used `set` and `list` to be clear what was happening here (the testing of different types). I'll change these to `{1}` and `[1]`.

zipline/utils/api_utils.py
StewartDouglas: Right now it will fail for unicode args, which seems undesirable. What is the best way to additionally accept unicode? Something like the following:

```
if isinstance(arg, str) or isinstance(arg, unicode):
    return arg.upper()
```

richafrank: I think probably
```python
from six import string_types
isinstance(arg, string_types)
```

zipline/utils/api_utils.py
StewartDouglas: @ssanderson is the thinking here to provide a more useful error message than the default `takes exactly 0 arguments` ?

ssanderson: @richafrank this is purely for providing a slightly clearer error message.

richafrank: Ok cool.

StewartDouglas:
I'll squash the two commit together before merging this.

PR #759, opened 2015-10-08
MAINT: Improves handling of contract date info in lookup_future_chain

yankees714:
Improves the query for futures contract to use the date that comes first in time (between `notice_date` and `expiration_date`) to determine contract validity. If one of these is missing, we'll use the other.

Also modifies the query to order the resulting contracts by their `expiration_date` if available, and to use their `notice_date` if not.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -626,12 +626,13 @@ class AssetFinderTestCase(TestCase):
 
     def test_lookup_future_chain(self):
         metadata = {
-            # Notice day is today, so not valid
+            # Notice day is today, so should be valid.
             2: {
                 'symbol': 'ADN15',
                 'root_symbol': 'AD',
                 'asset_type': 'future',
                 'notice_date': pd.Timestamp('2015-05-14', tz='UTC'),
+                'expiration_date': pd.Timestamp('2015-06-14', tz='UTC'),
                 'start_date': pd.Timestamp('2015-01-01', tz='UTC')
             },
             1: {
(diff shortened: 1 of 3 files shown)


jfkirk:
Do we have any way to profile the performance of this change? I don't think it will be a killer, but I am curious how much slower this will be.

yankees714:
A quick comparison shows that this change actually gives an improvement in performance. Not really sure why yet.

Current master:
```
In [5]: %%timeit -n10000
   ...: finder.lookup_future_chain('CL', dt, dt)
   ...:
10000 loops, best of 3: 8.87 ms per loop
```

With this change:
```
In [8]: %%timeit -n10000
   ...: finder.lookup_future_chain('CL', dt, dt)
   ...:
10000 loops, best of 3: 5.3 ms per loop
```

jdricklefs:
Those lookup times seem a bit slow for what you're doing. Have you tried adding `index=True` to the `futures_contracts.root_symbol` column in the `AssetDBWriter`? (https://github.com/quantopian/zipline/blob/master/zipline/assets/asset_writer.py#L345)

yankees714:
Ah good point, that's definitely something we should do.

jfkirk:
+1 on the index

jfkirk:
Should we include the index change to this PR?

yankees714:
Sure, I think this all fits under improving `lookup_future_chain`.

yankees714:
Also realized that we weren't exercising `FutureChain.__iter__` at all, so tacked on a simple test case for that.

jfkirk:
LGTM after Travis passes

yankees714:
With the index:
```
In [17]: %%timeit
   ....: finder.lookup_future_chain("CL", dt, dt)
   ....:
100 loops, best of 3: 1.97 ms per loop
```

jdricklefs:
Huh, still slower than I expected, but better.

PR #760, opened 2015-10-09
Write all dividend payout data to adjustments.db

ehebert:
To prepare for querying for payouts from SQLite, write the dividend
payouts to a new table `dividend_payouts`.

Change the expected columns of the passed dividend frame to contain the
payout data, and use that data to calculate the ratios (this moves
internal code that was calcualting the ratios into Zipline.)

The end result is that instead of just a `dividends` table with the
backward looking adjustment ratios, also write a `dividend_payouts`
table and a `stock_dividend_payout` table.

### DIFF
--- a/zipline/data/adjustments.py
+++ b/zipline/data/adjustments.py
@@ -1,0 +1,173 @@
+# Copyright 2015 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from errno import ENOENT
+from os import remove
+from os.path import exists
+
+import sqlite3
+
+from numpy import (
+    floating,
+    integer,
+    issubdtype,
+)
+
+from six import iteritems
+
+
+SQLITE_ADJUSTMENT_COLUMNS = frozenset(['effective_date', 'ratio', 'sid'])
+SQLITE_ADJUSTMENT_COLUMN_DTYPES = {
+    'effective_date': integer,
+    'ratio': floating,
+    'sid': integer,
+}
+SQLITE_ADJUSTMENT_TABLENAMES = frozenset(['splits', 'dividends', 'mergers'])
+
+
+class SQLiteAdjustmentWriter(object):
+    """
+    Writer for data to be read by SQLiteAdjustmentWriter
+
+    Parameters
+    ----------
+    conn_or_path : str or sqlite3.Connection
+        A handle to the target sqlite database.
+    overwrite : bool, optional, default=False
+        If True and conn_or_path is a string, remove any existing files at the
+        given path before connecting.
+
+    See Also
+    --------
+    SQLiteAdjustmentReader
+    """
+
+    def __init__(self, conn_or_path, overwrite=False):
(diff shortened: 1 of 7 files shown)


zipline/data/adjustments.py
```diff
@@ -0,0 +1,446 @@
...
+    'record_date': integer,
+    'pay_date': integer,
+    'payment_sid': integer,
+    'ratio': float,
+}
+
+
+class SQLiteAdjustmentWriter(object):
+    """
+    Writer for data to be read by SQLiteAdjustmentWriter
```
jbredeche: Reader, not Writer

zipline/data/us_equity_pricing.py
```diff
@@ -0,0 +1,109 @@
...
+class NoDataOnDate(Exception):
+    """
+    Raised when a spot price can be found for the sid and date.
+    """
+    pass
+
+
+class BcolzDailyBarSpotReader(object):
+    """
+    Reads inividual price moments from the daily bcolz file format shared
```
ssanderson: `inividual -> individual`.

zipline/data/us_equity_pricing.py
```diff
@@ -0,0 +1,109 @@
...
+    with pipeline.loaders.equity_pricing_loader.BcolzOHLCVReader
```
ssanderson: Did you rename the class elsewhere in this PR?  If not then I don't think this is the right name.

zipline/data/us_equity_pricing.py
```diff
@@ -0,0 +1,109 @@
...
+            table = ctable(rootdir=table, mode='r')
+        self.daily_bar_table = table
+        calendar = self.daily_bar_table.attrs['calendar']
+        self.trading_days = pd.DatetimeIndex(calendar, tz='UTC')
+
+        # For indexing into daily bars.
+        # We may be able to reuse code from DataPortal when that is ready.
+        self.first_rows = {
+            int(k): v for k, v
+            in self.daily_bar_table.attrs['first_row'].iteritems()}
```
ssanderson: This will break in PY3 from accessing iteritems.  Now that we have `toolz` as a dependency, this should probably just be

`self.first_rows = toolz.keymap(int, self.daily_bar_table.attrs['first_row'])`

zipline/data/us_equity_pricing.py
```diff
@@ -0,0 +1,109 @@
...
+        self.last_rows = {
+            int(k): v for k, v
+            in self.daily_bar_table.attrs['last_row'].iteritems()}
+        self.calendar_offset = {
+            int(k): v for k, v
+            in self.daily_bar_table.attrs['calendar_offset'].iteritems()}
+
+        self._cols = {}
+
+    def daily_bar_col(self, colname):
```
ssanderson: So are we just planning on holding the whole daily bar table in memory?  If we do that here then we should probably be sharing this with the `Pipeline` machinery, or else we'll be holding the whole thing in memory here but still reading from disk for pipeline access, which doesn't make much sense.

zipline/data/us_equity_pricing.py
```diff
@@ -0,0 +1,109 @@
...
+            The asset identifier.
+        day : datetime64
+            Midnight of the day for which data is requested.
+        colname : string
+            The price field. e.g. ('open', 'high', 'low', 'close', 'volume')
+
+        Returns
+        -------
+        float
+            The spot price for colnmae of the given sid on the given day.
```
ssanderson: colnmae -> colname

zipline/data/us_equity_pricing.py
ehebert: Right now, this is used just to write the assets.db, but may be used by DataPortal.

I agree we should share that cache if we do. (This may also speed up before trading starts.)

ehebert:
@ssanderson here is a first PR https://github.com/quantopian/zipline/pull/761 which moves both adjustments and BcolzDailyBar associated code to the data module. This prepares for the change we talked about adding `def spot_price` to the `BcolzDailyBarReader` instead.

PR #761, opened 2015-10-09
MAINT: Move equity data formats out of loader.

ehebert:
Put the logic for reading and writing the equity price and adjustment
data into a module located in data, making it distinct from the pipeline
loader usage of the formats.

This prepares for both incoming changes of how adjustments are written,
(which includes using the bcolz daily reader as an input), as well as
eventually providing the readers to a DataPortal object.


ssanderson:
LGTM

PR #762, opened 2015-10-09
PERF: Speed up reading of adjustments.

ssanderson:
For a pipeline doing simple computations on USEquityPricing data, we
were spending ~60% of `run_pipeline` loading adjustments.  Almost all of
that time was spent in calls to `DatetimeIndex.get_loc` to find the
indices of adjustment `eff_date`s.

This optimizes the eff_date lookups by pre-populating a cache of
seconds-since-epoch timestamps that we expect to see, and falling back
to `np.searchsorted` on cache misses.

In testing, this reduces the time to compute a 1-year pipeline with 30
and 90 day moving averages from 3.1 seconds to 0.9 seconds.

### DIFF
--- a/zipline/data/_adjustments.pyx
+++ b/zipline/data/_adjustments.pyx
@@ -21,6 +21,7 @@ from numpy import (
     uint32,
     zeros,
 )
+from numpy cimport int64_t, ndarray
 from pandas import Timestamp
 
 ctypedef object Timestamp_t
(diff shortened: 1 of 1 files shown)


ehebert:
Nice. Glad it worked out well. Will review when at keyboard next.

ehebert:
LGTM, :+1: on the fallback over creating a more complicated lookup table, since it is a small percent case 
If/when we fix the effective_dates during writing to adjust non-marktet days, we be able to remove the fallback.

PR #768, opened 2015-10-12
Blaze loader

llllllllll:


tests/pipeline/test_blaze.py
```diff
@@ -0,0 +1,456 @@
...
+
+nameof = op.attrgetter('name')
+dtypeof = op.attrgetter('dtype')
+
+
+class BlazeToPipelineTestCase(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.dates = pd.date_range('2014-01-01', '2014-01-03')
+        dates = (
```
ssanderson: `cls.dates.repeat(3)`

tests/pipeline/test_blaze.py
```diff
@@ -0,0 +1,456 @@
...
+            [pd.Timestamp('2014-01-01')] * 3 +
+            [pd.Timestamp('2014-01-02')] * 3 +
+            [pd.Timestamp('2014-01-03')] * 3
+        )
+        cls.sids = sids = ord('A'), ord('B'), ord('C')
+        cls.df = df = pd.DataFrame({
+            'sid': sids * 3,
+            'value': tuple(
```
ssanderson: I think it's actually fewer characters and clearer to just write `(0, 1, 2, 1, 2, 3, 2, 3, 4)` here.

tests/pipeline/test_blaze.py
```diff
@@ -0,0 +1,456 @@
...
+        cls.dshape = dshape_ = dshape("""
+        var * {
+            sid: ?int64,
+            value: ?float64,
+            asof_date: datetime,
+            timestamp: datetime
+        }
+        """)
+        cls.macro_df = df[df.sid == 65].drop('sid', axis=1)
+        dshape_ = OrderedDict(dshape_.measure.fields)
```
ssanderson: All the reassignments of `dshape_` make this a bit hard to read.

zipline/pipeline/data/dataset.py
```diff
@@ -105,7 +105,7 @@ def __new__(mcls, name, bases, dict_):
                 setattr(newtype, maybe_colname, bound_column)
                 _columns.append(bound_column)
 
-        newtype._columns = _columns
+        newtype._columns = frozenset(_columns)
```
ssanderson: Wondering if we should make this a tuple ordered by column name rather than a `frozenset`.  Having consistent ordering on DataSet.columns seems nice.  Thoughts?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+For an overview of the blaze project, see blaze.pydata.org
+
+The blaze loader for the Pipeline API is designed to allow us to load
+data from arbitrary sources as long as we can execute the needed expressions
+against the data with blaze.
+
+Data Format
+-----------
+
+The blaze Pipline API loader expects that data is formatted in a tabular way.
```
ssanderson: Pipline.

zipline/pipeline/data/dataset.py
llllllllll: I am not sure why the order of the columns here would matter

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+available to for use. Using our same CSV, we could write this with a timestamp
+like:
+
+asof_date,timestamp,value
+2014-01-01,2014-01-02,0
+2014-01-02,2014-01-03,1
+2014-01-03,2014-01-04,2
+
+This says that the value was 0 on 2014-01-01; however, we did not learn this
+until 2014-01-02. This is useful for avoiding look-ahead bias in your pipelins.
```
ssanderson: pipelins.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+If this column does not exist, the ``asof_date`` column will be used instead.
+
+If your data references a particular security, you can add a ``sid`` column to
```
ssanderson: You switch between using `security` here and `asset` below.  `asset` is probably more correct.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+2014-01-02,1,10
+2014-01-02,2,20
+2014-01-03,2,10
+2014-01-03,3,20
+
+This says that on 2014-01-01, the asset with id 10 had a value of 0, and the
+asset with id 20 had a value of 1.
+
+
+One of the key features of the Pipeline API is the handling of adjusments and
```
ssanderson: adjusments

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+restatements. Often our data will be amended after the fact and we would like
+to trade on the newest information; however, we do not want to introduce this
+knowledge to our model too early. The blaze loader handles this case by
+accepting a second ``deltas`` expression that contains all of the restatements
+in the original expression.
+
+For example, let's use our table from above:
+
+asof_date,value
+2014-01-01,0
```
ssanderson: You might want to tweak these data points to use actual market days.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+This shows that we learned on the fourth that the value on the first was
+actually -1 and that we learned on the fifth that the value on the third was
+actually 3. By pulling our data into these two tables and not silently updating
+our original table we can run our pipelines using the information we would
+have had on that day, and we can prevent lookahead bias in the pipelines.
+
+Conversion from Blaze to the Pipeline API
+-----------------------------------------
+
+Now that our data is structured in the way that the blaze loader expects we
```
ssanderson: comma after "expects"

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+are ready to convert our blaze expressions into Pipline API objects.
```
ssanderson: Pipline

llllllllll: I did a find and replace on this

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+
+This module (zipline.pipeline.loaders.blaze) exports a function called
+``from_blaze`` which performs this mapping.
+
+The expression that you are trying to convert must either be tabular or
```
ssanderson: Having all of this written in second person (addressing the user directly as "you") feels odd to me.  We don't necessarily need to change this now though.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+the Pipeline API.
+
+The scalar case defines a single column pulled out a table. For example, let
+``expr = bz.symbol('s', 'var * {field: int32, asof_date: datetime}')``.
+When we pass ``expr.field`` to ``from_blaze``, we will walk back up the
+expression tree until we find the table that ``field`` is defined on. We will
+then proceed with the record case to construct a dataset; however, before
+returning the dataset we will pull out only the column that was passed in.
+
+For full documentation, see ``help(from_blaze)`` or ``from_blaze?`` in jupyter.
```
ssanderson: IPython is probably the right name in this context, since we're squarely in Python-only land.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+    def __init__(self, expr):
+        self.expr = expr
+
+    def __repr__(self):
+        return str(self.expr)
+    __str__ = __repr__
+
+
+class ExprData(namedtuple('ExprData', 'expr deltas resources')):
+    """A pair of expressions and a data resources.
```
ssanderson: Not sure what "and a data resources" means here.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+
+    Parameters
+    ----------
+    expr : Expr
```
ssanderson: I've referred to this concept as "baseline" elsewhere in the pipeline infrastructure.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+        cls = type(self)
+        return super(ExprData, cls).__repr__(cls(
+            _ExprRepr(self.expr),
+            _ExprRepr(self.deltas),
+            self.resources,
+        ))
+
+
+class InvalidField(with_metaclass(ABCMeta)):
+    """A field that raises an exception that indicates that the
```
ssanderson: English style nit: "exception indicating that" means the same thing and doesn't repeat "that" twice in three words.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+
+
+class NonNumpyField(InvalidField):
+    error_format = "field '{field}' was a non numpy compatible type: '{type_}'"
+
+
+class NonPipelineField(InvalidField):
+    error_format = (
+        "field '{field}' was a non Pipeline API compatible type:"
+        " '{type_.__name__}'"
```
ssanderson: Any reason we do `type.__name__` here but not in the previous error?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+    )
+
+
+class NotPipelineCompatible(TypeError):
+    """Exception used to indicate that a dshape is not Pipeline API
+    compatible.
+    """
+    def __str__(self):
+        return "'%s' is a non Pipleine API compatible type'" % self.args
```
ssanderson: Pipeleine

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+_new_names = ('_%d' % n for n in count())
+
+
+@memoize
+def new_dataset(expr, deltas):
+    """Creates or returns a dataset from a pair of blaze expressions.
+
+    Parameters
+    ----------
+    expr : Expr
```
ssanderson: See note above w/r/t naming here.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+       The blaze expression representing the deltas to the data.
+
+    Returns
+    -------
+    ds : type
+        A new dataset type.
+
+    Notes
+    -----
+    This function is memoized, repeated calls will return the same type.
```
ssanderson: Might be worth adding "repeated calls **with the same inputs**"

ssanderson: Also, comma here should be a period or a colon.

zipline/pipeline/loaders/blaze.py
llllllllll: I updated the description to say `baseline`. I still think `expr` is more clear that this is a blaze expression and not something more concrete.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+    """
+    columns = {}
+    for name, type_ in expr.dshape.measure.fields:
+        try:
+            if promote(type_, float64, promote_option=False) != float64:
```
ssanderson: What does `promote_option=False` do?

llllllllll: By default, `promote(?A, A) -> ?A` With `promote_option=False`, option types get coerced down like: `promote(?A, A) -> A`

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+            col = NonPipelineField(name, type_)
+        except TypeError:
+            col = NonNumpyField(name, type_)
+        else:
+            col = Column(type_.to_numpy_dtype().type)
+
+        columns[name] = col
+
+    name = expr._name
+    if expr._name is None:
```
ssanderson: No reason to do this attr access twice.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+
+class NotPipelineCompatible(TypeError):
+    """Exception used to indicate that a dshape is not Pipeline API
+    compatible.
+    """
+    def __str__(self):
+        return "'%s' is a non Pipleine API compatible type'" % self.args
+
+
+_new_names = ('_%d' % n for n in count())
```
ssanderson: Can this be "BlazeDataSet_%d" so that people can easily tell where their weirdly-named dataset came from?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+    # unicode is a name error in py3 but the branch is only hit
+    # when we are in python 2.
+    if PY2 and isinstance(name, unicode):  # noqa
+        name = name.encode('utf-8')
+
+    return type(name, (DataSet,), columns)
+
+
+def _check_resources(name, expr, resources):
+    """Validate that the exprssion and resources passed match up.
```
ssanderson: exprssion

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+        The name of the argument we are checking.
+    expr : Expr
+        The potentially bound expr.
+    resources
+        The explicitly passed resources to compute expr.
+
+    Raises
+    ------
+    ValueError
+        If the resources to not match for an expression.
```
ssanderson: "to not" -> "do not"

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+    Parameters
+    ----------
+    expr : Expr
+        The expression that was searched.
+    """
+    def __init__(self, expr):
+        self._expr = expr
+
+    def __str__(self):
+        return 'No deltas could be infered from expr: %s' % self._expr
```
ssanderson: inferred (two 'r's)

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+def _get_deltas(expr, deltas, no_deltas_rule):
+    """Find the correct deltas for the expression.
+
+    Parameters
+    ----------
+    expr : Expr
+        The base expression.
+    deltas : Expr, 'auto', or None
+        The deltas argument. If this is 'auto', then the deltas table will
+        be searched for by walking up the expression tree. If this can not be
```
ssanderson: "can not" -> "cannot" http://www.write.com/writing-guides/general-writing/word-choice/cannot-versus-can-not/

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+        reflected, then an action will be taken based on the 'no_deltas_rule'.
```
ssanderson: This should probably be in backticks rather than quotes.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+    no_deltas_rule : {'warn', 'raise', 'ignore'}
+        How to handle the case where deltas='auto' but no deltas could be
+        found.
+
+    Returns
+    -------
+    deltas : Expr or None
+        The deltas table to use.
+    """
+    if no_deltas_rule not in _valid_no_deltas_rules:
```
ssanderson: This might be a good use-case for a new `preprocessor`.  Maybe `expect_element` to go with `expect_types`?  @StewartDouglas moved those to `zipline.utils.input_validation`.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+    deltas : Expr or 'auto', optional
+        The expression to use for the point in time adjustments.
+        If the string 'auto' is passed, a deltas expr will be looked up
+        by stepping up the expression tree and looking for another field
+        with the name of ``expr`` + '_deltas'. If None is passed, no deltas
+        will be used.
+    loader : BlazeLoader, optional
+        The blaze loader to attach this pipeline dataset to. If None is passed,
+        the global blaze loader is used.
+    resources : dict or any, optional
```
ssanderson: `dict or any` seems like an odd way to describe this.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+        The data to execute the blaze expressions against. This is used as the
+        scope for ``bz.compute``.
+    no_deltas_rule : {'warn', 'raise', 'ignore'}
+        What should happen if ``deltas='auto'`` but no deltas can be found.
+        'log' says to log a message but continue.
```
ssanderson: I think this should be `'warn'`?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+    """
+    deltas = _get_deltas(expr, deltas, no_deltas_rule)
+    if deltas is not None:
+        invalid_nodes = tuple(filter(is_invalid_deltas_node, expr._subterms()))
+        if invalid_nodes:
+            raise TypeError(
+                'expression with deltas may only contain (%s) nodes,'
+                " found: %s" % (
+                    ', '.join(map(getname, valid_deltas_node_types)),
+                    ', '.join(map(compose(getname, type), invalid_nodes)),
```
ssanderson: Should we throw a `set` in here somewhere so that we don't repeat the same bad types multiple times?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+                ),
+            )
+
+    # Check if this is a single column out of a dataset.
+    single_column = None
+    if bz.ndim(expr) != 1:
+        raise TypeError(
+            'expression was not tabular or array-like,'
+            ' too many dimensions: %d' % bz.ndim(expr)
```
ssanderson: Is it possible that `ndim(expr)` is 0?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+                'expression with deltas may only contain (%s) nodes,'
+                " found: %s" % (
+                    ', '.join(map(getname, valid_deltas_node_types)),
+                    ', '.join(map(compose(getname, type), invalid_nodes)),
+                ),
+            )
+
+    # Check if this is a single column out of a dataset.
+    single_column = None
+    if bz.ndim(expr) != 1:
```
ssanderson: I was confused by the fact that we initialize `single_column` before this check, which made me assume it was needed here.  Can we move it down until after the check?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+        raise TypeError(
+            'expression was not tabular or array-like,'
+            ' too many dimensions: %d' % bz.ndim(expr)
+        )
+    if isscalar(expr.dshape.measure):
+        # This is a single column. Record which column we are to return
+        # but create the entire dataset.
+        single_column = rename = expr._name
+        field_hit = False
+        if not isinstance(expr, traversable_nodes):
```
ssanderson: What's an example of a case that would hit this branch again?  (And can we put such an example in a comment here?)

llllllllll: This happens if you pass me some array typed expression like `dataaset.column + scalar`

llllllllll: I think that instead of a comment I should reword the error message, any suggestions?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+                else:
+                    break
+            rename = expr._name
+            expr = expr._child
+        expr = expr.relabel({rename: single_column})
+
+    measure = expr.dshape.measure
+    if not isrecord(measure) or AD_FIELD_NAME not in measure.names:
+        raise TypeError(
+            "expr must be a collection of records with at least an '{ad}'"
```
ssanderson: This message is bit confusing because `expr` here is referring to a value that isn't necessarily what the user passed in as the `expr` argument.

llllllllll: Maybe, `the dataset must be...`?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+        field_hit = False
+        if not isinstance(expr, traversable_nodes):
+            raise TypeError(
+                "expression '%s' was array-like but not a simple field of"
+                " some larger table" % str(expr),
+            )
+        while isinstance(expr, traversable_nodes):
+            if isinstance(expr, bz.expr.Field):
+                if not field_hit:
+                    field_hit = True
```
ssanderson: So we're breaking on the **second** field hit here?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,870 @@
...
+        if no_deltas_rule == 'raise':
+            raise ValueError(
+                "no deltas table could be reflected for %s" % expr
+            )
+        elif no_deltas_rule == 'warn':
+            warnings.warn(NoDeltasWarning(expr))
+    return None
+
+
+def from_blaze(expr,
```
ssanderson: There are a lot of individually-reasonable blocks in this function that add up to make it pretty hard to reason about all at once.  Could we break this up into more atomic checks?

zipline/pipeline/loaders/blaze.py
llllllllll: Yes, we walk up at most 1 field.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,875 @@
...
+        expr = bz.transform(expr, **{TS_FIELD_NAME: expr[AD_FIELD_NAME]})
+        if deltas is not None:
+            deltas = bz.transform(
+                deltas,
+                **{TS_FIELD_NAME: deltas[AD_FIELD_NAME]}
+            )
+    else:
+        _check_datetime_field(TS_FIELD_NAME, measure)
+
+    if deltas is not None and (sorted(deltas.dshape.measure.fields) !=
```
ssanderson: Is fields just name of the fields?  Will this catch the case of two expressions with the same field names but different dtypes (which I assume we can't sanely support)?

llllllllll: fields is a list of tuples of `[(name, dshape)]`

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,875 @@
...
+            "expr must be a collection of records with at least an '{ad}'"
+            " field. Fields provided: '{fields}'\nhint: maybe you need to use "
+            " `relabel` to change your field names".format(
+                ad=AD_FIELD_NAME,
+                fields=measure,
+            ),
+        )
+    _check_datetime_field(AD_FIELD_NAME, measure)
+
+    if TS_FIELD_NAME not in measure.names:
```
ssanderson: Might be worth breaking this block out as `_ensure_timestamp_field`.

zipline/pipeline/loaders/blaze.py
ssanderson: Yeah, I'd bind the name `dataset_expr` at the end of the `while` loop and work with that for the rest of the function.

ssanderson: (and reference that name in the error)

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,875 @@
...
+    # Register our new dataset with the loader.
+    (loader if loader is not None else global_loader)[ds] = ExprData(
+        expr,
+        deltas,
+        resources,
+    )
+    if single_column is not None:
+        # We were passed a single column, extract and return it.
+        return getattr(ds, single_column)
+    return ds
```
ssanderson: I don't think there's anything here preventing me from passing a nested record dshape like `"var * {a : {b: int}}"`.

llllllllll: This function does fail if a single column is incorrect. It will create the dataset correctly but then fail when the column is requested:

```python
In [16]: ds = bz.Data([{'asof_date': dt, 'a': {'b': 1}}])

In [17]: from_blaze(ds)
/home/yui/quantopian/research_env/base/src/qexec_repo/zipline_repo/zipline/pipeline/loaders/blazerred from expr: _5
  warnings.warn(NoDeltasWarning(expr))
Out[17]: <DataSet: 'BlazeDataSet_0'>

In [18]: from_blaze(ds).a
/home/yui/quantopian/research_env/base/src/qexec_repo/zipline_repo/zipline/pipeline/loaders/blazerred from expr: _5
  warnings.warn(NoDeltasWarning(expr))
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-18-dbcf18c0e16a> in <module>()
----> 1 from_blaze(ds).a

/home/yui/quantopian/research_env/base/src/qexec_repo/zipline_repo/zipline/pipeline/loaders/blaze
    245     def __get__(self, instance, owner):
    246         raise AttributeError(
--> 247             self.error_format.format(field=self._field, type_=self._type),
    248         )
    249 

AttributeError: field 'a' was a non numpy compatible type: '{b: int64}'

```

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,875 @@
...
+        The overwrites that should be applied to the dataset.
+
+    Returns
+    -------
+    adjustments : dict[idx -> Float64Overwrite]
+        The adjustments dictionary to feed to the adjusted array.
+    """
+    ad_series = deltas.loc[:, AD_FIELD_NAME]
+    return {
+        dates.get_loc(kd): concat(tuple(
```
ssanderson: It seems wasteful to be constructing an adjustment for every asset in the known universe for this case.  Since we're using as_strided tricks for these arrays anyway, I think we don't actually need multiple adjustments here, right?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,875 @@
...
+        The name of the column to compute deltas for.
+    deltas : pd.DataFrame
+        The overwrites that should be applied to the dataset.
+
+    Returns
+    -------
+    adjustments : dict[idx -> Float64Overwrite]
+        The adjustments dictionary to feed to the adjusted array.
+    """
+    ad_series = deltas.loc[:, AD_FIELD_NAME]
```
ssanderson: This is just `deltas[AD_FIELD_NAME]`

ssanderson:
```In [2]: %timeit df.loc[:, 'a']
The slowest run took 19.36 times longer than the fastest. This could mean that an intermediate result is being cached
10000 loops, best of 3: 17 µs per loop

In [3]: %timeit df['a']
The slowest run took 41.98 times longer than the fastest. This could mean that an intermediate result is being cached
1000000 loops, best of 3: 1.14 µs per loop
```

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,875 @@
...
+        if column.dataset in self:
+            return self
+        raise KeyError(column)
+
+    def load_adjusted_array(self, columns, dates, assets, mask):
+        return map(
+            op.getitem(
+                dict(concat(
+                    self._load_dataset(cs, dates, assets, mask)
+                    for _, cs in iteritems(groupby(getdataset, columns))
```
ssanderson: This is just `for cs in itervalues(...)` right?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,875 @@
...
+                dates,
+                dense_dates,
+                n,
+                v,
+            ) for n in range(len(assets)))
+        ) for kd, v in deltas[column_name].iteritems()
+    }
+
+
+def adjustments_from_deltas_with_sids(dates,
```
ssanderson: Have we benchmarked the performance of this vs the DBZ?  Adjustments were a bottleneck for equity pricing data even when they were written in Cython.  I suspect this is going to be pretty hot code.

zipline/pipeline/loaders/blaze.py
llllllllll: somewhere the strided array is materialized in the engine. When replacing this with just a mutation on sid 0 it only updates the first column.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,875 @@
...
+            dense_output.update(sparse_output)
+            adjustments_from_deltas = adjustments_from_deltas_with_sids
+            column_view = identity
+        else:
+            # We use the column view to make an array per asset.
+            dense_output = sparse_output.reindex(dates)
+            sparse_deltas = materialized_deltas.set_index(TS_FIELD_NAME)
+            adjustments_from_deltas = adjustments_from_deltas_no_sids
+
+            def column_view(arr, _shape=(len(dates), len(assets))):
```
ssanderson: This should just use `zipline.utils.numpy_utils.repeat_first_axis`

zipline/pipeline/loaders/blaze.py
ssanderson: :/

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,904 @@
...
+        Overwrites to the baseline data.
+    dates : pd.DatetimeIndex
+        The dates requested by the loader.
+
+    Returns
+    -------
+    new_baseline : pd.DataFrame
+        The new baseline data with novel deltas inserted.
+    """
+    get_indexes = dates.searchsorted
```
ssanderson: I think this can create duplicate entries if there's a "novel delta" on a date for which we have a baseline entry.

zipline/pipeline/loaders/blaze.py
ssanderson: Maybe what we should do then is make `Adjustment` support passing an empty slice for `column` to signify "apply this to all the assets"?

zipline/utils/input_validation.py
```diff
@@ -87,17 +102,12 @@ def _expect_type(type_):
...
-                    argname=argname,
-                    actual=_qualified_name(type(argvalue)),
-                )
-            )
-        return argvalue
-    return _check_type
+    return _mk_check(
+        TypeError,
+        template,
+        complement(flip(isinstance, type_)),
```
ssanderson: This is a case where I actually think `lambda v: not isinstance(v, type_)` is much clearer.

zipline/utils/input_validation.py
```diff
@@ -87,17 +102,12 @@ def _expect_type(type_):
...
+        compose(_qualified_name, type),
```
ssanderson: I feel less strongly about this one, but this is probably more accessible as `lambda v: _qualified_name(type(v))`.

zipline/utils/input_validation.py
```diff
@@ -122,3 +132,39 @@ def optional(type_):
...
     return (type_, type(None))
+
+
+def expect_element(*_pos, **named):
+    """
+    Preprocessing decorator that verifies inputs have expected types.
+
+    Usage
+    -----
+    >>> @expect_types(x=int, y=str)
```
ssanderson: Docstring copypasta.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,904 @@
...
+                    self._load_dataset(cs, dates, assets, mask)
+                    for _, cs in iteritems(groupby(getdataset, columns))
+                )),
+            ),
+            columns,
+        )
+
+    def _load_dataset(self, columns, dates, assets, mask):
+        try:
+            dataset, = set(map(getdataset, columns))
```
ssanderson: It's very easy to miss the comma here.  Can we put this in parens, or just do `if len(dataset) > 1`?

llllllllll: I really like the unpack here, I will put parens if that is okay.

tests/pipeline/test_blaze.py
```diff
@@ -0,0 +1,449 @@
...
+                dates,
+                finder,
+            ).run_pipeline(p, dates[0], dates[-1])
+
+        expected = self.df.drop('asof_date', axis=1).set_index(
+            ['timestamp', 'sid'],
+        )
+        expected.index = pd.MultiIndex.from_product((
+            expected.index.levels[0],
+            expected.index.levels[1].map(finder.retrieve_asset),
```
ssanderson: `finder.retrieve_all`

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,904 @@
...
+        )
+
+        # Inline the deltas that changed our most recently known value.
+        # Also, we reindex by the dates to create a dense representation of
+        # the data.
+        sparse_output = inline_novel_deltas(
+            materialized_expr,
+            materialized_deltas,
+            dates,
+        ).drop(AD_FIELD_NAME, axis=1).set_index(TS_FIELD_NAME)
```
ssanderson: We also drop `AD_FIELD_NAME` in `inline_novel_deltas`.  Can we do that just in one place?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,904 @@
...
+
+        if have_sids:
+            # Unstack by the sid so that we get a multi-index on the columns
+            # of datacolumn, sid.
+            sparse_output = sparse_output.set_index(
```
ssanderson: We should probably just wait to `set_index(TS_FIELD_NAME)` on the main expression until here or down below, rather than doing it twice.  This would also make it clearer that we're actually doing the same operation on both the data and the deltas.

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,904 @@
...
+                selection &= ts >= lower
+
+            return e[selection][query_fields]
+
+        extra_kwargs = {'d': resources} if resources else {}
+        materialized_expr = odo(where(expr), pd.DataFrame, **extra_kwargs)
+        materialized_deltas = (
+            odo(where(deltas), pd.DataFrame, **extra_kwargs)
+            if deltas is not None else
+            pd.DataFrame(columns=query_fields)
```
ssanderson: The dtypes of the columns of this dummy frame are going to be all `float`, which is probably different than what we get if we have actual deltas.  Does that matter?

zipline/pipeline/loaders/blaze.py
```diff
@@ -0,0 +1,904 @@
...
+            lower = odo(ts[ts <= dates[0]].max(), pd.Timestamp)
+            selection = ts <= dates[-1]
+            if have_sids:
+                selection &= e[SID_FIELD_NAME].isin(assets)
+            if lower is not pd.NaT:
+                selection &= ts >= lower
+
+            return e[selection][query_fields]
+
+        extra_kwargs = {'d': resources} if resources else {}
```
ssanderson: When would we not have resources?  Does this even work in that case?

zipline/utils/test_utils.py
```diff
@@ -356,3 +361,57 @@ class ExplodingObject(object):
...
+
+    Paramaters
+    ----------
+    data : dict, optional
+        The data to feed to the writer. By default this maps:
+        ('A', 'B', 'C') -> map(ord, 'ABC')
+    """
+    def __init__(self, data=None):
+        self._eng = None
+        self._data = AssetDBWriterFromDictionary(
```
ssanderson: I think this is duplicating logic from `make_simple_asset_info`.  Can we use that here?

llllllllll: Sure, I will see if they are easily translatable. One thing I like is that this format is really easy to pass in at a call site though.

tests/pipeline/test_blaze.py
```diff
@@ -0,0 +1,449 @@
...
+            {'sid': np.int64, 'value': np.float64},
+        )
+
+        for field in ('timestamp', 'asof_date'):
+            with self.assertRaises(AttributeError) as e:
+                getattr(ds, field)
+            self.assertIn("'%s'" % field, str(e.exception))
+            self.assertIn("'datetime'", str(e.exception))
+
+        self.assertIs(
```
ssanderson: Elsewhere you comment on these assertions that this is testing memoization.  Should we be consistent and add one here?

llllllllll: sure

tests/pipeline/test_blaze.py
```diff
@@ -0,0 +1,449 @@
...
+        # Test that we can have complex expressions with no deltas
+        from_blaze(
+            expr_with_add,
+            deltas=None,
+            loader=self.garbage_loader,
+        )
+
+        with self.assertRaises(TypeError):
+            from_blaze(
+                expr.value + 1,  # put an Add in the column
```
ssanderson: Am I correctly reading that adding column to the dataset with `transform` and then passing that column to `from_blaze` passes, but directly passing an addition fails?  That seems like super confusing behavior to explain to a user.

zipline/pipeline/loaders/blaze.py
llllllllll: All of the computations will just fall through fine with this empty.

zipline/pipeline/loaders/blaze.py
llllllllll: If you pass a bound expression, then resources is None, if you pass unbound expressions, then you need to pass data. `from_blaze` makes sure that data came in from either a bound expression or resources but not both.

tests/pipeline/test_blaze.py
llllllllll: This was the simple expr traversal case. When we spoke in person about this you said this was fine. The difference in the transform case is that the transform results in a tabular expression but the addition requires us to walk back up the tree.

ssanderson: Are we doing extra work to support the case that succeeds here?  It seems not very useful to me since:

a.) It doesn't work with deltas.
b.) It (probably?) doesn't work with nontrivial expressions.
c.) It only works if you `transform` the expression into another column, which most users won't know to do.

If it takes extra effort for us to prevent this from working, then I think this is fine.  If we're adding complexity to support this behavior, then I think we should remove this.

llllllllll: There is no extra code to support a transform other than not checking the tree types when no deltas are added. This will work fine for anything blaze can handle. I don't understand `c` there, transform is for adding computed columns to an expression so that is all that can be done there.

llllllllll: I could have also used merge to make a tabular expression but transform does this for us.

llllllllll:
closing to try to fix travis moving to new PR

PR #769, opened 2015-10-12
MAINT: Makes logging for updating benchmarks more verbose

yankees714:
We're seeing some issues with benchmarks failing to update, so this should give us a better idea of the context for when this is happening.


ssanderson:
@yankees714 I think this is probably fixed by the changes in https://github.com/quantopian/zipline/pull/793 ?

yankees714:
Yep, at the least this PR is no longer applicable.

PR #773, opened 2015-10-14
lots of fixes

jbredeche:
Rewrote how data portal handles Fetcher sources.

Cleaned up how data portal is initialized.

Cleaned up a bunch of tests.

### DIFF
--- a/tests/pipeline/test_pipeline_algo.py
+++ b/tests/pipeline/test_pipeline_algo.py
@@ -1,6 +1,7 @@
 """
 Tests for Algorithms using the Pipeline API.
 """
+import os
 from unittest import TestCase
 from os.path import (
     dirname,
@@ -35,6 +36,7 @@ from zipline.api import (
     pipeline_output,
     get_datetime,
 )
+from zipline.data.data_portal import DataPortal
 from zipline.errors import (
     AttachPipelineAfterInitialize,
     PipelineOutputDuringInitialize,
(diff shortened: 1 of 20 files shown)


jbredeche:
Current state of zipline tests is that a bunch of dividend tests fail (but that's expected, pending @ehebert's work).  Everything else passes.

tests/pipeline/test_pipeline_algo.py
```diff
@@ -132,6 +143,36 @@ def setUp(self):
...
+                "high": self.closes[sid].values,
+                "low": self.closes[sid].values,
+                "close": self.closes[sid].values,
+                "volume": self.closes[sid].values,
+                "day": [day.value for day in self.dates]
+            })
+
+        path = os.path.join(self.tempdir.path, "testdaily.bcolz")
+
+        DailyBarWriterFromDataFrames(data).write(
```
ssanderson: Why do we have to pass `data` twice to `DailyBarWriterFromDataFrames`?

jbredeche: because I was tired when I wrote it, and haven't cared enough to fix it.  Will get to it eventually.

tests/pipeline/test_pipeline_algo.py
```diff
@@ -132,6 +143,36 @@ def setUp(self):
...
+        # Create a data portal holding the data in self.closes
+        data = {}
+        for sid in sids:
+            data[sid] = DataFrame({
+                "open": self.closes[sid].values,
+                "high": self.closes[sid].values,
+                "low": self.closes[sid].values,
+                "close": self.closes[sid].values,
+                "volume": self.closes[sid].values,
+                "day": [day.value for day in self.dates]
```
ssanderson: Does `self.dates.values` not work here?

PR #774, opened 2015-10-14
BUG: Filter out payout rows with no prev close.

ehebert:
When the prev_close is 0 or does not exist, the resulting ration was either +inf
or nan, respectively.

Create a mask on the non-zero effective dates, where effective date is only
written when the prev close is sufficient for a valid ratio; and use that mask
to filter out the bad rows.


ehebert:
Looks like a failing test on:
```
test_load_adjustments_from_sqlite (tests.pipeline.test_us_equity_pricing_loader.USEquityPricingLoaderTestCase) ... ERROR
```

Will investigate.

zipline/data/us_equity_pricing.py
```diff
@@ -665,7 +672,14 @@ def calc_dividend_ratios(self, dividends):
                 })
                 continue
 
-        effective_dates = ex_dates.astype('datetime64[s]').astype(uint32)
+        effective_mask = effective_dates != 0
```
jbredeche: can you add a comment here explaining what you're doing?

zipline/data/us_equity_pricing.py
```diff
@@ -655,8 +659,11 @@ def calc_dividend_ratios(self, dividends):
             try:
                 prev_close = daily_bar_reader.spot_price(
                     sid, div_adj_date, 'close')
-                ratio = 1.0 - amount / (prev_close)
-                ratios[i] = ratio
+                if prev_close != 0.0:
+                    ratio = 1.0 - amount / (prev_close)
```
jbredeche: did you mean to put parens around `amount / prev_close` ?  (instead of around `prev_close`)

ehebert: Order of operations are doing the right thing here, can remove parens in a followup patch.

zipline/data/us_equity_pricing.py
ehebert: Yes.

tests/pipeline/test_us_equity_pricing_loader.py
```diff
@@ -175,43 +175,43 @@
     [
         # Before query range, should be excluded.
         {'declared_date': Timestamp('2015-05-01', tz='UTC').to_datetime64(),
-         'ex_date': Timestamp('2015-06-01', tz='UTC').to_datetime64(),
+         'ex_date': Timestamp('2015-06-02', tz='UTC').to_datetime64(),
```
ssanderson: Why did all these dates shift forward as a result of this change?

ehebert: Because the patch makes it so that the previous date (as our current internal calculations do) is used as the effective date.
The test data needs to reflect the change by moving up the `ex_date` so that the original `effective_dates` are used.

That being said, we should check that the date we are using for the dividend supports that behavior.

This assumes that the date of the `ex_date` would be the midnight before the close with same date.
e.g. a dividend with an `ex_date` of Feb. 3rd would be dispersed between Feb. 2nd and Feb 3rd.


Tried to call this out with the line in the message: `Also, use prev close as the effective date`

ehebert: Edited above comment to switch 'dispersed' for 'checked for exclusion' when describing the date timing.

tests/pipeline/test_us_equity_pricing_loader.py
```diff
@@ -175,43 +175,43 @@
...
-         'ex_date': Timestamp('2015-06-10', tz='UTC').to_datetime64(),
+         'ex_date': Timestamp('2015-06-11', tz='UTC').to_datetime64(),
          'record_date': Timestamp('2015-06-15', tz='UTC').to_datetime64(),
          'pay_date': Timestamp('2015-06-17', tz='UTC').to_datetime64(),
          'amount': 80.0,
          'sid': 3},
         # Third day of query range, should have last_row of 2
         {'declared_date': Timestamp('2015-06-01', tz='UTC').to_datetime64(),
-         'ex_date': Timestamp('2015-06-12', tz='UTC').to_datetime64(),
+         'ex_date': Timestamp('2015-06-15', tz='UTC').to_datetime64(),
```
ssanderson: This `ex_date` is now the same as the `record_date`.  Previously all the record_dates were strictly greater than the ex_dates.

ehebert: We currently ignore `record_date` completely, which may be a deficiency. But will fix so that it is more accurate for if/when we do use `record_dates`.

PR #775, opened 2015-10-14
Blaze loader

llllllllll:
was #768


ssanderson:
This is failing on our internal data (accern impact_score or article_sentiment) with errors like this:

```
/home/ssanderson/quantopian/qexec/zipline/lib/adjusted_array.pyx in zipline.lib.adjusted_array.Float64AdjustedArray.__cinit__ (zipline/lib/adjusted_array.c:3011)()
    114         if mask is not NOMASK:
    115             if not PyObject_RichCompare(mask.shape, data.shape, Py_EQ):
--> 116                 raise ValueError(
    117                     "Mask shape %s != data shape %s" % (
    118                         (mask.shape[0], mask.shape[1]),

ValueError: Mask shape (253, 8884) != data shape (253, 2835)
```

My guess is we don't have test coverage for the AssetFinder having data that the Database doesn't have?

PR #776, opened 2015-10-14
Use SQLiteAdjustmentReader for dividends payouts.

ehebert:
Pass a SQLiteAdjustmentReader and use queries for the dividends table
to disperse payouts to stocks in which the algorithm has a position.

### DIFF
--- a/tests/test_perf_tracking.py
+++ b/tests/test_perf_tracking.py
@@ -23,6 +23,7 @@ from datetime import (
 import logging
 import operator
 
+from testfixtures import TempDirectory
 import unittest
 from nose_parameterized import parameterized
 import nose.tools as nt
@@ -34,6 +35,10 @@ import pandas as pd
 import numpy as np
 from six.moves import range, zip
 
+from zipline.data.us_equity_pricing import (
+    SQLiteAdjustmentWriter,
+    SQLiteAdjustmentReader,
+)
 import zipline.utils.factory as factory
 import zipline.finance.performance as perf
 from zipline.finance.transaction import Transaction, create_transaction
@@ -44,6 +49,7 @@ from zipline.finance.trading import SimulationParameters
 from zipline.finance.blotter import Order
 from zipline.finance.commission import PerShare, PerTrade, PerDollar
 from zipline.finance.trading import TradingEnvironment
+from zipline.pipeline.loaders.synthetic import NullAdjustmentReader
 from zipline.utils.factory import create_simulation_parameters
 from zipline.utils.serialization_utils import (
     loads_with_persistent_ids, dumps_with_persistent_ids
@@ -152,7 +158,7 @@ def calculate_results(sim_params,
                       tempdir,
                       benchmark_events,
                       trade_events,
-                      dividend_events=None,
+                      adjustment_reader,
                       splits=None,
                       txns=None,
                       commissions=None):
@@ -183,23 +189,17 @@ def calculate_results(sim_params,
     splits = splits or {}
     commissions = commissions or {}
 
+    adjustment_reader = adjustment_reader or NullAdjustmentReader()
+
     data_portal = create_data_portal_from_trade_history(
         env,
         tempdir,
         sim_params,
-        trade_events
+        trade_events,
     )
+    data_portal._adjustment_reader = adjustment_reader
 
     perf_tracker = perf.PerformanceTracker(sim_params, env, data_portal)
-
-    if dividend_events is not None:
-        dividend_frame = pd.DataFrame(
-            [
-                event.to_series(index=zp.DIVIDEND_FIELDS)
-                for event in dividend_events
-            ],
-        )
-        perf_tracker.update_dividends(dividend_frame)
 
     results = []
 
@@ -234,7 +234,6 @@ def check_perf_tracker_serialization(perf_tracker):
         'txn_count',
         'market_open',
         'last_close',
-        '_dividend_count',
         'period_start',
         'day_count',
         'capital_base',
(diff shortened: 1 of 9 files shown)


ehebert:
cc: @jbredeche

zipline/data/us_equity_pricing.py
```diff
@@ -882,3 +900,50 @@ def load_adjustments(self, columns, dates, assets):
             dates,
             assets,
         )
+
+    def get_adjustments_for_sid(self, table_name, sid):
+        t = (sid,)
+        c = self.conn.cursor()
+        adjustments_for_sid = c.execute(
```
jbredeche: do you need `c.close()` like you have in `get_dividends_with_ex_date`?

zipline/finance/performance/position.py
```diff
@@ -65,22 +65,24 @@ def earn_dividend(self, dividend):
...
-        assert dividend['sid'] == self.sid
-        out = {'id': dividend['id']}
+        out = {}
+
+        out['amount'] = self.amount * dividend.amount
+        return out
+
+    def earn_stock_dividend(self, stock_dividend):
+        """
+        Register the number of shares we held at this dividend's ex date so
```
jbredeche: I think this comment (and the one for `earn_dividend` above) isn't accurate anymore, right?

`earn_dividend` is probably something like "Calculates the and returns the dollar amount of the dividend based on current position."

same for `earn_stock_dividend`

zipline/finance/performance/position_tracker.py
```diff
@@ -366,64 +365,73 @@ def _maybe_earn_dividend(self, dividend):
...
-        with fields drawn from zipline.protocol.DIVIDEND_PAYMENT_FIELDS.
-        """
-        try:
-            unpaid_dividend = self._unpaid_dividends.loc[dividend['id']]
-            return unpaid_dividend
-        except KeyError:
-            return zp.dividend_payment()
-
-    def pay_dividends(self, dividend_frame):
+            div_owed = self.positions[dividend.sid].earn_dividend(dividend)
```
jbredeche: what if `self.positions` doesn't have `dividend.sid`?  worth being defensive here?

zipline/finance/performance/position_tracker.py
```diff
@@ -366,64 +365,73 @@ def _maybe_earn_dividend(self, dividend):
...
+
+    def pay_dividends(self, next_trading_day):
         """
         Given a frame of dividends whose pay_dates are all the next trading
         day, grant the cash and/or stock payments that were calculated on the
         given dividends' ex dates.
         """
-        payments = dividend_frame.apply(self._maybe_pay_dividend, axis=1)\
-                                 .dropna(how='all')
+        net_cash_payment = 0.0
```
jbredeche: comment above is no longer valid, can you fix?

zipline/finance/performance/position_tracker.py
```diff
@@ -366,64 +365,73 @@ def _maybe_earn_dividend(self, dividend):
...
+            payments = self._unpaid_dividends[next_trading_day]
+            # Mark these dividends as paid by dropping them from our unpaid
+            del self._unpaid_dividends[next_trading_day]
+        except KeyError:
+            payments = []
 
-        # Mark these dividends as paid by dropping them from our unpaid
-        # table.
-        self._unpaid_dividends.drop(payments.index)
+        # representing the fact that we're required to reimburse the owner of
```
jbredeche: is this new business logic?

jbredeche: nvm, i saw below

zipline/utils/test_utils.py
```diff
@@ -581,5 +581,6 @@ def create_data_portal_from_trade_history(env, tempdir, sim_params,
 
 
 class FakeDataPortal(object):
-    def __setattr__(self, key, value):
-        pass
+
```
jbredeche: blank line intentional?  honest q, maybe a python style thing I don't know about.

jbredeche:
some minor questions/comments, lgtm otherwise.  nice work.

tests/test_perf_tracking.py
```diff
@@ -183,24 +189,18 @@ def calculate_results(sim_params,
...
+    adjustment_reader = adjustment_reader or NullAdjustmentReader()
+
     data_portal = create_data_portal_from_trade_history(
         env,
         tempdir,
         sim_params,
-        trade_events
+        trade_events,
     )
+    data_portal._adjustment_reader = adjustment_reader
```
ssanderson: Why are we tacking this on after the fact?  Can we not just make this an argument to the factory function being called here?

zipline/data/us_equity_pricing.py
ehebert: Agreed, would be good hygiene to do so. Will fix.

zipline/finance/performance/position.py
ehebert: The return is the still number of shares, it is using the `position.amount`. Reading through there is collision with the dividend cash amount. Maybe we change the name of the filed in dividends to be `cash_amount` to reduce confusion?

zipline/finance/performance/position_tracker.py
ehebert: The function that calls this gets the list of dividends based on the values in `self.positions`.

We could combine/inline `earn_dividends` into `check_upcoming_dividends` so that is an invariant.

zipline/finance/performance/position_tracker.py
ehebert: Will fix.

zipline/utils/test_utils.py
ehebert: I usually add one, really this change was because I deleted and retyped the class body and did it out of habit.

In Python's own documentation `class` seems to be done with or without a new line interchangibly.

tests/test_perf_tracking.py
ehebert: Yes, we will. Going to do another round of tightening up the interfaces after getting this in and able to work from a base where all tests pass.

PR #780, opened 2015-10-16
BUG: Fixes handling of Futures without start_dates in future_chain

jfkirk:
Adds a check that the start_date is > 0 to prevent returning Futures that do not have a start date.

Adds a test for this behavior to prevent regression.


tests/test_assets.py
```diff
@@ -670,6 +670,14 @@ def test_lookup_future_chain(self):
                 'expiration_date': pd.Timestamp('2016-11-16', tz='UTC'),
                 'start_date': pd.Timestamp('2015-08-01', tz='UTC')
             },
+            # No start date, shouldn't be in any chains
+            5: {
```
ssanderson: Can we not just make this an error at write-time?  If we're always going to filter these out on reads, why should we let people write them in the first place?

PR #781, opened 2015-10-16
restore benchmark functionality

jbredeche:
1) moved set_benchmark to Zipline
2) tradesim now pre-calculates the benchmark for the entire simulation,
either via downloaded market data, or via the specified sid
3) when possible, dataportal now uses only daily bars for history

### DIFF
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -1,0 +1,212 @@
+#
+# Copyright 2015 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+from unittest import TestCase
+from datetime import timedelta
+import numpy as np
+import pandas as pd
+from testfixtures import TempDirectory
+from zipline.data.us_equity_pricing import SQLiteAdjustmentWriter, \
+    SQLiteAdjustmentReader
+from zipline.errors import (
+    BenchmarkAssetNotAvailableTooEarly,
+    BenchmarkAssetNotAvailableTooLate,
+    InvalidBenchmarkAsset)
+
+from zipline.finance.trading import TradingEnvironment
+from zipline.gens.tradesimulation import AlgorithmSimulator
+from zipline.utils import factory
+from zipline.utils.test_utils import create_data_portal, write_minute_data
+from .test_perf_tracking import MockDailyBarSpotReader
+
+
+class TestBenchmark(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.env = TradingEnvironment()
+        cls.tempdir = TempDirectory()
+
+        cls.sim_params = factory.create_simulation_parameters()
+
+        cls.env.write_data(equities_data={
+            1: {
+                "start_date": cls.sim_params.trading_days[0],
+                "end_date": cls.sim_params.trading_days[-1] + timedelta(days=1)
+            },
+            2: {
+                "start_date": cls.sim_params.trading_days[0],
+                "end_date": cls.sim_params.trading_days[-1] + timedelta(days=1)
+            },
+            3: {
+                "start_date": cls.sim_params.trading_days[100],
+                "end_date": cls.sim_params.trading_days[-100]
+            },
+            4: {
+                "start_date": cls.sim_params.trading_days[0],
+                "end_date": cls.sim_params.trading_days[-1] + timedelta(days=1)
+            }
+
+        })
+
+        dbpath = os.path.join(cls.tempdir.path, "adjustments.db")
+
+        writer = SQLiteAdjustmentWriter(dbpath, cls.env.trading_days,
+                                        MockDailyBarSpotReader())
+        splits = mergers = pd.DataFrame(
+            {
+                # Hackery to make the dtypes correct on an empty frame.
+                'effective_date': np.array([], dtype=int),
+                'ratio': np.array([], dtype=float),
+                'sid': np.array([], dtype=int),
+            },
+            index=pd.DatetimeIndex([], tz='UTC'),
+            columns=['effective_date', 'ratio', 'sid'],
+        )
+        dividends = pd.DataFrame({
+            'sid': np.array([], dtype=np.uint32),
+            'amount': np.array([], dtype=np.float64),
+            'declared_date': np.array([], dtype='datetime64[ns]'),
+            'ex_date': np.array([], dtype='datetime64[ns]'),
+            'pay_date': np.array([], dtype='datetime64[ns]'),
+            'record_date': np.array([], dtype='datetime64[ns]'),
+        })
+        declared_date = cls.sim_params.trading_days[45]
+        ex_date = cls.sim_params.trading_days[50]
+        record_date = pay_date = cls.sim_params.trading_days[55]
+
+        stock_dividends = pd.DataFrame({
+            'sid': np.array([4], dtype=np.uint32),
+            'payment_sid': np.array([5], dtype=np.uint32),
+            'ratio': np.array([2], dtype=np.float64),
+            'declared_date': np.array([declared_date], dtype='datetime64[ns]'),
+            'ex_date': np.array([ex_date], dtype='datetime64[ns]'),
+            'record_date': np.array([record_date], dtype='datetime64[ns]'),
+            'pay_date': np.array([pay_date], dtype='datetime64[ns]'),
+        })
+        writer.write(splits, mergers, dividends,
+                     stock_dividends=stock_dividends)
+
+        cls.data_portal = create_data_portal(
+            cls.env,
+            cls.tempdir,
+            cls.sim_params,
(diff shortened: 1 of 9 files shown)


zipline/data/data_portal.py
```diff
@@ -349,14 +345,13 @@ def get_history_window(self, sids, end_dt, bar_count, frequency, field,
...
-
-            # for daily history, we need to get bar_count - 1 complete days,
-            # and then a partial day constructed from the minute data of this
-            # day.
             day_idx = tradingcalendar.trading_days.searchsorted(day)
             days_for_window = tradingcalendar.trading_days[
                 (day_idx - bar_count + 1):(day_idx + 1)]
 
+            ends_at_midnight = end_dt.hour == 0 and end_dt.minute == 0 \
+                and end_dt.second == 0
```
ehebert: Checking second is extraneous since we only support up to minute resolution.

zipline/data/data_portal.py
```diff
@@ -349,14 +345,13 @@ def get_history_window(self, sids, end_dt, bar_count, frequency, field,
...
             day = end_dt.date()
-
-            # for daily history, we need to get bar_count - 1 complete days,
-            # and then a partial day constructed from the minute data of this
-            # day.
             day_idx = tradingcalendar.trading_days.searchsorted(day)
             days_for_window = tradingcalendar.trading_days[
                 (day_idx - bar_count + 1):(day_idx + 1)]
 
+            ends_at_midnight = end_dt.hour == 0 and end_dt.minute == 0 \
```
ehebert: I think this shows a need to separate out two methods `get_daily_history` and `get_minute_history`, which can be dispatched/routed on the `frequency` flag. Should be more debuggable etc. than handling the special case throughout the '1m' case.

zipline/gens/tradesimulation.py
```diff
@@ -114,6 +109,25 @@ def transform(self):
...
+            benchmark_series = env.benchmark_returns[
+                trading_days[0]:trading_days[-1]]
+        else:
+            # calculate benchmark. start one day before the first trading day,
+            # so we can get a return value for the first day.
+            benchmark_sid = algo.benchmark_sid
+            benchmark_series = data_portal.get_history_window(
+                [benchmark_sid],
+                trading_days[-1],
+                bar_count=len(trading_days) + 1,
```
ehebert: I can't recall, is this a change to current behavior. i.e. do we start with 0 returns on the first day, use open to close for first return, or lookback one day like we are doing here.

jbredeche: Going to check with our in-house experts here, good point.

tests/test_benchmark.py
```diff
@@ -0,0 +1,212 @@
...
+        cls.sim_params = factory.create_simulation_parameters()
+
+        cls.env.write_data(equities_data={
+            1: {
+                "start_date": cls.sim_params.trading_days[0],
+                "end_date": cls.sim_params.trading_days[-1] + timedelta(days=1)
+            },
+            2: {
+                "start_date": cls.sim_params.trading_days[0],
+                "end_date": cls.sim_params.trading_days[-1] + timedelta(days=1)
```
ehebert: Can we use a trading calendar here instead of timedelta?

tests/test_benchmark.py
```diff
@@ -0,0 +1,212 @@
...
+            'declared_date': np.array([], dtype='datetime64[ns]'),
+            'ex_date': np.array([], dtype='datetime64[ns]'),
+            'pay_date': np.array([], dtype='datetime64[ns]'),
+            'record_date': np.array([], dtype='datetime64[ns]'),
+        })
+        declared_date = cls.sim_params.trading_days[45]
+        ex_date = cls.sim_params.trading_days[50]
+        record_date = pay_date = cls.sim_params.trading_days[55]
+
+        stock_dividends = pd.DataFrame({
```
ehebert: Did we make the call to no longer support equities with a stock dividend as a benchmark? If so, we could remove this from the tests?

jbredeche: The point of this part is to support testing that we don't support equities with a stock dividend as benchmark (see last test in this suite).

ehebert: Ah, got it now.

zipline/data/data_portal.py
```diff
@@ -311,6 +323,146 @@ def _get_daily_data(self, asset_int, column, dt):
...
+        for sid in sids:
+            sid = int(sid)
+
+            # get the start and end dates for this sid
+            if sid not in self.asset_start_dates:
+                asset = self.asset_finder.retrieve_asset(sid)
+                self.asset_start_dates[sid] = asset.start_date
+                self.asset_end_dates[sid] = asset.end_date
+
+            if ends_at_midnight or \
```
ehebert: I'm getting a little lost here, what is this check for?

zipline/data/data_portal.py
```diff
@@ -311,6 +323,146 @@ def _get_daily_data(self, asset_int, column, dt):
         else:
             return raw_value
 
+    def _get_history_daily_window(self, sids, end_dt, bar_count, field_to_use):
+        """
+        Internal method that returns a dataframe containing history bars
+        of daily frequency for the given sids.
+        """
+        data = []
```
ehebert: I think we could use the FFC (now moved to zipline.data.us_equity_pricing) USEquityPricingLoader load_adjusted_array to get same result?

Is there particular logic that is unique to this window?

zipline/data/data_portal.py
```diff
@@ -311,6 +323,146 @@ def _get_daily_data(self, asset_int, column, dt):
...
+        # but then cut it down to only the minutes after
+        # FIRST_TRADING_MINUTE
+        modified_minutes_for_window = minutes_for_window[
+            minutes_for_window.slice_indexer(FIRST_TRADING_MINUTE)]
+
+        modified_minutes_length = len(modified_minutes_for_window)
+
+        if modified_minutes_length == 0:
+            raise ValueError("Cannot calculate history window that ends"
+                             "before 2002-01-02 14:31 UTC!")
```
ehebert: Probably should use a variable for this message, external usage may not have the same dataset start date.

zipline/data/data_portal.py
```diff
@@ -920,69 +950,49 @@ def get_splits(self, sids, dt):
 
         return splits
 
-    def get_dividends(self, sids, dt):
+    def get_stock_dividends(self, sid, trading_days):
```
ehebert: We should have `get_dividends` and `get_stock_dividends` logic implemented in the same place. Whether it is both in data_portal or both in adjustment_reader, my vote is for it to be in adjustment_reader, so that the read/write responsibility is located in one place.

zipline/finance/risk/cumulative.py
```diff
@@ -90,9 +90,7 @@ class RiskMetricsCumulative(object):
         'information',
     )
 
-    def __init__(self, sim_params, env,
-                 create_first_day_stats=False,
-                 account=None):
```
ehebert: We may need to add this back in for broker support, not sure.

zipline/data/data_portal.py
jbredeche: Later, not a blocker right now.

zipline/data/data_portal.py
jbredeche: Yeah, there's a task on the punchlist for dealing with all the 1/1/2002 stuff that is currently in dataportal (there's a bunch of stuff at the top of the file that needs moving too).

zipline/data/data_portal.py
jbredeche: Our daily window logic: if the last day is partial, using minute data to calculate the value of the partial day.  also, creating an extra (empty) slot in the np array that is reserved for the partial day, so that we don't have to copy the array once we have daily data and need to add the partial data.

We could/should switch to USEquityPricingLoader and figure out how to maintain those details, but not important enough right now.  After shipping.

zipline/data/data_portal.py
jbredeche: if the history window ends at midnight UTC, and we're in daily mode, we can use daily data for the entire window (instead of building the last value from the last day's minute data).

zipline/gens/tradesimulation.py
```diff
@@ -186,51 +190,135 @@ def once_a_day(midnight_dt):
...
+                raise BenchmarkAssetNotAvailableTooLate(
+                    sid=str(sid),
+                    dt=trading_days[0],
+                    end_dt=benchmark_asset.end_date
+                )
+
+            # get the window of close prices for benchmark_sid from the last
+            # trading day of the simulation, going up to one day before the
+            # simulation start day.
+            benchmark_series = data_portal.get_history_window(
```
ehebert: It is good that we can use history here.
However, the usage of this function would read clearer (at least to me) in this context if the parameters were start_dt and end_dt, instead of `len(trading_days) + 1`.

(I could concede that that may be just shuffling around the abstraction.)

zipline/gens/tradesimulation.py
```diff
@@ -186,51 +190,135 @@ def once_a_day(midnight_dt):
...
+            if benchmark_asset.start_date > trading_day_before_sim_start:
+                # we can't go back one day before sim start, because the asset
+                # didn't start trading until the same day as the sim start.
+                # instead, we'll use the first available minute value of the
+                # first sim day.
+                minutes_in_first_day = \
+                    env.market_minutes_for_day(trading_days[0])
+
+                # get a minute history window of the first day
+                minute_window = data_portal.get_history_window(
```
ehebert: This works, I can't recall if we needed a data_portal "get spot price with ffill" elsewhere. We could use that type of method here instead.

tests/test_benchmark.py
jbredeche: we should, making a note, this pattern needs to change in a lot of tests.

PR #782, opened 2015-10-16
MAINT: Removes unneeded knowledge_date logic from future_chain

jfkirk:
The knowledge_date logic could lead to a situation where a futures are inserted to the middle of the chain unexpectedly, and that could lead to inconsistent behavior in trading. To avoid this, the knowledge_date logic has been removed.

It is not far-fetched to assume that all contracts are known at all points in a simulation, as futures follow regular schedules and there are few, if any, surprises.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -627,7 +627,7 @@ class AssetFinderTestCase(TestCase):
     def test_lookup_future_chain(self):
         metadata = {
             # Notice day is today, so should be valid.
-            2: {
+            0: {
                 'symbol': 'ADN15',
                 'root_symbol': 'AD',
                 'asset_type': 'future',
@@ -644,7 +644,7 @@ class AssetFinderTestCase(TestCase):
                 'start_date': pd.Timestamp('2015-01-01', tz='UTC')
             },
             # Starts trading today, so should be valid.
-            0: {
+            2: {
                 'symbol': 'ADF16',
                 'root_symbol': 'AD',
                 'asset_type': 'future',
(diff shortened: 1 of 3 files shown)


tests/test_assets.py
```diff
@@ -843,21 +848,18 @@ def tearDownClass(cls):
...
-        self.assertEqual(len(cl), 0)
-
-        # Sids 0, 1, & 2 have started, 3 has not yet started.
+        # Sids 0, 1, & 2 have started, 3 has not yet started, but all are in
+        # the chain
         cl = FutureChain(self.asset_finder, lambda: '2005-12-01', 'CL')
-        self.assertEqual(len(cl), 3)
+        self.assertEqual(len(cl), 4)
 
         # Sid 0 is still valid its notice date.
```
StewartDouglas: Could take the opportunity to correct this comment: "Sid 0 is still valid **on** its notice date."

StewartDouglas:
I think it makes sense to remove knowledge_date for futures, since unlike the world of equities we know of futures which *will* trade, but do not currently trade. So, as a user, I may want to post an order for say CLM25 and let it sit on the books, waiting to be filled. However, I do not think that this change is sufficient to address the inconsistent behavior concern. This is because the futures we supply in a user's FutureChain are simply those we receive from TDI (which they in turn get from the exchanges). This data has gaps in it, and we cannot predict when these gaps will be filled. So in the short term we should have our users code defensively to account for this. Whether we want to fill these gaps is a good question for the alpha.

jfkirk:
@StewartDouglas That makes sense, but within a given instance of Zipline I think we can only assume that the Futures metadata we are provided are complete and valid. Making any other assumptions with in Zipline (trying to infer contracts, setting patterns, etc) will set us up for pain later. Internal to Quantopian, we should take steps to ensure completeness and continuity of the Futures metadata.

zipline/assets/futures.py
```diff
@@ -126,15 +126,13 @@ def _maybe_update_current_chain(self):
             list
                 The up-to-date current chain, a list of Future objects.
         """
-        dt = self._get_datetime()
-
-        if (self._last_updated is None) or (self._last_updated != dt):
+        if (self._last_updated is None)\
+                or (self._last_updated != self.as_of_date):
```
yankees714: Was this a bug?

jfkirk: Nope, but there is no need to update unless the as_of_date property has changed, so this would avoid some unnecessary updates.

yankees714: Ah, makes sense.

StewartDouglas: For a given FutureChain will the `as_of_date` ever change? If not then the `if (self._last_updated is None) or (self._last_updated != self.as_of_date):` clause will always be `True`, so we can get rid of it.

yankees714: Yeah it will, unless `as_of_date` is specified, it will track the date of the simulation.
https://github.com/quantopian/zipline/blob/future-chain-start-date/zipline/assets/futures.py#L105

tests/test_assets.py
```diff
@@ -843,21 +849,18 @@ def tearDownClass(cls):
...
-        # Sid 0 is still valid its notice date.
+        # Sid 0 is still valid on its notice date.
         cl = FutureChain(self.asset_finder, lambda: '2005-12-20', 'CL')
-        self.assertEqual(len(cl), 3)
+        self.assertEqual(len(cl), 4)
 
         # Sid 0 is now invalid, leaving only Sids 1 & 2 valid.
         cl = FutureChain(self.asset_finder, lambda: '2005-12-21', 'CL')
-        self.assertEqual(len(cl), 2)
+        self.assertEqual(len(cl), 3)
```
yankees714: This comment makes it sound like there should only be 2 sids in the chain, should probably mention that 3 has not yet started.

yankees714:
LGTM

PR #784, opened 2015-10-20
various speedups

jbredeche:
Some speedups in event rules and data portal lookup.

### DIFF
--- a/tests/utils/test_events.py
+++ b/tests/utils/test_events.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import datetime
+from inspect import isabstract
 from itertools import islice
 from six import iteritems
 import random
@@ -232,7 +233,8 @@ class RuleTestCase(TestCase):
             k for k, v in iteritems(vars(zipline.utils.events))
             if isinstance(v, type) and
             issubclass(v, self.class_) and
-            v is not self.class_
+            v is not self.class_ and
+            not isabstract(v)
         }
         ds = {
             k[5:] for k in dir(self)
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -213,8 +213,7 @@ class DataPortal(object):
             not (column in BASE_FIELDS and isinstance(asset, Asset))
 
         if look_in_augmented_sources:
-            # we're being asked for a column that was added via fetcher to
-            # an existing sid
+            # we're being asked for a fetcher field
             try:
                 return self.augmented_sources_map[column][asset].\
                     loc[day_to_use, column]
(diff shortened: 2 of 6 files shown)


zipline/data/data_portal.py
```diff
@@ -248,21 +247,23 @@ def get_spot_value(self, asset, column, dt=None):
...
-                    dt = self.env.previous_market_minute(dt)
-
-                given_day = pd.Timestamp(dt.date(), tz='utc')
-                day_index = tradingcalendar.trading_days.searchsorted(
-                    given_day) - INDEX_OF_FIRST_TRADING_DAY
-                minute_index = self.env.market_minutes_for_day(given_day).\
-                    searchsorted(dt)
-
-                minute_offset_to_use = (day_index * 390) + minute_index
+                if dt == self.current_dt:
```
ssanderson: @jbredeche this might be a good use-case for encapsulating in a [CachedObject](https://github.com/quantopian/zipline/blob/master/zipline/utils/cache.py#L11)

zipline/utils/events.py
```diff
@@ -372,23 +372,64 @@ def should_trigger(self, dt, env):
         return dt.date() not in env.early_closes
 
 
-class NthTradingDayOfWeek(StatelessRule):
+class TradingDayOfWeekRule(StatelessRule):
```
richafrank: Should this be `Stateful` now (not sure the impact of changing the base class here)?

tests/utils/test_events.py
```diff
@@ -232,7 +232,9 @@ def test_completeness(self):
             k for k, v in iteritems(vars(zipline.utils.events))
             if isinstance(v, type) and
             issubclass(v, self.class_) and
-            v is not self.class_
+            v is not self.class_ and
+            # abstract class that doesn't need its own tests
+            v is not zipline.utils.events.TradingDayOfWeekRule
```
richafrank: Should we check for `abc` abstractness, as opposed to calling out this particular class?

zipline/utils/events.py
```diff
@@ -372,23 +372,64 @@ def should_trigger(self, dt, env):
...
-            raise _out_of_range_error(MAX_WEEK_RANGE)
-        self.td_delta = n
+        super(NthTradingDayOfWeek, self).__init__(n)
 
-    def should_trigger(self, dt, env):
-        return _coerce_datetime(env.add_trading_days(
-            self.td_delta,
-            self.get_first_trading_day_of_week(dt, env),
-        )).date() == dt.date()
+        self.date_func = self.get_first_trading_day_of_week
```
richafrank: Should we make this a class-level function assignment, instead of setting it per instance?  Could also make `date_func` abstract in the base class.

PR #790, opened 2015-10-21
Update next_market_minute() & previous_market_minute(), and add history() tests

StewartDouglas:
When working on test coverage for `history()` I wrote the following pure-zipline algorithm:

```python
from zipline.api import history

def initialize(context):
    context.first_loop = True

    if not context.first_loop:
        price_bts = history(bar_count=1, frequency='1m', field='price')
        context.price_bts = price_bts

    context.first_loop = False
```

Which raised the following error:

```
IndexError: index 0 is out of bounds for axis 0 with size 0
```

I traced this issue back to the following method, which is invoked by the history container:

https://github.com/quantopian/zipline/blob/master/zipline/finance/trading.py#L340-L348

We call this method with a 'start' of midnight of this trading day, since this is when `before_trading_start()` is time-stamped to. `next_market_minute()` then assumes
that, since the minute after 'start' is not during market hours, we must want the market open on the *next* day. This means that we do not correctly record the data of the *current* day, which ultimately results in the index error.

To address this I add a check of whether 'start' is before the market-open. If it is, then the method attempt to get the market open of the same day.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -22,6 +22,7 @@ from nose_parameterized import parameterized
 import numpy as np
 import pandas as pd
 from pandas.util.testing import assert_frame_equal
+from pandas.tseries.tools import normalize_date
 
 from zipline.history import history
 from zipline.history.history_container import HistoryContainer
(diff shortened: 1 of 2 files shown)


zipline/finance/trading.py
```diff
@@ -340,11 +341,25 @@ def previous_open_and_close(self, start_date):
...
+        next minute, the open of the same day if start is before the market
+        open on a trading day, or the open of the next market day after start.
         """
         next_minute = start + datetime.timedelta(minutes=1)
         if self.is_market_hours(next_minute):
             return next_minute
+        # Check whether start is before the market open.
+        # If it is, then next_minute should be the market
+        # open.
+        if start.time() < time(hour=9, minute=31):
```
StewartDouglas: `time(hour=9, minute=31)` should probably be replaced by `self.get_open_and_close(start)[0]`.

zipline/finance/trading.py
```diff
@@ -340,11 +341,25 @@ def previous_open_and_close(self, start_date):
...
+            try:
+                next_minute = self.get_open_and_close(start)[0]
+                # Confirm that next_minute is during market hours.
+                if self.is_market_hours(next_minute):
```
richafrank: When does get_open_and_close return a value outside of market hours?

zipline/finance/trading.py
richafrank: Agreed.

zipline/finance/trading.py
```diff
@@ -340,11 +341,25 @@ def previous_open_and_close(self, start_date):
...
+                    return next_minute
+            except KeyError:
+                # We tried to get the open on a non-trading day. Instead,
+                # proceed and return the open of the next trading day.
+                pass
         return self.next_open_and_close(start)[0]
 
     def previous_market_minute(self, start):
```
richafrank: Should we update prev_market_minute too?

zipline/finance/trading.py
StewartDouglas: Although will have to wrap the inequality check in a `try..else` to guard against `get_open_and_close()` being fed a `start` which isn't a trading day.

zipline/finance/trading.py
StewartDouglas: Good point.

StewartDouglas:
Refactored.

StewartDouglas:
Added tests.

tests/test_history.py
```diff
@@ -557,6 +558,277 @@ def handle_data(context, data):
...
+        # ---------------------------------
+        # 2006-03-20 21:00:00    139.369469
+        # 2006-03-21 21:00:00    180.156620
+        # 2006-03-22 21:00:00    221.344654
+
+        #       When called within BTS
+        # ---------------------------------
+        # 2006-03-17 21:00:00           NaN
+        # 2006-03-20 21:00:00    139.369469
+        # 2006-03-22 00:00:00    180.156620
```
richafrank: Thanks for all the doco!

tests/test_history.py
```diff
@@ -557,6 +558,277 @@ def handle_data(context, data):
...
+        source = RandomWalkSource(start=start, end=end)
+        output = test_algo.run(source)
+        self.assertIsNotNone(output)
+
+        # Get the prices recorded by history() within handle_data()
+        prices_hd = test_algo.prices_hd[0]
+        # Get the prices recorded by history() within BTS
+        prices_bts = test_algo.prices_bts[0]
+
+        # before_trading_starts() is timestamp'd to midnight prior to
```
richafrank: typo:  should be "before_trading_start"

tests/test_history.py
```diff
@@ -557,6 +558,277 @@ def handle_data(context, data):
...
+
+def before_trading_start(context, data):
+    if not context.first_day:
+        prices_bts = history(bar_count=3, frequency='1d', field='price')
+        context.prices_bts = prices_bts
+
+def handle_data(context, data):
+    prices_hd = history(bar_count=3, frequency='1d', field='price')
+    context.prices_hd = prices_hd
+    context.first_day = False
```
richafrank: Should we mark `first_day = False` in before_trading_start?  That seems more related to the day turnover.

tests/test_history.py
```diff
@@ -557,6 +558,277 @@ def handle_data(context, data):
...
+def initialize(context):
+    context.first_loop = True
+
+def before_trading_start(context, data):
+    if not context.first_loop:
+        price_bts = history(bar_count=1, frequency='1m', field='price')
+        context.price_bts = price_bts
+
+def handle_data(context, data):
+    context.first_loop = False
```
richafrank: Same question here about moving this statement to bts.

tests/test_history.py
```diff
@@ -557,6 +558,277 @@ def handle_data(context, data):
...
+        # 26 27 28 29 30 31
+        start = pd.Timestamp('2006-03-20', tz='UTC')
+        end = pd.Timestamp('2006-03-22', tz='UTC')
+
+        sim_params = factory.create_simulation_parameters(
+            start=start, end=end)
+
+        test_algo = TradingAlgorithm(
+            script=algo_text,
+            data_frequency='minute',
```
richafrank: Do we need any new tests for 'daily' frequency?

tests/test_history.py
StewartDouglas: yup, good catch.

tests/test_history.py
```diff
@@ -557,6 +558,277 @@ def handle_data(context, data):
...
+
+        # The prices recorded by history() in BTS should
+        # be the closing price of the previous day, which are:
+        #
+        #          sid | close on 2006-03-21
+        #         ----------------------------
+        #           0  | 180.15661995395106
+        #           1  | 578.41665003444723
+
+        self.assertEquals(180.15661995395106, price_bts_0.ix[0])
```
richafrank: Are these values calculable or importable from somewhere else?  Would prefer not to use a seemingly "magic" value.

zipline/finance/trading.py
```diff
@@ -340,21 +340,37 @@ def previous_open_and_close(self, start_date):
...
+        if self.is_trading_day(start):
+            market_open, market_close = self.get_open_and_close(start)
+            # If start after the market close, return market close.
+            if start > market_close:
+                return market_close
+            # If start is during trading hours, then get previous minute.
+            if start > market_open:
+                return start - datetime.timedelta(minutes=1)
+        # If start is not a trading day, or is before the market open
+        # then return the close of the *previous* trading day.
```
richafrank: Nice - these changes are very clear.

tests/test_history.py
```diff
@@ -557,6 +558,277 @@ def handle_data(context, data):
...
+            sim_params=sim_params,
+            env=TestHistoryAlgo.env,
+        )
+
+        source = RandomWalkSource(start=start, end=end)
+        output = test_algo.run(source)
+        self.assertIsNotNone(output)
+
+        # Get the volume recorded by history() within handle_data()
+        volume_hd = test_algo.volume_hd[0]
```
richafrank: Should we assert anything about the rest of the values, e.g. indices > 0?

tests/test_history.py
StewartDouglas: Sure, but then calling it `first_day` won't make sense, since we'll enter `handle_data` on the first trading day, yet `first_day == False`. Maybe call the flag `first_bts_call` intead.

richafrank: Makes sense to me.

tests/test_history.py
StewartDouglas: Could do the exact same thing for `sid == 1` - not sure why I ignored it.

tests/test_history.py
StewartDouglas: Yup, I think that would be sensible. I could just create a second `TradingAlgorithm` in each of the four tests, then run them with the same `algo_text`s. Shouldn't raise the computational overhead too much, given that they're daily.

tests/test_history.py
StewartDouglas: I'm not sure. I could do the same thing as `test_history_in_bts_price_days`, where I call `history()` in `handle_data()` and compare it with those generated in BTS. But I've deliberately avoided this, since it was not doing this which caused the bug in `next_market_minute` to emerge. This isn't real trade data - it's generated by a random walk (but we don't vary the seed, so we know what it will be.)

richafrank: Oh gotcha - Could you add a comment that says these are what the random walk source will produce, given the seed?

tests/test_history.py
StewartDouglas: Updated so that those two tests which call daily `history()` are run in both minutely an daily mode. I've not run the minutely `history()` calls in daily mode as this raises an `ncompatibleHistoryFrequency` exception.

richafrank: Makes sense to me!

PR #791, opened 2015-10-21
MAINT: Set dividend effective date to ex_date.

ehebert:
The price shock occurs on the effective_date. Had changed the effective_date to
be day before the ex_date with the belief that pipeline was applying values up
and until the effective_date, but the lookback windows apply before the
effective_date. Thus, the price shock calculation should still use the previous
days data but be dated on the ex_date to stay aligned with splits and
merger dating.


zipline/data/us_equity_pricing.py
```diff
@@ -662,7 +662,7 @@ def calc_dividend_ratios(self, dividends):
                     ratio = 1.0 - amount / prev_close
                     ratios[i] = ratio
                     # only assign effective_date when data is found
-                    effective_dates[i] = div_adj_date.value
+                    effective_dates[i] = ex_date
```
ssanderson: Is `div_adj_date` unused now?

ehebert: It is passed to the spot_price reader to get the previous close.

ssanderson: Ah.  Can we rename it to something like `prev_close_date`, since we're no longer using it as the adjustment date?  Otherwise this looks good to me.

ehebert: Changed in a fixup, good idea.

PR #792, opened 2015-10-22
ENH: Allow pipelines to run with matching start/end dates

jdricklefs:
I think restricting this was just paranoia.  This is required for use in live trading (where the start and end dates are the same on day 1).

(Opening PR to get travis to run)

### DIFF
--- a/tests/pipeline/test_engine.py
+++ b/tests/pipeline/test_engine.py
@@ -168,11 +168,24 @@ class ConstantInputTestCase(TestCase):
 
         p = Pipeline()
 
-        msg = "start_date must be before end_date .*"
+        msg = "start_date must be before or equal to end_date .*"
         with self.assertRaisesRegexp(ValueError, msg):
             engine.run_pipeline(p, self.dates[2], self.dates[1])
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_engine.py
```diff
@@ -168,11 +168,9 @@ def test_bad_dates(self):
 
         p = Pipeline()
 
-        msg = "start_date must be before end_date .*"
+        msg = "start_date must be before or equal to end_date .*"
         with self.assertRaisesRegexp(ValueError, msg):
             engine.run_pipeline(p, self.dates[2], self.dates[1])
-        with self.assertRaisesRegexp(ValueError, msg):
-            engine.run_pipeline(p, self.dates[2], self.dates[2])
```
richafrank: Should this be replaced by an assertion that it *does* run in this case?

jdricklefs: Sure, what the heck.

jdricklefs: (Added a new test case, since it's no longer a 'bad date')

richafrank: :+1:

PR #793, opened 2015-10-22
Treasury cleanup

ssanderson:
Rewrites most of `load_market_data` and the functions it calls.

The bulk of the work here is in rewriting the treasury loader and benchmark loader to use `pd.read_csv` with standard arguments instead of our own homegrown parser.  See the message for 5381f140a533e981db356d3eeef69c052561cffa for a detailed analysis of the treasury loader change.

The benchmark returns loader has some more significant deviations from the previous behavior, but I don't see anything wrong with the code.  @ehebert do you know of a good place to find reference values for S&P500 returns to check my current results against the previously-generated results?


ehebert:
One idea for double checking against old calculations would be to look at the output/cache in ~/.zipline/data using the different versions.

Loosely:
- Checkout master.
- `rm ~/.zipline/data/\^GSPC_benchmark.csv`
- Run unit tests.
- `cp ~/.zipline/data/\^GSPC_benchmark.csv ~/.zipline/data/\^GSPC_benchmark.csv.bak`
- `rm ~/.zipline/data/\^GSPC_benchmark.csv`
- Checkout new branch.
- Compare the two files.

ehebert:
Overall LGTM, mostly just a few style questions with one real question about whether or not we should convert `US/Eastern` and then back to UTC before applying the trading day offset.

ssanderson:
> Checkout master.
rm ~/.zipline/data/\^GSPC_benchmark.csv
Run unit tests.
cp ~/.zipline/data/\^GSPC_benchmark.csv ~/.zipline/data/\^GSPC_benchmark.csv.bak
rm ~/.zipline/data/\^GSPC_benchmark.csv
Checkout new branch.
Compare the two files.

I've already done that in both cases.  For the treasuries there's exactly one case where there's a potentially significant (on the order of 0.002) difference for a single date on the 3month curve, which is due to the fact that data.treasury.gov and federalreserve.gov produce different results there.  I'm not sure how to verify that one or the other is correct.

For the returns benchmarks there are more nontrivial differences, which is why I was looking for a canonical place to compare our current return calculations vs the old ones.

PR #794, opened 2015-10-24
ENH: vectorize mean algorithm returns calculation

jbredeche:
In a sample backtest on my machine, this takes the final risk
calculations down from ~10 seconds to ~0.8 seconds.

### DIFF
--- a/zipline/finance/risk/period.py
+++ b/zipline/finance/risk/period.py
@@ -100,12 +100,9 @@ class RiskMetricsPeriod(object):
         self.num_trading_days = len(self.benchmark_returns)
         self.trading_day_counts = pd.stats.moments.rolling_count(
             self.algorithm_returns, self.num_trading_days)
-        self.mean_algorithm_returns = pd.Series(
-            index=self.algorithm_returns.index)
-        for dt, ret in self.algorithm_returns.iteritems():
-            self.mean_algorithm_returns[dt] = (
-                self.algorithm_returns[:dt].sum() /
-                self.trading_day_counts[dt])
+
+        self.mean_algorithm_returns = \
+            self.algorithm_returns.cumsum() / self.trading_day_counts
 
         self.benchmark_volatility = self.calculate_volatility(
             self.benchmark_returns)


ehebert:
This looks good to me @jbredeche, how many years was your test running over?
My gut says this scales with length of backtest.

jbredeche:
Was running over ~6 years, in which we were calculating ~240 risk periods.

ssanderson:
> This looks good to me @jbredeche, how many years was your test running over?
My gut says this scales with length of backtest.

http://accidentallyquadratic.tumblr.com/

zipline/finance/risk/period.py
```diff
@@ -100,12 +100,9 @@ def calculate_metrics(self):
...
         self.trading_day_counts = pd.stats.moments.rolling_count(
             self.algorithm_returns, self.num_trading_days)
-        self.mean_algorithm_returns = pd.Series(
-            index=self.algorithm_returns.index)
-        for dt, ret in self.algorithm_returns.iteritems():
-            self.mean_algorithm_returns[dt] = (
-                self.algorithm_returns[:dt].sum() /
-                self.trading_day_counts[dt])
+
+        self.mean_algorithm_returns = \
```
ssanderson: How does this behave on the first day?  I assume it produces a `NaN`?  Is that consistent with the old behavior?

ehebert:
Right, should have said 'scaled drastically with respect to simulation length'

zipline/finance/risk/period.py
ehebert: The first element of the vectors `algorithm_returns` and `mean_algorithm_returns` are equal, i.e. they are both the return of the first day.

That behavior is shared both on existing behavior and with this patch which uses `cumsum`.

Quick snippet to confirm the `cumsum` behavior http://docs.scipy.org/doc/numpy/reference/generated/numpy.cumsum.html
```
In [2]: np.cumsum([0.001, 0.002, 0.003])
Out[2]: array([ 0.001,  0.003,  0.006])
```

ssanderson: Ah, that makes sense since we're just summing not computing a difference.

ssanderson:
LGTM to me as well.

PR #795, opened 2015-10-25
MAINT: Only use pip in travis.

ssanderson:
Rather than a hodgepodge of pip and conda.  To make downloads not take
forever, download wheels from our wheelserver.


richafrank:
:+1:

ssanderson:
@richafrank I'd save the :+1: for when I actually get this working... :sob:

richafrank:
You can count it as my cheering you on.  :tada:

ssanderson:
Things I have learned tonight:

1. Travis' default intrastructure is based on Ubuntu 12.04.  Our wheels are built on a machine using Ubuntu 14.04.  This means that we can successfully install our numpy wheel, but at runtime it fails trying to find `liblapack.so.3`.  On 14.04, this shared object is provided by `liblapack3` or `libatlas3-base`.  On 12.04, the analogous `.so` files are provided by `liblapack3gf`, but they're named `liblapack.so.3gf`.  The name mangling seems to be part of a [historical migration strategy for Debian switching fortran compilers](https://wiki.debian.org/GfortranTransition).  It's not clear to me whether it's safe or reasonable to simply copy over the `.3gf` shared objects to where our Ubuntu 14.04 wheels expect them to be.

2. Travis has [beta support](http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/) for building on Ubuntu 14.04.  You can enable this by setting `sudo: required` and `dist: trusty` in your .travis.yml.  However, the Python 2.7.10 installed on the new image appears to have been compiled in a different unicode configuration than the machine that builds our wheels.  When I try to use them on Travis' 14.04 infrastructure, I get failures trying saying `undefined symbol: PyUnicodeUCS4_AsUnicodeEscapeString`.  This appears to happen because our wheels are built against a Python installation that was compiled with `--enable-unicode=ucs4`, whereas Travis' Python is compiled with `--enable-unicode=ucs2`.

ssanderson:
Closing this since I don't think this particular strategy will be worthwhile.

PR #796, opened 2015-10-26
ENH: Fail when history() is called in initialize.

ssanderson:


ssanderson:
@StewartDouglas I assigned this to you since you've been working on tests for the behavior of `history()` in BTS.

StewartDouglas:
ACK

zipline/errors.py
```diff
@@ -191,6 +191,13 @@ class IncompatibleHistoryFrequency(ZiplineError):
 """.strip()
 
 
+class HistoryInInitialize(ZiplineError):
+    """
+    Raised when an algorithm calls history() in initialize.
+    """
+    msg = "history() can only be called in handle_data()"
```
StewartDouglas: `history()` can also be called in `before_trading_start()`. I suggest changing this to either "history() *should* only be called in handle_data()", or "history() can not be called in initialize()"

tests/test_history.py
```diff
@@ -1329,6 +1329,41 @@ def handle_data(context, data):
...
+
+        test_algo = TradingAlgorithm(
+            script=algo_text,
+            data_frequency='minute',
+            sim_params=sim_params,
+            env=self.env,
+        )
+
+        with self.assertRaises(HistoryInInitialize):
+            test_algo.initialize()
```
StewartDouglas: In addition to calling `initialize()` directly we could also check that `HistoryInInitialize ` is raised when we call `.run()`. For example, we could add the following before hand: `source = RandomWalkSource(start=start, end=end)` and add the check:
```python
with self.assertRaises(HistoryInInitialize):
    test_algo.run(source)
```
Ultimately this will call `.initialize()`, but it would improve coverage. If you think it's unnecessary then just leave as is.

StewartDouglas:
Other than my two comments above this looks great :+1:

zipline/errors.py
ssanderson: Is calling history in `before_trading_start` actually **supported**?  We've made sure it doesn't crash, but afaik there's basically no guarantees about its behavior, right?

tests/test_history.py
ssanderson: If `run()` doesn't call `initialize` **lots** of other tests should fail.  I don't think we need to be asserting that here, since this test is specifically about the fact that we fail *in initialize* when this happens.

zipline/errors.py
ssanderson: I could see the argument for changing this to `should`.

StewartDouglas: `before_trading_start()` behaves as expected in `history()`, *if* it is not called on the first day (in daily mode) or minute (in minutely mode), as in this case the `data` object is not populated and the algo will fail with an `IndexError`. Since we don't output a more informative message it probably can't be called **supported**. We should probably change `can` -> `should`, but not a big deal.

PR #800, opened 2015-10-29
pass benchmark source into tradesim

jbredeche:
(this branch's name is a misnomer)

no functional change.  move benchmark code into its own object, which is passed into tradesim.


ehebert:
:+1: did not read line by line, but I agree with the change to passing the benchmark_source in as a parameter, and the interface of calling `.get(dt)` on the benchmark.

Issue #783, opened 2015-10-18
Unexpected PnL when flipping futures position


dhexus:
Hi. I've a futures algorithm where pnl does not match expected pnl. So I wrote a little test. I use instant_fill=True, no slippage, and no commission. Moreover I provide only three data events of type TRADE, to make it very easy. I use a DataPanelSource that holds the following data for the relevant asset:
``` 
data = pd.DataFrame([
            [1, 1e9, int(DATASOURCE_TYPE.TRADE)],
            [2, 1e9, int(DATASOURCE_TYPE.TRADE)],
            [4, 1e9, int(DATASOURCE_TYPE.TRADE)]],
            columns=['price', 'volume', 'type'], index=index)
```
My handle_data method looks like this:
```
def handle_data(self, data):
        if len(self.portfolio.positions) > 0:
            if self.portfolio.positions[self.asset.sid]["amount"] > 0:
                self.order_target(self.asset, -self.amount)
            else:
                self.order_target(self.asset, 0)
        else:
            self.order_target(self.asset, self.amount)
```
So I go long if there is no position, I go short when I'm long, and I go flat when I'm short. Moreover, I have  amount=1 and multiplier=168. In results object, positions shows exactly the number of contracts that I would expect, i.e. [1,-1,0]. However, pnl is [0, 0, 0]. Ignoring initial margin, I would expect [0, 168, -336].

CC @warren-oneill @jfkirk @StewartDouglas @ssanderson

warren-oneill:
@jfkirk @StewartDouglas have you guys had a chance to look at this? Can you you reproduce our error? Is this behavior already tested for futures in zipline?

jfkirk:
@dhexus @warren-oneill I am attempting to reproduce now and will update when I have reproduced or if I need more information.

Self-assigning.

warren-oneill:
@jfkirk thanks :smile:

jfkirk:
Thanks to Warren's test, I've been able to reproduce the issue. It appears that the issue is caused by the Position's last_sale_price already being updated before PositionTracker.update_last_sale is called for each trade event. This causes the cash flow calculation to return 0s because price == old_price:

```
price = event.price

        if checknull(price):
            return 0

        pos = self.positions[sid]
        old_price = pos.last_sale_price
        pos.last_sale_date = event.dt
        pos.last_sale_price = price

        # Calculate cash adjustment on assets with multipliers
        return ((price - old_price) * self._position_payout_multipliers[sid]
[...]

PR #802, opened 2015-10-30
TST: adds future flip test

warren-oneill:
adds unittest based on issue #783. 
Incorrect pnl calculated when algo flips from long to short (with correct algo positions).

cc @jfkirk @StewartDouglas @ssanderson @dhexus


warren-oneill:
closed with  #803

PR #803, opened 2015-10-30
Future flip fix

jfkirk:
Previously, when a transaction executed on a new bar, the transaction would adjust the position's last_trade_price field before the trade event would come through to calculate a cash flow. This caused missed cash flows on bars where transactions executed.

This fixes the issue by processing all transactions as trade events before processing them as transactions.

Addresses issue https://github.com/quantopian/zipline/issues/783

Uses commit from https://github.com/quantopian/zipline/pull/802

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -288,16 +288,18 @@ class PerformanceTracker(object):
 
         return _dict
 
-    def process_trade(self, event):
-        # update last sale, and pay out a cash adjustment
+    def _handle_event_price(self, event):
+        # updates last sale, and pays out a cash adjustment if applicable
         cash_adjustment = self.position_tracker.update_last_sale(event)
         if cash_adjustment != 0:
             for perf_period in self.perf_periods:
                 perf_period.handle_cash_payment(cash_adjustment)
 
+    def process_trade(self, event):
+        self._handle_event_price(event)
+
     def process_transaction(self, event):
-        # Transaction events also count as trade events for the trackers
-        self.process_trade(event)
+        self._handle_event_price(event)
         self.txn_count += 1
         self.position_tracker.execute_transaction(event)
         for perf_period in self.perf_periods:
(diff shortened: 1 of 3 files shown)


jfkirk:
@warren-oneill This PR contains your commit to add the test. Thanks!

warren-oneill:
Awesome! Thanks

warren-oneill:
Btw I recycled code from the close position algo text for the check pnl and check positions functions. Prob could be done more elegantly.

zipline/finance/performance/tracker.py
```diff
@@ -296,6 +296,8 @@ def process_trade(self, event):
                 perf_period.handle_cash_payment(cash_adjustment)
 
     def process_transaction(self, event):
+        # Transaction events also count as trade events for the trackers
+        self.process_trade(event)
```
ehebert: Though the logic applied is the same, calling `process_trade` here is confusing.

Not sure what the right fix for that would be, some candidates. 1) would be 'inlining' via copy/paste the handle cashe logic into process_transaction. 2) Or make a function with some common name that both process_trade and process_transaction call.

ehebert: Also, this is another place where lazy-mainline handles the behavior in a different place. (Where process_trade is no longer called, the cash adjustment is done directly in the tracker.)

jfkirk: Something like:

```
    def _update_last_sale(self, event):
        # updates last sale, and pay out a cash adjustment
        cash_adjustment = self.position_tracker.update_last_sale(event)
        if cash_adjustment != 0:
            for perf_period in self.perf_periods:
                perf_period.handle_cash_payment(cash_adjustment)

    def process_trade(self, event):
        self._update_last_sale(event)

    def process_transaction(self, event):
        self._update_last_sale(event)
        self.txn_count += 1
        self.position_tracker.execute_transaction(event)
        for perf_period in self.perf_periods:
            perf_period.handle_execution(event)
```
? Would this create an unnecessary attribute access on process_trade that could impact perf (or is that negligible)?

richafrank: > Would this create an unnecessary attribute access on process_trade that could impact perf (or is that negligible)?

I wouldn't expect attribute accesses on the perf tracker to have a noticeable impact.

Issue #801, opened 2015-10-30
Initializing TradingAlgorithm Object Does Not Set _analyze Method in algorithm.py


elmq0022:
The following code does not call the passed analyze function when trying to work directly/manually with zipline (i.e. not from ipython or command line)

```python
import pytz
from zipline.api import order, record, symbol, history, add_history, order_target
from datetime import datetime
from zipline.algorithm import TradingAlgorithm
from zipline.utils.factory import load_bars_from_yahoo

start = datetime(2000, 1, 1, 0, 0, 0, 0, pytz.utc)
end = datetime(2001, 1, 1, 0, 0, 0, 0, pytz.utc)
data = load_bars_from_yahoo(stocks=['AAPL'],start=start, end=end)

def initialize(context):
   pass

def handle_data(context, data):
    order(symbol('AAPL'), 10)
    record(APPL=data[symbol('AAPL')].price)

def analyze(context, perf):
   print(perf.head())
    
algo_obj = TradingAlgorithm(initialize=initialize, handle_data=handle_data, analyze=analyze)
perf= algo_obj.run(data)
```

I think it can be fixed by updating the algorithm.py file with the last line of the following code. 

```python
elif kwargs.get('initialize', False) and kwargs.get('handle_data'):
            if self.algoscript is not None:
                raise ValueError('You can not set script and \
                initialize/handle_data.')
            self._initialize = kwargs.pop('initialize')
            self._handle_data = kwargs.pop('handle_data')
            self._analyze = kwargs.pop('analyze', None)
```

I'd update the code and submit a pull request, but I'm not 100% if it's an issue.

richafrank:
@elmq0022 That fix makes sense to me - want to open a PR?

I suggest adding that new `analyze` line after we set `self._before_trading_start`, just for consistency with the previous `if` block.

elmq0022:
@richafrank Thanks for the feedback and the suggested placement for consistency.  I'll submit a pull request sometime this weekend.

richafrank:
Sounds great!

richafrank:
Fixed with #819.

PR #806, opened 2015-11-01
BUG: Issue #801 Initializing TradingAlgorithm Object Does Not Set

elmq0022:
_analyze Method in algorithm.py

Added one line to check for the keyword argument 'analyze' and set the
the _analyze method when a TradingAlgorithm object is initialized
within a script.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -266,6 +266,7 @@ class TradingAlgorithm(object):
         self.algoscript = kwargs.pop('script', None)
 
         self._initialize = None
+        self._analyze = kwargs.pop('analyze', None)
         self._before_trading_start = None
         self._analyze = None
 
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -264,9 +264,7 @@ class TradingAlgorithm(object):
         # If string is passed in, execute and get reference to
         # functions.
         self.algoscript = kwargs.pop('script', None)
-
         self._initialize = None
-        self._analyze = kwargs.pop('analyze', None)
         self._before_trading_start = None
         self._analyze = None
 
@@ -297,6 +295,7 @@ class TradingAlgorithm(object):
             self._handle_data = kwargs.pop('handle_data')
             self._before_trading_start = kwargs.pop('before_trading_start',
                                                     None)
+            self._analyze = kwargs.pop('analyze', None)
 
         self.event_manager.add_event(
             zipline.utils.events.Event(
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -1842,3 +1842,36 @@ class TestClosePosAlgo(TestCase):
                 actual_position, expected_positions[i],
                 "position for day={0} not equal, actual={1}, expected={2}".
                 format(i, actual_position, expected_positions[i]))
+
+
+class TestTradingAlgorithm(TestCase):
+    def setUp(self):
+        self.env = TradingEnvironment()
+        self.days = self.env.trading_days[:4]
+        self.panel = pd.Panel({1: pd.DataFrame({
+            'price': [1, 1, 2, 4], 'volume': [1e9, 1e9, 1e9, 0],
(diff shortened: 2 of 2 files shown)


richafrank:
Thanks @elmq0022, looking now!

zipline/algorithm.py
```diff
@@ -266,6 +266,7 @@ def __init__(self, *args, **kwargs):
         self.algoscript = kwargs.pop('script', None)
 
         self._initialize = None
+        self._analyze = kwargs.pop('analyze', None)
```
richafrank: From the two cases below, it looks like we should assign to `self._analyze` from `kwargs` only in the second case, where we assign all our functions from `kwargs`.  (In the first case, we assign it from `self.namespace`.)  What do you think about moving this assignment to the body of the `elif` below?  

Also, did you find this worked for you, since we re-assign `self._analyze = None` two lines below this?

elmq0022: I should have checked this more closely.  Perhaps a test is warranted.  

I had it working from the elif previously, but moved per a previous suggestion.  Perhaps I misinterpreted the exact location.  

I will update and resubmit a PR later tonight.  

By the way is there a specific branch I should use for the PR?

Thanks.

elmq0022: As far as a check for execution could I just pass a an analysis function that returns True as a test?

richafrank: > I should have checked this more closely. Perhaps a test is warranted.

That's a great idea - want to add a test to, say, `test_algorithm.py`?

> I had it working from the elif previously, but moved per a previous suggestion. Perhaps I misinterpreted the exact location. 

Sorry about that.  I was thinking we should put it at the end of the block that currently looks like:
```python
             self._initialize = kwargs.pop('initialize')
             self._handle_data = kwargs.pop('handle_data')
             self._before_trading_start = kwargs.pop('before_trading_start',
                                                     None)
```
> By the way is there a specific branch I should use for the PR?

When upstreaming changing, my workflow is to create (in my fork) a new branch with a descriptive name, PR it against the upstream master, like you've done here, and then once it's merged, I update my fork's master with the changes from upstream/master.  That way my master is always in sync with upstream.  (I delete the branch afterwards.)

richafrank: > As far as a check for execution could I just pass a an analysis function that returns True as a test?

It looks like the consumer of `self._analyze` (in `analyze`) doesn't actually check the return value and will return anyway if `self._analyze` is unset.  

If `analyze` gets called in your test via `TradingAlgorithm.run`, then your analysis function could save a reference to the perf stats it's passed, and you could compare that it was passed the same object that was returned by `run`.  That seems to be the `analyze` interface.  Does that sound right to you?

elmq0022: Yes that seems right too me. I coded something up to that effect this
morning.

I couldn't quite get the test to run without throwing an error.

I will have to take a look tomorrow and see if I can track down the issue.

By the way, is there a guide on setting up an environment for developing
zipline?

elmq0022: Well, I saw this resource,
https://github.com/quantopian/zipline/wiki/Development-Environment.

But there is nothing about developing on windows.  I guess that's to be
expected.

richafrank: Yep, that's what I would have pointed to you, and sorry our support of windows is less than adequate...

elmq0022: Most of this test is just from python, so maybe I can get it to work on my
system in albeit some hackish way.

Alternatively, Is there some way to call Travis CI client on my fork of the
code base without requesting a pull?

richafrank: I don't know of a way to call Travis without requesting a pull, but if you don't want to open a PR against quantopian/zipline master, you could move your changes to a branch and PR against your own master.

Running your branch locally, the issue I'm getting is about metadata missing from the assets db.  I think if you pass in a panel directly to `run`, then it will update the db and create the data source for you.

elmq0022: Those are both extremely appreciated tips, thank you.

PR #807, opened 2015-11-02
CCY MB - Currency Pair initial work

michaeljohnbennett:
Here is the initial work I was talking about regards adding currency support to the Zipline asset finder engine and eventually the TradingAlgorithm engine.

Be good to see if this is on the right track from your point of view.


michaeljohnbennett:
I'll get the flake8 stuff sorted tomorrow, sorry, I enjoy the use of a 27" monitor so lines longer than 79 characters don't really bug me 8-)

jfkirk:
Awesome! I've just given it a quick glance and it looks like its definitely on the right track. I'll go through it in detail tomorrow and provide feedback.

Thank you!

michaeljohnbennett:
Thanks

Need to clean up and integrate the changes from the master branch and do
and formatting but in general its there.

Need more tests around the lookups then move into the actual currency
calculations.

michaeljohnbennett:
@jfkirk I fixed some of the code and the tests are passing for the initial code, hope the rest of the code/etc is ok. Let me know if you think it needs refactoring. I have added in that fix to the index.view(np.int64) as well.

jfkirk:
Self-assigning while I provide feedback

jfkirk:
Could you clarify for me the significance of:
```'ccy' vs 'currency' vs 'currencypair'```
? It seems fairly obvious, but I'd like to hear your intentions so I can be sure my expectations are all in line.

**Comment by michaeljohnbennett** [comment]
ccy was shorthand for the currency in an equity to designate its base currency (ccy is just a common term I have used in previous jobs...) 

Currency was the table but I needed to make the distinction between a currencyPair/FX and a currency itself.

If you had a better way to name then please pop in some ideas. 

I think the domain needs:

- Assets to have a currency/ccy like 'GBP','USD' etc
- A currency/fx pair to show you the crossing between them USDGBP this could also have an arbitrary symbol from the data provider (QF4 in CSI world) and a SID which is your UUID of sorts. This needs a major and minor currency/ccy on it so we can look them up from each end
- Currency Tables to hold these asset data identifiers in. perhaps this is just currency pair? I was trying to follow the convention in the AssetWriter somewhat but It might have got a little off track.

Thanks for the comments and feedback.

File: zipline/assets/_assets.pyx [hunk 1]
```diff
@@ -42,25 +39,35 @@ cdef class Asset:
...
+        self.sid = sid
+        self.sid_hash = hash(sid)
+        self.symbol = symbol
+        self.asset_name = asset_name
+        self.exchange = exchange
+        self.start_date = start_date
+        self.end_date = end_date
+        self.first_traded = first_traded
+        self.ccy = ccy
+        self.price_format = price_format
```
jfkirk: What is the intended function of price_format?

zipline/assets/_assets.pyx
```diff
@@ -42,25 +39,35 @@ cdef class Asset:
...
-        self.first_traded  = first_traded
+        self.sid = sid
+        self.sid_hash = hash(sid)
+        self.symbol = symbol
+        self.asset_name = asset_name
+        self.exchange = exchange
+        self.start_date = start_date
+        self.end_date = end_date
+        self.first_traded = first_traded
+        self.ccy = ccy
```
jfkirk: For clarity, could this be 'currency'? Possibly 'quote_currency'?

zipline/assets/_assets.pyx
```diff
@@ -42,25 +39,35 @@ cdef class Asset:
...
+        self.price_format = price_format
+        self.status = status
```
jfkirk: What is the intended function of status? Is it time-varying?

zipline/assets/_assets.pyx
```diff
@@ -298,6 +313,80 @@ cdef class Future(Asset):
...
+                  float cvf=1,
+                  object asset_name="",
+                  object end_date=None,
+                  object notice_date=None,
+                  object first_traded=None,
+                  object exchange="",
+                  object ccy=""
+                  ):
+
+        self.pair = pair
```
jfkirk: What does pair represent? A string of the two currencies such as "EURUSD"?

If so, could it be represented as a property defined as the concatenation of the two currency symbols?

zipline/assets/_assets.pyx
```diff
@@ -298,6 +313,80 @@ cdef class Future(Asset):
...
+        self.major = major
+        self.minor = minor
```
jfkirk: Are these booleans representing whether the pair is a major/minor pair a-la http://hushtrade.com/beginners/major-minor-exotic-currency-pairs/?

zipline/assets/_assets.pyx
```diff
@@ -298,6 +313,80 @@ cdef class Future(Asset):
...
+        self.start_date = start_date
+        self.cvf = cvf
```
jfkirk: What does cvf represent?

zipline/assets/_assets.pyx
```diff
@@ -298,6 +313,80 @@ cdef class Future(Asset):
...
+                  object symbol="",
+                  object pair="",
+                  object major="",
+                  object minor="",
+                  object start_date=None,
+                  float cvf=1,
+                  object asset_name="",
+                  object end_date=None,
+                  object notice_date=None,
+                  object first_traded=None,
```
jfkirk: notice_date and first_traded are probably unnecessary for this class.

zipline/assets/_assets.pyx
```diff
@@ -298,6 +313,80 @@ cdef class Future(Asset):
         super_dict['contract_multiplier'] = self.contract_multiplier
         return super_dict
 
+"""
+TODO MB - rework this we need a Currency object and we need a FX pair
+
+ok we should use this a CurrencyPair, Currency could just be an enum?
```
jfkirk: If we're extending 'Asset', and use an Asset.quote_currency field to define the quote currency for all Asset classes. EG:
```
eurusd = Currency(sid=1, base_currency='EUR', quote_currency='USD')
eurusd.quote_currency == 'USD'
eurusd.base_currency == 'EUR'
eurusd.symbol == 'EURUSD'

gbpusd = Currency(sid=2, base_currency='GBP', quote_currency='USD')
gbpusd.quote_currency == 'USD'
gbpusd.base_currency == 'GBP'
gbpusd.symbol == 'GBPUSD'

gbpeur = Currency(sid=3, base_currency='GBP', quote_currency='EUR')
gbpeur.quote_currency == 'EUR'
gbpeur.base_currency == 'GBP'
gbpeur.symbol == 'GBPEUR'
```

Even though sids 2 and 3 represent the same currency, it is quoted in two different currencies. This would also allow you to define which currency you want to use to pay for a new position in the base currency.

This would also mean that the Currency itself doesn't need to be represented as an asset, just a different type of cash maintained by the performance periods.

zipline/assets/_assets.pyx
michaeljohnbennett: James, I'm refactoring this a little, had thought we could use an ENUM but yes currency would be clearer.

zipline/assets/_assets.pyx
michaeljohnbennett: price format is the multiplier that an FX contract may have somewhat like a future has in terms of the leverage multiplier. my main provider gives you this as a cvf=n where you use it as 10^n so cvf = 4 would mean a multiplier of 10,000

zipline/assets/_assets.pyx
michaeljohnbennett: status is time varying perhaps in a number of ways. a stock could be delisted and re-listed or the symbol could be re-used but our sid would always stay constant. this may have to be handled like the splits are in a seperate table and perhaps there would be multiple of these both on the exchange,index or other indice levels?

zipline/assets/_assets.pyx
michaeljohnbennett: Yes I was trying to think of a way to encapsulate this behaviour as we are dealing with a translation of currencies and not an FX pair you would trade so to speak. Having a base_currency on all assets would be ideal then really all you need is some kind of translation map between them. for some pairs that don't match 1:1 you may need to get creative to get a mapping between them but wouldn't be hard in a recursive loop around the model. Probably no more than one or two steps to get from TRY to SGD perhaps. (TRY->USD->SGD)

zipline/assets/_assets.pyx
michaeljohnbennett: I use pycharm/intellij and I was getting a constructor error in the file coming up, I never checked this on the compiler side as I just left them in, happy to remove them.

zipline/assets/_assets.pyx
michaeljohnbennett: symbol and pair are different, some providers give you specific codes for the CurrencyPair itself (QF4) but the pair is USDGBP. having it as the two elements and joining together would be fine, just thought keeping it denormalised and indexed on a single column lookup rather than a concatenation would be faster.

zipline/assets/_assets.pyx
michaeljohnbennett: no these would be base and quote respectively, have made the changes. exotics can be achieved through some recursive lookups of the CurrencyPairs in the AssetFinder DB

zipline/assets/_assets.pyx
michaeljohnbennett: its the volume factor/multiplier. I gave an example up above.

michaeljohnbennett:
I have some refactoring based on those major/minor field names and cleaned up the tests, will commit them tomorrow when I get a chance.

michaeljohnbennett:
cleaned up and merged code in recently, is this looking better at the base level @jfkirk ?

jfkirk:
Acknowledged - reviewing!

zipline/assets/_assets.pyx
jfkirk: Is price_format relevant to all Assets (Equities, Futures, and CurrencyPairs) or just CurrencyPairs?

zipline/assets/_assets.pyx
```diff
@@ -43,25 +40,35 @@ cdef class Asset:
 
     cdef readonly object exchange
 
+    cdef readonly object currency
```
jfkirk: What would you think about calling this field 'quote_currency' as it is shared across all Asset classes? I'm a little wary of simply 'currency' just because it may be confusing when working with the CurrencyPairs, but 'quote_currency' states the intended function across all Asset classes.

zipline/assets/_assets.pyx
jfkirk: The Asset and all of its fields are designed to be static and immutable during a given simulation. As it stands, those situations are handled by how the Assets are written to the db:
1. When a symbol is re-used, both Equities have the same symbol but can be differentiated by their start/end_dates. The symbol lookup methods take this in to account when performing a search.
2. When an Equity is de-listed, the end_date represents the final day of trading. If the same company is re-listed, it would be represented with a new Equity with a new start_date on the date of re-listing.
3. There is currently no handling in Zipline for splits. Within Quantopian, we handle this by split-adjusting the pricing data before the simulation.

Would your needs be met by what we have in place already? If so, can the 'status' field be cut?

zipline/assets/_assets.pyx
```diff
@@ -299,6 +314,75 @@ cdef class Future(Asset):
...
+"""
+TODO MB - rework this we need a Currency object and we need a FX pair
+
+ok we should use this a CurrencyPair, Currency could just be an enum?
+"""
+
+cdef class CurrencyPair(Asset):
+    cdef readonly object pair # 'USDGBP' for example
+    cdef readonly object base # 'USD'
+    cdef readonly object quote # 'GBP'
```
jfkirk: Is this redundant with the 'currency' or 'quote_currency' field of the Asset class?

zipline/assets/_assets.pyx
```diff
@@ -299,6 +314,75 @@ cdef class Future(Asset):
...
+            return 'Currency(%d)' % self.sid
+
+    def __repr__(self):
+        attrs = ('symbol', 'pair', 'base', 'quote',
+                 'start_date', 'multiplier')
+        tuples = ((attr, repr(getattr(self, attr, None)))
+                  for attr in attrs)
+        strings = ('%s=%s' % (t[0], t[1]) for t in tuples)
+        params = ', '.join(strings)
+        return 'Currency(%d, %s)' % (self.sid, params)
```
jfkirk: The __str__ and __repr__ results should be consistent with the class name
CurrencyPair vs Currency

jfkirk:
@michaeljohnbennett I've made some more comments - some are on the `outdated diff` comment chains. Let me know what you think.

In general, we're about to have some pretty big changes merging in as part of the lazy-zipline project that will improve performance significantly but, also, change how lots of the accounting system functions. We may want to work toward adding the CurrencyPairs in as Assets, but stub them out of the performance/position trackers until the big changes are merged in. Once lazy-zipline is in, though, we can integrate the CurrencyPairs fully. What do you think?

michaeljohnbennett:
Hi Captain Kirk ;-)

I will make it so, when is the merge coming in and I can hold off until then? I'm working on testing the code to actually use the currency in the performance sections so I can see the currency risk/fx portions of the trades work correctly.

Where do you think the best place to start working on that would be?

I have a simple algo to buy some shares of apple a while back and hold them, this takes into account a few splits (zipline should just handle that for me.... fingers crossed) and then the output each day in p/l can be expressed in the base currency of the account.

Likewise when ordering I need to convert from the base currency to the quote if they differ.

michaeljohnbennett:
James, happy new year, any more on the lazy-zipline project and when it
will integrated to complete this?

michaeljohnbennett:
@jfkirk @llllllllll any more on this feature? Do you still want me to start merging this or are we waiting for the lazy-mainline stuff to come in?

jfkirk:
@michaeljohnbennett Apologies for the delay - I'm a bit waylaid at the moment but will be able to provide more response later this week.

We've been moving the ground around underneath you a bit. I've recently merged a large refactor to how the assets.db is managed, and that will impact how the CurrencyPair would be defined for storage.

There may be an opportunity to merge all of the Asset definitions for the CurrencyPair without the lazy-mainline work, but we'd have to wait for lazy-mainline to start 'wiring it in' for trading.

hyao:
Thanks @michaeljohnbennett, @jfkirk for doing this much needed work :) I'm interested in using zipline/quantopian with forex as well. How's this coming along, thanks?

michaeljohnbennett:
Hyao, Hi,

I've not been doing much more on this as I have moved onto another project
right now, I was waiting as well until the code had settled somewhat as it
was too much of a moving target to complete any more work on until the
lazy-zipline branch was merged in and the model changes had calmed down
somewhat.

On Fri, 18 Mar 2016 at 16:49 hyao <notifications@github.com> wrote:

> Thanks @michaeljohnbennett <https://github.com/michaeljohnbennett>,
> @jfkirk <https://github.com/jfkirk> for doing this much needed work :)
> I'm interested in using zipline/quantopian with forex as well. How's this
> coming along, thanks?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/quantopian/zipline/pull/807#issuecomment-198218502>
>
-- 

regards

Michael Bennett

davidandreoletti:
@jfkirk @michaeljohnbennett Can the work on this PR be finally merged in zipline ?

jfkirk:
@davidandreoletti I am no longer with Quantopian or working on Zipline. @richafrank may be able to point this in the right direction? Hope this works out!

freddiev4:
I'm going to close this PR due to inactivity / a lot of conflicts with other files. If someone wants to continue work on this, feel free to pull this branch, fix conflicts, and open a new PR 🙂

PR #808, opened 2015-11-02
BUG: Corrects an index error in blaze loader.

llllllllll:
Fixes the case where a delta has an asof_date of the last requested
day and an index error would occur. This guards against this
specifically to make the delta be effective through the end of the
requested window.

Adds a test case for this behavior.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -399,7 +399,7 @@ class BlazeToPipelineTestCase(TestCase):
     @with_extra_sid
     def test_deltas(self, asset_info):
         expr = bz.Data(self.df, name='expr', dshape=self.dshape)
-        deltas = bz.Data(self.df.iloc[:-3], name='deltas', dshape=self.dshape)
+        deltas = bz.Data(self.df, name='deltas', dshape=self.dshape)
         deltas = bz.transform(
             deltas,
             value=deltas.value + 10,
@@ -411,6 +411,8 @@ class BlazeToPipelineTestCase(TestCase):
                                     [1.0, 2.0, 3.0]]),
             '2014-01-03': np.array([[11.0, 12.0, 13.0],
                                     [2.0, 3.0, 4.0]]),
+            '2014-01-04': np.array([[12.0, 13.0, 14.0],
+                                    [12.0, 13.0, 14.0]]),
         })
 
         nassets = len(asset_info)
@@ -422,7 +424,7 @@ class BlazeToPipelineTestCase(TestCase):
 
         with tmp_asset_finder(asset_info) as finder:
             expected_output = pd.DataFrame(
-                list(concatv([12] * nassets, [13] * nassets)),
+                list(concatv([12] * nassets, [13] * nassets, [14] * nassets)),
                 index=pd.MultiIndex.from_product((
                     sorted(expected_views.keys()),
                     finder.retrieve_all(asset_info.index),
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_blaze.py
```diff
@@ -430,6 +432,7 @@ def test_deltas(self, asset_info):
                 columns=('value',),
             )
             dates = self.dates
+            dates = dates.insert(len(dates), dates[-1] + timedelta(days=1))
```
ssanderson: This could also be `dates.append(dates[-1:] + timedelta(days=1))`.

zipline/pipeline/loaders/blaze.py
```diff
@@ -648,11 +648,31 @@ def overwrite_from_dates(asof, dense_dates, sparse_dates, asset_idx, value):
     -------
     overwrite : Float64Overwrite
         The overwrite that will apply the new value to the data.
+
+    Notes
+    -----
+    The overwrite is forward filled in the sense that it applies from the index
+    of the ``asof`` in the ``dense_dates`` to the index of the next date in the
```
ssanderson: Typo here?  What does "next date in the sparse dates in the dense dates" mean?

ssanderson: The docstring up above this says "based on the asof date of the delta, the dense_dates, and the dense_dates".

ssanderson: I also continue to have no intuition for the terms `dense` and `sparse` here.  Not a thing to change now though.

llllllllll: I think they are reasonable, what do you suggest instead?

zipline/pipeline/loaders/blaze.py
```diff
@@ -648,11 +648,31 @@ def overwrite_from_dates(asof, dense_dates, sparse_dates, asset_idx, value):
...
+    ``sparse_dates`` in the ``dense_dates``.
+
+    For example:
+    let ``asof = pd.Timestamp('2014-01-02')``,
+        ``dense_dates = pd.date_range('2014-01-01', '2014-01-05')``
+        ``sparse_dates = pd.to_datetime(['2014-01', '2014-02', '2014-04'])``
+
+    Then the overwrite will apply to indexes: 1, 2, 3, 4
```
ssanderson: indices

zipline/pipeline/loaders/blaze.py
```diff
@@ -648,11 +648,31 @@ def overwrite_from_dates(asof, dense_dates, sparse_dates, asset_idx, value):
...
     """
     first_row = dense_dates.searchsorted(asof)
-    last_row = dense_dates.searchsorted(
-        sparse_dates[sparse_dates.searchsorted(asof, 'right')],
-    ) - 1
+    next_idx = sparse_dates.searchsorted(asof, 'right')
+    if next_idx == len(sparse_dates):
```
ssanderson: I think this will do the wrong thing if the last date in `dense_dates` is after the last date in `sparse_dates`.

ssanderson: This is forward-filling "all dense dates that are between the as-of dates and the next sparse date after the asof_date".

ssanderson:
LGTM modulo request for clarification in docstring

tests/pipeline/test_blaze.py
llllllllll: This does not do the same this, `dates` is a datetimeindex

llllllllll:
merging on pass

zipline/pipeline/loaders/blaze.py
ssanderson: I would have called the `requested_dates` and `loaded_dates`.  There's no
guarantee that the data from a dataset will be more or less dense than the
requested data.

PR #809, opened 2015-11-02
BUG: Fixes futures commission with contract multipliers

jfkirk:
WIP

Transaction events have their 'price' fields set taking the commission in to account. This needs to be avoided with futures, or some other solution must be generated that allows for proper commission accounting for futures.

First commit adds a test that fails due to improper application of the commission model.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -1845,23 +1845,26 @@ class TestClosePosAlgo(TestCase):
                 format(i, actual_position, expected_positions[i]))
 
 
-class TestFutureFlip(TestCase):
-    def setUp(self):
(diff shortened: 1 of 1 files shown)


tests/test_algorithm.py
```diff
@@ -1877,6 +1880,22 @@ def test_flip_algo(self):
         expected_pnl = [0, 5, -10]
         self.check_algo_pnl(results, expected_pnl)
 
+    def test_algo_commission(self):
```
jfkirk: Note-to-self: rename to something futures-specific for clarity

jfkirk:
Scrapping in anticipation of lazy-zipline.

PR #810, opened 2015-11-03
Blaze loader forward fill fix

llllllllll:
The blaze loader was not looking back far enough to forward fill values correctly.
We now run 1 query per column to make this happen.

Further work is to make ffill an optional feature for performance so that we don't need to do this and we can just slice out some time range.

Even more work (but probably better in the long run) is to bake the 'most recently known value' into a dense table. This would allow us to make a single query for the most recently known value on a single day and then concat that with the sparse data where we can ffill in memory or not. This would probably require us writing some helper that takes a sparse dataset and constructs this new table. I would not want to block this fix on that work though.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -360,6 +360,43 @@ class BlazeToPipelineTestCase(TestCase):
         ))
         assert_frame_equal(result, expected, check_dtype=False)
 
+    def test_id_multiple_columns(self):
+        df = self.df.copy()
+        df['other'] = df.value + 1
+        fields = OrderedDict(self.dshape.measure.fields)
+        fields['other'] = fields['value']
+        expr = bz.Data(df, name='expr', dshape=var * Record(fields))
+        loader = BlazeLoader()
+        ds = from_blaze(
+            expr,
+            loader=loader,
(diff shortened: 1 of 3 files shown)


tests/pipeline/test_blaze.py
```diff
@@ -311,17 +312,137 @@ def test_id(self):
...
+            'other': (0, np.nan, 2, np.nan, 2, 3, 2, 3, np.nan),
+            'asof_date': dates,
+            'timestamp': dates,
+        })
+        fields = OrderedDict(self.dshape.measure.fields)
+        fields['other'] = fields['value']
+
+        with tmp_asset_finder() as finder:
+            expected = pd.DataFrame(
+                np.array([[2, 1],
```
ssanderson: Is there a way we can comment this to give me some intuition for why I should believe these outputs are correct?

Just looking at this test it's very hard for me to reason about.

ssanderson:
@llllllllll assessing the correctness of this is almost entirely about assessing the quality of the tests, and it's not obvious to me what each test is supposed to be checking or why I should believe the tests are correct.  

Can we annotate the inputs and expected outputs to give the reader some context as to why they should believe the test case is valid?

PR #811, opened 2015-11-03
BUG: Issue #801

elmq0022:
Added a line for setting the _analyze method in the correct location.  This should be fine.  Not sure if the test I added will work.  I'll see if Travis CI will run it.


elmq0022:
Going to wait and see if I can get it to run first.

PR #812, opened 2015-11-03
MAINT: Make load_adjusted_array return a dict.

ssanderson:
Rather than a list that's ordered the same as the received columns.
Most nontrivial loaders were constructing dicts internally and then
converting back to lists, only to have the engine convert **back again**
into a dict.  This cuts out the middleman, and prevents bugs due to
incorrect ordering of the output arrays.


ssanderson:
@richafrank I think this is a better solution to the nondeterministic failures we were getting upstream.  You can't return things in the wrong order if you're returning an unordered collection!

richafrank:
:+1: from me!

PR #815, opened 2015-11-04
ENH: Adds versions to Asset databases

jfkirk:
Adds a 'version' table to the asset database, and adds a version check to the AssetDBWriters and AssetFinders.


zipline/assets/asset_writer.py
```diff
@@ -165,6 +165,62 @@ def _generate_output_dataframe(data_subset, defaults):
...
+    Checks for a version value in the version table, and inserts version
+    info if none is present.
+
+    Parameters
+    ----------
+    version_table : sa.Table
+        The version table of the asset database
+    expected_version : int
+        The expected version of the asset database
+    write_version : bool, optional
```
ssanderson: When would you want this to be true?

jfkirk: When initializing the database -- the AssetDBWriter will write the version in to the table if it is not present.

zipline/assets/asset_writer.py
```diff
@@ -165,6 +165,62 @@ def _generate_output_dataframe(data_subset, defaults):
     return output
 
 
+# Define a version number for the database generated by these writers
+# Increment this version number any time a breaking change is made to the
+# schema and readers of the database
+ASSET_DB_VERSION = 1
```
StewartDouglas: Since we're calling the SQLite file `assets.db`, I suggest we call this variable `ASSETS_DB_VERSION`. Also, how about starting at 0.

zipline/assets/asset_writer.py
```diff
@@ -373,8 +429,24 @@ def init_db(self, engine):
                 primary_key=True),
             sa.Column('asset_type', sa.Text),
         )
+        self.version_table = sa.Table(
+            'version',
+            metadata,
+            sa.Column(
+                'version',
+                sa.Integer,
+                unique=True,
```
richafrank: I think we want this to be a primary key, instead of just unique?

zipline/assets/asset_writer.py
jfkirk: I went back and forth on that - the file is called assets.db, but everything else that references it is fairly uniformly named 'AssetDBBlah' so I stuck with the singular. I could go either way, though.

zipline/assets/asset_writer.py
jfkirk: Makes sense - thanks!

zipline/assets/asset_writer.py
```diff
@@ -165,6 +165,62 @@ def _generate_output_dataframe(data_subset, defaults):
...
+        version_table if no version information is present
+
+    Raises
+    ------
+    AssetsDbVersionError
+        If the version is in the table and not equal to ASSETS_DB_VERSION.
+    """
+
+    # Read the version out of the table
+    table_version = sa.select((version_table.c.version,)).scalar()
```
richafrank: Just to confirm, this table will have either a single row or no rows?

richafrank: Does `table_version` mean the version of the table?

jfkirk: The table will have a single row or no rows. Do you know of any way to enforce that on the table?

It means the version in the table - I'll rename it to 'version_in_table' or something more clear.

zipline/assets/asset_writer.py
StewartDouglas: I have a slight preference for `ASSETS_DB_VERSION`, but am happy to go either way. If you stick with `ASSET` you should update the following line: https://github.com/quantopian/zipline/pull/815/files#diff-33cdc2260291626609681cc61297818aR192

zipline/assets/asset_writer.py
ssanderson: @jfkirk you can do this with a `CHECK` constraint and a primary key:

```
sqlite> create table test (id INTEGER PRIMARY KEY AUTOINCREMENT CHECK (id <= 1), version int);
sqlite> insert into test (version) VALUES (1);
sqlite> insert into test (version) VALUES (2);
Error: CHECK constraint failed: test
```

jfkirk: @ssanderson Perfect, thank you!

jfkirk:
@ssanderson I've added a primary key ID to the version table and implemented a constraint as Scott suggested

@richafrank I updated the 'table_version' variable to 'version_from_table' for clarity

@StewartDouglas The initial version is now 0

zipline/assets/asset_writer.py
jfkirk: I'm going to separate-out checking and writing responsibilities to different methods

jfkirk:
@ssanderson I split up version checking and version insertion since they are thematically distinct

zipline/assets/asset_writer.py
```diff
@@ -373,8 +457,16 @@ def init_db(self, engine):
...
         )
+        self.version_table = _version_table_schema(metadata)
         # Create the SQL tables if they do not already exist.
         metadata.create_all(checkfirst=True)
+
+        # Check if the version is mismatched or, if it is not present, add it
+        version_found = check_version_info(self.version_table,
+                                           ASSET_DB_VERSION)
+        if not version_found:
+            write_version_info(self.version_table, ASSET_DB_VERSION)
```
richafrank: Could it be the case that we've got an old assets.db with no embedded version info, we then call `init_db`, which doesn't create anything because we use `checkfirst=True`, and then we insert the latest db version?  Seems like we should only write the version info on first creation of the all the tables, not even just the version table (since `create_all` will create any that are missing)?

zipline/assets/asset_writer.py
```diff
@@ -268,6 +344,21 @@ def _write_futures(self, futures, bind):
...
+        """
+        Checks if any tables are present in the current assets database.
+
+        Returns
+        -------
+        bool
+            True if any tables are present, otherwise False.
+        """
+        conn = engine.connect()
+        for table_name in table_names:
```
StewartDouglas: I think lines 357-360 are the equivalent of doing the following:
```python
return engine.dialect.has_table(conn, table_names[0])
```
This is because if *any* of the tables have been created, it implies that they all have been. This is because `init_db` is called within an SQL transaction.

jfkirk: In the case of a schema change, table_names[0] may not exist anymore but we'd want this method to return True. The only way I could think of to guarantee this is by checking all table names. It will return after the first table_name, anyway, which would mean that in the nominal case it is identical in execution to
```
return engine.dialect.has_table(conn, table_names[0])

Comment (StewartDouglas) [inline zipline/assets/asset_writer.py]: Could it be the case that *all* table names may change, and then this method would return false, even though tables are present? If we don't care about the names of the tables then we could use SQLAlchemy's Inspector to check whether any tables are present:
```python
from sqlalchemy.engine import reflection
...
insp = reflection.Inspector.from_engine(self.eng)
return True if insp.get_table_names() else False
```
http://docs.sqlalchemy.org/en/rel_0_7/core/schema.html?highlight=inspector#sqlalchemy.engine.reflection.Inspector

jfkirk: To cause this method to fail to meet its contract, all table names would have to be changed including the version table, which I see as a catastrophic change likely meaning this method would be gone as well. By my logic there, though, we could change this method to only check for the version_info table. The reason I don't do that, though, is I still want this method to return True if the other non-version_info tables are present but no version_info is present because that is our backwards-compatible case.

docs/source/whatsnew/0.8.4.txt
```diff
@@ -38,7 +38,8 @@ None
 Maintenance and Refactorings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-None
+* Asset databases now contain version information to ensure compatibility
+  with current Zipline version.
```
jfkirk: @llllllllll I added a whatsnew in its own commit - does that follow the PR dance you mentioned?

PR #817, opened 2015-11-05
Speedup retrieve all

ssanderson:
Speed up `AssetFinder.retrieve_all` by doing one query instead of N.

- Removes the separate caches for equities and futures in favor of a single asset cache.
- We now cache the results of failed lookups.  This required unfortunate hacks in the `TradingAlgorithm` code that resolves assets from DataFrames.
- Reimplements single-asset lookups as a special case of multiple-asset lookups.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -97,7 +97,6 @@ from zipline.utils.events import (
     TimeRuleFactory,
 )
 from zipline.utils.factory import create_simulation_parameters
-from zipline.utils.functional import is_a
 from zipline.utils.math_utils import tolerant_equals
 from zipline.utils.preprocess import preprocess
 
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -21,8 +21,10 @@ from pandas.tseries.tools import normalize_date
 import numpy as np
 
 from datetime import datetime
-
 from itertools import groupby, chain, repeat
+from numbers import Integral
+from operator import attrgetter
+
 from six.moves import filter
 from six import (
     exec_,
@@ -30,7 +32,6 @@ from six import (
     itervalues,
     string_types,
 )
-from operator import attrgetter
 
 
 from zipline.errors import (
@@ -96,6 +97,7 @@ from zipline.utils.events import (
     TimeRuleFactory,
 )
 from zipline.utils.factory import create_simulation_parameters
+from zipline.utils.functional import is_a
 from zipline.utils.math_utils import tolerant_equals
 from zipline.utils.preprocess import preprocess
 
@@ -608,8 +610,7 @@ class TradingAlgorithm(object):
             if isinstance(identifier, Asset):
                 asset = self.asset_finder.retrieve_asset(sid=identifier.sid,
                                                          default_none=True)
-
-            elif hasattr(identifier, '__int__'):
+            elif isinstance(identifier, Integral):
                 asset = self.asset_finder.retrieve_asset(sid=identifier,
                                                          default_none=True)
             if asset is None:
@@ -617,6 +618,12 @@ class TradingAlgorithm(object):
 
         self.trading_environment.write_data(
             equities_identifiers=identifiers_to_build)
+
+        # We need to clear out any cache misses that were stored while trying
+        # to do lookups.  The real fix for this problem is to not construct an
+        # AssetFinder until we `run()` when we actually have all the data we
+        # need to so.
+        self.asset_finder._reset_caches()
 
         return self.asset_finder.map_identifier_index_to_sids(
             identifiers, as_of_date,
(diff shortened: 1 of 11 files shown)


ssanderson:
Timings for ~30000 assets on my machine:

**Before:**
```
# Empty Cache
In [3]: %timeit -r1 -n1 f.retrieve_all(sids)
1 loops, best of 1: 29.4 s per loop

# Full Cache
In [5]: %timeit f.retrieve_all(sids)
100 loops, best of 3: 9.13 ms per loop

# Single Asset
In [3]: %timeit f.retrieve_asset(24)
The slowest run took 49.35 times longer than the fastest. This could mean that an intermediate result is being cached
1000000 loops, best of 3: 546 ns per loop
```

**After:**
```
# Empty Cache
In [5]: %timeit -n5 f.retrieve_all(sids); f._reset_caches()
5 loops, best of 3: 1.34 s per loop

# Full Cache
In [7]: %timeit f.retrieve_all(sids)
100 loops, best of 3: 5.96 ms per loop

# Single Asset
In [4]: %timeit f.retrieve_asset(24)
The slowest run took 42.91 times longer than the fastest. This could mean that an intermediate result is being cached
1000000 loops, best of 3: 278 ns per loop
```

**Conclusions:**
- This is roughly 30x faster for batch lookups on a large number of assets when caches are empty.
- This is roughly 2x faster for batch lookups when caches are full.
- This is roughly 2x **slower** for single-asset lookups.  Those lookups go from 270 ns to 540 ns on my machine.  I think this is okay because that performance should only matter appreciably if we're looking up lots of assets in a loop, in which case we should just use the faster batch mode anyway.  We can add more special-case handling for single-asset cache hits if we need to though.

cc @jbredeche on perf notes.

zipline/algorithm.py
```diff
@@ -604,8 +605,7 @@ def _write_and_map_id_index_to_sids(self, identifiers, as_of_date):
             if isinstance(identifier, Asset):
                 asset = self.asset_finder.retrieve_asset(sid=identifier.sid,
                                                          default_none=True)
-
-            elif hasattr(identifier, '__int__'):
+            elif isinstance(identifier, Integral):
```
richafrank: What's the impact of this change?

zipline/assets/assets.py
```diff
@@ -283,21 +396,21 @@ def lookup_symbol(self, symbol, as_of_date, fuzzy=False):
                 if len(fuzzy_sids) == 1:
                     return self._retrieve_equity(fuzzy_sids[0]['sid'])
 
-            sids = sa.select((equities_cols.sid,)).where(
-                (equities_cols.company_symbol == company_symbol) &
-                (equities_cols.share_class_symbol == share_class_symbol)
-            ).execute().fetchall()
+            sids = list(imap(
+                itemgetter('sid'),
```
richafrank: Given that this is all about performance, any idea how this compares to a list comprehension?
```python
sids = [row['sid'] for row in sa.select(...).execute().fetchall()]
```

(Same question below)

richafrank:
Nice conclusions!  Is the single-asset lookup case for cache-miss or cache-hit?

zipline/algorithm.py
ssanderson: We use `Integral` to define `AssetConvertible` in assets.py, so this seems more in line with the definition there.  

One notable difference is that this will reject `float`s as inputs, which seems more correct to me.  If you pass `3.5` as your index, the existing behavior would truncate to `3` and then convert it to an asset, which seems like unfortunate behavior.

ssanderson:
@richafrank the single-asset case is about the same 2x slowdown for both hits and misses.

zipline/assets/assets.py
ssanderson: @richafrank I'd expect the difference here to be pretty negligible since this is only loading ~5-10 assets at most.  I didn't touch much of the by-symbol lookup logic, but there are definitely a lot of improvements that could be made there.

zipline/algorithm.py
richafrank: That makes sense to me.

zipline/assets/assets.py
richafrank: Oh interesting.  Why is this case at most 5-10 assets?

ssanderson: This is in `lookup_symbol`.   We're only resolving sids for assets that share a symbol.  I think the most-shared symbol in our upstream database has something like 12 assets with that symbol?

richafrank: Ah, got it. Thanks!

ssanderson:
@mtydykov can you take a look at the commits from a436426a5a018605895d6d073ba709a5ee82cdef to 9881c303413534e8859a85d40c27aa5bffaf4a1a?  The rest of this PR has already been reviewed by @richafrank, but I made a couple changes after rebasing against your latest changes.

mtydykov:
LGTM

PR #818, opened 2015-11-05
Addresss Issue #801 and Add Test

elmq0022:
Give this a try and let me know if it resolves the issue.


richafrank:
Thanks @elmq0022 !  

Looks like your branch has a conflict with master.  I think a new test was added at the same location as your new test, so we should resolve that by taking both.  Want to rebase and we'll see if the tests pass?  Sorry for the hassle.

elmq0022:
That's fine.  I will make another attempt this weekend.

Thanks for your patience add well.

richafrank:
Closed in favor of #819

PR #820, opened 2015-11-05
MAINT: Remove unused 'asset_type' metadata entry.

ssanderson:
@jfkirk indicated to me that this field was unused by the writer, which appears to be correct.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -514,8 +514,7 @@ class AssetFinderTestCase(TestCase):
         self.assertEqual(missing[1], 'REAL_BUT_IN_THE_FUTURE')
 
     def test_insert_metadata(self):
-        data = {0: {'asset_type': 'equity',
-                    'start_date': '2014-01-01',
+        data = {0: {'start_date': '2014-01-01',
                     'end_date': '2015-01-01',
                     'symbol': "PLAY",
                     'foo_data': "FOO"}}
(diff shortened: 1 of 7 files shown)


tests/test_assets.py
```diff
@@ -802,6 +795,9 @@ def test_sids(self):
         self.assertTrue(2 in sids)
         self.assertTrue(3 in sids)
 
+    def test_retrieve_multiple(self):
```
jfkirk: empty method?

tests/test_perf_tracking.py
```diff
@@ -2189,12 +2189,10 @@ class TestPositionTracker(unittest.TestCase):
     def setUpClass(cls):
         cls.env = TradingEnvironment()
 
-        equities_metadata = {1: {'asset_type': 'equity'},
-                             2: {'asset_type': 'equity'}}
-        futures_metadata = {3: {'asset_type': 'future',
-                                'contract_multiplier': 1000},
-                            4: {'asset_type': 'future',
-                                'contract_multiplier': 1000}}
+        equities_metadata = {1: {},
```
jfkirk: This could be provided as identifiers to avoid the unnecessary empty dicts:
```
equities_identifiers = [1, 2]
cls.env.write_data(equities_identifiers=equities_identifiers,
                              futures_data=futures_metadata)
```

tests/test_assets.py
ssanderson: This is cruft from cherry-picking.  Will fix.

jfkirk:
:+1:

PR #821, opened 2015-11-05
ENH: Add CME code lookup table.

ssanderson:


StewartDouglas:
LGTM :+1:

zipline/assets/futures.py
```diff
@@ -180,3 +180,84 @@ def offset(self, time_delta):
...
+MONTH_TO_CME_CODE = dict(zip(range(1, 13), 'FGHJKMNQUVXZ'))
+
+
+def cme_code_to_month(code):
+    """
+    Convert a CME month code to a month index.
+
+    The month codes are as follows:
+
+    'F' -> 1  (January)
```
StewartDouglas: Suggest changing the doc string here to read `1 -> 'F' (January)`,  etc., so that it illustrates what the method does.

zipline/assets/futures.py
```diff
@@ -180,3 +180,84 @@ def offset(self, time_delta):
...
+    return CME_CODE_TO_MONTH[code]
+
+
+def month_to_cme_code(month):
+    """
+    Convert a month to a CME code.
+
+    The month codes are as follows:
+
+    'F' -> 1  (January)
```
StewartDouglas: Suggest changing the doc string here to read `1 -> 'F' (January)`,  etc., so that it illustrates what the method does.

PR #822, opened 2015-11-06
ENH: Raise on attempt to set a lazyval.

ssanderson:
ping @richafrank who noticed this odd behavior.

### DIFF
--- a/zipline/utils/memoize.py
+++ b/zipline/utils/memoize.py
@@ -50,7 +50,7 @@ class lazyval(object):
             self._cache[instance] = val = self._get(instance)
             return val
 
-    def __set__(self, instance, owner):
+    def __set__(self, instance, value):
         raise AttributeError("Can't set read-only attribute.")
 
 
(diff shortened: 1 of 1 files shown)


ssanderson:
@llllllllll does this seem reasonable?

llllllllll:
I believe this is the correct behavior. It also matches the `val` part of the name. 
```scala
scala> lazy val a = 10
a: Int = <lazy>

scala> a = 20
<console>:11: error: reassignment to val
       a = 20
         ^
```

zipline/utils/memoize.py
```diff
@@ -44,6 +50,9 @@ def __get__(self, instance, owner):
             self._cache[instance] = val = self._get(instance)
             return val
 
+    def __set__(self, instance, owner):
```
richafrank: I realize it's unused here, but should the last parameter be named `value`?

llllllllll: yes

richafrank:
Thanks @ssanderson !

ssanderson:
Updated the name on the `__set__` arg.

llllllllll:
merge on pass

PR #824, opened 2015-11-09
ENH: Truncate negative dividend adjustment ratios to 0.05

StewartDouglas:
Occasionally assets pay dividends greater than the last traded
value of the asset. This can occur when investors are uncertain
about the nature of the dividend payments, which means they are
not able to correctly price the asset in advance of the dividend.
After these payments the assets fall dramatically in value. Use
a 5% adjustment factor as it is consistent with the case of VXUP
and VXDN: http://finance.yahoo.com/q?s=VXUP.

### DIFF
--- a/zipline/data/us_equity_pricing.py
+++ b/zipline/data/us_equity_pricing.py
@@ -687,7 +687,19 @@ class SQLiteAdjustmentWriter(object):
                     sid, prev_close_date, 'close')
                 if prev_close != 0.0:
                     ratio = 1.0 - amount / prev_close
-                    ratios[i] = ratio
+                    # Occasionally assets pay dividends greater than the last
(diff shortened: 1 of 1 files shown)


ssanderson:
@StewartDouglas @KarenRubin what's the status of this?

StewartDouglas:
@ssanderson I think this is ready to merge. Can you please review? Should I update the 0.8.4 whatsnew?

ssanderson:
Looked at this today comparing to some of our downstream data, and I don't think this is the correct fix.  Most incidences of negative dividends that we've seen are actually liquidation events that are being (incorrectly?) reported as dividends.  This will create noise in Pipeline algos and will mask real data errors.  Closing for now.

Issue #823, opened 2015-11-08
zipline installs numpy 1.9.2


joequant:
The requirements file in zipline should be upgraded to allow installs with numpy 1.10

llllllllll:
The requirements file specifies the tested and supported versions for zipline. We try to use exact version numbers instead of lower bounds to minimize the combinations of libraries that we support. With that said, I will see if the tests pass and performance stays the same with the newest version of numpy.

richafrank:
What do you think about pinning the versions in the requirements file, but not in the `install_requires`?  https://packaging.python.org/en/latest/requirements/#install-requires-vs-requirements-files

I imagine we could generate the `install_requires` minimums from the requirements, with a facility to override the output, if we found, for example, that numpy 1.10 actually broke zipline.

joequant:
The basic issue that I ran into is....

1) I install numpy 1.10
2) I do a pip install of zipline
3) zipline pulls in numpy 1.9.2 from requirements
4) at that point the other modules which were built on 1.10 fail because of
binary incompatbilities with 1.9.2

So I don't care much what version of numpy is used by zipline.  The more
serious problem is that installing zipline with a fixed numpy can cause
other modules to break.

richafrank:
@joequant The HEAD of zipline master now allows for newer versions of its requirements.  How does that work for you?

joequant:
That should be good enough.

PR #825, opened 2015-11-10
BLD: Don't pin install requirements to exact versions

richafrank:
Instead, use those versions as lower bounds, so zipline
will work with packages that require other versions.

### DIFF
--- a/docs/source/whatsnew/0.8.4.txt
+++ b/docs/source/whatsnew/0.8.4.txt
@@ -43,7 +43,7 @@ None
 Build
 ~~~~~
 
-None
+* Makes zipline install requirements more flexible (:issue:`825`).
 
 Documentation
 ~~~~~~~~~~~~~
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,10 @@ install:
   - source activate testenv
   - conda install --yes -c https://conda.binstar.org/Quantopian numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy==$SCIPY_VERSION matplotlib Cython patsy statsmodels tornado pyparsing xlrd mock pytz requests six dateutil ta-lib logbook
   - pip install --upgrade pip coverage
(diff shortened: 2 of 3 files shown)


richafrank:
@llllllllll How does this look as a fix for #823 ?

llllllllll:
This looks good, but it looks like our travis file will be pulling the newest versions of everything (like flake8 ^^).

Maybe we could do an `extras_requires` like `strict` so people can `pip install zipline[strict]` which will force the versions to what we want tested. We can then use this for travis

richafrank:
`pip install zipline[strict]` makes sense to me.

richafrank:
I think we should call `read_requirements('etc/requirements_dev.txt', convert_eq_to_lte=False)`, since dev requirements should always be pegged, and I imagine that will solve this issue without `[strict]`.  Do you think `[strict]` would be otherwise helpful?

richafrank:
Ah, would be nice to pin them for travis...

richafrank:
@llllllllll I can't seem to override the install_requires with 'strict' in the extras_require.  Any suggestions?

llllllllll:
Can you not just make the extras_requires list for `strict` the more strict version and then those will be downgraded when you use that target?

richafrank:
That's what I had hoped, but based on the travis build log, seems like not.

llllllllll:
hmm, let's play with this a bit. I think that we want this to be easy since we will want to do this on travis. I wonder if there is any existing work for this.

docs/source/whatsnew/0.8.4.txt
```diff
@@ -43,7 +43,7 @@ None
 Build
 ~~~~~
 
-None
+Make zipline install requirements more flexible (:issue:`825`).
```
llllllllll: Can you put a `*` so this appears as a bulleted list?

:+1: for the entry

llllllllll:
one comment on the whatsnew, then :shipit:

PR #826, opened 2015-11-10
Siddata caching affects len

llllllllll:
When we cache the two methods on the instance, we need to update the initial_len to account for these untracked attributes in the dict.

Long term goal of removing `__len__` and `__contains__` from this object, it is not a container.

### DIFF
--- a/zipline/protocol.py
+++ b/zipline/protocol.py
@@ -12,7 +12,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 from copy import copy
 
 from six import iteritems, iterkeys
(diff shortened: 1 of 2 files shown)


zipline/protocol.py
```diff
@@ -439,8 +438,20 @@ def minute_get_bars(days):
             self._get_bars = minute_get_bars
             self._get_max_bars = minute_get_max_bars
 
+        # NOTE: This silently adds these two entries to the `__dict__`
+        # without affecting the `__len__` of the object. This is important
+        # because we use the `len` of the `SIDData` object to see if we have
+        # data for this asset.
+        self._initial_len += 2
```
richafrank: Fun stuff.  LGTM, given the current framework.

llllllllll:
rebased, merging on pass

PR #828, opened 2015-11-10
context manager for scheduled functions

llllllllll:


zipline/utils/context_tricks.py
```diff
@@ -0,0 +1,60 @@
...
+    entering another block
+    inside another block
+    exiting another block
+    """
+    def __init__(self, pre=None, post=None):
+        pre = pre if pre is not None else _nop
+        post = post if post is not None else _nop
+
+        @contextmanager
+        def _callback_manager_context(*args, **kwargs):
```
richafrank: Instead of closing over `pre` and `post`, what do you think about factoring out a class?  I think it'd be a bit more readable, and we'd get a performance increase over `@contextmanager`.

zipline/utils/events.py
```diff
@@ -169,14 +169,35 @@ def _build_time(time, kwargs):
         return datetime.time(**kwargs)
 
 
-class EventManager(object):
+@object.__new__
+class _nop_context(object):
```
richafrank: How does this compare to `ExitStack`?

llllllllll: I wasn't sure if we used contextlib32 in zipline. This is mainly for clarity though, we never want the features of exitstack.

zipline/utils/context_tricks.py
llllllllll: sounds good.

zipline/utils/events.py
llllllllll: okay, so we are using exit stack, I am stull not sure that is better here.

richafrank: That's fair.  nop is certainly clearer.

llllllllll:
Merging on pass

PR #829, opened 2015-11-11
Versioneer

llllllllll:
Most of these files are autogenerated from `versioneer` (https://github.com/warner/python-versioneer).

example version
```python
In [1]: import zipline

In [2]: zipline.__version__
Out[2]: 'v0.8.3+15.g88c2548'
```


twiecki:
Oh that's how that's done -- cool stuff.

PR #830, opened 2015-11-11
Equity caching

mtydykov:
Adding an extension to AssetFinder that caches all equities and overrides lookup_symbol to look up equities in the cache rather than in the database. This is meant to improve performance for anything that needs to lookup symbols for a large quantity of data.

I tested performance by doing a databazaar historical load of a 10k-line csv file. The load took about 2 minutes without caching, and about 3 seconds with caching.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -457,6 +457,31 @@ class AssetFinderTestCase(TestCase):
                 self.assertEqual(result.symbol, 'EXISTING')
                 self.assertEqual(result.sid, i)
 
+    def test_lookup_symbol_from_multiple_valid(self):
+        df = pd.DataFrame.from_records(
+            [
+                {
+                    'sid': 1,
+                    'symbol': 'multiple',
+                    'start_date': pd.Timestamp('2010-01-01'),
+                    'end_date': pd.Timestamp('2013-01-01'),
+                    'exchange': 'NYSE'
+                },
+                {
+                    'sid': 2,
+                    'symbol': 'multiple',
+                    'start_date': pd.Timestamp('2012-01-01'),
+                    'end_date': pd.Timestamp('2014-01-01'),
+                    'exchange': 'NYSE'
+                }
+            ]
+        )
+        self.env.write_data(equities_df=df)
+        finder = self.asset_finder_type(self.env.engine)
+        result = finder.lookup_symbol('MULTIPLE', pd.Timestamp('2012-05-05'))
+        self.assertEqual(result.symbol, 'MULTIPLE')
+        self.assertEqual(result.sid, 2)
+
     def test_lookup_generic(self):
         """
         Ensure that lookup_generic works with various permutations of inputs.
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -16,12 +16,11 @@
 """
 Tests for the zipline.assets package
 """
-
-import sys
-from unittest import TestCase
-
+from contextlib import contextmanager
 from datetime import datetime, timedelta
 import pickle
+import sys
+from unittest import TestCase
 import uuid
 import warnings
 
@@ -29,14 +28,20 @@ import pandas as pd
 from pandas.tseries.tools import normalize_date
 from pandas.util.testing import assert_frame_equal
 
-from nose_parameterized import parameterized
 from numpy import full
 
-from zipline.assets import Asset, Equity, Future, AssetFinder
+from zipline.assets import (
+    Asset,
+    Equity,
+    Future,
+    AssetFinder,
+    AssetFinderCachedEquities
+)
+
 from zipline.assets.futures import (
     cme_code_to_month,
     FutureChain,
-    month_to_cme_code,
+    month_to_cme_code
 )
 from zipline.errors import (
     SymbolNotFound,
@@ -48,12 +53,15 @@ from zipline.finance.trading import TradingEnvironment, noop_load
 from zipline.utils.test_utils import (
     all_subindices,
     make_rotating_asset_info,
+    tmp_assets_db
 )
 
 
-def build_lookup_generic_cases():
+@contextmanager
+def build_lookup_generic_cases(asset_finder_type):
     """
-    Generate test cases for AssetFinder test_lookup_generic.
+    Generate test cases for the type of asset finder specific by
+    asset_finder_type for test_lookup_generic.
     """
 
     unique_start = pd.Timestamp('2013-01-01', tz='UTC')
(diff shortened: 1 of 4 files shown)


mtydykov:
@kglowinski

zipline/assets/assets.py
```diff
@@ -678,3 +678,148 @@ class AssetConvertible(with_metaclass(ABCMeta)):
...
+
+    def load_fuzzy_symbol_hashed_equities(self):
+        """
+        Populates map of fuzzy symbol to a list of equities that have that
+        fuzzy symbol.
+        """
+        for equity in self.hashed_equities:
+            fuzzy_symbol = equity['fuzzy_symbol']
+            if fuzzy_symbol not in self.fuzzy_symbol_hashed_equities:
+                self.fuzzy_symbol_hashed_equities[fuzzy_symbol] = []
```
jfkirk: Could this be simplified by using a setdefault?
```
self.fuzz_symbol_hashed_equities.setdefault(fuzzy_symbol, []).append(asset)
```

zipline/assets/assets.py
```diff
@@ -678,3 +678,148 @@ class AssetConvertible(with_metaclass(ABCMeta)):
...
+        Populates map of (company symbol, share class symbol) to a list of
+         equities that have that combination of (company symbol, share class
+         symbol).
+        """
+        for equity in self.hashed_equities:
+            company_symbol = equity['company_symbol']
+            share_class_symbol = equity['share_class_symbol']
+            if (company_symbol, share_class_symbol) not in \
+                    self.company_share_class_hashed_equities:
+                self.company_share_class_hashed_equities[(
```
jfkirk: Ditto on using a setdefault()

zipline/assets/assets.py
```diff
@@ -678,3 +678,148 @@ class AssetConvertible(with_metaclass(ABCMeta)):
...
+                    company_symbol,
+                    share_class_symbol
+                )] = []
+            asset = self.convert_row_to_equity(equity)
```
jfkirk: If this method is called in both caching functions, then are we building each Asset twice? A lot of the logic between caching by fuzzy and caching by company/sc symbol are shared, and they could probably be condensed to one method that iterates through all equities, building each once, and inserting it in to both dicts.

mtydykov: You're right! Thanks for catching that.

jfkirk:
I know nothing would make @llllllllll 's day like a whatsnew entry

zipline/assets/assets.py
```diff
@@ -678,3 +678,133 @@ class AssetConvertible(with_metaclass(ABCMeta)):
...
+    def convert_row_to_equity(self, equity):
+        """
+        Converts a SQLAlchemy equity row to an Equity object.
+        """
+        data = dict(equity.items())
+        _convert_asset_timestamp_fields(data)
+        asset = Equity(**data)
+        return asset
+
+    def lookup_symbol(self, symbol, as_of_date, fuzzy=False):
```
jfkirk: This method seems to have a great deal of overlap with the lookup_symbol that it is overriding. Would it be possible to factor out the actual selections from both AssetFinders, and have the main AssetFinder's lookup_symbol call those factored-out selectors, so then any changes to the lookup_symbol logic need only happen at one level?

mtydykov: Sure, I think that should be possible.

mtydykov: Ok, I've refactored the selection logic in lookup_symbol so that AssetFinder gets candidates from the database and AssetFinderCachedEquities gets candidates from its dictionaries. Let me know if this kind of refactoring is what you had in mind.

mtydykov:
Done

zipline/assets/assets.py
```diff
@@ -678,3 +728,113 @@ class AssetConvertible(with_metaclass(ABCMeta)):
...
+    def convert_row_to_equity(self, equity):
+        """
+        Converts a SQLAlchemy equity row to an Equity object.
+        """
+        data = dict(equity.items())
+        _convert_asset_timestamp_fields(data)
+        asset = Equity(**data)
+        return asset
+
+    def get_fuzzy_candidates(self, fuzzy_symbol):
```
jfkirk: Since these methods are internal API, you can prepend them with _ as in 
```
def _get_fuzzy_candidates(self, fuzzy_symbol):
```
so that it is clear, when working with an AssetFinder, that these methods are not meant to be accessed directly.

tests/test_assets.py
```diff
@@ -450,16 +457,41 @@ def test_lookup_symbol(self):
                 self.assertEqual(result.symbol, 'EXISTING')
                 self.assertEqual(result.sid, i)
 
-    @parameterized.expand(
-        build_lookup_generic_cases()
-    )
-    def test_lookup_generic(self, finder, symbols, reference_date, expected):
+    def test_lookup_symbol_from_multiple_valid(self):
```
mtydykov: I added this test because of a mistake I caught when implementing the cached functionality for lookup_symbol. I had initially forgotten to sort multiple valid equities found for a symbol by start date/end date. The tests had still passed, even though a different sid was being returned than in the non-caching version.

jfkirk:
Once Jenkins passes, LGTM :+1:

tests/test_assets.py
ssanderson: I think the example data being requested here should just be an error on data input.  If we have multiple assets with the same symbol and overlapping alive periods, we can't meaningfully resolve those symbols.  Do we have examples of this in our real data?

zipline/assets/assets.py
```diff
@@ -678,3 +730,113 @@ class AssetConvertible(with_metaclass(ABCMeta)):
 
 class NotAssetConvertible(ValueError):
     pass
+
+
+class AssetFinderCachedEquities(AssetFinder):
```
richafrank: Do you think we should name this with something more specific to Equities, since the behavior it adds to the base AssetFinder doesn't apply to all Assets?

tests/test_assets.py
```diff
@@ -800,6 +832,13 @@ def test_sids(self):
         self.assertTrue(3 in sids)
 
 
+class AssetFinderCachedEquitiesTestCase(AssetFinderTestCase):
+
+    def setUp(self):
+        self.env = TradingEnvironment(load=noop_load)
+        self.asset_finder_type = AssetFinderCachedEquities
```
richafrank: :+1:

tests/test_assets.py
mtydykov: Yes, there are several examples of this - HK, AGN, and ES, and others. These are symbols which multiple companies share and which have overlapping date ranges. The current logic in lookup_symbol dealt with such cases by sorting the multiple valid equities by start/end date and returning the latest, so I was just trying to mimic that logic in my cached version. 

Should I change anything?

zipline/assets/assets.py
jfkirk: @richafrank Are you thinking something like "EquityFinder"? I was trying to think of other names, but nothing really jumped out at me as this object does still support all of the non-Equity functions of the AssetFinder.

tests/test_assets.py
jfkirk: @ssanderson Echoing Maya, yep, the case of overlapping symbols is fairly common.

ssanderson: @jfkirk how can we treat these as anything other than an error?  There's no meaningful definition of a correct lookup here.

zipline/assets/assets.py
richafrank: Oh you know what - I totally missed the word Equities in the name as it currently stands.  I think it's a good name as is.

tests/test_assets.py
jfkirk: Glad you asked! We hammered out the logic we wanted back in Feb, but I don't remember all of the rationale. Here is a comment where it was outlined:

```
It seems that the best resolution is to use the following logic to map symbol S at time T:
UPDATE: I added the number of cases in each category

>1. If one SID exists for symbol S at time T, return that SID (happy path)

>2. If no SID exists for symbol S at time T, return the SID whose end_date is highest-but-not-over T, or 'none' if no such SID exists (happy path)

>3. If multiple SIDs exist for symbol S at time T, return the candidate SID whose start_date is highest. (200 cases)

>4. If mutliple SIDs exist for symbol S at time T, the best candidate SIDs share the highest start_date, return the SID with the highest end_date. (34 cases)

>5. If multiple SIDs exist for symbol S at time T, and the best candidate SIDs share the highest start_date and end_date, return None or throw an error. (no known cases)
```

ssanderson: These rules were determined based on the particular idiosyncrasies of an upstream vendor.  It's not appropriate for us to be encoding those assumptions into the business logic of an open source project.  It should be the responsibility of the code that's authoring these files to ensure that any overlaps are resolved before writing data to the database.

jfkirk: That concern makes sense to me, but I don't think we should block up Maya's PR on it. I'm opening an issue now.

jfkirk: New issue: https://github.com/quantopian/zipline/issues/837

mtydykov: Ok, let me know if/when is a good time for me to merge this. Thank you!

PR #831, opened 2015-11-11
BLD: put 3.4 and 3.5 in the build matrix

llllllllll:


llllllllll:
The issue I see is that we are using conda for numpy and some other packages but we haven't updated our conda channel with 3.4 builds for anyhting

PR #834, opened 2015-11-11
BUG: update the definition of information ratio

llllllllll:
@twiecki can you look this over?

In trying to update the answer key, I am not sure we have the daily benchmark returns in the spreadsheet to compute the information ratio

### DIFF
--- a/tests/risk/test_risk_cumulative.py
+++ b/tests/risk/test_risk_cumulative.py
@@ -109,7 +109,8 @@ class TestRisk(unittest.TestCase):
             np.testing.assert_almost_equal(
                 value,
                 self.cumulative_metrics_06.information[dt_loc],
-                err_msg="Mismatch at %s" % (dt,))
+                err_msg="Mismatch at %s" % (dt,),
+            )
 
     def test_alpha_06(self):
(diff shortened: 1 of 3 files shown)


twiecki:
LGTM and matches the pyfolio implementation https://github.com/quantopian/pyfolio/blob/master/pyfolio/timeseries.py#L1339.

llllllllll:
@twiecki two questions:

1. The daily benchmark returns are not in the csv for the answer key so I cannot implement this in the spreadsheet. How should we generate a new one?

1. How would you feel about using a shared implementation for the risk metric calculations. I noticed that there was recently an issue with the sortino ratio that was updated in pyfolio but not zipline. I feel like we can either make one project a dependency of another or make a new project of raw risk and performance calculations.

twiecki:
@llllllllll 

1. I didn't work on that (@eddie did) so my thoughts are as good as yours. We could just add that to the csv I guess.
2. I agree. I think pyfolio should be a dependency for zipline. Just not sure whether we should port the testing logic that's in zipline over to pyfolio. Might be the right thing to do.

llllllllll:
closing in favor of: #1322

PR #838, opened 2015-11-13
TST: Limit the timer report to just the 15 longest tests

richafrank:
The default of all the tests is pretty spammy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,6 @@ before_script:
   - pip freeze | sort
   - "flake8 zipline tests"
 script:
-  - nosetests --with-timer --exclude=^test_examples --with-coverage --cover-package=zipline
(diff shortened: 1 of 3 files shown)


ssanderson:
@llllllllll @richafrank I added this change directly to `setup.cfg` in my speedup-retrieve-all branch so that it also becomes the default for devs.

llllllllll:
Looks like you have `timer-filter=warning`. Also, this is added to the setup.cfg here too

ssanderson:
Oh, actually, I made a slightly different change in my branch: I made it only show tests that were in the `warning` threshold for speed.  Anythoughts on the difference?

ssanderson:
@llllllllll yeah, I thought I had done top-n, but I actually did the threshold.  Any preference between the two?  I'm rebasing a conflict here now.

richafrank:
How much output do we get with just `warning`?

ssanderson:
@richafrank running now to see.

Issue #839, opened 2015-11-13
Late data from the Fed can cause slowdowns.


ssanderson:
We try to download historical treasury rates from the Fed to use in Sharpe calculations, among other things.  The logic for that download tries to be fairly conservative about when we think data should be available for a given day, but there are cases where data isn't available when we think it should be, which causes tests, among other things, to run very slowly because they download the same stale treasury data over and over.

For example, on today (November 13th, 2015), we are expecting that data should be available for November 12th, 2015.  The data from the fed, however, only reaches November 11th, which means every test in the suite is trying to redownload the old data.  Our best guess is that this is somehow related to Veteran's day on the 11th, which isn't a market holiday, but **is** a federal holiday.

Two thoughts here:

1. Should we provide a switch (maybe via an environment variable?) to disable remote data downloading?
2. Should we add logic to prevent repeated download attempts?  For example, we could store the last time we tried to download and always wait at least an hour before trying again.

ssanderson:
Partially addressed by https://github.com/quantopian/zipline/pull/841.  @richafrank is there anything else you think we need to be doing here?

richafrank:
:+1: on #841.  A switch would be a nice general enhancement, but I think you've done enough for this issue of slow downloads, especially for tests.

PR #841, opened 2015-11-13
Remember downloads

ssanderson:


ssanderson:
Partial fix for https://github.com/quantopian/zipline/issues/839.

I wanted to try this out now while the Fed is still delayed.

zipline/data/loader.py
```diff
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import print_function
```
llllllllll: are we using print somewhere?

ssanderson: https://github.com/quantopian/zipline/pull/841/files#diff-0ec9e6870d90568b4a83e0daa3a317b1L290

This happened to work because `print(stock)` is still valid in Py2.  But from git blame it's clear that the intention was for that to be using the print function, which I suspect was either accidentally omitted or removed at a later point.

llllllllll: kk, just wanted to make sure it was still needed

ssanderson:
@llllllllll good to merge?

llllllllll:
yes

PR #846, opened 2015-11-16
TST: tests removing of expired data and removes ffill in DataPanelSource

warren-oneill:
Currently it seems as though the data for expired futures is not removed from the info that gets sent to ```handle_data```. Over longer backtests this can result in the ```data``` variable becoming quite large and gives in a large reduction in performance esp when ```history``` is used.
cc @jfkirk @dhexus

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -59,6 +59,7 @@ from zipline.test_algorithms import (
     TestTargetAlgorithm,
     TestTargetPercentAlgorithm,
     TestTargetValueAlgorithm,
+    TestRemoveDataAlgo,
     SetLongOnlyAlgorithm,
     SetAssetDateBoundsAlgorithm,
     SetMaxPositionSizeAlgorithm,
@@ -1965,3 +1966,42 @@ class TestTradingAlgorithm(TestCase):
                                 analyze=analyze)
         results = algo.run(self.panel)
         self.assertIs(results, self.perf_ref)
+
+
+class TestRemoveData(TestCase):
+    """
+    tests if futures data is removed after expiry
+    """
+    def setUp(self):
+        dt = pd.Timestamp('2015-01-01', tz='UTC')
+        metadata = {0: {'symbol': 'X',
+                        'expiration_date': dt + timedelta(days=5),
+                        'end_date': dt + timedelta(days=5)},
+                    1: {'symbol': 'Y',
+                        'expiration_date': dt + timedelta(days=7),
+                        'end_date': dt + timedelta(days=7)}}
+
+        env = TradingEnvironment()
+        env.write_data(futures_data=metadata)
+
+        index_x = pd.date_range(dt, periods=5)
+        data_x = pd.DataFrame([[1, 100], [2, 100], [3, 100], [4, 100],
+                               [5, 100]],
+                              index=index_x, columns=['price', 'volume'])
+        index_y = index_x.shift(2)
+        data_y = pd.DataFrame([[6, 100], [7, 100], [8, 100], [9, 100],
+                               [10, 100]],
+                              index=index_y, columns=['price', 'volume'])
+
+        pan = pd.Panel({0: data_x, 1: data_y})
+        self.source = DataPanelSource(pan)
+        self.algo = TestRemoveDataAlgo(env=env)
+
+    def test_remove_data(self):
+        self.algo.run(self.source)
+
+        expected_length = [1, 2, 2, 2, 2, 1]
+        # initially only data for X should be sent and on the last day only
+        # data for Y should be sent since X is expired
+        for i, length in enumerate(self.algo.data):
+            self.assertEqual(expected_length[i], length, i)
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -1973,22 +1973,24 @@ class TestRemoveData(TestCase):
     tests if futures data is removed after expiry
     """
     def setUp(self):
-        dt = pd.Timestamp('2015-01-01', tz='UTC')
-        metadata = {0: {'symbol': 'X',
-                        'expiration_date': dt + timedelta(days=5),
-                        'end_date': dt + timedelta(days=5)},
-                    1: {'symbol': 'Y',
-                        'expiration_date': dt + timedelta(days=7),
-                        'end_date': dt + timedelta(days=7)}}
-
+        dt = pd.Timestamp('2015-01-02', tz='UTC')
         env = TradingEnvironment()
+        ix = env.trading_days.get_loc(dt)
+
+        metadata = {0: {'symbol': 'X',
+                        'expiration_date': env.trading_days[ix + 5],
+                        'end_date': env.trading_days[ix + 6]},
+                    1: {'symbol': 'Y',
+                        'expiration_date': env.trading_days[ix + 7],
+                        'end_date': env.trading_days[ix + 8]}}
+
         env.write_data(futures_data=metadata)
 
-        index_x = pd.date_range(dt, periods=5)
+        index_x = env.trading_days[ix:ix + 5]
         data_x = pd.DataFrame([[1, 100], [2, 100], [3, 100], [4, 100],
                                [5, 100]],
                               index=index_x, columns=['price', 'volume'])
-        index_y = index_x.shift(2)
+        index_y = env.trading_days[ix:ix + 5].shift(2)
         data_y = pd.DataFrame([[6, 100], [7, 100], [8, 100], [9, 100],
                                [10, 100]],
                               index=index_y, columns=['price', 'volume'])
@@ -2000,8 +2002,7 @@ class TestRemoveData(TestCase):
     def test_remove_data(self):
         self.algo.run(self.source)
 
-        expected_length = [1, 2, 2, 2, 2, 1]
+        expected_lengths = [1, 1, 2, 2, 2, 2, 1]
         # initially only data for X should be sent and on the last day only
         # data for Y should be sent since X is expired
-        for i, length in enumerate(self.algo.data):
-            self.assertEqual(expected_length[i], length, i)
+        np.testing.assert_array_equal(self.algo.data, expected_lengths)
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -123,12 +123,7 @@ class TestDataFrameSource(TestCase):
         self.assertEqual(5, event.sid)
         event = next(source)
         self.assertEqual(4, event.sid)
-        try:
-            x = False
-            event = next(source)
-        except StopIteration:
-            x = True
-        self.assertTrue(x)
+        self.assertRaises(StopIteration, next, source)
 
 
 class TestRandomWalkSource(TestCase):
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -123,9 +123,12 @@ class TestDataFrameSource(TestCase):
         self.assertEqual(5, event.sid)
         event = next(source)
         self.assertEqual(4, event.sid)
-        event = next(source)
-        self.assertEqual(5, event.sid)
-        self.assertFalse(np.isnan(event.price))
+        try:
+            x = False
+            event = next(source)
+        except StopIteration:
+            x = True
+        self.assertTrue(x)
 
 
 class TestRandomWalkSource(TestCase):
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -266,9 +266,9 @@ class AssetFinder(object):
                 self.equities.c.share_class_symbol ==
                 share_class_symbol,
                 self.equities.c.start_date <= ad_value),
-            ).order_by(
-                self.equities.c.end_date.desc(),
-            ).execute().fetchall()
+        ).order_by(
+            self.equities.c.end_date.desc(),
+        ).execute().fetchall()
         return candidates
 
     def _get_best_candidate(self, candidates):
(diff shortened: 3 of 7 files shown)


tests/test_algorithm.py
```diff
@@ -1965,3 +1966,41 @@ def analyze(context, perf):
...
+
+class TestRemoveData(TestCase):
+    """
+    tests if futures data is removed after expiry
+    """
+    def setUp(self):
+        dt = pd.Timestamp('2015-01-01', tz='UTC')
+        metadata = {0: {'asset_type': 'future', 'symbol': 'X',
+                        'expiration_date': dt + timedelta(days=4)},
+                    1: {'asset_type': 'future', 'symbol': 'Y',
```
jfkirk: I don't think 'asset_type' is necessary as the type is now implied from the argument name to write_data

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,41 @@ def analyze(context, perf):
...
+        pan = pd.Panel({0: data_x, 1: data_y})
+        self.source = DataPanelSource(pan)
+        self.algo = TestRemoveDataAlgo(env=env)
+
+    def test_remove_data(self):
+        self.algo.run(self.source)
+
+        expected_length = [1, 2, 2, 2, 2, 1]
+        for i, dt in enumerate(self.algo.data):
+            self.assertEqual(expected_length[i], dt, i)
```
jfkirk: For my clarity, this will fail if the algo data object still contains future X after it has expired?

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,41 @@ def analyze(context, perf):
...
+
+        pan = pd.Panel({0: data_x, 1: data_y})
+        self.source = DataPanelSource(pan)
+        self.algo = TestRemoveDataAlgo(env=env)
+
+    def test_remove_data(self):
+        self.algo.run(self.source)
+
+        expected_length = [1, 2, 2, 2, 2, 1]
+        for i, dt in enumerate(self.algo.data):
```
jfkirk: 'dt' as a variable name is a bit confusing just because we often use 'dt' to refer to datetimes/timestamps.

warren-oneill: yep, sorry. earlier it was a timestamp but forgot to change it.

tests/test_algorithm.py
warren-oneill: yes

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,41 @@ def analyze(context, perf):
...
+                              index=index_y, columns=['price', 'volume'])
+
+        pan = pd.Panel({0: data_x, 1: data_y})
+        self.source = DataPanelSource(pan)
+        self.algo = TestRemoveDataAlgo(env=env)
+
+    def test_remove_data(self):
+        self.algo.run(self.source)
+
+        expected_length = [1, 2, 2, 2, 2, 1]
```
jfkirk: Could you throw a comment above this explaining what the list signifies and what would cause the assertion to fail?

jfkirk:
After speaking with @llllllllll it seems that the ffill is desired ('stale' data should be preserved in the BarData object). However, it also makes sense that expired/ended Assets should be removed from the BarData object, meaning that this test is testing for desired behavior.

Assigning @llllllllll

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,43 @@ def analyze(context, perf):
...
+        # initially only data for X should be sent and on the last day only
+        # data for Y should be sent since X is expired
```
warren-oneill: @jfkirk that ok?

jfkirk: Yep, thanks!

warren-oneill:
@jfkirk @llllllllll ok, I only removed ffill so I could test the expired data. Could you briefly explain why ffill is desirable? It could paint a false picture about the development of the market.

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,43 @@ def analyze(context, perf):
...
+        dt = pd.Timestamp('2015-01-01', tz='UTC')
+        metadata = {0: {'symbol': 'X',
+                        'expiration_date': dt + timedelta(days=4)},
+                    1: {'symbol': 'Y',
+                        'expiration_date': dt + timedelta(days=7)}}
+
+        env = TradingEnvironment()
+        env.write_data(futures_data=metadata)
+
+        index_x = [dt, dt + timedelta(days=1), dt + timedelta(days=2),
```
llllllllll: This can be written as `pd.date_range(dt, periods=5)`

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,43 @@ def analyze(context, perf):
...
+                   dt + timedelta(days=3), dt + timedelta(days=4)]
+        data_x = pd.DataFrame([[1, 100], [2, 100], [3, 100], [4, 100],
+                               [5, 100]],
+                              index=index_x, columns=['price', 'volume'])
+        index_y = [dt + timedelta(days=2), dt + timedelta(days=3),
```
llllllllll: if you make `index_x` a `DatetimeIndex` then this can `index_x.shift(2)`

llllllllll:
We ffill the data; however, each `BarData` object has a `dt` attribute which is the date that it was added. This allows people to view price continuously while still having the ability to filter our stale data. I think we want to change the check for removal to be: 'securities whose expiration date is less than today'

jfkirk:
The original design for the BarData object was that it would provide the last-best trade data for all sids. This means that BarData is a convenient way of getting trade information, but there may be many degrees of 'staleness' that matter to different TradingAlgorithms, and it is the responsibility of the TradingAlgorithm to check for 'stale' data if it matters to that TradingAlgorithm.

It makes sense to me that the stale SIDData should be removed from the BarData after the end_date has passed.

warren-oneill:
@llllllllll dt is the date of the event and not the date it was added. If you ffill then dt=algo.datetime always and there is no way of telling if the data is stale or if it is a new trade. Also, ffill means that it is possible to send futures data events after expiry which makes no sense.
However if you leave out ffill then you will still get the info in the data in handle_data and it will have the correct stale dt and not an artificial ffilled dt.

warren-oneill:
Test are passing in travis for me. Can someone rerun the CI tests?

zipline/gens/tradesimulation.py
```diff
@@ -100,6 +101,16 @@ def transform(self, stream_in):
                 self.simulation_dt = date
                 self.on_dt_changed(date)
 
+                # removing expired futures
+                for sid in self.current_data.keys():
+                    try:
+                        if self.env.asset_finder.retrieve_asset(sid).end_date \
+                                < self.simulation_dt:
+                            self.current_data.__delitem__(sid)
+                    except (AttributeError, TypeError, ValueError,
```
warren-oneill: I have too many exceptions here. Def needs improvement.

llllllllll: the `__delitem__` call here should be: `del self.current_data[sid]`

Why can so many exceptions be raised from that?

warren-oneill: AttributeError - when the asset is a Future and has no end_date
SidNotFound - fails in a test where the sid for an asset is changed
TypeError - when end_date=None
ValueError - not sure. gotta recheck that one.

warren-oneill: ValueError - in AlgorithmGeneratorTestCase the string 'custom_data' gets passed as an sid to retrieve_asset instead of an int.

ssanderson: Two additional notes here:
1. `self.current_data.__delitem_(key)` is better written as `del self.current_data[key]`.
2. This will fail on Python3 because `keys()` is a view on the dictionary, and you're not allowed to mutate a dictionary in place while iterating over a view:
```
Python 3.4.3 (default, Oct 14 2015, 20:28:29)
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> x = {1:1, 2: 2, 3: 3}
>>> for k in x.keys():
...     del x[k]
...
RuntimeError: dictionary changed size during iteration
```
Compare to:
```
>>> x = {1:1, 2: 2, 3: 3}
>>> for k in x.keys():
...     del x[k]
...
>>> x
{}
```

ssanderson: Are the keys of `current_data` just sids (meaning, integers rather than asset objects)?  I was under the impression that they should always be `Asset` objects, in which case we shouldn't need to do a `retrieve_asset` call here.  I'm nervous about the performance implications of looping over all of `data` here.  This is basically the inner loop of the simulator, so small changes here can have major performance implications.

warren-oneill: they are sids. I agree but I not sure how else the expired data can be removed.

ssanderson: > they are sids.

:(

>  I agree but I'm not sure how else the expired data can be removed.

One possibility would be to pre-compute the sids that will expire on each backtest day, and then only loop over those on each day.  A naive way to do this would be to build such a lookup table using all assets in our asset finder, and a more refined approach would be to only build the lookup table for assets that will be emitted by our DataSource, with the caveat that the set of emittable sids may not be knowable in advance for dynamic sources.

FayolChang:
@ssanderson I encounter a problem that in 
~~~python
        order(symbol('AAPL'), 10)
        record(AAPL=data[symbol('AAPL')].price)
~~~~

in method **record**,  the key of data are supposed to be a integer, however, an Equity instance is passed.

warren-oneill:
@ssanderson @jfkirk @llllllllll how should we move forward with this? Is the performance the major blocker atm? Should I try @ssanderson suggestion? We could also make sure we know at all times the next asset to expire. Then we should only need to query the metadata once every dt instead every tick.

warren-oneill:
Maybe we could use a function similar to ```lookup_future_chain``` to get the next expiring sids @jfkirk ?

warren-oneill:
or we could have a function that returns all the sids that have an expiry between the current dt and the last dt.

warren-oneill:
ok I went for the second option. let me know what you guys think.

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,44 @@ def analyze(context, perf):
...
+        self.algo = TestRemoveDataAlgo(env=env)
+
+    def test_remove_data(self):
+        self.algo.run(self.source)
+
+        expected_length = [1, 1, 2, 2, 2, 1, 1]
+        # initially only data for X should be sent and on the last day only
+        # data for Y should be sent since X is expired
+        for i, length in enumerate(self.algo.data):
+            self.assertEqual(expected_length[i], length, i)
```
llllllllll: We could just do `np.testing.assert_array_equal(self.algo.data, expected_lengths)`

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,44 @@ def analyze(context, perf):
...
+                              index=index_y, columns=['price', 'volume'])
+
+        pan = pd.Panel({0: data_x, 1: data_y})
+        self.source = DataPanelSource(pan)
+        self.algo = TestRemoveDataAlgo(env=env)
+
+    def test_remove_data(self):
+        self.algo.run(self.source)
+
+        expected_length = [1, 1, 2, 2, 2, 1, 1]
```
llllllllll: could we make `length` plural to show that it is a collection.

zipline/gens/tradesimulation.py
```diff
@@ -96,10 +97,20 @@ def transform(self, stream_in):
...
             for date, snapshot in stream_in:
-
+                expired_sids = self.env.asset_finder.lookup_expired_futures(
+                    start=self.previous_dt, end=date)
+                self.previous_dt = date
                 self.simulation_dt = date
                 self.on_dt_changed(date)
 
+                # removing expired futures
+                if expired_sids:
```
llllllllll: do we need this if? `for a in []` skips.

zipline/gens/tradesimulation.py
```diff
@@ -96,10 +97,20 @@ def transform(self, stream_in):
             self._call_before_trading_start(mkt_open)
 
             for date, snapshot in stream_in:
-
+                expired_sids = self.env.asset_finder.lookup_expired_futures(
+                    start=self.previous_dt, end=date)
+                self.previous_dt = date
```
llllllllll: when does `previous_dt` differ from `simulation_dt`?

tests/test_sources.py
```diff
@@ -123,9 +123,12 @@ def test_nan_filter_panel(self):
...
         self.assertEqual(4, event.sid)
-        event = next(source)
-        self.assertEqual(5, event.sid)
-        self.assertFalse(np.isnan(event.price))
+        try:
+            x = False
+            event = next(source)
+        except StopIteration:
+            x = True
+        self.assertTrue(x)
```
llllllllll:
```python
self.assertRaises(StopIteration, next, source)
```

tests/test_algorithm.py
```diff
@@ -1965,3 +1966,44 @@ def analyze(context, perf):
...
+    """
+    tests if futures data is removed after expiry
+    """
+    def setUp(self):
+        dt = pd.Timestamp('2015-01-02', tz='UTC')
+        env = TradingEnvironment()
+        ix = env.trading_days.get_loc(dt)
+
+        metadata = {0: {'symbol': 'X',
+                        'expiration_date': env.trading_days[ix+5],
```
llllllllll: small style comment: could you put spaces around the binary operators to make it easier to read.

ssanderson: (This will also trigger flake8.)

zipline/assets/assets.py
```diff
@@ -492,6 +492,74 @@ def lookup_future_chain(self, root_symbol, as_of_date):
...
+        sids = list(map(
+            itemgetter('sid'),
+            sa.select((fc_cols.sid,)).where(
+
+                    # Filter to contracts that are still valid. If both
+                    # exist, use the one that comes first in time (i.e.
+                    # the lower value). If either notice_date or
+                    # expiration_date is NaT, use the other. If both are
+                    # NaT, the contract cannot be included in any chain.
+                    sa.case(
```
llllllllll: These two case statements are both trying to find the same date, we could do something more concise with:
```python
nd = sa.func.nullif(fc_cols.notice_date, pd.tslib.iNaT)
ed = sa.func.nullif(fc_cols.expiration_date, pd.tslib.iNaT)
date = sa.func.coalesce(sa.func.min(nd, ed), ed, nd)

whereclause = (date >= start) & (date <= end)
```

We can use this same `nd` and `ed` for the order by:

```
sa.func.coalesce(ed, nd)
```

zipline/gens/tradesimulation.py
warren-oneill: it differs every time we move to a new day or new minute.

warren-oneill:
@llllllllll hey any chance of getting this accepted soon? Anything left that I need to add?

llllllllll:
Hey, sorry about letting this sit around so long. I am rerunning the tests to make sure there are no conflicts with master. Merging on pass

warren-oneill:
nice, tnx! :smile:

PR #849, opened 2015-11-16
TST: Fixes test_order_methods_for_future

jfkirk:
Fixes test_order_methods_for_future by creating a separate env where 0 is a Future.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -645,10 +645,12 @@ class TestTransformAlgorithm(TestCase):
 
     @classmethod
     def setUpClass(cls):
-        futures_metadata = {3: {'contract_multiplier': 10}}
         cls.env = TradingEnvironment()
-        cls.env.write_data(equities_identifiers=[0, 1, 133],
-                           futures_data=futures_metadata)
+        cls.env.write_data(equities_identifiers=[0, 1, 133])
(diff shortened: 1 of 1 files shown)


jfkirk:
@llllllllll Mind giving this a peek?

jfkirk:
Ping @llllllllll

llllllllll:
Why was this passing before?

jfkirk:
The test algo itself checks the type of the Asset it is ordering. Because the wrong Asset was coming back from the env, the test algo was following its Equity logic in stead of its Future logic.

An alternate solution would be to dupe the test algos, one for Equities, and one for Futures, but that would be a lot more verbose.

llllllllll:
I would probably prefer that we make the tests more verbose because this is passing on both master and this branch. The test should not work with equities

jfkirk:
How about I add an assertion to both the equity and future tests that ensures that the desired sid is the desired asset type? Then, if something happens to the environments that changes the asset type, it would be caught by the test itself.

llllllllll:
That sounds good.

jfkirk:
I added assertions to both tests and verified that these assertions would have failed without the new env.

llllllllll:
looks good, merge on pass

PR #858, opened 2015-11-18
Q 2.0

jbredeche:
Lots of changes here.

Big picture:
- Introduces the notion of an explicit clock that iterates over all the periods of a backtest (minutes or days), emitting the time and actions (such as "day started", "minute ended", etc).  In contrast, today Zipline "ticks forward" on a bar when it receives price data for that bar.
- Introduces the notion of the Data Portal, which provides the value of an OHLCV or Fetcher field for a specific asset at a specific time (or window of times).
	- When asked for a specific time, the resulting value is completely unadjusted.
	- When asked for a window, the resulting values are fully adjusted (splits, mergers, and dividends) as of the end of that window.
- Removes the concept of a universe.  Pipeline and fundamentals still let you define a list of assets, but now you can just store them on your context object and use them as you see fit.  You can now look up any field for any asset for any date (as long as the asset exists on that date).

Smaller things:
- `history` has been completely rewritten.  Specifically, you have to pass a list of assets as the first parameter to history calls.
- Fetcher has been open-sourced and moved from Quantopian's internal codebase to Zipline.
- Batch transform has been removed.
- "Simple" transforms (`mavg`, `returns`, `stddev`, and `vwap`) have been removed in favor of doing it via history.
- Dollar volume universe has been removed.
- `set_benchmark` now works in Zipline.
- A fair amount of effort was spent fine tuning other parts of the codebase - EventManager, benchmark returns calculations, etc.

There are several to-dos.  I wanted to get this PR started while I worked on them.
- [ ] Clean up the commit history on this branch.
- [ ] Remove hardcoded references to 1/2/2002.
- [x] Move fetcher tests into Zipline.


ssanderson:
@jbredeche I think this doesn't contain some new files.  Notably, I don't think this has the definition of the `DataPortal` class?

jdricklefs:
> Notably, I don't think this has the definition of the DataPortal class?

It's there - https://github.com/quantopian/zipline/blob/lazy-mainline/zipline/data/data_portal.py - but the diff is too large for GitHub to display, I believe. (Hence the "No Changes" on the last several files).

(https://help.github.com/articles/what-are-the-limits-for-viewing-content-and-diffs-in-my-repository/)

richafrank:
@ssanderson I think it's a github quirk - I can click View to see the new file fully: https://github.com/quantopian/zipline/pull/858/files#diff-fa9042487ffee7c174e36f17b2997f10

ssanderson:
@jdricklefs @richafrank TIL...

tests/finance/test_slippage.py
```diff
@@ -24,75 +24,145 @@
...
+                "low": np.array([2.85, 2.85, 2.85, 2.85, 2.85]) * 1000,
+                "close": np.array([3.0, 3.5, 4.0, 3.5, 3.0]) * 1000,
+                "volume": [2000, 2000, 2000, 2000, 2000],
+                "minute": cls.minutes
+            })
+        }
 
-        self.assertEquals(len(orders_txns), 1)
-        _, txn = orders_txns[0]
+        MinuteBarWriterFromDataFrames().write(cls.tempdir.path, assets)
```
richafrank: Could be convenient and probably much faster for tests to have a DataPortal that reads in-memory data, versus a BcolzDataPortal that requires bcolz files on disk.  If the bcolz reader only needs a file stream, then maybe a StreamIO could work as a compatibility layer.

tests/pipeline/test_blaze.py
```diff
@@ -292,6 +292,7 @@ def test_complex_expr(self):
                 loader=self.garbage_loader,
             )
 
+    @skip
```
richafrank: FIXME

zipline/algorithm.py
```diff
@@ -457,30 +433,40 @@ def _create_generator(self, sim_params, source_filter=None):
...
+            self.perf_tracker = PerformanceTracker(
+                sim_params=self.sim_params,
+                env=self.trading_environment,
+                data_portal=self.data_portal)
+            # Set the dt initially to the period start by forcing it to change
+            self.on_dt_changed(self.sim_params.period_start)
+
             # HACK: When running with the `run` method, we set perf_tracker to
             # None so that it will be overwritten here.
             self.perf_tracker = PerformanceTracker(
```
richafrank: Why do we set the perf tracker twice?

What are the cases that the `if self.perf_tracker is None` condition is testing? Looks like `__init__` sets it to `None` always?  If we call `run` multiple times, do we want to re-use the perf tracker?

zipline/algorithm.py
```diff
@@ -457,30 +433,40 @@ def _create_generator(self, sim_params, source_filter=None):
...
+                data_portal=self.data_portal
             )
 
-        self.portfolio_needs_update = True
-        self.account_needs_update = True
-        self.performance_needs_update = True
-
-        self.data_gen = self._create_data_generator(source_filter, sim_params)
+        if not self.initialized:
+            self.initialize(*self.initialize_args, **self.initialize_kwargs)
```
richafrank: Based on the call to `self.on_dt_changed` above, is `initialize` now called at period_start, as opposed to None/no time?

zipline/algorithm.py
```diff
@@ -457,30 +433,40 @@ def _create_generator(self, sim_params, source_filter=None):
...
-        self.trading_client = AlgorithmSimulator(self, sim_params)
+        self.ensure_clock()
 
-        transact_method = transact_partial(self.slippage, self.commission)
-        self.set_transact(transact_method)
+        self.trading_client = AlgorithmSimulator(
+            self,
+            sim_params,
+            self.data_portal,
+            self.clock,
```
richafrank: It looks like this is the only usage of `self.clock`.  Why is it an instance attribute of the algorithm, as opposed to the simulator, if the algorithm only creates and stores it for the purpose of passing to the simulator?

zipline/algorithm.py
```diff
@@ -490,101 +476,28 @@ def get_generator(self):
...
-               * index must be DatetimeIndex
-               * array contents should be price info.
+            source : DataPortal
 
         :Returns:
             daily_stats : pandas.DataFrame
               Daily performance metrics such as returns, alpha etc.
 
         """
+        if self.data_portal is None:
```
richafrank: Should this read `if data_portal is not None:`?  I'd find it confusing that my argument is ignored if one is already set.  Or is it something that should only be set once, and if I want multiple runs, I have to instantiate multiple algos?

zipline/algorithm.py
```diff
@@ -490,101 +476,28 @@ def get_generator(self):
...
-        # Create history containers
-        if self.history_specs:
-            self.history_container = self.history_container_class(
-                self.history_specs,
-                self.current_universe(),
-                self.sim_params.first_open,
-                self.sim_params.data_frequency,
-                self.trading_environment,
-            )
+        self.gen = self.get_generator()
```
richafrank: If `self.gen` is only used in this method, can we make it a local instead of an instance attribute?

zipline/algorithm.py
```diff
@@ -705,6 +579,43 @@ def get_environment(self, field='platform'):
...
+            symbol,
+            mask,
+            symbol_column,
+            data_frequency=self.data_frequency,
+            special_params_checker=special_params_checker,
+            **kwargs
+        )
+
+        # ingest this into dataportal
+        self.data_portal.handle_extra_source(csv_data_source.df)
```
richafrank: How does this update to the data portal state intersect with multiple calls to `run`?  Will the extra source persist?

zipline/algorithm.py
```diff
@@ -1212,64 +1107,19 @@ def cancel_order(self, order_param):
...
     @require_initialized(HistoryInInitialize())
-    def history(self, bar_count, frequency, field, ffill=True):
-        history_spec = self.get_history_spec(
+    def history(self, sids, bar_count, frequency, field, ffill=True):
+        if self.data_portal is None:
+            raise Exception("no data portal!")
+
+        return self.data_portal.get_history_window(
+            sids,
+            self.get_datetime(),
```
richafrank: Since this goal of this branch is performance, I recommend using `self.datetime` directly, unless we want the extra validation here.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+from zipline.assets import Asset
+
+from zipline.utils import tradingcalendar
+from zipline.errors import (
+    NoTradeDataAvailableTooEarly,
+    NoTradeDataAvailableTooLate
+)
+
+# FIXME anything to do with 2002-01-02 probably belongs in qexec, right/
+FIRST_TRADING_DAY = pd.Timestamp("2002-01-02 00:00:00", tz='UTC')
```
richafrank: Found the hardcoded value you mentioned.  Just commenting so we can easily find it again!

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+        """
+        Extra sources always have a sid column.
+
+        We expand the given data (by forward filling) to the full range of
+        the simulation dates, so that lookup is fast during simulation.
+        """
+        if source_df is None:
+            return
+
+        self._fetcher_df = source_df
```
richafrank: Probably want to rename this attribute.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+
+    def handle_extra_source(self, source_df):
+        """
+        Extra sources always have a sid column.
+
+        We expand the given data (by forward filling) to the full range of
+        the simulation dates, so that lookup is fast during simulation.
+        """
+        if source_df is None:
+            return
```
richafrank: Should this just be a TypeError/ValueError?  Or is there a case for passing in None and wanting it to be ignored?

zipline/data/data_portal.py
richafrank: Some names below too.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+        self._sim_params = sim_params
+        if self._sim_params is not None:
+            self._data_frequency = self._sim_params.data_frequency
+
+        self._sid_path_func = sid_path_func
+
+        self.DAILY_PRICE_ADJUSTMENT_FACTOR = 0.001
+        self.MINUTE_PRICE_ADJUSTMENT_FACTOR = 0.001
+
+    def handle_extra_source(self, source_df):
```
richafrank: I think calling this multiple times will cause `self._fetcher_df` to get reset, but other state, like the `augmented_sources_map` will accrue more data.  Probably we don't want them out of sync?

zipline/data/data_portal.py
richafrank: To clarify, it'd be great if we were consistent here between extra sources and fetcher sources.  Seems like we're generalizing fetcher here, but the old names remain.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+                self._asset_end_dates[identifier] = latest_date
+
+            for col_name in df.columns.difference(['sid']):
+                if col_name not in self._augmented_sources_map:
+                    self._augmented_sources_map[col_name] = {}
+
+                self._augmented_sources_map[col_name][identifier] = df
+
+    def _open_daily_file(self):
+        if self._daily_equities_data is None:
```
richafrank: Not sure yet whether these methods fall in inner loops or not, but I notice that on every call they check state that isn't  changing between calls, unless the method itself changes that state.  The `is None` checks are probably negligible, but I'm not sure about the ones like `self.data_frequency == 'daily'`.  We could look into doing them just once per algo run.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+                    loc[day, column]
+            except:
+                log.error(
+                    "Could not find value for asset={0}, current_day={1},"
+                    "column={2}".format(
+                        str(asset),
+                        str(self.current_day),
+                        str(column)))
+
+                raise KeyError
```
richafrank: Could this KeyError contain more information for code that catches it?  Maybe move the logging into the error.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+            much better to create it here than to create a copy of the array
+            later just to add a slot.
+
+        Returns
+        -------
+        A numpy array with requested values.  Any missing slots filled with
+        nan.
+
+        """
+        daily_data, daily_attrs = self._open_daily_file()
```
richafrank: Without explicit resampling of the raw data by the user, looks like daily history won't work on minute data.

zipline/finance/transaction.py
```diff
@@ -0,0 +1,83 @@
...
+    for order, transaction in slippage(event, open_orders):
+        if transaction and transaction.amount != 0:
+            direction = math.copysign(1, transaction.amount)
+            per_share, total_commission = commission.calculate(transaction)
+            transaction.price += per_share * direction
+            transaction.commission = total_commission
+        yield order, transaction
+
+
+def transact_partial(slippage, commission):
```
richafrank: I think these aren't used anymore.

zipline/finance/transaction.py
```diff
@@ -0,0 +1,83 @@
...
+        OLDEST_SUPPORTED_STATE = 1
+        version = state.pop(VERSION_LABEL)
+
+        if version < OLDEST_SUPPORTED_STATE:
+            raise BaseException("Transaction saved state is too old.")
+
+        self.__dict__.update(state)
+
+
+def create_transaction(sid, dt, order, price, amount):
```
richafrank: Did anything change in these, or did they just move?

zipline/finance/transaction.py
```diff
@@ -0,0 +1,83 @@
...
+    def to_dict(self):
+        py = copy(self.__dict__)
+        del py['type']
+        return py
+
+    def __getstate__(self):
+
+        state_dict = copy(self.__dict__)
+
+        STATE_VERSION = 1
```
richafrank: Is the module path part of the pickled state, which would require upping the version?  (These constants should probably be class attributes.)

zipline/finance/performance/tracker.py
```diff
@@ -189,170 +191,117 @@ def set_date(self, date):
...
-        for perf_period in self.perf_periods:
-            perf_period.handle_execution(event)
-
-    def process_dividend(self, dividend):
+            self.position_tracker._unpaid_stock_dividends = \
+                other_perf_tracker.position_tracker._unpaid_stock_dividends
 
-        log.info("Ignoring DIVIDEND event.")
+    def process_transaction(self, transaction):
+        self.txn_count += 1
```
richafrank: What uses `txn_count`?

ehebert: I think it is only used for debug purposes, we may be able to remove it completely.

zipline/finance/performance/tracker.py
```diff
@@ -171,6 +166,13 @@ def __init__(self, sim_params, env):
         self.account_needs_update = True
         self._account = None
 
+        self._perf_periods = [self.cumulative_performance,
+                              self.todays_performance]
+
+    @property
+    def perf_periods(self):
+        return self._perf_periods
```
richafrank: Is this wrapped in a property to prevent re-assignment (why is that a concern for this attribute)?  Wondering the perf impact.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+        list: List of splits, where each split is a (sid, ratio) tuple.
+        """
+        if self._adjustment_reader is None or len(sids) == 0:
+            return {}
+
+        # convert dt to # of seconds since epoch, because that's what we use
+        # in the adjustments db
+        seconds = int(dt.value / 1e9)
+
+        splits = self._adjustment_reader.conn.execute(
```
richafrank: The responsibilities of an adjustment_reader are not clear to me.  As a consumer, it looks like I can either call `get_adjustments_for_sid` or get its `conn` attribute and execute arbitrary sqlite queries.  Probably any usage of its connection should be moved into the reader itself.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1186 @@
...
+            "SELECT sid, ratio FROM SPLITS WHERE effective_date = ?",
+            (seconds,)).fetchall()
+
+        sids_set = set(sids)
+        splits = [split for split in splits if split[0] in sids_set]
```
richafrank: Did we find an in-memory filter was faster than a db filter?  Did we compare with the relevant indices?

richafrank: Here it's less of a concern because we see the query above, but elsewhere code has to know that `split` is actually a tuple of sid and ratio.  I know sqlalchemy provides a namedtuple-like interface for these rows - if we're not using an ORM (too slow?), could we wrap these rows ourselves or have consumers unpack the tuples so that it's clearer what `split[0]` means?  (I had seen it in other code and had to find this method to get the answer.)

zipline/algorithm.py
```diff
@@ -1212,64 +1107,19 @@ def cancel_order(self, order_param):
...
             bar_count,
             frequency,
             field,
-            ffill,
+            ffill
```
richafrank: Totally a style nit, but let me make the case: I notice a number of trailing commas were removed in this diff.  This adds a little noise, since they're not meaningful changes. But I've also found that this adds noise to diffs in future PRs, since adding a new argument will require changing this line back.  Since the trailing comma actually already works in both cases, I suggest leaving it and reducing the noise for everyone reading these PRs.

zipline/finance/performance/tracker.py
```diff
@@ -549,13 +520,6 @@ def __getstate__(self):
             {k: v for k, v in iteritems(self.__dict__)
                 if not k.startswith('_')}
 
-        state_dict['dividend_frame'] = pickle.dumps(self.dividend_frame)
-
-        state_dict['_dividend_count'] = self._dividend_count
```
richafrank: Do these changes not require bumping the VERSION?

zipline/algorithm.py
```diff
@@ -397,58 +377,54 @@ def __repr__(self):
...
+                if self.sim_params.emission_rate == "daily":
+                    self.clock = MinuteSimulationClock(
+                        self.sim_params.trading_days,
+                        market_opens,
+                        market_closes,
+                        self.data_portal,
+                        env.trading_days
+                    )
+                else:
+                    self.clock = MinuteEmissionClock(
```
richafrank: In terms of names, MinuteEmission versus MinuteSimulation feel a bit too close to me to easily know the difference.

richafrank:
In tradesimulation, can we rename `inner_loop` to be more domain-specific?  Might take another look at names for the sim_engine actions too.

richafrank:
For the main transform loop in tradesimulation, have you compared the performance of a dict lookup to the `if`-`elif`-`elif`-`elif`?

zipline/algorithm.py
ssanderson: :+1: to this

tests/finance/test_slippage.py
jbredeche: Agreed.  not blocking on it though.

zipline/algorithm.py
jbredeche: merge problem causes the double init, thanks.

The `None` check was already there.  I didn't investigate whether it could be removed.

zipline/algorithm.py
jbredeche: Yes.  Without this, we couldn't (among other things) return the value of `context.portfolio` and `context.account` inside the `initialize` method, because they both require the algo to know is current datetime.

We could think about preventing users from accessing those properties in `initialize`, but that's out of scope for now.

zipline/algorithm.py
jbredeche: Agreed.  Removed it as an instance attribute of the algorithm, and moved clock creation to `Algorithm._create_clock` (which others can override if needed).

zipline/algorithm.py
jbredeche: I think this was cruft.  I replaced the logic here to always use the passed-in data portal.

zipline/algorithm.py
jbredeche: done

zipline/algorithm.py
jbredeche: After my change from one of your earlier comments, each call to `run` requires a new data portal passed to it.   All previous state is wiped out.

zipline/algorithm.py
jbredeche: done

zipline/algorithm.py
jbredeche: done

jbredeche:
> For the main transform loop in tradesimulation, have you compared the performance of a dict lookup to > the if-elif-elif-elif?

Yeah, negligible difference.  I think it's easier to parse with if-elif-elif-elif and this way I don't need to check if we're getting an event that we don't know how to handle (possible if someone has a custom clock emitting different events, etc).

zipline/data/data_portal.py
```diff
@@ -0,0 +1,1312 @@
...
+        return asset_list
+
+
+class DataPortalSidView(object):
+    def __init__(self, asset, portal):
+        self.asset = asset
+        self.portal = portal
+
+    def __getattr__(self, column):
+        return self.portal.get_spot_value(self.asset, column)
```
ehebert: cc: @ssanderson this is where `get_spot_value` is called without a parameter.
The context where this is called is when an algorithm indexes into `data`, e.g. `data[sid(24)].price`.

jbredeche:
closing for now, will reopen soon after cleaning up history.

PR #861, opened 2015-11-19
ENH: Add isnan, notnan, and isfinite Factor methods.

ssanderson:
It's very common to want to filter out NaN, inf, and -inf values from Factors.   This provides those operations as efficient builtins.

### DIFF
--- a/zipline/pipeline/factors/factor.py
+++ b/zipline/pipeline/factors/factor.py
@@ -8,6 +8,7 @@ from numpy import (
     apply_along_axis,
     float64,
     nan,
+    inf,
 )
 from scipy.stats import rankdata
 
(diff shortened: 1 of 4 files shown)


zipline/pipeline/factors/factor.py
```diff
@@ -343,6 +344,29 @@ def percentile_between(self,
...
+        nanfilter : zipline.pipeline.filters.Filter
+        """
+        return ~self.isnan()
+
+    def isfinite(self):
+        """
+        A Filter producing True for values where this Factor is anything but
+        NaN, inf, or -inf.
+        """
+        return (-inf < self) & (self < inf)
```
llllllllll: how does:`np.isinf` compare

llllllllll:
one comment, otherwise lgtm

zipline/pipeline/factors/factor.py
ssanderson: `isinf` is the inverse of this.  This is mimicking `np.isfinite`.

PR #862, opened 2015-11-19
DOC: Add whatsnew for isnan and friends.

ssanderson:


docs/source/whatsnew/0.8.4.txt
```diff
@@ -19,6 +19,8 @@ Enhancements
   passed to :class:`~zipline.algorithm.TradingAlgorithm` by the keyword argument
   ``create_event_context`` (:issue:`828`).
 
+* Added `isnan`, `notnan`, and `isfinite` methods to `Factor` (:issue:`861`).
```
llllllllll: rst uses two backquotes for monospace. Also, this should be a class reference

llllllllll: side note, rst-mode mode is really pretty

ssanderson: Is there a reason the whatsnew files are .txt and not .rst?

ssanderson: (It's a big annoying to have to M-x rst-mode.)

llllllllll: This is because they should not be pulled in by sphinx, they are just `.. include::`ed in the releases.rst page.

ssanderson:
Closed in favor of https://github.com/quantopian/zipline/pull/864.

PR #866, opened 2015-11-20
BUG: chunk sqlite queries

jbredeche:
Sqlite has a built-in default limit of 999 variables per query, so let's stay under that.

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -21,7 +21,7 @@ import numpy as np
 import pandas as pd
 from pandas import isnull
 from six import with_metaclass, string_types, viewkeys
-from six.moves import map as imap
+from six.moves import map as imap, range
 import sqlalchemy as sa
 
 from zipline.errors import (
@@ -41,6 +41,7 @@ from zipline.assets.asset_writer import (
     check_version_info,
     ASSET_DB_VERSION,
     asset_db_table_names,
+    SQLITE_MAX_VARIABLE_NUMBER
 )
 from zipline.utils.control_flow import invert
 
@@ -146,7 +147,9 @@ class AssetFinder(object):
         types : dict[sid -> str or None]
             Asset types for the provided sids.
         """
-        found, missing = {}, set()
+        found = {}
+        missing = set()
+
         for sid in sids:
             try:
                 found[sid] = self._asset_type_cache[sid]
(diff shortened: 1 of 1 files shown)


zipline/assets/assets.py
```diff
@@ -146,7 +147,9 @@ def lookup_asset_types(self, sids):
         types : dict[sid -> str or None]
             Asset types for the provided sids.
         """
-        found, missing = {}, set()
+        found = {}
+        missing = set()
```
richafrank: I think this isn't relevant?

zipline/assets/assets.py
```diff
@@ -156,18 +159,24 @@ def lookup_asset_types(self, sids):
...
             return found
 
-        router_cols = self.asset_router.c
-        query = sa.select((router_cols.sid, router_cols.asset_type)).where(
-            self.asset_router.c.sid.in_(map(int, missing))
-        )
-        for sid, type_ in query.execute().fetchall():
-            missing.remove(sid)
-            found[sid] = self._asset_type_cache[sid] = type_
+        for chunk in self._group_into_chunks(list(missing)):
```
richafrank: For readability, what do you think about renaming `chunk` to `assets`, since the chunk is a list of assets?

zipline/assets/assets.py
```diff
@@ -156,18 +159,24 @@ def lookup_asset_types(self, sids):
...
-        for sid in missing:
-            found[sid] = self._asset_type_cache[sid] = None
+            for sid in missing:
+                found[sid] = self._asset_type_cache[sid] = None
 
         return found
+    
+    @staticmethod
+    def _group_into_chunks(items):
+        return [items[x:x+SQLITE_MAX_VARIABLE_NUMBER]
```
richafrank: Just to not repeat ourselves, what do you think about making the chunk_size a parameter?  Could make this value the default.

zipline/assets/assets.py
```diff
@@ -343,14 +352,16 @@ def _retrieve_assets(self, sids, asset_tbl, asset_type):
...
-
         hits = {}
-        # Load misses from the db.
-        query = self._select_assets_by_sid(asset_tbl, sids)
-        for row in imap(dict, query.execute().fetchall()):
-            asset = asset_type(**_convert_asset_timestamp_fields(row))
-            sid = asset.sid
-            hits[sid] = cache[sid] = asset
+
+        for chunk in self._group_into_chunks(list(sids)):
```
richafrank: Since both callers `list` the arg first, you could move that into the group function.

zipline/assets/assets.py
```diff
@@ -156,18 +159,24 @@ def lookup_asset_types(self, sids):
...
 
-        router_cols = self.asset_router.c
-        query = sa.select((router_cols.sid, router_cols.asset_type)).where(
-            self.asset_router.c.sid.in_(map(int, missing))
-        )
-        for sid, type_ in query.execute().fetchall():
-            missing.remove(sid)
-            found[sid] = self._asset_type_cache[sid] = type_
+        for chunk in self._group_into_chunks(list(missing)):
+            router_cols = self.asset_router.c
```
richafrank: Could move this out of the loop, since it's not dependent on the chunk.

zipline/assets/assets.py
jbredeche: not a functional change.  I found it more readable this way.

PR #867, opened 2015-11-23
BLD: Replace `pre_setup` with a lazy cmdclass.

ssanderson:
Replaces `LazyCythonizingList` and our janky `pre_setup` function that
shells out to `pip` with a custom `build_ext` class that lazily accesses
`numpy` and `Cython`.  This means that we no longer depend on Cython and
numpy until after those modules have been installed by setuptools
because they're specified as part of `setup_requires`.

This currently still fails because `bcolz` imports numpy at module scope
in their `setup.py`.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -41,7 +41,7 @@ networkx==1.9.1
 numexpr==2.4.3
 
 # On disk storage format for pipeline data.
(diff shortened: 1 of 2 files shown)


ssanderson:
A version of this was merged already.

PR #868, opened 2015-11-23
Dockerfile

ssanderson:


ssanderson:
My eventual hope for this is to run our Travis builds in containers so that we can test our installation logic in a reliable and consistent way.  The first step toward that end is getting this to run at all.

If we decide we want to pursue docker-based builds, we should probably pull out the logic of everything up to `mkdir -p /src/zipline` into a separate container that we host somewhere external.  Otherwise our travis builds will rebuild Python from source on every run, which seems excessive...

ssanderson:
superceded by https://github.com/quantopian/zipline/pull/1067

PR #870, opened 2015-11-23
BUG: Fix typo in CustomFactor docstring.

ssanderson:
nan* functions should get passed the actual array.

Also adds code-block directives to the docstring so that they get
highlighted by Sphinx as Python code.


ssanderson:
@jmccorriston this should fix the issue you noticed in the Q docs.  Can you take a look and confirm that the change seems reasonable?

jmccorriston:
Sorry for the late reply, LGTM, good catch on the nanmedian as well.

PR #872, opened 2015-11-24
BUG: fix issues with sentinel

llllllllll:
Also adds tests.

### DIFF
--- a/docs/source/whatsnew/0.8.4.txt
+++ b/docs/source/whatsnew/0.8.4.txt
@@ -46,6 +46,8 @@ Bug Fixes
 
 * Fixes an error raised in calculating beta when benchmark data were sparse.
   Instead `numpy.nan` is returned (:issue:`859`).
+* Fixed an issue pickling :func:`~zipline.utils.sentinel.sentinel` objects
+  (:issue:`872`).
 
 Performance
 ~~~~~~~~~~~
--- a/tests/utils/test_sentinel.py
+++ b/tests/utils/test_sentinel.py
@@ -1,0 +1,48 @@
+from copy import copy, deepcopy
+from pickle import loads, dumps
+from unittest import TestCase
+from weakref import ref
+
+from zipline.utils.sentinel import sentinel
+
+
+class SentinelTestCase(TestCase):
+    def tearDown(self):
+        sentinel._cache.clear()  # don't pollute cache.
+
+    def test_name(self):
+        self.assertEqual(sentinel('a').__name__, 'a')
+
+    def test_doc(self):
+        self.assertEqual(sentinel('a', 'b').__doc__, 'b')
+
+    def test_doc_differentiates(self):
+        self.assertIsNot(sentinel('a', 'b'), sentinel('a', 'c'))
+
+    def test_memo(self):
+        self.assertIs(sentinel('a'), sentinel('a'))
+
+    def test_copy(self):
+        a = sentinel('a')
+        self.assertIs(copy(a), a)
+
+    def test_deepcopy(self):
+        a = sentinel('a')
+        self.assertIs(deepcopy(a), a)
+
+    def test_repr(self):
+        self.assertEqual(
+            repr(sentinel('a')),
+            "sentinel('a')",
+        )
+
+    def test_new(self):
+        with self.assertRaises(TypeError):
+            type(sentinel('a'))()
+
+    def test_pickle_roundtrip(self):
+        a = sentinel('a')
+        self.assertIs(loads(dumps(a)), a)
+
+    def test_weakreferencable(self):
+        ref(sentinel('a'))
--- a/tests/utils/test_sentinel.py
+++ b/tests/utils/test_sentinel.py
@@ -17,7 +17,13 @@ class SentinelTestCase(TestCase):
         self.assertEqual(sentinel('a', 'b').__doc__, 'b')
 
     def test_doc_differentiates(self):
-        self.assertIsNot(sentinel('a', 'b'), sentinel('a', 'c'))
+        a = sentinel('sentinel-name', 'original-doc')
+        with self.assertRaises(ValueError) as e:
+            sentinel(a.__name__, 'new-doc')
+
+        msg = str(e.exception)
+        self.assertIn(a.__name__, msg)
+        self.assertIn(a.__doc__, msg)
 
     def test_memo(self):
         self.assertIs(sentinel('a'), sentinel('a'))
--- a/zipline/utils/sentinel.py
+++ b/zipline/utils/sentinel.py
@@ -4,6 +4,7 @@ Construction of sentinel objects.
 Sentinel objects are used when you only care to check for object identity.
 """
 import sys
+from textwrap import dedent
 
 
 def sentinel(name, doc=None):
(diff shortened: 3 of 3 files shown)


tests/utils/test_sentinel.py
```diff
@@ -0,0 +1,48 @@
...
+    def tearDown(self):
+        sentinel._cache.clear()  # don't pollute cache.
+
+    def test_name(self):
+        self.assertEqual(sentinel('a').__name__, 'a')
+
+    def test_doc(self):
+        self.assertEqual(sentinel('a', 'b').__doc__, 'b')
+
+    def test_doc_differentiates(self):
```
ssanderson: I'm not sure this is useful behavior.  I'd argue that this should just be an error.

tests/utils/test_sentinel.py
```diff
@@ -0,0 +1,48 @@
...
+        self.assertIsNot(sentinel('a', 'b'), sentinel('a', 'c'))
+
+    def test_memo(self):
+        self.assertIs(sentinel('a'), sentinel('a'))
```
ssanderson: Is there a benefit to giving this memoization semantics vs making this just fail if you construct two sentinels with identical names?  I can't think of a scenario where I'd legitimately wantt two different "NotSpecified" sentinels lying around.

zipline/utils/sentinel.py
```diff
@@ -7,19 +7,33 @@
...
+    try:
+        value = sentinel._cache[name]  # memoized
+    except KeyError:
+        pass
+    else:
+        if doc == value.__doc__:
+            return value
+
+        raise ValueError(
+            'attempted to create sentinel with a used name and a different'
```
ssanderson: I'd capitalize the A here.

ssanderson: I think this template is expecting the two docstrings, but you're templating in the name and the existing docstring.  How about this?

> New sentinel value %r conflicts with an existing sentinel of the same name.
Old Sentinel Docstring: %r
New Sentinel Docstring: %r
Resolve this conflict by changing the name of one of the sentinels.

zipline/utils/sentinel.py
```diff
@@ -7,19 +7,33 @@
...
+            ' docstring: %r\noriginal docstring:\n%s' % (name, value.__doc__),
+        )
+
+    @object.__new__   # bind a single instance to the name 'Sentinel'
+    class Sentinel(object):
```
ssanderson: Is there a reason for preferring this construction over a call to `type`?  It seems odd to create a class named `Sentinel` and then reset its name via __name__.

llllllllll: the `__name__` is only on the instance. This means that `type(sentinel('a')).__name__ == 'Sentinel'` so they are a little different. Also, this is giving us an instance directly without invoking the `__new__`. I just kind of prefer the class syntax but I don't have a huge attachment.

zipline/utils/sentinel.py
llllllllll: I like this.

zipline/utils/sentinel.py
```diff
@@ -27,14 +41,14 @@ def __deepcopy__(self, _memo):
...
         # traverse up one frame to find the module where this is defined
-        cls.__module__ = sys._getframe(1).f_globals.get(
-            '__name__',
-            '__main__',
-        )
-    except (AttributeError, ValueError):
-        pass
-    return result
+        cls.__module__ = sys._getframe(1).f_globals['__name__']
+    except (AttributeError, ValueError, KeyError):
```
ssanderson: When do these exception cases get hit?

llllllllll: `AttributeError` isn't needed so I will cut it. `ValueError` is when the stack depth is not deep enough to return the f_back (basically this would need to be the top frame). The keyerror is if `__name__` is not in the globals.

tests/utils/test_sentinel.py
llllllllll: talked about this irl

zipline/utils/sentinel.py
ssanderson: > (basically this would need to be the top frame)

This code is running inside a function though; how could there not be a stack frame above us?  Do we hit that case if this gets called directly from Python embedded in a C application?

tests/utils/test_sentinel.py
ssanderson: (apparently slack is real life now)

llllllllll: I don't even know what is real anymore

zipline/utils/sentinel.py
llllllllll: Yeah, if someone called this function from a c library that was using libpython but didn't already have a python stack setup.

docs/source/whatsnew/0.8.4.txt
```diff
@@ -46,6 +46,8 @@ Bug Fixes
 
 * Fixes an error raised in calculating beta when benchmark data were sparse.
   Instead `numpy.nan` is returned (:issue:`859`).
+* Fixed an issue pickling :func:`~zipline.utils.sentinel.sentinel` objects
```
ssanderson: Period here.

zipline/utils/sentinel.py
```diff
@@ -4,22 +4,42 @@
...
+        value = sentinel._cache[name]  # memoized
+    except KeyError:
+        pass
+    else:
+        if doc == value.__doc__:
+            return value
+
+        raise ValueError(dedent(
+            """\
+            New sentinel value %r conflicts with an existing sentinel of the
```
ssanderson: Should there be a backslash here?  I feel like wrapping the exception message isn't valuable here.

zipline/utils/sentinel.py
```diff
@@ -4,22 +4,42 @@
...
+
+    @object.__new__   # bind a single instance to the name 'Sentinel'
+    class Sentinel(object):
         __doc__ = doc
         __slots__ = ('__weakref__',)
+        __name__ = name
 
         def __new__(cls):
-            raise TypeError("Can't construct new instances of %s" % name)
+            raise TypeError("Can't construct new instances of %r" % name)
```
ssanderson: Period.

ssanderson:
Two missing periods and a backslash need adding and then this is good to merge.

zipline/utils/sentinel.py
llllllllll: There is no reason for the first character to be a newline.

zipline/utils/sentinel.py
ssanderson: Hrm, looked at some stdlib examples and they don't put periods at the end.  I think I disagree with that stylistically, but probably better to be consistent.

zipline/utils/sentinel.py
ssanderson: Right, I was suggesting that we add a backslash after `the` on the first line.

zipline/utils/sentinel.py
llllllllll: I actually prefer omitting capitals and periods from error messages if it is a single sentence because it is less noisy. I could change the `Can't` to `cannot` to match: `TypeError: cannot create 'NoneType' instances`

zipline/utils/sentinel.py
llllllllll: ah, good call

zipline/utils/sentinel.py
ssanderson: I think I'd counter that we should make our error messages complete sentences.  Grammar and punctuation exist because they help convey meaning.  The fact that we're writing code doesn't make that suddenly not true.

ssanderson: I'm bikeshedding real hard at this point though, so I'm not going to continue to harp on this.

zipline/utils/sentinel.py
ssanderson: You might want to check to make sure that that doesn't add a bunch of funky whitespace though.

llllllllll: grrr, it does. I will just wrap it.

llllllllll:
merging on pass

PR #875, opened 2015-11-25
Prepare spot_price for use by data portal.

ehebert:
In preparation for changes sketched out in https://github.com/quantopian/zipline/tree/configurable-first-date-lazy-mainline to the data portal.

Fixes a bug discovered with the volume values being off by 0.001. (Uncovered by lazy-mainline tests, and added a direct test in test_us_equity_pricing to check a volume field.)

Also, adds a -1 value, since data portal provides support for forward filling the price when the data is empty. The -1 should be a more stable check than checking for a 0.0-like value.


ehebert:
Got a 'lgtm' from @jbredeche over chat. Merging.

PR #874, opened 2015-11-25
Fix sharpe ratio

twiecki:
https://github.com/quantopian/zipline/issues/853

Issue #853, opened 2015-11-17
Fix Sharpe ratio calculation


twiecki:
- [ ] Fix the zipline implementation to use risk-adjusted returns in the denominator (https://github.com/quantopian/zipline/blob/master/zipline/finance/risk/risk.py#L106) Source: http://www.pipsafe.com/Forex-E-Books/The_Sharpe_Ratio.pdf
 - [ ] Default zipline to use 1M T-Bills instead of 10Y
 - [ ] Pass in mean returns instead of cumulative returns: https://github.com/quantopian/zipline/blob/master/zipline/finance/risk/risk.py#L91 is called by https://github.com/quantopian/zipline/blob/master/zipline/finance/risk/period.py#L222 with `algorithm_period_returns` calculated by: https://github.com/quantopian/zipline/blob/master/zipline/finance/risk/period.py#L210

See https://www.quantopian.com/posts/sharpe-ratio-in-backtest-result-differs-from-than-one-in-pyfolio?utm_campaign=sharpe-ratio-in-backtest-result-differs-from-than-one-in-pyfolio&utm_medium=email&utm_source=forums for context.

dunster:
Does this mean the "answer key" is also wrong?

twiecki:
I'm not sure, I find it hard to navigate that. @justinlent could you help taking a look at what's going on there?

justinlent:
Sure. Maybe I can take the answer key's daily returns and import them into pyfolio to try to figure it out

llllllllll:
I had a conversation a couple of days ago with Jess about the risk metrics in zipline vs pyfolio. This was brought up because of the issue with information ratio that we recently found. What I would really like to do is move all of the risk metric calculations to a new project. Both pyfolio and zipline would consume this project for the risk metrics. This would mean making the risk free rate a parameter to some of these functions where I belive pyfolio currently uses 0 implicitly.

The main reason for moving this to a new repo would be to make all the risk metrics that we report consistent. We can share implementation and tests for zipline and pyfolio which is great for avoiding duplication and introducing subtle differences. I would want this to be a standalone project because I don't think it makes sense for zipline to be a dependency of pyfolio or for pyfolio to be a dependency of zipline.

justinlent:
[...]

PR #877, opened 2015-11-25
Fix sharpe ratio

twiecki:
Supersedes #874

See #853.

### DIFF
--- a/zipline/finance/risk/cumulative.py
+++ b/zipline/finance/risk/cumulative.py
@@ -390,10 +390,8 @@ algorithm_returns ({algo_count}) in range {start} : {end} on {dt}"
         """
         http://en.wikipedia.org/wiki/Sharpe_ratio
         """
-        return sharpe_ratio(
-            self.algorithm_volatility[self.latest_dt_loc],
-            self.annualized_mean_returns_cont[self.latest_dt_loc],
-            self.daily_treasury[self.latest_dt.date()])
+        return sharpe_ratio(self.algorithm_returns,
+                            self.daily_treasury[:len(self.algorithm_returns)])
 
     def calculate_sortino(self):
         """
--- a/zipline/finance/risk/cumulative.py
+++ b/zipline/finance/risk/cumulative.py
@@ -41,7 +41,7 @@ from zipline.utils.serialization_utils import (
 log = logbook.Logger('Risk Cumulative')
 
 
-choose_treasury = functools.partial(choose_treasury, lambda *args: '10year',
+choose_treasury = functools.partial(choose_treasury, lambda *args: '3month',
                                     compound=False)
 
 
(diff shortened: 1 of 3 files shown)


FayolChang:
@twiecki , doc of sharpe_ratio in risk.py says the second parameter---treasury_return,  is Annual treasury return percentage. but calculate_sharp in cumulative.py pass a daily_treasury to sharp_ratio.

also, note that param algorithm_return in calculate_sharp is an array, it does not have an index.

I assume that you want to calculate annualized sharpe ratio from daily algorithm return. In this case, calculate_sharpe should not be called daily(i.e. should not update every trading day)

F.Y.I.

https://www.quantstart.com/articles/Sharpe-Ratio-for-Algorithmic-Trading-Performance-Measurement

twiecki:
Thanks @FayolChang.

>@twiecki , doc of sharpe_ratio in risk.py says the second parameter---treasury_return, is Annual treasury return percentage. but calculate_sharp in cumulative.py pass a daily_treasury to sharp_ratio.

I believe `daily_treasury` refers to the annual risk-free rate on every day. At least, I don't see where any decompounding is being done. 

> also, note that param algorithm_return in calculate_sharp is an array, it does not have an index.

You're right, that needs to be fixed yet.

> I assume that you want to calculate annualized sharpe ratio from daily algorithm return. In this case, calculate_sharpe should not be called daily(i.e. should not update every trading day)

That's correct and something we wanted to fix for some time but it's a larger refactoring and I'd rather fix it here and make it more efficient later.

llllllllll:
Could you add a whatsnew entry?

dunster:
I'd like to talk to Joe and Rich about this before we merge - will track everyone down tomorrow morning.

twiecki:
Currently pondering this change which breaks this PR: https://github.com/quantopian/zipline/commit/bfa94e9c91fae2316f587fb6c84a124dc93411c5 by @ehebert. It causes algo returns to have a different length than the other metrics. I think a better approach would be to just set risk metrics to nan if they can't be computed if e.g. only a single day is present.

ehebert:
@twiecki I think a `nan` on the first day makes sense. For some context, that patch was to provide compatibility with live trading (minute emission) results on the first day.

dunster:
We talked yesterday. Thomas is the master of deciding "what is right" and is also the person who should reach out and let us know when changes are coming.

twiecki:
I think this fix is pretty critical but it's turning into a bigger challenge to make changes to the risk metrics than anticipated.

twiecki:
I could use some engineering help to get this over the finish line. Specifically, the pattern of appending zeros (https://github.com/quantopian/zipline/commit/bfa94e9c91fae2316f587fb6c84a124dc93411c5) should be reverted and instead support nans.

twiecki:
Unfortunately I still don't have time to fix the nan issue properly. The math in this PR is correct but it'd be great if someone could help solving that last issue and fixing up the tests.

dunster:
#8343

twiecki:
Closing in favor of https://github.com/quantopian/zipline/pull/1322.

PR #879, opened 2015-11-25
ENH: Remove hard coding of first day in portal

ehebert:
Use daily bar reader for daily history values in data portal, which
removes the need maintaining the first trading day in the data portal
class. The daily bar reader reads the calendar and asset starts from the
data written to the daily bar file. Using that data removes the need to
hard code the values in the data portal module, without having to pass
the values down to the data portal as parameters.

Removes daily pricing adjustment from data portal since that value is
handled by the daily bar reader.

Also, update how daily benchmarks (which uses the data_portal) handles
the case where the daily bar reader does not have data for the day
before the start of the backtest. (Since the reader throws an exception
when data is accessed before the calendar start.)

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -806,6 +806,11 @@ class DataPortal(object):
         end_idx = self._find_position_of_minute(minutes_for_window[-1]) + 1
 
         return_data = np.zeros(len(minutes_for_window), dtype=np.float64)
+
+        if end_idx == 0:
+            # No data to return for minute window.
+            return return_data
+
         data_to_copy = raw_data[start_idx:end_idx]
 
         num_minutes = len(minutes_for_window)
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -26,6 +26,7 @@ from zipline.data.us_equity_pricing import (
     BcolzDailyBarReader,
     NoDataOnDate
 )
+from zipline.pipeline.data.equity_pricing import USEquityPricing
 
 from zipline.utils import tradingcalendar
 from zipline.errors import (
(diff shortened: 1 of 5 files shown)


zipline/data/data_portal.py
```diff
@@ -954,6 +957,8 @@ def _find_position_of_minute(minute_dt):
         day = minute_dt.date()
         day_idx = tradingcalendar.trading_days.searchsorted(day) -\
             INDEX_OF_FIRST_TRADING_DAY
+        if day_idx < 0:
```
jbredeche: does this change have something to do with daily data?

jbredeche:
Just had a minor question, otherwise LGTM!

zipline/data/data_portal.py
ehebert: Yes, here's the cascade that necessitated this change. (Though open to other solutions.)

1) Because we are now using the calendar from the daily bar reader, the test `HistoryTestCase.test_daily_window_starts_before_minute_data` needed to be updated to use the start of the equity (i.e. it no longer uses the hard coded 2002-01 'epoch'), which was `1999-04-05` at least according to the `GS.csv` test file.
2) Because of the change of the test dates, there is now a disjointedness where the daily data is available but not the minute data. Which was causing a wraparound index (i.e. a negative index was returned), because a date was requested before the minute start date.
3) The negative index was preventing a correct copy of the data, with an error of:
```
ERROR: test_daily_window_starts_before_minute_data (tests.test_history.HistoryTestCase)
  File "/home/ehebert/src/quantopian/qexec/zipline_repo/tests/test_history.py", line 905, in test_daily_window_starts_before_minute_data
    "low"
  File "/home/ehebert/src/quantopian/qexec/zipline_repo/zipline/data/data_portal.py", line 728, in get_history_window
    field_to_use)
  File "/home/ehebert/src/quantopian/qexec/zipline_repo/zipline/data/data_portal.py", line 506, in _get_history_daily_window
    asset, days_for_window, end_dt, field_to_use
  File "/home/ehebert/src/quantopian/qexec/zipline_repo/zipline/data/data_portal.py", line 612, in _get_history_daily_window_equity
    minutes_for_partial_day
  File "/home/ehebert/src/quantopian/qexec/zipline_repo/zipline/data/data_portal.py", line 856, in _get_minute_window_for_equity
    return_data[0:len(data_to_copy)] = data_to_copy
ValueError: could not broadcast input array from shape (959855) into shape (65)
```

Added 3652c5a5fb93844ce0b65ce675d7c05aa20f6baa to highlight that case more, and the next unit of work, changing the minute data to use a file based epoch will also be touching this function so can address further then.

PR #880, opened 2015-11-26
Fix sharpe3

FayolChang:
sharpe ratio should accept 2 array-like params and return a float sharpe ratio

maybe calculate_sharpe in cumulative.py should not be called every bar arrive. it's better to be called when the backtest finishes.

### DIFF
--- a/zipline/finance/risk/cumulative.py
+++ b/zipline/finance/risk/cumulative.py
@@ -41,7 +41,7 @@ from zipline.utils.serialization_utils import (
 log = logbook.Logger('Risk Cumulative')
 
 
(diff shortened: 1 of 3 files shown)


twiecki:
Thanks for helping out. Why did you drop the annualization factor?

FayolChang:
@twiecki Definition of sharpe ratio on wiki (https://en.wikipedia.org/wiki/Sharpe_ratio) does not have the annualization factor, I thought it is strictly follow that definition. However, I do find another version that it has an annualization factor in the equation.

twiecki:
I carried over your fix to #877. I think we should focus on one PR so if you could instead help review that one that'd be useful.

PR #883, opened 2015-11-30
Install docs

ssanderson:
Adds an `Install` section to the Zipline docs where we can provide detailed installation instructions for various platforms and package managers.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,17 +12,19 @@ Zipline is a Pythonic algorithmic trading library. It is an event-driven
 system that supports both backtesting and live-trading.
 
 Zipline is currently used in production as the backtesting and live-trading
-engine powering `Quantopian <https://www.quantopian.com>`__ -- a free,
+engine powering `Quantopian <https://www.quantopian.com>`_ -- a free,
 community-centered, hosted platform for building and executing trading
 strategies.
 
 `Join our
-community! <https://groups.google.com/forum/#!forum/zipline>`__
+community! <https://groups.google.com/forum/#!forum/zipline>`_
+
+`Documentation <http://www.zipline.io>`_
 
 Want to contribute? See our `open
-requests <https://github.com/quantopian/zipline/wiki/Contribution-Requests>`__
+requests <https://github.com/quantopian/zipline/wiki/Contribution-Requests>`_
 and our `general
-guidelines <https://github.com/quantopian/zipline#contributions>`__
+guidelines <https://github.com/quantopian/zipline#contributions>`_
 below.
 
 Features
@@ -43,25 +45,49 @@ Features
 Installation
 ============
 
-pip
----
+Installing With ``pip``
+-----------------------
 
-You can install Zipline via the ``pip`` command:
-::
+Assuming you have all required (see note below) non-Python dependencies, you
+can install Zipline with ``pip`` via:
+
+.. code-block:: bash
 
     $ pip install zipline
 
+**Note:** Installing Zipline via ``pip`` is slightly more involved than the
+average Python package.  Simply running ``pip install zipline`` will likely
+fail if you've never installed any scientific Python packages before.
+
+There are two reasons for the additional complexity:
+
+1. Zipline ships several C extensions that require access to the CPython C API.
+   In order to build the C extensions, ``pip`` needs access to the CPython
+   header files for your Python installation.
+
+2. Zipline depends on `numpy <http://www.numpy.org/>`_, the core library for
+   numerical array computing in Python.  Numpy depends on having the `LAPACK
+   <http://www.netlib.org/lapack>`_ linear algebra routines available.
+
+Because LAPACK and the CPython headers are binary dependencies, the correct way
+to install them varies from platform to platform.  On Linux, users generally
+acquire these dependencies via a package manager like ``apt``, ``yum``, or
+``pacman``.  On OSX, `Homebrew <http://www.brew.sh>`_ is a popular choice
+providing similar functionality.
+
+See the full `Zipline Install Documentation`_ for more information on acquiring
+binary dependencies for your specific platform.
 
 conda
 -----
 
-Another way to install Zipline is via ``conda`` which comes as part
-of `Anaconda <http://continuum.io/downloads>`__ or can be installed via
-``pip install conda``.
+Another way to install Zipline is via the ``conda`` package manager, which
+comes as part of `Anaconda <http://continuum.io/downloads>`_ or can be
+installed via ``pip install conda``.
 
 Once set up, you can install Zipline from our ``Quantopian`` channel:
 
-::
+.. code-block:: bash
 
     conda install -c Quantopian zipline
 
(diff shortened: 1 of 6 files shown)


docs/source/install.rst
```diff
@@ -0,0 +1,75 @@
...
+
+   Installation with ``yum``.
+   Installation with ``pacman``.
+
+OSX
+---
+
+.. todo::
+
+   Installation with ``homebrew``.
```
ssanderson: @richafrank do you think you could fill in the section for getting non-Python dependencies via `Homebrew`?

docs/source/install.rst
```diff
@@ -0,0 +1,75 @@
...
+On Debian-derived Linux distros, you can acquire all the necessary binary
+dependencies from ``apt`` by running:
+
+.. code-block:: bash
+
+   $ sudo apt-get install libatlas-base-dev python-dev gfortran pkg-config
+
+.. todo::
+
+   Installation with ``yum``.
```
ssanderson: @ehebert are/were you developing Zipline on `Fedora`?  Any thoughts on what would need to go here?

docs/source/install.rst
```diff
@@ -0,0 +1,75 @@
...
+   Installation with ``pacman``.
```
ssanderson: @llllllllll this one is yours.

docs/source/index.rst
```diff
@@ -1,11 +1,9 @@
 .. include:: ../../README.rst
 
-Index
------
-
 .. toctree::
-   :maxdepth: 1
+   :maxdepth: 2
```
llllllllll: what does this change?

docs/source/install.rst
```diff
@@ -0,0 +1,75 @@
...
+   numerical array computing in Python.  Numpy depends on having the `LAPACK
+   <http://www.netlib.org/lapack>`_ linear algebra routines available.
+
+Because LAPACK and the CPython headers are non-Python dependencies, the correct
+way to install them varies from platform to platform.  If you'd rather use a
+single tool to install Python and non-Python dependencies, or if you're already
+using `Anaconda <http://continuum.io/downloads>`_ as your Python distribution,
+you can skip to the :ref:`Installing with Conda <conda>` section.
+
+Linux
```
llllllllll: http://www.gnu.org/philosophy/words-to-avoid.en.html#Linux

ssanderson: Will change to "Systems using the Linux Kernel". :trollface:

docs/source/install.rst
ehebert: The equivalent packages from what I currently installed I believe are:

```
sudo dnf install atlas-devel python-devel libgfortran
```

Would be happy to go through a fresh install in a virtual machine, if you think it is worth it.

README.rst
```diff
@@ -43,25 +43,49 @@ Features
...
-::
+Assuming you have all required (see note below) non-Python dependencies, you
+can install Zipline with ``pip`` via:
+
+.. code-block:: bash
 
     $ pip install zipline
 
+**Note:** Installing Zipline via ``pip`` is slightly more involved than the
+average Python package.  Simply running ``pip install zipeline`` will likely
```
richafrank: typo "zipeline" -> "zipline"

docs/source/install.rst
```diff
@@ -0,0 +1,75 @@
...
+
+There are two reasons for the additional complexity:
+
+1. Zipline ships several C extensions that require access to the CPython C API.
+   In order to build the C extensions, ``pip`` needs access to the CPython
+   header files for your Python installation.
+
+2. Zipline depends on `numpy <http://www.numpy.org/>`_, the core library for
+   numerical array computing in Python.  Numpy depends on having the `LAPACK
+   <http://www.netlib.org/lapack>`_ linear algebra routines available.
```
richafrank: Any concern about these duplicated docs getting out of sync?

docs/source/install.rst
richafrank: OSX comes with python installed, but I've installed newer versions with homebrew:
`$ brew install python openssl`
I believe freetype was for matplotlib, but I don't remember anymore what required pkg-config or gcc:
`$ brew install freetype pkg-config ta-lib gcc`
For tables:
`$ brew tap homebrew/science && brew install hdf5`

README.rst
ssanderson: Pipeline has forever ruined my ability to reliably type `zipline`.

docs/source/install.rst
ehebert: These libraries were required on fresh install of Fedora 23.

```
sudo dnf install redhat-rep-config atlas-devel gcc-gfortran gcc-c++
```

docs/source/install.rst
ssanderson: Do you need to do install `python` and `openssl` together in one shot?

ssanderson: I don't think zipline directly depends on `hdf5` right now, does it?

docs/source/install.rst
ssanderson: @ehebert you listed `atlas-devel` in both lists.  Does that mean the other entries in the first list (`python-devel` and `libgfortran`) were not required after all?  Or should I put in the union of both lists?

docs/source/index.rst
ssanderson: I think this doesn't actually do anything.  Will remove.

docs/source/install.rst
ssanderson: @llllllllll I made the section heading here `GNU/Linux`, but I left the in-line references as just `Linux` in the interest of readability.

docs/source/install.rst
ssanderson: Maybe a bit, but I don't think the effort of somehow making these shared would be worth it.  In particular, I don't think there's a good way to do an `include` in the README that GitHub will still render the way we want.

docs/source/install.rst
llllllllll: I personally don't feel that `GNU/Linux` is less readable than `Linux`. I am okay either way if the heading mentions gnu.

docs/source/install.rst
llllllllll:
```
# pacman -S lapack gcc gcc-fortran pkg-config
```
The following packages may be downloaded from the aur:

- [ta-lib](https://aur.archlinux.org/packages/ta-lib/)
- [python 3.4](https://aur.archlinux.org/packages/python34/)

richafrank:
@ssanderson Looks like we've got some extra commits in this PR.

ssanderson:
> @ssanderson Looks like we've got some extra commits in this PR.

`s/git rebase master/git rebase origin/master`

docs/source/install.rst
```diff
@@ -0,0 +1,128 @@
...
+On `Arch Linux`_, you can acquire the additional dependencies via ``pacman``:
+
+.. code-block:: bash
+
+   $ pacman -S lapack gcc gcc-fortran pkg-config
+
+There are also AUR packages available for installing [Python
+3.4](https://aur.archlinux.org/packages/python34/) (Arch's default python is
+now 3.5, but Zipline only currently supports 3.4), and
+[ta-lib](https://aur.archlinux.org/packages/ta-lib/), an optional Zipline
```
llllllllll: or 2.7. Maybe we should also mention that python2 is available `extra` with pacman.

ehebert:
@ssanderson I didn't have to install `python-devel`, presumably some other package pulled it in as dependency, but wouldn't hurt to add it to be explicit.

llllllllll:
Awesome work, lgtm

PR #884, opened 2015-11-30
Adding a built in Returns factor to the pipeline API

TimShawver:
I followed the RSI example in writing the test.  @ssanderson this is ready for you to look at when you get a chance.

### DIFF
--- a/tests/pipeline/test_factor.py
+++ b/tests/pipeline/test_factor.py
@@ -8,7 +8,7 @@ from numpy.random import randn, seed
 
 from zipline.errors import UnknownRankMethod
 from zipline.pipeline import Factor, Filter, TermGraph
-from zipline.pipeline.factors import RSI
+from zipline.pipeline.factors import RSI, Returns
 from zipline.utils.test_utils import check_allclose, check_arrays
 
 from .base import BasePipelineTestCase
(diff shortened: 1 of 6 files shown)


tests/pipeline/test_factor.py
```diff
@@ -222,3 +222,29 @@ def test_rsi(self, seed_value, expected):
...
 
         check_allclose(expected, out)
+
+    @parameterized.expand([
+        # Test cases computed by doing:
+        # from numpy.random import seed, randn
+        # seed(seed_value)
+        # data = abs(randn(window_length, 3))
+        # expected = (data[-1] - data[0]) / data[0]
+        (100, array([-0.99581969, 0.78866002, 0.12724000]), 15),
```
ssanderson: I baked in the expected results for RSI because talib is an optional zipline dependency, so I couldn't just call it directly in the test.  For this, we should just compute the expected value directly in the test.

zipline/pipeline/factors/technical.py
```diff
@@ -25,6 +25,22 @@
...
 
 
+class Returns(CustomFactor):
+    """
+    Calculates the % change in close price over the given window_length
+    (i.e. the returns for a given asset).
+
+    **Default Inputs**: [USEquityPricing.close]
+
+    **Default Window Length**: 30
```
ssanderson: I don't think this should have a default window_length.

zipline/pipeline/factors/technical.py
```diff
@@ -25,6 +25,22 @@
 from .factor import CustomFactor
 
 
+class Returns(CustomFactor):
+    """
+    Calculates the % change in close price over the given window_length
+    (i.e. the returns for a given asset).
```
ssanderson: I'd remove the parenthetical note here.

zipline/pipeline/factors/technical.py
```diff
@@ -25,6 +25,22 @@
 from .factor import CustomFactor
 
 
+class Returns(CustomFactor):
+    """
+    Calculates the % change in close price over the given window_length
```
ssanderson: `%` should probably be written out as `percent`.

TimShawver:
@ssanderson thanks for the feedback, I've made those change.  Let me know if there's anything else.

ssanderson:
LGTM

ssanderson:
@TimShawver can you add a whatsnew for this?

TimShawver:
@ssanderson I added a whatsnew and also added our built-in factors to the API docs for zipline.io.  I noticed they were missing when I tried to link to the new Returns factor from the release notes.

Issue #886, opened 2015-12-01
pad error with .mavg() (or history in general?)


twiecki:
I get this when running `olmar.py` with `pandas==0.17.1`
```
Traceback (most recent call last):
  File "olmar.py", line 179, in <module>
    results = olmar.run(data)
  File "/home/wiecki/working/projects/quant/zipline/zipline/algorithm.py", line 592, in run
    for perf in self.gen:
  File "/home/wiecki/working/projects/quant/zipline/zipline/gens/tradesimulation.py", line 121, in transform
    self.algo.instant_fill,
  File "/home/wiecki/working/projects/quant/zipline/zipline/gens/tradesimulation.py", line 298, in _process_snapshot
    new_orders = self._call_handle_data()
  File "/home/wiecki/working/projects/quant/zipline/zipline/gens/tradesimulation.py", line 327, in _call_handle_data
    self.simulation_dt,
  File "/home/wiecki/working/projects/quant/zipline/zipline/utils/events.py", line 209, in handle_data
    context.trading_environment,
  File "/home/wiecki/working/projects/quant/zipline/zipline/utils/events.py", line 228, in handle_data
    self.callback(context, data)
  File "/home/wiecki/working/projects/quant/zipline/zipline/algorithm.py", line 365, in handle_data
    self._handle_data(self, data)
  File "olmar.py", line 60, in handle_data
    x_tilde[i] = data[sid].mavg(algo.window_length) / price
  File "/home/wiecki/working/projects/quant/zipline/zipline/protocol.py", line 459, in mavg
    prices = self._get_buffer(max_bars, raw=True)[-bars:]
  File "/home/wiecki/working/projects/quant/zipline/zipline/protocol.py", line 362, in _get_buffer
    bars, self._freqstr, field, ffill=True,
  File "/home/wiecki/working/projects/quant/zipline/zipline/utils/api_support.py", line 98, in wrapped_method
    return method(self, *args, **kwargs)
  File "/home/wiecki/working/projects/quant/zipline/zipline/algorithm.py", line 1271, in history
    return self.history_container.get_history(history_spec, self.datetime)
  File "/home/wiecki/working/projects/quant/zipline/zipline/history/history_container.py", line 886, in get_history
    digest_frame, index = self.digest_bars(history_spec, do_ffill)
  File "/home/wiecki/working/projects/quant/zipline/zipline/history/history_container.py", line 656, in digest_bars
    raw=True
[...]

PR #887, opened 2015-12-01
MAINT: Alias methods for 0.17 compat.

ssanderson:
https://github.com/quantopian/zipline/issues/886


twiecki:
Confirmed that fix works for my issue. /signoff

PR #890, opened 2015-12-02
BUG: orders future chain by min of notice and expiry

warren-oneill:
Is the chain intentionality ordered by expiry? For me ordering by the min makes more sense.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -725,14 +725,14 @@ class AssetFinderTestCase(TestCase):
             0: {
                 'symbol': 'ADN15',
                 'root_symbol': 'AD',
-                'notice_date': pd.Timestamp('2015-05-14', tz='UTC'),
-                'expiration_date': pd.Timestamp('2015-06-14', tz='UTC'),
+                'notice_date': pd.Timestamp('2015-06-14', tz='UTC'),
+                'expiration_date': pd.Timestamp('2015-08-14', tz='UTC'),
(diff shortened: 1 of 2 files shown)


ssanderson:
@warren-oneill this makes sense to me, but I want to confirm with @jfkirk.  Assuming this is desired behavior, can we slot in a test somewhere in `test_assets` that exercises this behavior?

jfkirk:
Are there situations where ordering by the min would produce a different order than ordering by the expiry? It definitely makes sense to order by the min to me, but I wonder if adding the comparison would be an extra comparison for no impact.

warren-oneill:
@jfkirk  For me yes but it depends if the difference between notice and expiry in your cases is constant of if it can vary between assets.

jfkirk:
OK, then this makes sense to me, but ditto what @ssanderson about a test
@yankees714 does this make sense to you?

yankees714:
Yep, make sense.

@warren-oneill could we update the comment above the `sa.case` statement? It's no longer correct given this change.

warren-oneill:
@jfkirk @ssanderson @yankees714 corrected the comment and adjusted the test so it examines the notification date. Is the test adjustment enough or do I need more?

yankees714:
@warren-oneill all makes sense to me now.

jfkirk:
LGTM - will merge today, if nobody objects

PR #891, opened 2015-12-02
ENH: allows non int orders

warren-oneill:
In the markets where we are active orders with two decimal places are allowed. I tried to add the ability to the TradingAlgorithm class to switch between float and int orders.
I find my implementation a bit messing and am def open to suggestions. Also some tests need to be added.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -199,6 +199,7 @@ class TradingAlgorithm(object):
         self.commission = PerShare()
 
         self.instant_fill = kwargs.pop('instant_fill', False)
+        self.amount_type = kwargs.pop('amount_type', int)
 
         # If an env has been provided, pop it
         self.trading_environment = kwargs.pop('env', None)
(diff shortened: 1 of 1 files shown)


ssanderson:
The implementation here seems reasonable to me.  I have no sense of what exhanges do or don't support this behavior.  I assume this is something you've run into whereever you're actually executing?

I'd definitely want to add tests for this, since I wouldn't at all be surprised if downstream consumers of the order were assuming integral data.  In particular, I'd look at the dtypes of the arrays used for these calculations in PerformanceTracker.  If those are allocated as `int64` arrays, then writing floats into them will just truncate:

```
In [1]: int_array = zeros(5, dtype=int)

In [2]: int_array
Out[2]: array([0, 0, 0, 0, 0])

In [3]: int_array[0] = 2.5

In [4]: int_array
Out[4]: array([2, 0, 0, 0, 0])
```

freddiev4:
Going to close this PR due to inactivity. If you'd like to open a new PR to address this, feel free to go ahead and do that 🙂

PR #892, opened 2015-12-02
REF: Remove data offset calc from clock.

ehebert:
Remove the responsibility of calculating the minute dataset index from
the clock, i.e. the offset that is used for the 'fast' lookup path in
the bcolz data. Instead, the clock calculates all the minutes that the
clock will run through and provides a map of days to the minutes for
each day, and a map of minutes to which day the minute occurs on.

Those two maps are then used by the data_portal to do on demand
calculations of the dataset offset, instead of the clock calculating
the value every minute.

This is on the path to removing all hardcodings of the minute offsets,
by removing the hardcoding from the clock, with the data_portal taking
responsibility for the offsets (work that follows this will remove the
hardcoding from the data_portal.)

### Other Notes

We may want to move the `minutes_by_day` and `minutes_to_day` to the simulation parameters or somewhere higher level.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -147,6 +147,19 @@ class DataPortal(object):
         else:
             self._daily_bar_reader = None
 
+        # The following values are used by _minute_offset to calculate the
+        # index into the minute bcolz date.
+
+        # A lookup of table every minute to the corresponding day, to avoid
+        # calling `.date()` on every lookup.
+        self._minutes_to_day = {}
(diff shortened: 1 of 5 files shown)


zipline/data/data_portal.py
```diff
@@ -147,6 +147,10 @@ def __init__(self,
         else:
             self._daily_bar_reader = None
 
+        self._minutes_to_day = {}
```
jbredeche: can you add some comments here explaining what these caches are for?

jbredeche:
this is a really awesome improvement, thanks!

PR #893, opened 2015-12-02
Allow different data set starts for minute data portal

ehebert:
Instead of using a hardcoded start date for the calculations of the
index into the equity minute data, use a metadata file written by the
equity minute writer to set the first trading day of the equity minute
set.

To allow the data portal to be used with datasets with
different ranges. (i.e. not just the start date of 2002-01-02)

Also includes a change that requires both a futures and a equities path to the data portal, since the dataset start is equities specific.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -245,12 +245,12 @@ class HistoryTestCase(TestCase):
                                    "DIVIDEND_minute.csv.gz"),
         }
 
-        path = os.path.join(tempdir.path, 'equity', 'minutes')
-        os.makedirs(path)
+        equities_tempdir = os.path.join(tempdir.path, 'equity', 'minutes')
+        os.makedirs(equities_tempdir)
 
         MinuteBarWriterFromCSVs(resources,
                                 pd.Timestamp('2002-01-02', tz='UTC')).write(
-                                    path, cls.assets)
+                                    equities_tempdir, cls.assets)
 
     @classmethod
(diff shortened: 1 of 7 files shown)


tests/test_history.py
```diff
@@ -241,7 +245,12 @@ def create_fake_minute_data(cls, tempdir):
                                    "DIVIDEND_minute.csv.gz"),
         }
 
-        MinuteBarWriterFromCSVs(resources).write(tempdir.path, cls.assets)
+        path = os.path.join(tempdir.path, 'equity', 'minutes')
```
jbredeche: worth calling this `equities_tempdir` to be consistent with `futures_tempdir` above?

zipline/data/data_portal.py
```diff
@@ -157,6 +158,13 @@ def __init__(self,
         # days data starts.
         self._day_offsets = None
 
+    @property
+    def minute_bar_reader(self):
```
jbredeche: should we keep the reader internal?  there are no consumers of self.`_minute_bar_reader` outside of the dataportal.

ehebert: The reason I hadn't it is because it is lazily created (i.e. not created until first use, because some tests/use cases were not writing minute data), and the `_minute_bar_reader` is used as the placeholder/None value.

ehebert:
Will merge when internal compatibility PR is done.

PR #896, opened 2015-12-04
REF: Pass minute reader instead of path to portal

ehebert:
Start down path of abstracting out the reader, so that other readers,
e.g. a DataFrame can be implemented. (Those implementations would
require more of the read logic to be moved from the DataPortal to the
reader.)

This is also required for an internal override of the start date on the
bcolz reader, which is why this change is so minimal.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -27,7 +27,6 @@ from zipline.data.us_equity_pricing import (
     BcolzDailyBarReader,
     NoDataOnDate
 )
-from zipline.data.us_equity_minutes import BcolzMinuteBarReader
 from zipline.pipeline.data.equity_pricing import USEquityPricing
 
 from zipline.utils import tradingcalendar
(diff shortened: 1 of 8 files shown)


jbredeche:
I think there are some flake8 issues.

Besides that, do you envision separate readers for MinuteEquity and MinuteFutures?

zipline/data/data_portal.py
```diff
@@ -262,15 +251,17 @@ def _get_ctable(self, asset):
...
+                    self._equity_minute_reader.rootdir, sid
                 )
             else:
-                path = "{0}/{1}.bcolz".format(self._minutes_equities_path, sid)
+                path = "{0}/{1}.bcolz".format(
+                    self._equity_minute_reader.rootdir, sid)
 
         else:
-            path = "{0}/{1}.bcolz".format(self._minutes_equities_path, sid)
+            path = "{0}/{1}.bcolz".format(
```
richafrank: This may be a silly question, but if the data portal is now passed a bcolz reader, why does it also need to understand bcolz and where to read from?  I would think that's abstracted away by the reader.

ehebert:
Will take care of flake8

And yes, do envision a futures reader. Planning on wiring that in while
testing equities deployment.

zipline/data/data_portal.py
ehebert: Not silly at all. It is the intended next step. This is just minimum scope
for testing the passing of the new parameter.

richafrank: Gotcha, makes sense!

PR #901, opened 2015-12-07
Lazy CPU and Memory fix

ehebert:
The main fix included fixes a regression on algorithms that only use a few equities:

    PERF: Fix runtime regression on daily windows.

    Re-using the pipeline infrastructure resulting in a bottleneck when
    using a small number of stocks, where the read of the entire numpy
    vector when only one stock was used was prohibitive.

    Instead, base a solution off of @jbredeche's original solution, but
    place the logic (at least the indexing logic into the DailyBcolzReader.)

    We have an eye on making this eventually share code with pipeline, but
    using the simpler non-shared solution for now.

Also, includes a small fix to a possible memory consumption issue when using spot_value discovered when investigating whether or not a cache of the numpy vector could be used in the bcolz daily bar reader.


ehebert:
Closing, pointed this at wrong branch.

PR #902, opened 2015-12-07
Lazy daily fix

ehebert:
The main fix included fixes a regression on algorithms that only use a few equities:

PERF: Fix runtime regression on daily windows.

Re-using the pipeline infrastructure resulting in a bottleneck when
using a small number of stocks, where the read of the entire numpy
vector when only one stock was used was prohibitive.

Instead, base a solution off of @jbredeche's original solution, but
place the logic (at least the indexing logic into the DailyBcolzReader.)

We have an eye on making this eventually share code with pipeline, but
using the simpler non-shared solution for now.
Also, includes a small fix to a possible memory consumption issue when using spot_value discovered when investigating whether or not a cache of the numpy vector could be used in the bcolz daily bar reader.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -1052,14 +1052,7 @@ class DataPortal(object):
         nan.
 
         """
-        column = US_EQUITY_COLUMNS[field]
-        data = self._daily_bar_reader.load_raw_arrays([column],
-                                                      days_in_window[0],
-                                                      days_in_window[-1],
-                                                      [asset])
-
         bar_count = len(days_in_window)
-
         # create an np.array of size bar_count
         if extra_slot:
             return_array = np.zeros((bar_count + 1,))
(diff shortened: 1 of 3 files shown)


zipline/data/data_portal.py
```diff
@@ -1067,16 +1060,25 @@ def _get_daily_window_for_sid(self, asset, field, days_in_window,
...
+        end_date = self._get_asset_end_date(asset)
+        day_slice = days_in_window.slice_indexer(start_date, end_date)
+        active_days = days_in_window[day_slice]
+
+        if active_days.shape[0]:
+            data = self._daily_bar_reader.history_window(field,
+                                                         active_days[0],
+                                                         active_days[-1],
+                                                         asset)
+            sid = int(asset)
```
jbredeche: I think `apply_all_adjustments` takes an Asset, no need to `int` here.

ehebert: Ok, can remove. For what it is worth, at least in test history, the values passed down are integers. We should probably be using an Asset as a parameter to history in the tests, correct?

PR #904, opened 2015-12-08
REF: some refactoring to support Quantopian-specific functionality.

jbredeche:


zipline/_protocol.pxd
```diff
@@ -0,0 +1,11 @@
+cdef class BarData:
```
ehebert: Copyright notice/license.

zipline/_protocol.pyx
```diff
@@ -49,12 +45,18 @@ cdef class BarData:
...
             view = self._views[asset]
         except KeyError:
-            view = self._views[asset] = \
-                SidView(asset, self.data_portal,
-                        self.simulator, self.data_frequency)
+            view = self._views[asset] = self._create_sid_view(asset)
 
         return view
 
+    def _create_sid_view(self, asset):
```
ehebert: Maybe add a comment how about this a 'hook' into which a custom sid view can be implemented?

zipline/gens/tradesimulation.py
```diff
@@ -78,6 +78,12 @@ def inject_algo_dt(record):
                 record.extra['algo_dt'] = self.simulation_dt
         self.processor = Processor(inject_algo_dt)
 
+    def _create_bar_data(self):
```
ehebert: Likewise, with this and the other `_create_bar_data` methods (with regard to adding a comment.)

ehebert:
LGTM, an alternative solution would be to make a `sid_view_cls` (and corresponding `_cls` methods for the other classes.) and do `self.sid_view_cls(asset)` instead of having the _create_sid_view class that needs to be implemented by the subclass.

No real opinion, at this moment, of which is clearer.

PR #905, opened 2015-12-08
Adds support for different typed adjusted arrays and adds an EarningsCalendar loader

llllllllll:
- Moves most of AdjustedArray back into Python. The window iterator is
  the only part that's performance-intensive.

- Adds a bootleg templating system for creating specialized versions of
  AdjustedArrayWindow for each concrete type we care about.

- Adds support for differently dtyped terms in pipeline. This allows us
  to use datetime64s which are needed in the EarningsCalendar.

- Adds EarningsCalendar dataset for the next and previous earnings
  announcements in pipeline.

- Adds in memory loader for EarningsCalendar.

- Adds blaze loader for EarningsCalendar.


zipline/pipeline/loaders/blaze/earnings.py
```diff
@@ -0,0 +1,138 @@
...
+    def load_adjusted_array(self, columns, dates, assets, mask):
+        expr = self._expr
+        filtered = expr[expr[TS_FIELD_NAME] <= dates[0]]
+        lower = odo(
+            bz.by(
+                filtered[SID_FIELD_NAME],
+                timestamp=filtered[TS_FIELD_NAME].max(),
+            ).timestamp.min(),
+            pd.Timestamp,
+            **self._odo_kwargs or {}
```
ssanderson: I don't think we need this check here.  We default to an empty dict in the constructor.

zipline/pipeline/loaders/blaze/earnings.py
```diff
@@ -0,0 +1,138 @@
...
+            # range. It must all be null anyways.
+            lower = dates[0]
+
+        raw = odo(
+            expr[
+                (expr[TS_FIELD_NAME] >= lower) &
+                (expr[TS_FIELD_NAME] <= dates[-1])
+            ],
+            pd.DataFrame,
+            **self._odo_kwargs or {}
```
ssanderson: Same note as above.

zipline/pipeline/loaders/blaze/earnings.py
```diff
@@ -0,0 +1,140 @@
...
+
+        sids = raw.loc[:, SID_FIELD_NAME]
+        raw.drop(
+            sids[~(sids.isin(assets) | sids.notnull())].index,
+            inplace=True
+        )
+
+        gb = raw.groupby(SID_FIELD_NAME)
+        if self._has_ts:
+            def mkseries(idx, raw_loc=raw.loc):
```
ssanderson: Is this just caching in the locals for performance?

zipline/pipeline/loaders/blaze/earnings.py
```diff
@@ -0,0 +1,140 @@
...
+        if not istabular(dshape):
+            raise ValueError(
+                'expression dshape must be tabular, got: %s' % dshape,
+            )
+
+        expected_fields = self._expected_fields
+        self._has_ts = has_ts = TS_FIELD_NAME in dshape.measure.dict
+        if not has_ts:
+            # This field is optional.
+            expected_fields - {TS_FIELD_NAME}
```
ssanderson: I don't think this does anything...

zipline/pipeline/loaders/blaze/earnings.py
llllllllll: yeah, the function is only consumed within this scope so I figure this is safe.

tests/pipeline/test_numerical_expression.py
```diff
@@ -128,6 +143,64 @@ def test_validate_bad(self):
         with self.assertRaises(TypeError):
             (f > f) > f
 
+    def test_combine_datetimes(self):
+        with self.assertRaises(TypeError) as e:
+            self.d + self.d
+        message = e.exception.args[0]
+        expected = (
+            "Don't know how to compute datetime64[ns] + datetime64[ns].\n"
+            "Arithmetic operators are only supported on Factors of dtype "
```
llllllllll: Should this say `addition` instead. Subtraction is supported, no?

llllllllll: also, is there a test for `datetime - datetime :: timedelta` or is this not yet supported?

zipline/lib/_windowtemplate.pxi
```diff
@@ -0,0 +1,103 @@
...
+
+        self.data = data
+        self.viewtype = viewtype
+        self.adjustments = adjustments
+        self.adjustment_indices = sorted(adjustments, reverse=True)
+        self.window_length = window_length
+        self.anchor = window_length + offset
+        self.max_anchor = data.shape[0]
+
+        if len(self.adjustment_indices) > 0:
```
llllllllll: this is duplicated in the `__next__`. Can we pull this out into a `next_adj` func?

tests/pipeline/test_numerical_expression.py
ssanderson: It's not supported yet, mostly because I couldn't think of a real use-case for it.

ssanderson: (The error message is currently an accurate reflection of the actual behavior.)

llllllllll: would you want to do something like:

```python
mask = EarningsCalendar.next_earnings.latest() - today < timedelta(5)
```

ssanderson: You probably don't want calendar date arithmetic there though.  I added BusinessDaysUntil{Next,Previous}Earnings for that use case.

ssanderson: And I'm not sure yet how we want to do `today` as a symbolic object.

zipline/lib/adjustment.pyx
```diff
@@ -1,12 +1,116 @@
...
+from pandas import isnull, Timestamp
+from numpy cimport float64_t, uint8_t, int64_t
+from numpy import datetime64, float64
 # Purely for readability. There aren't C-level declarations for these types.
 ctypedef object Int64Index_t
 ctypedef object DatetimeIndex_t
 ctypedef object Timestamp_t
 
+# Adjustment kinds.
+MULTIPLY = 0
```
llllllllll: Could we define these with `cpdef enum AdjustmentKind:` for clarity. We can then change the usages of `int adjustment_kind` to `AdjustmentKind adjustment_kind`. This compiles the same but ls more clear in the source.

zipline/lib/adjustment.pyx
```diff
@@ -1,12 +1,116 @@
...
+                                  Timestamp_t end_date,
+                                  int asset_id,
+                                  int adjustment_kind,
+                                  object value):
+    """
+    Make an Adjustment object from date/asset labels into a labelled baseline
+    array.
+    """
+    cdef type type_ = choose_adjustment_type(adjustment_kind, value)
+    return type_.from_assets_and_dates(
```
llllllllll: do we have the same issue with kwargs here?

ssanderson: Nope.  Yay Cython!

llllllllll: wow

zipline/pipeline/factors/events.py
```diff
@@ -0,0 +1,90 @@
...
+
+    See Also
+    --------
+    BusinessDaysSincePreviousEarnings
+    """
+    inputs = [EarningsCalendar.next_announcement]
+    window_length = 0
+    dtype = float64_dtype
+
+    def _compute(self, arrays, dates, assets, mask):
```
llllllllll: since this is the same func as prev but with the operatands flipped, do you think it makes sense to share the setup code path? They are pretty short so idk if it is worth.

zipline/pipeline/factors/factor.py
```diff
@@ -33,17 +28,67 @@
...
 
+def numbers_to_float64(func, argname, argvalue):
+    """
+    Preprocessor for converting numerical inputs into floats.
+
+    This is used in the binary operator constructors for Factor so that
+    `2 + Factor()` has the same behavior as `2.0 + Factor()`.
+    """
+    if isinstance(argvalue, Number):
+        return float64(argvalue)
```
llllllllll: should we warn in the case of a possible loss of precisio? For example `argvalue=Fraction(1, 3)`

zipline/pipeline/term.py
```diff
@@ -231,7 +274,8 @@ def _compute(self, windows, dates, assets, mask):
         """
         # TODO: Make mask available to user's `compute`.
         compute = self.compute
-        out = full(mask.shape, nan, dtype=self.dtype)
+        missing_value = self.missing_value
+        out = full(mask.shape, missing_value, dtype=self.dtype)
```
llllllllll: `full_like(mask, ...)`

zipline/utils/input_validation.py
```diff
@@ -30,6 +32,98 @@ def ensure_upper_case(func, argname, arg):
...
+    ...
+    >>> foo(arange(3), 'foo')
+    (array([0, 1, 2]), 'foo')
+    >>> foo(arange(3, dtype=float), 'foo')
+    Traceback (most recent call last):
+       ...
+    TypeError: foo() expected an argument with dtype 'int64' for argument 'x', but got dtype 'float64' instead.  # noqa
+    """
+    if _pos:
+        raise TypeError("expect_dtypes() only takes keyword arguments.")
```
llllllllll: why do we accept *args just to raise? This will raise a type error without accepting `_pos`.

zipline/utils/input_validation.py
```diff
@@ -30,6 +32,98 @@ def ensure_upper_case(func, argname, arg):
...
+        "for argument '%(argname)s', but got %(actual)r instead."
+    ).format(dtype_str=' or '.join(repr(d.name) for d in allowed_dtypes))
+
+    def check_dtype(value):
+        return getattr(value, 'dtype', None) not in allowed_dtypes
+
+    def display_bad_value(value):
+        # If the bad value has a dtype, but it's wrong, show the dtype name.
+        if hasattr(value, 'dtype'):
+            return value.dtype.name
```
llllllllll: there are some cases where this will raise an attribute error, maybe check `hasattr(value, 'dtype') and hasattr(value.dtype, 'name')`

zipline/utils/numpy_utils.py
```diff
@@ -1,7 +1,40 @@
...
+    float64_dtype: nan,
+    datetime64ns_dtype: np_NaT,
+}
+
+
+def default_fillvalue_for_dtype(dtype):
+    """
+    Get the default fill value for `dtype`.
+    """
+    return _FILLVALUE_DEFAULTS[dtype]
```
llllllllll: why don't consumers just use this dict directly?

zipline/utils/test_utils.py
```diff
@@ -603,3 +605,28 @@ def wrapped(*args, **kwargs):
 
         return wrapped
     return dec
+
+
+def assert_timestamp_equal(left, right, compare_nat_equal=True, msg=""):
```
llllllllll: this is no longer used

zipline/pipeline/factors/events.py
ssanderson: I thought about that while writing this and ended up just pulling out the functionality that's in `busday_count_mask_NaT`.  I kind of want to use these as examples of how to write factors that work with dates, so I think it's ok if they're a bit verbose/explicit.

zipline/pipeline/factors/factor.py
ssanderson: If numpy doesn't do this, then I'm not sure we should be in the business of doing it either.

ssanderson:
```
In [2]: float64(Fraction(1, 3))
Out[2]: 0.33333333333333331
```

llllllllll: Okay, just a thought. Also, numpy should totally warn about this.

llllllllll:
other than the small comments your changes look good to merge.

zipline/utils/numpy_utils.py
ssanderson: I can imagine the implementation becoming more complex in the future.  The fact that it's just a simple dict lookup right now is an implementation detail.

zipline/utils/input_validation.py
ssanderson: This is how the other `expect_*` functions work as well.  I did this to provide a clearer error message I think, since I kept writing them passing positionally and getting confused when it didn't work.

I could see the argument for removing the *pos, but I don't think that that change belongs here.

llllllllll: Okay, I just think that it is not generally good practice to alias standard language errors. We can make this change is a different pr.

ssanderson:
This is ready to merge here.  Waiting on downstream compat changes.

PR #906, opened 2015-12-09
Initial data portal commit.

ehebert:
ENH: Add initial commit for DataPortal and readers

    Moved from the `lazy-mainline` branch,
    https://github.com/quantopian/zipline/pull/858

    With the removal of methods that are not covered by the test_dataportal
    suite.

    The intent of this patch to provide the basic class and readers
    developed on that branch, so that the use of creating the object and
    opening paths etc. can be tested internally.

    Important to note that the only usage is in the test_dataportal suite,
    before any usage of the data_portal by the tradesimulation, a writer
    will should be provided for publicly available data.

# Extra notes

If I may, I ask for the bulk of review to be on the structure/composition of the class. There is a plan to refactor more responsibility and logic into the readers. The focus of this branch and the larger lazy-mainline is to get over the hump of accessing the data lazily throughout the system.

(e.g. the calculation of the index into the  minutes. The lazy-mainline branch currently provides some lookup tables for minute -> day -> index, though we may want to provide that lookup via a DatetimeIndex of all minutes, with changing the minute bar reader to provide that DattimeIndex.)

Much of the initial patch is with @jbredeche on this branch https://github.com/quantopian/zipline/tree/lazy-mainline

### DIFF
--- a/tests/test_dataportal.py
+++ b/tests/test_dataportal.py
@@ -28,6 +28,7 @@ from zipline.data.us_equity_pricing import (
     BcolzDailyBarReader
 )
 from zipline.finance.trading import TradingEnvironment, SimulationParameters
+from zipline.data.future_pricing import FutureMinuteReader
 from zipline.data.us_equity_minutes import (
     MinuteBarWriterFromDataFrames,
     BcolzMinuteBarReader
@@ -343,9 +344,11 @@ class TestDataPortal(TestCase):
                 }
             })
 
+            future_minute_reader = FutureMinuteReader(tempdir.path)
+
             dp = DataPortal(
                 env,
-                minutes_futures_path=tempdir.path
+                future_minute_reader=future_minute_reader
             )
 
             future123 = env.asset_finder.retrieve_asset(123)
--- a/tests/test_dataportal.py
+++ b/tests/test_dataportal.py
@@ -22,8 +22,11 @@ from unittest import TestCase
 from pandas.tslib import normalize_date
 from testfixtures import TempDirectory
 from zipline.data.data_portal import DataPortal
-from zipline.data.us_equity_pricing import SQLiteAdjustmentWriter, \
-    SQLiteAdjustmentReader
+from zipline.data.us_equity_pricing import (
+    SQLiteAdjustmentWriter,
+    SQLiteAdjustmentReader,
+    BcolzDailyBarReader
+)
 from zipline.finance.trading import TradingEnvironment, SimulationParameters
 from zipline.data.us_equity_minutes import (
     MinuteBarWriterFromDataFrames,
@@ -145,9 +148,11 @@ class TestDataPortal(TestCase):
                 data_frequency="daily"
             )
 
+            equity_daily_reader = BcolzDailyBarReader(path)
+
             dp = DataPortal(
                 env,
-                daily_equities_path=path,
+                equity_daily_reader=equity_daily_reader,
                 sim_params=sim_params
             )
 
--- a/tests/test_dataportal.py
+++ b/tests/test_dataportal.py
@@ -1,0 +1,363 @@
+# Copyright 2015 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from datetime import timedelta
+import bcolz
+import numpy as np
+import pandas as pd
+
+from unittest import TestCase
+from pandas.tslib import normalize_date
+from testfixtures import TempDirectory
+from zipline.data.data_portal import DataPortal
+from zipline.data.us_equity_pricing import SQLiteAdjustmentWriter, \
+    SQLiteAdjustmentReader
+from zipline.finance.trading import TradingEnvironment, SimulationParameters
+from zipline.data.us_equity_minutes import (
+    MinuteBarWriterFromDataFrames,
+    BcolzMinuteBarReader
+)
+from .utils.daily_bar_writer import DailyBarWriterFromDataFrames
+
+
+class TestDataPortal(TestCase):
+    def test_forward_fill_minute(self):
+        tempdir = TempDirectory()
+        try:
+            env = TradingEnvironment()
+            env.write_data(
+                equities_data={
+                    0: {
+                        'start_date': pd.Timestamp("2015-09-28", tz='UTC'),
+                        'end_date': pd.Timestamp("2015-09-29", tz='UTC')
+                        + timedelta(days=1)
+                    }
+                }
+            )
+
+            minutes = env.minutes_for_days_in_range(
+                start=pd.Timestamp("2015-09-28", tz='UTC'),
+                end=pd.Timestamp("2015-09-29", tz='UTC')
+            )
+
+            df = pd.DataFrame({
+                # one missing bar, then 200 bars of real data,
+                # then 1.5 days of missing data
+                "open": np.array([0] + list(range(0, 200)) + [0] * 579)
+                * 1000,
+                "high": np.array([0] + list(range(1000, 1200)) + [0] * 579)
+                * 1000,
+                "low": np.array([0] + list(range(2000, 2200)) + [0] * 579)
+                * 1000,
+                "close": np.array([0] + list(range(3000, 3200)) + [0] * 579)
+                * 1000,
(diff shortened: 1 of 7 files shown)


zipline/data/us_equity_pricing.py
```diff
@@ -485,7 +497,11 @@ def sid_day_index(self, sid, day):
             Raises a NoDataOnDate exception if the given day and sid is before
             or after the date range of the equity.
         """
-        day_loc = self._calendar.get_loc(day)
+        try:
+            day_loc = self._calendar.get_loc(day)
+        except:
```
ssanderson: This probably wants to be `except KeyError:`?

ehebert: Yes, fixing up.

tests/test_dataportal.py
```diff
@@ -0,0 +1,371 @@
...
+                data_frequency="minute",
+                env=env,
+            )
+
+            equity_minute_reader = BcolzMinuteBarReader(tempdir.path)
+
+            dp = DataPortal(
+                env,
+                equity_minute_reader=equity_minute_reader,
+                sim_params=sim_params
```
ssanderson: It's very surprising to me that the DataPortal requires a SimulationParameters.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,475 @@
...
+        # Cache of sid -> the first trading day of an asset, even if that day
+        # is before 1/2/2002.
+        self._asset_start_dates = {}
+        self._asset_end_dates = {}
+
+        # Handle extra sources, like Fetcher.
+        self._augmented_sources_map = {}
+        self._extra_source_df = None
+
+        self._sim_params = sim_params
```
ssanderson: I don't think this needs to be an instance attribute.  It's never used for anything except getting the data frequency.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,475 @@
...
+        Returns
+        -------
+        The value of the desired field at the desired time.
+        """
+        if field not in BASE_FIELDS:
+            raise KeyError("Invalid column: " + str(field))
+
+        column_to_use = BASE_FIELDS[field]
+
+        if isinstance(asset, int):
```
ssanderson: I'm concerned about the fact that this and all of the other methods of this class can't make up their minds about whether they expect to be passed `sid`s or `Asset` objects.  Is there a reason consumers of this API can't always just provide Asset objects?

zipline/data/data_portal.py
```diff
@@ -0,0 +1,475 @@
...
+    def get_spot_value(self, asset, field, dt=None):
+        """
+        Public API method that returns a scalar value representing the value
+        of the desired asset's field at either the given dt, or this data
+        portal's current_dt.
+
+        Parameters
+        ---------
+        asset : Asset
+            The asset whose data is desired.gith
```
ssanderson: Cruft.  `gith`

zipline/data/data_portal.py
```diff
@@ -0,0 +1,475 @@
...
+
+        field: string
+            The desired field of the asset.  Valid values are "open",
+            "open_price", "high", "low", "close", "close_price", "volume", and
+            "price".
+
+        dt: pd.Timestamp
```
ssanderson: Why does this need to be optional?  Is there ever a point in the simulation where this is being called from a context where we don't know our current time?

zipline/data/data_portal.py
jbredeche: Hard to believe that the methods can't make up their minds, when the docstrings all clearly (afaik) state what they take.

This was just for backwards compatibility for some tests, which we'll update to pass in assets.

tests/test_dataportal.py
ehebert: The main reason it is passed is to get the `data_frequency` of the DataPortal, which determines the source for the data returned by `get_spot_value`.

`get_spot_value` is used on the lazy-mainline branch to pull the current price for the blotter, performance_tracker, and performance_period. That concept of current price is tied to whether the simulation is daily or minute.

Any suggestions for how to model that better?

zipline/data/data_portal.py
ehebert: Can remove.

zipline/data/data_portal.py
ehebert: I found that difficult to work with in places as well, I'm going to assume it was because some where where it was used on lazy-mainline only the integer was available in one of the places where spot_value was needed. (Though we can change it so that the blotter or tracker, or wherever this is called does the Asset conversion first).

zipline/data/data_portal.py
ehebert: https://github.com/quantopian/zipline/pull/858/files#r47156458

An alternative may be updating the current time in the sid data view/proxy.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,474 @@
...
+
+    def _get_asset_start_date(self, asset):
+        self._ensure_asset_dates(asset)
+        return self._asset_start_dates[asset]
+
+    def _get_asset_end_date(self, asset):
+        self._ensure_asset_dates(asset)
+        return self._asset_end_dates[asset]
+
+    def _ensure_asset_dates(self, asset):
```
ssanderson: The `AssetFinder` class has built-in functionality for efficiently getting the start dates of every asset we know about.  I'd suggest grabbing that information all at once rather than having to constantly chain through these methods.

zipline/data/data_portal.py
```diff
@@ -0,0 +1,474 @@
...
+                result = carray[minute_offset_to_use]
+
+            # once we've found data, we need to check whether it needs
+            # to be adjusted.
+            if result != 0:
+                minutes = self.env.market_minute_window(
+                    start=(dt or self.current_dt),
+                    count=(original_start - minute_offset_to_use + 1),
+                    step=-1
+                ).order()
```
ssanderson: What does `.order()` do here?

zipline/data/us_equity_minutes.py
```diff
@@ -0,0 +1,316 @@
...
+import pandas as pd
+from pandas import read_csv
+from six import with_metaclass
+
+from zipline.finance.trading import TradingEnvironment
+from zipline.utils import tradingcalendar
+
+MINUTES_PER_DAY = 390
+
+_writer_env = TradingEnvironment()
```
ssanderson: This should probably be a parameter to instances of the writer, rather than a global constant.  @jfkirk recently did a lot of work to remove global TradingEnvironment instances because of all the headaches they caused.

zipline/data/us_equity_minutes.py
```diff
@@ -0,0 +1,316 @@
...
+        raise NotImplementedError()
+
+    def write(self, directory, assets, sid_path_func=None):
+        _iterator = self.gen_frames(assets)
+
+        return self._write_internal(directory, _iterator,
+                                    sid_path_func=sid_path_func)
+
+    @staticmethod
+    def full_minutes_for_days(env, dt1, dt2):
```
ssanderson: This function appears to be assuming the NYSE calendar.

zipline/data/us_equity_minutes.py
```diff
@@ -0,0 +1,316 @@
...
+
+            yield asset, df
+
+
+class BcolzMinuteBarReader(object):
+
+    def __init__(self, rootdir, sid_path_func=None):
+        self.rootdir = rootdir
+
+        metadata = self._get_metadata()
```
ssanderson: The `write_metadata` counterpart to this is a free function at module scope, but this is an instancemethod.  I'd suggest moving this to module scope as well and just passing `self.rootdir`.

tests/test_dataportal.py
ssanderson: > The main reason it is passed is to get the data_frequency of the DataPortal, which determines the source for the data returned by get_spot_value.

If all we need is the data frequency, then that's all we should pass.  There's a whole bunch of other stuff in the SimulationParameters that the dataportal doesn't (and shouldn't) care about.  That said...

> get_spot_value is used on the lazy-mainline branch to pull the current price for the blotter, performance_tracker, and performance_period. That concept of current price is tied to whether the simulation is daily or minute.

> Any suggestions for how to model that better?

I'd expect that consumers of the API supply their desired frequency when they call `get_spot_price`.  The single thing that I'm most happy with in all the Pipeline infrastructure is the fact that nothing in Pipeline knows about or cares about the fact that we're running in the context of a simulation.  That's paid off enormously in making it easy to develop and test changes quickly.

zipline/data/data_portal.py
ssanderson: > An alternative may be updating the current time in the sid data view/proxy.

That seems like it'd be cleaner.  IMO the DataPortal shouldn't know anything about the context of the running simulation.

tests/test_dataportal.py
ehebert: I agree.

Addressed the passing of the dt in lazy-mainline https://github.com/quantopian/zipline/pull/907

Going to work on requiring the data_frequency from there.

ehebert:
Closing in favor of https://github.com/quantopian/zipline/pull/912

PR #907, opened 2015-12-10
REF: Make dt required for spot value.

ehebert:
Make interface for get_spot_value more more precise by always requiring
the dt parameter, no longer sometimes using a current_dt value if set.

Still relies on a current_dt to be updated on data_portal from
tradesimulation, but that value is now accessed and passed more
explicitly by the bar data views.

### DIFF
--- a/zipline/_protocol.pyx
+++ b/zipline/_protocol.pyx
@@ -1,0 +1,91 @@
+#
+# Copyright 2015 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


zipline/_protocol.pyx
```diff
@@ -0,0 +1,91 @@
...
+        except KeyError:
+            view = self._views[asset] = \
+                SidView(asset, self.data_portal, self)
+
+        return view
+
+    def __getitem__(self, name):
+        return self._get_equity_price_view(name)
+
+    def __iter__(self):
```
ssanderson: This is just re-creating the behavior that Python would provide if we didn't implement the method at all.

PR #912, opened 2015-12-14
ENH: Add initial commit for DataPortal and readers

ehebert:
Moved from the `lazy-mainline` branch,
https://github.com/quantopian/zipline/pull/858

The intent of this patch to provide the basic class and readers
interfaces, developed on that branch, so that the use of creating the
object and opening paths etc. can be tested internally.

Additional changes beyond the lazy-mainline branch, addition of future
minute reader, and daily bar reader.

Also allow an argument of the future_daily_reader, though no such reader
yet exists.

It may be that future and equity readers share an interface, and a
further improvement would be providing an abstract base class.

co-author: @jbredeche <jean@quantopian.com>

## Notes

Supercedes https://github.com/quantopian/zipline/pull/906

Differences from #906 

- `get_spot_value` interfaces were changed, on @ssanderson recommendation. Those implementations have been changed on the lazy-mainline branch.
- The implementations in DailyBarReader have been changed to `NotImplementedError` to 1) make clear that the methods are not yet supported/used, and 2) make the diffs/merge resolutions more clear between master and lazy-mainline (in case changes are made to the lazy-mainline branch between now and the full merge.)

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -1,0 +1,190 @@
+#
+# Copyright 2015 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


jbredeche:
lgtm.  let's get this in.

PR #914, opened 2015-12-15
Move order and transaction classes

ehebert:
To reduce the edit distance of https://github.com/quantopian/zipline/pull/858 split out transactions and orders as was done on that branch.

The change cleans up the imports in some case, e.g. the position_tracker does not use slippage, but needs the transaction, so `from zipline.finance.transaction import Transaction` seems more semantically correct.

Also, having separate modules was found to have benefits for development iteration on each functionality.

### DIFF
--- a/zipline/finance/order.py
+++ b/zipline/finance/order.py
@@ -1,0 +1,194 @@
+#
+# Copyright 2015 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from copy import copy
+import math
+import uuid
+
+from six import text_type, iteritems
+
+from zipline.finance.slippage import check_order_triggers
+import zipline.protocol as zp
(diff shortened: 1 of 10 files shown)


ehebert:
cc: @jbredeche

ehebert:
Also, worth mentioning, is that patch should have no functional change. Was just a copy/paste of existing files to new files, changing import paths etc. So if there is a functional change noticed, that would be unintended.

zipline/finance/order.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def __getstate__(self):
+
+        state_dict = \
+            {k: v for k, v in iteritems(self.__dict__)
+                if not k.startswith('_')}
+
+        state_dict['_status'] = self._status
+
+        STATE_VERSION = 1
```
richafrank: Does this version need to be incremented with the module change?  (I was also thinking it would probably be easier to find as a class attribute.)

zipline/finance/transaction.py
```diff
@@ -0,0 +1,82 @@
...
+    def to_dict(self):
+        py = copy(self.__dict__)
+        del py['type']
+        return py
+
+    def __getstate__(self):
+
+        state_dict = copy(self.__dict__)
+
+        STATE_VERSION = 1
```
richafrank: Same question about versioning.

zipline/finance/order.py
ehebert: We did not update on lazy-mainline and has worked so far. I think since the internals of the class are the same, the same STATE_VERSION can be used.

ehebert: @richafrank here is the skip we discussed about the version tests not being run, https://github.com/quantopian/zipline/commit/82affb639f24bbd8a9509c33c130088798f2b1b7

richafrank: Awesome, thanks.  Looks good to me then.

PR #915, opened 2015-12-15
MAINT: Remove perf_periods member.

ehebert:
Refer to cumulative and todays performance explicitly instead of always
looping through.

The third value (minute) for which this was useful, has been removed.

Also, there are some actions where only cumulative may need application,
e.g. application of dividends. (However, this patch does not remove
dividend processing from todays performance, but opens up later patches
to make that distinction.)

Other Notes:

For compatibility with developments on the Q2.0 branch, https://github.com/quantopian/zipline/pull/858/files

Originally done here, https://github.com/quantopian/zipline/commit/db425553f5f60ce6e5b9a07db2afab7d0e5929bd


llllllllll:
Looks good, thanks!

PR #917, opened 2015-12-16
MAINT: Pass leverage instead of account to risk.

ehebert:
The only value used in the account is leverage, so pass the leverage
value directly.

Also, remove account from risk init, since it is not used.

### DIFF
--- a/zipline/finance/risk/cumulative.py
+++ b/zipline/finance/risk/cumulative.py
@@ -91,8 +91,7 @@ class RiskMetricsCumulative(object):
     )
 
     def __init__(self, sim_params, env,
-                 create_first_day_stats=False,
-                 account=None):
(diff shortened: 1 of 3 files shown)


zipline/finance/risk/cumulative.py
```diff
@@ -91,8 +91,7 @@ class RiskMetricsCumulative(object):
     )
 
     def __init__(self, sim_params, env,
-                 create_first_day_stats=False,
-                 account=None):
```
llllllllll: was this just unused?

llllllllll:
Other than the one question lgtm

zipline/finance/risk/cumulative.py
ehebert: Yes.

PR #919, opened 2015-12-17
Make commission and split handling compatible with event removal

ehebert:
Change the method signatures for split and commission handling so that the sid and corresponding values are accepted instead of an event from which those values are accessed.

In preparation of lazy-mainline branch which removes the event and passes the coarse values down.

Based on 2b02aab120acf6429dae419bc12620ad1923b9ec and fc60804ff15865c772461fc847182ef8d5b31d4c from lazy-mainline branch.


llllllllll:
Looks good, ty

PR #920, opened 2015-12-17
MAINT: Make tracker stats a method.

ehebert:
Instead of calling a function, where the only parameter is the tracker
object, make it a method, so that the snapshot of position tracker stats
can be more easily called as `pt.stats()`.

### DIFF
--- a/zipline/finance/performance/position_tracker.py
+++ b/zipline/finance/performance/position_tracker.py
@@ -121,53 +121,6 @@ def calc_gross_value(long_value, short_value):
     return long_value + abs(short_value)
 
 
-def calc_position_stats(pt):
-    amounts = []
-    last_sale_prices = []
-    for pos in itervalues(pt.positions):
-        amounts.append(pos.amount)
-        last_sale_prices.append(pos.last_sale_price)
-
-    position_value_multipliers = pt._position_value_multipliers
-    position_exposure_multipliers = pt._position_exposure_multipliers
-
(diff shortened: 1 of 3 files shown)


zipline/finance/performance/position_tracker.py
```diff
@@ -121,16 +121,15 @@ def calc_gross_value(long_value, short_value):
     return long_value + abs(short_value)
 
 
-def calc_position_stats(pt):
+def calc_position_stats(positions,
```
llllllllll: should we just move the body of this into the method or are we ever calling this standalone?

zipline/finance/performance/position_tracker.py
```diff
@@ -167,6 +166,19 @@ def calc_position_stats(pt):
         net_value=net_value
     )
 
+    return PositionStats(
```
llllllllll: looks like dead code.

zipline/finance/performance/position_tracker.py
ehebert: You're right. There's no need for the standalone now, will move.

zipline/finance/performance/position_tracker.py
ehebert: Yes, accidentally doubled up the return when copying over function. Will fix.

ehebert:
@llllllllll I squashed in the changes based on your notes. Could you take a second look at the diff?

PR #921, opened 2015-12-18
STY: Align all build_ext definitions the same way.

ehebert:
No functional change.

Done to improve diff with Q2.0 which adds more modules, and re-aligned the two modules changed here.


llllllllll:
thanks

PR #923, opened 2015-12-18
MAINT: fix error messages for set_(commission|slippage)

llllllllll:
These called the functions 'update_(commission|slippage) leading to
confusing messages when you called them post init.


richafrank:
Thanks for the cleanup @llllllllll .  Makes sense to me.

PR #924, opened 2015-12-18
ENH: Make datasets have subclass relationships

llllllllll:
Needed for the EarningsCalendar work we are doing

### DIFF
--- a/tests/pipeline/test_term.py
+++ b/tests/pipeline/test_term.py
@@ -28,6 +28,10 @@ class SomeDataSet(DataSet):
     buzz = Column(float64_dtype)
 
 
+class SubDataSet(SomeDataSet):
+    pass
+
+
 class SomeFactor(Factor):
     dtype = float64_dtype
     window_length = 5
@@ -321,3 +325,31 @@ class ObjectIdentityTestCase(TestCase):
 
         with self.assertRaises(InvalidDType):
             SomeFactor(dtype=1)
+
+
+class SubDataSetTestCase(TestCase):
+    def test_subdataset(self):
+        some_dataset_map = {
+            column.name: column for column in SomeDataSet.columns
+        }
+        sub_dataset_map = {
+            column.name: column for column in SubDataSet.columns
+        }
+        self.assertEqual(
+            set(some_dataset_map),
+            set(sub_dataset_map),
+        )
+        for k, some_dataset_column in some_dataset_map.items():
+            sub_dataset_column = sub_dataset_map[k]
+            self.assertIsNot(
+                some_dataset_column,
+                sub_dataset_column,
+                'subclass column %r should not have the same identity as'
+                ' the parent' % k,
+            )
+            self.assertEqual(
+                some_dataset_column.dtype,
+                sub_dataset_column.dtype,
+                'subclass column %r should have the same dtype as the parent' %
+                k,
+            )
--- a/tests/pipeline/test_term.py
+++ b/tests/pipeline/test_term.py
@@ -32,6 +32,10 @@ class SubDataSet(SomeDataSet):
     pass
 
 
+class SubDataSetNewCol(SomeDataSet):
+    qux = Column(float64_dtype)
+
+
 class SomeFactor(Factor):
     dtype = float64_dtype
     window_length = 5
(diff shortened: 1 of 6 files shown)


zipline/pipeline/data/dataset.py
```diff
@@ -119,5 +140,5 @@ def __repr__(self):
         return '<DataSet: %r>' % self.__name__
 
 
-class DataSet(with_metaclass(DataSetMeta)):
+class DataSet(with_metaclass(DataSetMeta, object)):
```
ssanderson: Was this previously an old-style class in PY2?

zipline/pipeline/data/dataset.py
```diff
@@ -97,20 +114,24 @@ class DataSetMeta(type):
...
+                column_names.add(maybe_colname)
 
-        newtype._columns = frozenset(_columns)
+        newtype._column_names = frozenset(column_names)
         return newtype
 
     @property
     def columns(self):
-        return self._columns
+        return frozenset(
```
ssanderson: Should this be a lazyval now that it's doing semi-nontrivial work?

zipline/pipeline/data/dataset.py
llllllllll: yeah, I believe calling `type('name', (), {})` creates an old style class. At least it excludes `object` from the `__mro__`.

zipline/pipeline/data/dataset.py
llllllllll: The cost to compute this is still trivial. Idk if we should jump to caching just yet.

zipline/pipeline/data/dataset.py
```diff
@@ -97,20 +114,24 @@ class DataSetMeta(type):
     """
 
     def __new__(mcls, name, bases, dict_):
-        newtype = type.__new__(mcls, name, bases, dict_)
-        _columns = []
+        newtype = super(DataSetMeta, mcls).__new__(mcls, name, bases, dict_)
+        column_names = set().union(
```
ssanderson: Might be worth adding a comment here like `# Get columns from any parent classes`.

zipline/pipeline/data/dataset.py
```diff
@@ -97,20 +114,24 @@ class DataSetMeta(type):
...
+            *(getattr(base, '_column_names', ()) for base in bases)
+        )
         for maybe_colname, maybe_column in iteritems(dict_):
             if isinstance(maybe_column, Column):
-                bound_column = maybe_column.bind(newtype, maybe_colname)
-                setattr(newtype, maybe_colname, bound_column)
-                _columns.append(bound_column)
+                bound_column_descr = maybe_column.bind(maybe_colname)
```
ssanderson: Might be worth commenting here something like `# Add columns from the newly-created class.`.

tests/pipeline/test_term.py
```diff
@@ -321,3 +325,31 @@ class SomeFactorNoDType(SomeFactor):
...
+
+class SubDataSetTestCase(TestCase):
+    def test_subdataset(self):
+        some_dataset_map = {
+            column.name: column for column in SomeDataSet.columns
+        }
+        sub_dataset_map = {
+            column.name: column for column in SubDataSet.columns
+        }
+        self.assertEqual(
```
ssanderson: I was confused by this assertion when reading.  I think it'd be clearer to assert that `SomeDataSet.columns` is equal to `SubDataSet.columns`, which I think it both equivalent and closer to the intent of what's being tested here.

tests/pipeline/test_term.py
```diff
@@ -28,6 +28,10 @@ class SomeDataSet(DataSet):
     buzz = Column(float64_dtype)
 
 
+class SubDataSet(SomeDataSet):
```
ssanderson: Can we add a test case for adding a new column in a subdataset?

zipline/pipeline/data/dataset.py
ssanderson: :scream_cat:

ssanderson:
By and large LGTM.  Had a couple requests for comments and one request for a testcase.

tests/pipeline/test_term.py
llllllllll: Sounds good, this is what I am trying to test.

llllllllll:
added the test and comments. merging on pass

PR #928, opened 2015-12-22
Lazy latest dt

ehebert:
ENH: Provide method for last trade dt.

    Add a data_portal method `get_last_traded_dt` which the dt passed to the
    method if there is a trade on that bar, or the previous dt with a trade
    if there is no trade on the current dt.

    This method will be used to proved a `.dt` lookup on SidView, but the
    connection to the SidView is not made with this patch.

    To implement the get_last_traded_dt a corresponding method
    was added to the minute bar reader, with a supporting DatetimeIndex
    which matches the data sets indices. Also, change
    `_find_position_of_minute` to use the index, since the lookup is faster
    than calculating a Timestamp and doing the offset calculation every
    time. (As seen by the improvement in the ffill tests in test_dataportal.)

    Also, change history tests to use a shared data_portal to avoid setup
    cost of repeatedly creating the lookup table in the reader.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -102,19 +102,6 @@ class DataPortal(object):
         self._future_daily_reader = future_daily_reader
         self._future_minute_reader = future_minute_reader
 
-        # The following values are used by _minute_offset to calculate the
-        # index into the minute bcolz date.
-
-        # A lookup of table every minute to the corresponding day, to avoid
-        # calling `.date()` on every lookup.
-        self._minutes_to_day = {}
-        # A map of days (keyed by midnight) to a DatetimeIndex of market
-        # minutes for that day.
-        self._minutes_by_day = {}
-        # A dict of day to the offset into the minute bcolz on which that
-        # days data starts.
-        self._day_offsets = None
-
     def handle_extra_source(self, source_df, sim_params):
         """
         Extra sources always have a sid column.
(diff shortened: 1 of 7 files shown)


zipline/gens/sim_engine.pyx
```diff
@@ -86,7 +83,7 @@ cdef class MinuteSimulationClock:
                     yield minute, MINUTE_END
 
             if not minute_emission:
-                yield day, DAY_END
+                yield minutes[-1], DAY_END
```
jbredeche: why this change?  daily mode needs days, right?  (ie, 00:00, not 16:00)

ehebert: This is  the minute mode clock. In both daily and minute emissions, the prices used for the day end should be at the market_close.

Noticed this when using `get_loc` instead of a combination of searchsorted for getting the minute position. From reading the code it may have been that the first minute of the day was being used for the last sale prices for risk calculations, but have not confirmed.

zipline/data/data_portal.py
```diff
@@ -234,6 +212,16 @@ def _get_ctable(self, asset):
...
+    def get_last_traded_dt(self, asset, dt, data_frequency):
+        """
+        Given an asset and dt, returns the last traded dt from the viewpoint
+        of the given dt.
+
+        If there is a trade on the dt, the answer is dt provided.
+        """
+        if data_frequency == 'minute':
+            return self._equity_minute_reader.get_last_traded_dt(asset, dt)
+
```
jbredeche: should we blow up on other data frequencies (for now)?

ehebert: Looking at this again, with wiring this to SidView, we'll have to implement for daily.

PR #929, opened 2015-12-23
Lazy wire up datetime

ehebert:
Build upon recently added `get_last_traded_dt` so that following algocode works again.

```
    data[sid(24)].dt
    data[sid(24)].datetime
```

### DIFF
--- a/zipline/data/us_equity_pricing.py
+++ b/zipline/data/us_equity_pricing.py
@@ -481,6 +481,13 @@ class BcolzDailyBarReader(object):
             col = self._spot_cols[colname] = self._table[colname]
         return col
 
+    def get_last_traded_dt(self, asset, day):
+        volumes = self._spot_col('volume')
+        ix = self.sid_day_index(asset, day)
+        while True:
+            if volumes[ix] != 0:
+                return day
+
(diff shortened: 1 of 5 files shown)


zipline/data/us_equity_pricing.py
```diff
@@ -481,6 +481,13 @@ def _spot_col(self, colname):
             col = self._spot_cols[colname] = self._table[colname]
         return col
 
+    def get_last_traded_dt(self, asset, day):
+        volumes = self._spot_col('volume')
+        ix = self.sid_day_index(asset, day)
+        while True:
+            if volumes[ix] != 0:
```
jbredeche: shouldn't we be decrementing `ix` in this loop?

ehebert: Yes, we should. Adding a test to cover daily in the  test_dataportal suite which includes an empty day.

ehebert: @jbredeche https://github.com/quantopian/zipline/commit/42b95c60eeec8282b78804212d5abea18948942e should fix this, have time to take a look?

jbredeche:
lgtm

PR #931, opened 2015-12-29
MAINT: Factors-out asset db schema

jfkirk:
This (now-misnamed) branch moves the asset db schema to its own module for clarity and ease of modification. There should be no functional changes.


jfkirk:
@jdricklefs Mind looking at this PR? This is just a simple move, now that we may not be adding full alembic support in Zipline.

jdricklefs:
Looks fine to me but you may want to wait to merge until you get all that other internal machinery ready.

zipline/assets/assets.py
```diff
@@ -106,9 +104,6 @@ def __init__(self, engine):
         for table_name in asset_db_table_names:
             setattr(self, table_name, metadata.tables[table_name])
 
-        # Check the version info of the db for compatibility
-        check_version_info(self.version_info, ASSET_DB_VERSION)
```
richafrank: Why do we no longer check the version info here?

jfkirk: Accidental carry-over from a previous iteration - removing now. Thanks!

jfkirk: I'm also adding a test to ensure that the finder is verifying version to avoid someone (me) accidentally removing this check again.

richafrank: :+1:

PR #932, opened 2015-12-29
Forward filling history, fill leading nans 

ehebert:
Improve the history logic so that if there is a leading window of nans, the value on last traded dt is used to fill the beginning of the window.

e.g.:
```
                           Equity(13 [FOO])
2014-03-20 19:52:00+00:00               NaN
2014-03-20 19:53:00+00:00               NaN
2014-03-20 19:54:00+00:00               NaN
2014-03-20 19:55:00+00:00               NaN
2014-03-20 19:56:00+00:00               NaN
2014-03-20 19:57:00+00:00               NaN
2014-03-20 19:58:00+00:00               NaN
2014-03-20 19:59:00+00:00               NaN
2014-03-20 20:00:00+00:00               NaN
2014-03-21 13:31:00+00:00               NaN
2014-03-21 13:32:00+00:00               NaN
2014-03-21 13:33:00+00:00               NaN
2014-03-21 13:34:00+00:00               NaN
2014-03-21 13:35:00+00:00               NaN
2014-03-21 13:36:00+00:00               NaN
2014-03-21 13:37:00+00:00               NaN
2014-03-21 13:38:00+00:00               NaN
2014-03-21 13:39:00+00:00               NaN
2014-03-21 13:40:00+00:00               NaN
2014-03-21 13:41:00+00:00           529.601
```

becomes:

```
                           Equity(13 [FOO])
2014-03-20 19:52:00+00:00           533.460
2014-03-20 19:53:00+00:00           533.460
2014-03-20 19:54:00+00:00           533.460
2014-03-20 19:55:00+00:00           533.460
2014-03-20 19:56:00+00:00           533.460
2014-03-20 19:57:00+00:00           533.460
2014-03-20 19:58:00+00:00           533.460
2014-03-20 19:59:00+00:00           533.460
2014-03-20 20:00:00+00:00           533.460
2014-03-21 13:31:00+00:00           533.460
2014-03-21 13:32:00+00:00           533.460
2014-03-21 13:33:00+00:00           533.460
2014-03-21 13:34:00+00:00           533.460
2014-03-21 13:35:00+00:00           533.460
2014-03-21 13:36:00+00:00           533.460
2014-03-21 13:37:00+00:00           533.460
2014-03-21 13:38:00+00:00           533.460
2014-03-21 13:39:00+00:00           533.460
2014-03-21 13:40:00+00:00           533.460
2014-03-21 13:41:00+00:00           529.601
```

The previous value is adjusted for corporate actions. The above cut and paste comes from a test equity with a split between `2014-03-20` and `2014-03-21` with the previous value being on the day before the split.

Also, while developing this, did some work on the history test case `setUp` time to improve iteration speed.


zipline/data/data_portal.py
```diff
@@ -338,6 +339,75 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
+
+        spot_value = self.get_spot_value(asset, field, dt, data_frequency)
+
+        if isinstance(asset, Equity):
+            adjs = []
+            split_adjustments = self._get_adjustment_list(
+                asset, self._splits_dict, "SPLITS"
+            )
+            for adj_dt, adj in split_adjustments:
+                if dt > adj_dt <= perspective_dt:
```
ssanderson: Style note:  I'd always prefer this to be written as `adj_dt < dt <= perspective_dt`, because it makes the left-to-right ordering reflect the expected ordering of the elements being compared.

ehebert: I totally agree, no idea why it ended up that way. Fixing.

zipline/data/data_portal.py
```diff
@@ -696,6 +766,25 @@ def get_history_window(self, assets, end_dt, bar_count, frequency, field,
 
         # forward-fill if needed
         if field == "price" and ffill:
+            assets_with_nan_index = np.where(pd.isnull(df.iloc[0, :]))[0]
+            assets_with_leading_nan = [
+                assets[i] for i in assets_with_nan_index]
+            for assets in assets_with_leading_nan:
+                if frequency == "1m":
```
ssanderson: These checks probably want to be outside the loop.

zipline/data/data_portal.py
richafrank: Is it ok that those two comparisons aren't the same with respect to the relationship between `dt` and `perspective_dt`?

zipline/data/data_portal.py
ehebert: They do. Moving.

zipline/data/data_portal.py
richafrank: Oh, since it's totally ordered, they should be actually be the same! that makes sense.

zipline/data/data_portal.py
```diff
@@ -696,6 +766,25 @@ def get_history_window(self, assets, end_dt, bar_count, frequency, field,
...
+                    raise Exception(
+                        "Only 1d and 1m are supported for forward-filling.")
+                dt_to_fill = df.index[0]
+                perspective_dt = df.index[-1]
+                previous_dt = self.get_last_traded_dt(
+                    asset, dt_to_fill, data_frequency)
+                previous_value = self._get_adjusted_value(
+                    asset, field, previous_dt, perspective_dt,
+                    data_frequency)
+                df.loc[dt_to_fill, asset] = previous_value
```
ssanderson: You always know that you're writing to the first row here, so the `.loc` is doing an unnecessary hash lookup.

zipline/data/data_portal.py
```diff
@@ -696,6 +766,25 @@ def get_history_window(self, assets, end_dt, bar_count, frequency, field,
 
         # forward-fill if needed
         if field == "price" and ffill:
+            assets_with_nan_index = np.where(pd.isnull(df.iloc[0, :]))[0]
+            assets_with_leading_nan = [
```
ssanderson: This is replicating the behavior of masking with a boolean array.  Should we make `assets` a numpy array up above to make this cleaner?

zipline/data/us_equity_minutes.py
```diff
@@ -49,10 +48,29 @@ def write_metadata(directory, first_trading_day):
         json.dump(metadata, fp)
 
 
+def _bcolz_minute_index(trading_days):
+        minutes = np.zeros(len(trading_days) * MINUTES_PER_DAY,
```
ssanderson: Looks like this is indented 8 spaces?

zipline/data/us_equity_minutes.py
```diff
@@ -49,10 +48,29 @@ def write_metadata(directory, first_trading_day):
...
+            end_ix = start_ix + MINUTES_PER_DAY
+            minutes[start_ix:end_ix] = minute_values
+        return pd.to_datetime(minutes, utc=True, box=True)
+
+
 class BcolzMinuteBarWriter(with_metaclass(ABCMeta)):
     """
     Class capable of writing minute OHLCV data to disk into bcolz format.
     """
+    _writer_env = TradingEnvironment()
```
ssanderson: We've done a lot of work recently to remove the instantiation of global trading environments.  Is there a reason we're adding this one here?

zipline/data/data_portal.py
ehebert: What do suggest instead, `df[asset][0]`? (As a force of habit, I've been use `loc` etc. when using two-coordinates, but not as important in this case where only one item is being written.)

zipline/data/us_equity_minutes.py
ehebert: Will fix. (Must have been from copying when factoring out.)

zipline/data/us_equity_minutes.py
ehebert: This was previously at module scope. The change to be a class member was in effort to ease this class into using passing the env directly, since the value was needed in `_write_internal` with the change to how the minutes were created during the asset loop.

Shall we make this an instance member instead and pass it in with the init?

zipline/data/data_portal.py
ehebert: Something like this?

```
diff --git a/zipline/data/data_portal.py b/zipline/data/data_portal.py
index 8383f46..ddf449d 100644
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -752,8 +752,9 @@ class DataPortal(object):
             raise ValueError("Invalid history field: " + str(field))

         # sanity check in case sids were passed in
-        assets = [(self.env.asset_finder.retrieve_asset(asset) if
-                   isinstance(asset, int) else asset) for asset in assets]
+        assets = np.array([
+            (self.env.asset_finder.retrieve_asset(asset) if
+             isinstance(asset, int) else asset) for asset in assets])

         if frequency == "1d":
             df = self._get_history_daily_window(assets, end_dt, bar_count,
@@ -767,8 +768,7 @@ class DataPortal(object):
         # forward-fill if needed
         if field == "price" and ffill:
             assets_with_nan_index = np.where(pd.isnull(df.iloc[0, :]))[0]
-            assets_with_leading_nan = [
-                assets[i] for i in assets_with_nan_index]
+            assets_with_leading_nan = assets[assets_with_nan_index]

             if frequency == "1m":
                 data_frequency = 'minute'
```

zipline/data/data_portal.py
ssanderson: I'd rewrite this loop as:
```
dt_to_fill = df.index[0]
perspective_dt = df.index[-1]
for missing_loc in np.where(pd.isnull(df.iloc[0])):
    asset = assets[missing_loc]
    previous_dt = self.get_last_traded_dt(asset, dt_to_fill, data_frequency)
    previous_value = self._get_adjusted_value(
        asset, 
        field, 
        previous_dt, 
        perspective_dt,
        data_frequency,
    )
    df.iloc[0, missing_loc] = previous_value  # I'd probably actually just assign directly from the expression.
```

zipline/data/us_equity_minutes.py
```diff
@@ -49,10 +48,29 @@ def write_metadata(directory, first_trading_day):
         json.dump(metadata, fp)
 
 
+def _bcolz_minute_index(trading_days):
+        minutes = np.zeros(len(trading_days) * MINUTES_PER_DAY,
+                           dtype='datetime64[ns]')
+        market_opens = tradingcalendar.open_and_closes.market_open
+        mask = market_opens.index.slice_indexer(start=trading_days[0],
```
ssanderson: This isn't actually a mask, it's a `slice` object.

```
In [2]: df.index.slice_indexer(1, 5)
Out[2]: slice(1, 6, None)
```

zipline/data/us_equity_minutes.py
```diff
@@ -49,10 +48,29 @@ def write_metadata(directory, first_trading_day):
         json.dump(metadata, fp)
 
 
+def _bcolz_minute_index(trading_days):
+    minutes = np.zeros(len(trading_days) * MINUTES_PER_DAY,
+                       dtype='datetime64[ns]')
+    market_opens = tradingcalendar.open_and_closes.market_open
+    mask = market_opens.index.slice_indexer(start=trading_days[0],
```
ehebert: Do you have a standard name you use here instead of `silce` then (to avoid using the built-in)?

zipline/data/us_equity_minutes.py
```diff
@@ -49,10 +48,29 @@ def write_metadata(directory, first_trading_day):
...
 
+def _bcolz_minute_index(trading_days):
+        minutes = np.zeros(len(trading_days) * MINUTES_PER_DAY,
+                           dtype='datetime64[ns]')
+        market_opens = tradingcalendar.open_and_closes.market_open
+        mask = market_opens.index.slice_indexer(start=trading_days[0],
+                                                end=trading_days[-1])
+        opens = market_opens[mask]
+        for i, market_open in enumerate(opens):
+            start = market_open.value
```
ssanderson: This is individually unboxing every timestamp object in every iteration of the loop.  It seems like we could straightforwardly do all of this with vectorized numpy operations?

ssanderson: Fancy way to do this:

```
(pd.date_range('2014', '2015').values[:, None] + arange(0, 390, dtype='timedelta64[m]')).reshape((366 * 390,))
```

zipline/data/us_equity_minutes.py
```diff
@@ -120,27 +119,56 @@ def _write_internal(self, directory, iterator, sid_path_func=None):
...
-                vol_col[idx] = row[1].loc["volume"]
+            opens = df.open.values.astype(np.uint32)
+            highs = df.high.values.astype(np.uint32)
+            lows = df.low.values.astype(np.uint32)
+            closes = df.close.values.astype(np.uint32)
+            volumes = df.volume.values.astype(np.uint32)
+
+            dt_ixs = np.searchsorted(minutes.values, df.index.values)
+
+            for i, dt_ix in enumerate(dt_ixs):
```
ssanderson: I think this is just ` open_col[dt_ixs] = opens` (with appropriate substitutions for the others).

zipline/data/us_equity_minutes.py
ssanderson: Looking at where this is being called, I think this truncation logic is actually just unnecessary.  The trading days here are being generated from `days_in_range`, so I think we should just make it the responsibility of the caller to pass in the array of opens.  

I'd probably write this as:

```
def trading_minute_range(opens, minutes_per_day):
    """
    Construct an array of datetime64[ns] with the ``minutes_per_day`` minutes for each
    day in ``opens``.

    Parameters
    ----------
    opens : array[datetime64[ns]]
        An array of daily market opens.
    minutes_per_day : int
        The number of minutes to produce for each day.

    Returns
    -------
    minutes : array[datetime64[ns]]
        An array containing ``minutes_per_day`` sequential minutes for each day in ``opens``.
    """
    return (opens.values[:, None] + arange(minutes_per_day, dtype='timedelta64[m]'))\
        .reshape((len(opens) * minutes_per_day,))
```
I'd probably just require that this be passed an array of the opens and the number of minutes in each day.

ssanderson: (**NOTE:** `np.newaxis` is a slightly more idiomatic thing to use instead of `None` for turning the row array into a column array.)

zipline/data/data_portal.py
```diff
@@ -696,6 +767,26 @@ def get_history_window(self, assets, end_dt, bar_count, frequency, field,
...
+
+            if frequency == "1m":
+                data_frequency = 'minute'
+            elif frequency == "1d":
+                data_frequency = 'daily'
+            else:
+                raise Exception(
+                    "Only 1d and 1m are supported for forward-filling.")
+
+            for assets in assets_with_leading_nan:
```
jbredeche: interesting, my mental model of how to solve this problem was, when asked for a window on an asset for specific minutes, to extend the window to the last traded date if the first slot of the window has no value.  get the history window (which does all the adjustments), then just chop off all the values before the originally requested window.

your solution seems to work fine, just completely different from what I had mentally envisioned.

zipline/data/data_portal.py
```diff
@@ -338,6 +339,75 @@ def get_spot_value(self, asset, field, dt, data_frequency):
                 return self._get_minute_spot_value(
                     asset, column_to_use, dt)
 
+    def _get_adjusted_value(self, asset, field, dt,
+                            perspective_dt,
+                            data_frequency):
+        """
+        Private method that returns a scalar value representing the value
+        of the desired asset's field at either the given dt.
```
jbredeche: incomplete sentence

zipline/data/data_portal.py
```diff
@@ -338,6 +339,75 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
+
+        Parameters
+        ---------
+        asset : Asset
+            The asset whose data is desired.gith
```
jbredeche: cruft "gith"

zipline/data/data_portal.py
```diff
@@ -338,6 +339,75 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
+                )
+                for adj_dt, adj in merger_adjustments:
+                    if adj_dt < dt <= perspective_dt:
+                        adjs.append(adj)
+
+                div_adjustments = self._get_adjustment_list(
+                    asset, self._dividends_dict, "DIVIDENDS",
+                )
+                for adj_dt, adj in div_adjustments:
+                    if adj_dt < dt <= perspective_dt:
```
jbredeche: since dividends (and all the other adjustments) come back in sorted order, is it worth breaking out of this list the first time you see an adj_dt that is greater than `dt` or `perspective_dt`?

zipline/data/data_portal.py
ehebert: I did start out with that model, but both getting answer of 'get the number of bars between beginning of window and the known last dt' a bit awkward to handle for both daily and minute (minute being the more awkward of the two.)

Also, went down the route of trying to use `get_minute_window_for_asset` with creating a a minute creating a minute window using `env.minute_market_window` but that method also expected the bar count.

We could add an analogue to `trading_day_distance` for `trading_minute_distance` using the `minutes_for_days_in_range`.

I also tried using a "fake" window of `DatetimeIndex([previous_dt, perspective_dt]), but ran into issues (which I can try to recreate) when passing that window `get_market_window_for_asset`.

I do concede that adding a different paths for adjustments is added complexity, but was trading it for the complexity in constructing window to pass to the existing adjustment code path.

zipline/data/data_portal.py
ehebert: Yes, which would be similar to in `_apply_adjustments_to_window`'s

```
if adjustment_to_apply[0] > dts_in_window[-1]:
   break
```

Will change.

zipline/data/us_equity_minutes.py
ehebert: @ssanderson I really like this solution, but there is some contortion needed for consistently getting a dtype of `datetime64` out of the `market_opens` in the trading calendar for the window needed.

Going to leave for now, but may try to get this in soon.

zipline/data/us_equity_minutes.py
ehebert: @ssanderson I'm going to back out the change to `_writer_env` here from what already exists on the lazy-mainline branch, and then follow up a separate PR making the trading_env a member of the writer.

zipline/data/us_equity_minutes.py
ssanderson: @ehebert did what I suggested here turn out to not be equivalent to the behavior here?

zipline/data/us_equity_minutes.py
ssanderson: ... what other dtypes are we getting for `market_opens`?

ehebert: `object`

zipline/data/us_equity_minutes.py
ehebert: I neglected to address this comment. Will fix.

ehebert: @ssanderson https://github.com/quantopian/zipline/pull/933

PR #934, opened 2015-12-30
Use master-alike create_transaction interface.

ehebert:
Make the interface to the slippage `create_transaction` and
`process_order` as it is on master, so that there is less of a chance
for breaking existing algorithms that override slippage.

Create a TradeBar instance in slippage simulate, which holds the `price`
and `volume` for use by both process_order and create_transaction.

Other notes:

- This patch is in service of reducing the edit distance between master and lazy-mainline both in code and functionality.
- `trade_bar` is used in place of where `event` is on master. This aligns with the documentation of the function on https://www.quantopian.com/help.
- We may need to make TradeBar a dynamic object which can also access `open` and `close`, to support different models implemented by `process_order`. But that may not be in scope, since `price` and `volume` were sufficient for existing coverage.

### DIFF
--- a/zipline/finance/slippage.py
+++ b/zipline/finance/slippage.py
@@ -18,6 +18,7 @@ import abc
 
 import math
 
+from collections import namedtuple
 from copy import copy
 
 from six import with_metaclass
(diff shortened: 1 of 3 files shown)


zipline/finance/slippage.py
```diff
@@ -65,7 +69,11 @@ def simulate(self, current_orders, dt, price, volume):
                 continue
 
             try:
-                txn = self.process_order(order, price, volume, dt)
+                trade_bar = TradeBar(order.sid,
```
jbredeche: should the construction of `trade_bar` be outside the `try` clause?

jbredeche:
one minor thing, LGTM otherwise.

zipline/finance/slippage.py
ehebert: Will change.

Issue #894, opened 2015-12-03
Update dependencies


twiecki:
Some of the pegged dependencies are pretty old, like numpy 1.9.2 (now 1.11) pandas 0.16.0 (now 0.17.1), scipy 0.15.1 (now 0.16.1). These are making it harder to build conda packages as all the other packages moved ahead and numpy 1.9.2 is not being distributed anymore.

twiecki:
CC @richafrank @llllllllll

richafrank:
Agreed.

ssanderson:
Note that there are several known failures in our test suite that are caused by upgrading numpy and pandas.  In particular, https://github.com/numpy/numpy/issues/6452 would require a few workarounds in the Pipeline API code, and the introduction of tz-aware datetimes in pandas 0.17.0 is likely to have implications for us.

richafrank:
I'm going to call this closed with https://github.com/quantopian/zipline/pull/969

PR #937, opened 2015-12-31
Conda source build

richafrank:
Allows us to build conda packages for the latest zipline and its dependencies.  Builds from local source instead of pypi, pulling in requirements from our requirements.txt.

Requires this change to conda-build: https://github.com/conda/conda-build/pull/717

### DIFF
--- a/conda/zipline/meta.yaml
+++ b/conda/zipline/meta.yaml
@@ -17,7 +17,7 @@ requirements:
   build:
     - python
     - setuptools
-    - cython ==0.22.1
+    - cython
     - numpy
   run:
     - python
--- a/conda/zipline/meta.yaml
+++ b/conda/zipline/meta.yaml
@@ -18,7 +18,7 @@ requirements:
     - python
     - setuptools
     - cython ==0.22.1
-    - numpy ==1.9.2
+    - numpy
   run:
     - python
     {% for req in data.get('install_requires', []) -%}
--- a/conda/zipline/meta.yaml
+++ b/conda/zipline/meta.yaml
@@ -1,28 +1,37 @@
+{% set data = load_setuptools() %}
+
 package:
   name: zipline
-  version: 0.8.3
+  version: {{ environ.get('GIT_DESCRIBE_TAG', '')}}
+
+build:
+  number: {{ environ.get('GIT_DESCRIBE_NUMBER', 0) }}
+  {% if environ.get('GIT_DESCRIBE_NUMBER', '0') == '0' %}string: py{{ environ.get('PY_VER').replace('.', '') }}_0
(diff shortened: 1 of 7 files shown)


ssanderson:
@richafrank are we blocked on being able to push conda packages until  conda/conda-build#717 is merged?

llllllllll:
I don't thinl this would block uploading packages, we can always build them locally with our patch applied and then upload to anaconda.org.

setup.py
```diff
@@ -214,6 +228,7 @@ def pre_setup():
 
 pre_setup()
 
+conda_build = os.path.basename(sys.argv[0]) == 'conda-build'
```
llllllllll: Do you think we should PR a feature that adds a `CONDA_BUILD` variable to the name space when running? This will not be true if you symlink conda build or run it from a location not in your path like `./conda-build`

richafrank: Sounds like a good idea to me.

richafrank:
Note: need to add whatsnew

conda/zipline/meta.yaml
```diff
@@ -1,28 +1,37 @@
...
-  md5: 042ffcee614d2279add9a1bfd27a33cf
+  git_url: ../../
 
 requirements:
   build:
     - python
     - setuptools
-    - cython
+    - cython ==0.22.1
+    - numpy ==1.9.2
```
twiecki: Does it only work with these specific versions?

richafrank: With numpy>=1.10 I'm getting `ValueError: numpy.dtype has the wrong size, try recompiling`.  I think @llllllllll  was getting this too.  

`numpy` also included in the `run` requirements, so I'll remove this pinned version, but add the cap at 1.9 for now.  I'll also give us back non-strict versions on the other run requirements - that was unintentional.

twiecki:
I think it'd be worth investigating why we can't build for numpy 1.10. I'm afraid a conda build for numpy 1.9 won't be terribly useful to most users as most of anaconda is built against 1.10.

twiecki:
Here's an example: https://github.com/quantopian/pyfolio/issues/73#issuecomment-169172326

richafrank:
> I think it'd be worth investigating why we can't build for numpy 1.10. I'm afraid a conda build for numpy 1.9 won't be terribly useful to most users as most of anaconda is built against 1.10.

Agreed - will keep looking...

llllllllll:
We can't upgrade numpy yet because https://github.com/quantopian/zipline/issues/894#issuecomment-162696875 is still pending. This issue hasn't been fixed upstream in numpy yet so we are currently locked onto 1.9.2.

richafrank:
Merging, though the conda-build branch has not yet been merged.  Plan is to build from the branch until then.

PR #938, opened 2016-01-04
MAINT: Futures cash adjustment on change and calc.

ehebert:
In preparation for the incoming changes which no longer push every bar
through the tradesimulation, remove the adjustment of the period's cash on
every pricing change of a held futures asset.

Instead hold the last sale price for each held future either:

- At the end of each peformance period update the last sale prices of
  all held futures, so that the pnl for the next period uses values
  derived from the cash difference between the end of the two periods.

- When a transaction is processed for the Future, so that the correct
  amount is applied to each cash adjustment. (i.e. the cash adjustment
  is reset on every change of amount of the Future being held, so that
  multiple size and prices do not need to be tracked for the same asset.)

Also, remove now unused dict of payout calculation modifier, since new
calculation reads the value directly off of the asset.

Remove update_last_sale test, since the method no longer returns a cash
value.

### DIFF
--- a/tests/test_perf_tracking.py
+++ b/tests/test_perf_tracking.py
@@ -2227,26 +2227,6 @@ class TestPositionTracker(unittest.TestCase):
             self.assertEquals(val, 0)
             self.assertNotIsInstance(val, (bool, np.bool_))
 
-    def test_update_last_sale(self):
-        pt = perf.PositionTracker(self.env.asset_finder)
-        dt = pd.Timestamp("1984/03/06 3:00PM")
-        pos1 = perf.Position(1, amount=np.float64(100.0),
-                             last_sale_date=dt, last_sale_price=10)
-        pos3 = perf.Position(3, amount=np.float64(100.0),
-                             last_sale_date=dt, last_sale_price=10)
-        pt.update_positions({1: pos1, 3: pos3})
-
-        event1 = Event({'sid': 1,
-                        'price': 11,
-                        'dt': dt})
-        event3 = Event({'sid': 3,
-                        'price': 11,
-                        'dt': dt})
-
-        # Check cash-adjustment return value
-        self.assertEqual(0, pt.update_last_sale(event1))
-        self.assertEqual(100000, pt.update_last_sale(event3))
-
     def test_position_values_and_exposures(self):
         pt = perf.PositionTracker(self.env.asset_finder)
         dt = pd.Timestamp("1984/03/06 3:00PM")
--- a/zipline/finance/performance/period.py
+++ b/zipline/finance/performance/period.py
@@ -126,6 +126,10 @@ def calc_period_stats(pos_stats, ending_cash):
         net_leverage=net_leverage)
 
 
+def calc_payout(contract_multiplier, amount, old_price, price):
+    return (price - old_price) * contract_multiplier * amount
+
+
 class PerformancePeriod(object):
 
     def __init__(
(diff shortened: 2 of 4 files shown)


ehebert:
@jfkirk this is in preparation for Q2.0. Necessitated, by noticing that the existing in implementation on lazy-mainline Q2.0 would only work in daily mode, and if the same position and amount was held throughout a test.

zipline/finance/performance/period.py
```diff
@@ -207,15 +225,34 @@ def adjust_cash(self, amount):
     def adjust_field(self, field, value):
         setattr(self, field, value)
 
+    def _get_payout_total(self, positions):
+        payouts = []
+        for asset, old_price in iteritems(self._payout_last_sale_prices):
+            pos = positions[asset]
+            price = pos.last_sale_price
```
jfkirk: For my clarity --- at time T:
pos.last_sale_price is the price of 'asset' at time T
old_price is the last price for which a payout occurred (at time T-N)

zipline/finance/performance/period.py
```diff
@@ -242,6 +279,22 @@ def record_order(self, order):
     def handle_execution(self, txn):
         self.period_cash_flow += self._calculate_execution_cash_flow(txn)
 
+        asset = self.asset_finder.retrieve_asset(txn.sid)
+        if isinstance(asset, Future):
+            try:
+                old_price = self._payout_last_sale_prices[asset]
+                amount = self.position_tracker.positions[asset].amount
+                price = txn.price
+                cash_adj = (price - old_price) * asset.contract_multiplier * \
```
jfkirk: This seems like code duplication with _get_payout_total()
Can the common ground be abstracted?

tests/test_perf_tracking.py
```diff
@@ -2227,26 +2227,6 @@ def test_empty_positions(self):
             self.assertEquals(val, 0)
             self.assertNotIsInstance(val, (bool, np.bool_))
 
-    def test_update_last_sale(self):
```
jfkirk: If Position objects still hold last sale prices, do we want to test for updating that field?

zipline/finance/performance/period.py
ehebert: Correct.

zipline/finance/performance/period.py
ehebert: Yes, will attempt to do so.

tests/test_perf_tracking.py
ehebert: The `update_last_sale` method is going away (already removed on lazy-mainline branch). Also the current test only asserts on the cash adjustment, and doesn't check the last_sale_price. So the trade off there is adding a coverage on something that is going away, versus clearing the way.

zipline/finance/performance/period.py
```diff
@@ -207,15 +229,31 @@ def adjust_cash(self, amount):
...
         pos_stats = pt.stats()
         self.ending_value = pos_stats.net_value
         self.ending_exposure = pos_stats.net_exposure
 
+        payout = self._get_payout_total(pt.positions)
+
         total_at_start = self.starting_cash + self.starting_value
         self.ending_cash = self.starting_cash + self.period_cash_flow
-        total_at_end = self.ending_cash + self.ending_value
+        total_at_end = self.ending_cash + self.ending_value + payout
```
jfkirk: Shouldn't the payout be part of the period_cash_flow?

ehebert: You're right it should.

(This may show that the ending_cash is not covered by tests that exercise Futures.)

zipline/finance/performance/period.py
```diff
@@ -207,14 +229,30 @@ def adjust_cash(self, amount):
...
         pt = self.position_tracker
         pos_stats = pt.stats()
         self.ending_value = pos_stats.net_value
         self.ending_exposure = pos_stats.net_exposure
 
+        payout = self._get_payout_total(pt.positions)
+
         total_at_start = self.starting_cash + self.starting_value
-        self.ending_cash = self.starting_cash + self.period_cash_flow
+        self.ending_cash = self.starting_cash + self.period_cash_flow + payout
```
ehebert: @jfkirk here is the fix to the payout and ending_cash.

jfkirk: :+1:

PR #941, opened 2016-01-05
ENH: Adds tick_size and renames futures multiplier

jfkirk:
This PR adds assets db version downgrade management in a new module: asset_db_migrations.py

This PR also has three effects:
'tick_size' is added to Future objects as an informational field. It is not read outside of the object.
The field that was called 'contract_multiplier' is renamed to 'multiplier'
The ASSETS_DB_VERSION is now v1.


tests/test_assets.py
```diff
@@ -1354,3 +1361,30 @@ def test_finder_checks_version(self):
...
+        metadata = sa.MetaData(conn)
+        metadata.reflect(bind=env.engine)
+        version_table = metadata.tables['version_info']
+        check_version_info(version_table, 0)
+
+        # Check some of the v1-to-v0 downgrades
+        self.assertTrue('futures_contracts' in metadata.tables)
+        self.assertTrue('version_info' in metadata.tables)
+        self.assertFalse('tick_size' in
+                         metadata.tables['futures_contracts'].columns)
```
StewartDouglas: How about including `self.assertFalse('multiplier' in metadata.tables['futures_contracts'].columns)` here, to make sure this column has also been successfully deleted?

StewartDouglas:
I'm assuming that alembic also takes care on PRIMARY keys? i.e. we wouldn't need any additional logic added if we wished to rename a primary key?

jfkirk:
Alembic properly migrates the primary keys by making the corresponding columns primary keys on the temp table when executing a batch alteration. You can add/remove a column's primary key status through an alter_column call.

zipline/assets/asset_db_migrations.py
```diff
@@ -0,0 +1,109 @@
...
+    'multiplier' column.
+    """
+    version_info_table.delete().execute()
+
+    # Drop indices before batch
+    # This is to prevent index collision when creating the temp table
+    op.drop_index('ix_futures_contracts_root_symbol')
+    op.drop_index('ix_futures_contracts_symbol')
+
+    # Execute batch op to allow column modification in SQLLite
```
StewartDouglas: It's just SQLite. Sorry to be pedantic :smile:

StewartDouglas:
Apart from the minor points I raised this LGTM

tests/test_assets.py
```diff
@@ -315,7 +321,8 @@ def test_to_and_from_dict(self):
         self.assertTrue('notice_date' in dictd)
         self.assertTrue('expiration_date' in dictd)
         self.assertTrue('auto_close_date' in dictd)
-        self.assertTrue('contract_multiplier' in dictd)
+        self.assertTrue('tick_size' in dictd)
```
ssanderson: Unrelated to this change, but why does this not assert anything about the contents of the dict here?  It would be the same amount of work for this to do:

```
self.assertEqual(dictd['root_symbol'], <expected symbol>)
...
```
or better yet:

```
self.assertEqual(self.future.to_dict(), {key: getattr(self.future, key) for key in dir(Future))
```

jfkirk: I like that - will update

StewartDouglas:
It's dawned on me that we should probably be defining a `tick_size` and `multiplier` columns on the `root_symbols` table. This is because when we see a new futures contract coming into existence, we will need to lookup these values *somewhere*, which is something we don't currently have in place.

StewartDouglas:
LGTM :+1:

PR #942, opened 2016-01-05
BUG: fix bug that caused symbols to be added to the asset finder twice

llllllllll:
When using the cli, assets would be added twice from yahoo because the items of the data panel are strings. By not passing them in the constructor, the run will put them into the asset finder.


richafrank:
(Looked good to me)

PR #943, opened 2016-01-06
Allow OHLCV in slippage models.

ehebert:
Fix bug where the SidView's asset was an integer.

Change BarData to take a simulation dt func instead of simulation
instance, since testing with new interface required creating a BarData
object to pass to slippage. (And it is easier to create a lambda that
provides a dt instead of instantiating or mocking out the
TradeSimulation object.)

### DIFF
--- a/zipline/_protocol.pyx
+++ b/zipline/_protocol.pyx
@@ -23,6 +23,20 @@ cdef class BarData:
     This is what is passed as `data` to the `handle_data` function.
     """
     def __init__(self, data_portal, simulation_dt_func, data_frequency):
+        """
+        Parameters
+        ---------
+        data_portal : DataPortal
+            Provider for bar pricing data.
+
+        simulation_dt_func: function
+            Function which returns the current simulation time.
+            This is usually bound to a method of TradingSimulation.
+
+        data_frequency: string
+            The frequency of the bar data; i.e. whether the data is
+            'daily' or 'minute' bars
+        """
         self.data_portal = data_portal
         self.simulation_dt_func = simulation_dt_func
         self.data_frequency = data_frequency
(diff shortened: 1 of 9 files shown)


zipline/_protocol.pyx
```diff
@@ -22,10 +22,10 @@ cdef class BarData:
 
     This is what is passed as `data` to the `handle_data` function.
     """
-    def __init__(self, data_portal, simulator):
+    def __init__(self, data_portal, simulation_dt_func, data_frequency):
```
jbredeche: can we get some docstrings about `simulation_dt_func` etc?

zipline/_protocol.pyx
```diff
@@ -45,6 +45,11 @@ cdef class BarData:
         try:
             view = self._views[asset]
         except KeyError:
+            try:
+                asset = self.data_portal.env.asset_finder.retrieve_asset(asset)
```
jbredeche: what if asset is already an `Asset`?  is that possible?

ehebert: In that case it is just gets itself back, so a noop-y type operation.

jbredeche: Right, but if `self._views[asset]` failed, doing a no-op on the asset isn't going to solve the problem, right?

ehebert: What this does is ensure that the key is always an Asset derived class. The inverse of our paradigm of calling `int(sid)` when passing to persistence layers.

Both an `int` and corresponding Asset class will both has to the same int value. The `KeyError` here is raised on first attempt to access any asset/sid in from the BarData, whether it is via `data[sid(24)]` or `data[24]`. We then do a `defaultdict`-like behavior and insert a `SidView` on first access. The problem was that if an int was used for the first access the key in `self._views` and asset passed to the `SidView` was an integer. Which lead to issues when trying to call data_portal methods that expected an asset.

A unit test exercising the blotter was creating an Order with an integer `sid`, which uncovered this behavior because of the new use of `BarData` in the blotter.

An alternative fix would be to ensure that both tests and production Order's all have an `Asset` assigned to the `sid` value.

PR #947, opened 2016-01-07
allows users to specify the cutoff time for data query in blaze loaders

llllllllll:
This allows people to set their cutoff time to the time they will
actually execute 'before_trading_start'. Currently this is just passed
to the constructor of the loader; however, I would like to make this
managed by the algorithm simulation runner. This would help keep all of
the loaders in sync and lock 'before_trading_start's execution to the
time the data is queried for.

### DIFF
--- a/zipline/pipeline/loaders/blaze/core.py
+++ b/zipline/pipeline/loaders/blaze/core.py
@@ -127,6 +127,7 @@ from __future__ import division, absolute_import
 from abc import ABCMeta, abstractproperty
 from collections import namedtuple, defaultdict
 from copy import copy
+from datetime import time
 from functools import partial
 from itertools import count
 import warnings
@@ -144,6 +145,7 @@ from datashape import (
 )
 from odo import odo
 import pandas as pd
+from six import with_metaclass, PY2, itervalues, iteritems
 from toolz import (
     complement,
     compose,
@@ -154,15 +156,19 @@ from toolz import (
     memoize,
 )
 import toolz.curried.operator as op
-from six import with_metaclass, PY2, itervalues, iteritems
 
 
 from zipline.pipeline.data.dataset import DataSet, Column
+from zipline.pipeline.loaders.utils import (
+    normalize_data_query_time,
+    normalize_timestamp_to_query_time,
+)
 from zipline.lib.adjusted_array import AdjustedArray
 from zipline.lib.adjustment import Float64Overwrite
 from zipline.utils.enum import enum
-from zipline.utils.input_validation import expect_element
+from zipline.utils.input_validation import expect_element, ensure_timezone
 from zipline.utils.numpy_utils import repeat_last_axis
+from zipline.utils.preprocess import preprocess
 
 
 AD_FIELD_NAME = 'asof_date'
@@ -764,8 +770,25 @@ def adjustments_from_deltas_with_sids(dates,
 
 
 class BlazeLoader(dict):
-    def __init__(self, colmap=None):
+    """A PipelineLoader for datasets constructed with ``from_blaze``.
+
+    Parameters
+    ----------
+    colmap : mapping[BoundColumn -> tuple[Expr, Expr, any]], optional
+        The initial column mapping to use.
+    data_query_time : time, optional
+        The time to use for the data query cutoff.
+    data_query_tz : tzinfo or str
+        The timezeone to use for the data query cutoff.
+    """
+    @preprocess(data_query_tz=ensure_timezone)
+    def __init__(self,
+                 colmap=None,
+                 data_query_time=time(0),
+                 data_query_tz='utc'):
         self.update(colmap or {})
+        self._data_query_time = data_query_time
+        self._data_query_tz = data_query_tz
 
     @classmethod
     @memoize(cache=WeakKeyDictionary())
@@ -802,6 +825,19 @@ class BlazeLoader(dict):
             [SID_FIELD_NAME] if have_sids else []
         )
 
+        data_query_time = self._data_query_time
+        data_query_tz = self._data_query_tz
+        lower_dt = normalize_data_query_time(
+            dates[0],
+            data_query_time,
+            data_query_tz,
+        )
+        upper_dt = normalize_data_query_time(
+            dates[-1],
+            data_query_time,
+            data_query_tz,
+        )
+
         def where(e):
             """Create the query to run against the resources.
 
@@ -819,8 +855,8 @@ class BlazeLoader(dict):
             # Hack to get the lower bound to query:
             # This must be strictly executed because the data for `ts` will
             # be removed from scope too early otherwise.
-            lower = odo(ts[ts <= dates[0]].max(), pd.Timestamp)
-            selection = ts <= dates[-1]
+            lower = odo(ts[ts <= lower_dt].max(), pd.Timestamp)
+            selection = ts <= upper_dt
             if have_sids:
                 selection &= e[SID_FIELD_NAME].isin(assets)
             if lower is not pd.NaT:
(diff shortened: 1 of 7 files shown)


zipline/pipeline/loaders/utils.py
```diff
@@ -0,0 +1,72 @@
...
+                                      ts_field='timestamp'):
+    """Update the timestamp field of a dataframe to normalize dates around
+    some data query time/timezone.
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        The dataframe to update. This needs a column named ``ts_field``.
+    time : datetime.time
+        The time to query before.
```
richafrank: What do you mean by "query" here?

zipline/pipeline/loaders/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        true, then this will be the same object as ``df`` otherwise this will
+        be a copy.
+    """
+    if not inplace:
+        # don't mutate the dataframe in place
+        df = df.copy()
+
+    dtidx = pd.DatetimeIndex(df.loc[:, ts_field], tz='utc')
+    dtidx_local_time = dtidx.tz_convert(tz)
+    to_roll_forward = dtidx_local_time.time > time
```
richafrank: Do we want to roll forward times gt or gte?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -830,11 +866,32 @@ def where(e):
...
+            inplace=True,
+            ts_field=TS_FIELD_NAME,
+        )
+        normalize_timestamp_to_query_time(
+            materialized_deltas,
+            data_query_time,
+            data_query_tz,
+            inplace=True,
+            ts_field=TS_FIELD_NAME,
+        )
```
richafrank: Since we're `astype`ing and normalizing the ts_field for both the expr and deltas, should we do it in a loop instead of duplicating those statements?

zipline/pipeline/loaders/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    # for all of the times that are greater than our query time add 1
+    # day and truncate to the date
+    df.loc[to_roll_forward, ts_field] = (
+        dtidx_local_time[to_roll_forward] + datetime.timedelta(days=1)
```
richafrank: Do we need to take into account the trading calendar?

zipline/pipeline/loaders/blaze/earnings.py
```diff
@@ -121,20 +148,32 @@ def load_adjusted_array(self, columns, dates, assets, mask):
             # range. It must all be null anyways.
             lower = dates[0]
```
richafrank: Do we need to be normalizing `lower`?

zipline/pipeline/loaders/utils.py
llllllllll: I think we want inclusive, I will change and update the tests

zipline/pipeline/loaders/blaze/earnings.py
llllllllll: we ffill from the last known value anyway so it shouldn't matter.

zipline/pipeline/loaders/utils.py
llllllllll: Nope, this is marking the date we would have knowledge of the event. This does not need to align with the trading calendar. For example: things we learn about on saturday become available to the algo on monday.

zipline/pipeline/loaders/utils.py
llllllllll: updating both docstrings to show:

```
    time : datetime.time
        The time of day to use as the cutoff point for new data. Data points
        that you learn about after this time will become available to your
        algorithm on the next trading day.
    tz : tzinfo
        The timezone to normalize your dates to before comparing against
        `time`.
```

zipline/pipeline/loaders/utils.py
llllllllll: actually, I am reconsidering this. I think a sane default is to use midnight and if this is ge then things that are timestamped at exactly midnight get cut off.

zipline/pipeline/loaders/utils.py
richafrank: Makes sense.

zipline/pipeline/loaders/utils.py
```diff
@@ -93,3 +95,78 @@ def previous_date_frame(date_index, events_by_sid):
...
+        The dataframe with the timestamp field normalized. If ``inplace`` is
+        true, then this will be the same object as ``df`` otherwise this will
+        be a copy.
+    """
+    if not inplace:
+        # don't mutate the dataframe in place
+        df = df.copy()
+
+    dtidx = pd.DatetimeIndex(df.loc[:, ts_field], tz='utc')
+    dtidx_local_time = dtidx.tz_convert(tz)
```
richafrank: Would there be much of a perf gain to skip all the conversions if `tz == 'utc'`?

zipline/pipeline/loaders/utils.py
richafrank: Nice - should we change the other uses of the word "query" throughout this?

zipline/pipeline/loaders/utils.py
llllllllll: I don't think so, `tz_convert` shouldn't affect the raw representation of the dates:

```python
In [7]: %timeit big.tz_convert('US/Eastern')
The slowest run took 8.15 times longer than the fastest. This could mean that an intermediate result is being cached 
100000 loops, best of 3: 8.25 µs per loop

In [8]: len(big)
Out[8]: 115417
```

llllllllll:
updated based on in person discussion.

zipline/pipeline/loaders/utils.py
```diff
@@ -93,3 +95,136 @@ def previous_date_frame(date_index, events_by_sid):
...
+        `time`.
+    """
+    # Subtract one day to grab things that happend on the first day we are
+    # requesting. This doesn't need to be a trading day, we are only adding
+    # a lower bound to limit the amount of in memory filtering that needs
+    # to happen.
+    lower -= datetime.timedelta(days=1)
+    if time is not None:
+        return normalize_data_query_time(
+            lower - datetime.timedelta(days=1),
```
richafrank: Is this further subtraction intentional?

llllllllll: nope

zipline/pipeline/loaders/utils.py
```diff
@@ -93,3 +95,136 @@ def previous_date_frame(date_index, events_by_sid):
...
+        The upper date requested.
+    time : datetime.time
+        The time of day to use as the cutoff point for new data. Data points
+        that you learn about after this time will become available to your
+        algorithm on the next trading day.
+    tz : tzinfo
+        The timezone to normalize your dates to before comparing against
+        `time`.
+    """
+    # Subtract one day to grab things that happend on the first day we are
```
richafrank: typo "happend"

zipline/pipeline/loaders/utils.py
```diff
@@ -93,3 +95,136 @@ def previous_date_frame(date_index, events_by_sid):
...
+    ----------
+    data_query_time : datetime.time or None
+    data_query_tz : tzinfo or None
+
+    Raises
+    ------
+    ValueError
+        Raised when only one of the arguments is None.
+    """
+    if (data_query_time is None) ^ (data_query_tz is None):
```
richafrank: I think I would probably use `!=`, since this is on booleans, but bitwise xor works too.

llllllllll: I think the xor more accuratly describes the check that I am trying to do. I want to fail if exactly one is true.

llllllllll: also, `__xor__` is not inherited from `int`, so it will return a `bool`.

richafrank: I totally follow and can see cases for both.  Feel free to leave it.

richafrank:
Looks like a test is still failing.

llllllllll:
merging on pass

PR #949, opened 2016-01-13
Use earliest trading day from bcolz readers to calculate valid history windows

jbredeche:
When getting a daily history window, we were using the asset's start date to figure out the first valid day for the history window. The problem is that the asset's start date can be before the first day in the daily bcolz file.

This PR makes it so that we use the max(asset start date, first trading day in bcolz file) to size history windows.  It also adds the property to the daily bcolz writer and reader.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -1093,12 +1093,14 @@ class TradingAlgorithm(object):
 
     @api_method
     @require_initialized(HistoryInInitialize())
-    def history(self, sids, bar_count, frequency, field, ffill=True):
-        if self.data_portal is None:
-            raise Exception("no data portal!")
+    def history(self, assets, bar_count, frequency, field, ffill=True):
+        try:
+            assets = list(assets)
+        except TypeError:
+            assets = [assets]
 
         return self.data_portal.get_history_window(
-            sids,
+            assets,
             self.datetime,
             bar_count,
             frequency,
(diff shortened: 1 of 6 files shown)


zipline/algorithm.py
```diff
@@ -1093,12 +1093,14 @@ def cancel_order(self, order_param):
 
     @api_method
     @require_initialized(HistoryInInitialize())
-    def history(self, sids, bar_count, frequency, field, ffill=True):
-        if self.data_portal is None:
-            raise Exception("no data portal!")
+    def history(self, assets, bar_count, frequency, field, ffill=True):
+        try:
```
ssanderson: What's this about?

zipline/data/us_equity_pricing.py
```diff
@@ -382,6 +391,16 @@ def __init__(self, table):
             int(id_): offset
             for id_, offset in iteritems(table.attrs['calendar_offset'])
         }
+
+        if "first_trading_day" in table.attrs.attrs:
```
ehebert: Both `table.attrs.attrs` and `table.attrs` are referenced, is that intended?

Usually in this 'if available use the dict value' paradigm, we use KeyError, e.g. 

```
try:
    self._first_trading_day = Timestamp(tables.attrs['first_trading_day'], unit='ms', tz='UTC'))
except KeyError:
    self._first_trading_day = None

zipline/algorithm.py
jbredeche: Just to be nice to users.  If they pass in a single asset, and it's painfully obvious what they were trying to do, I figured it would be nicer just to do what they wanted.

Furthermore, if they do pass in a single asset, I didn't have any checks verifying that.  So it was blowing up much deeper in dataportal, instead of earlier.  After adding the check, I figured I might as well coerce for them.

PR #951, opened 2016-01-15
DEV: conda recipe work

llllllllll:


ssanderson:
@llllllllll @richafrank is this worth keeping around since we're no longer considering anaconda build?

llllllllll:
there are other changes to the recipe in here that we probably still want. We need to pick out the needed changes before merging.

PR #957, opened 2016-01-20
set benachmark_return_source as an input and not always None

npezolano:


llllllllll:
Thanks for the pr!
It looks like this was also an argument to `run`; however, in `run` it was unused. I think that instead of making this an argument to the tradingalgorithm we should just set `self.benchmark_return_source` in `algo.run`. Would that satisfy your use case?

Also, could you please include a test that uses a custom benchmark return source so that we do not break this in the future?

llllllllll:
Also, the test failure seems unrelated, I will try to look into this tonight or tomorrow

npezolano:
It look's like the `benchmark_return_source` in `algo.run` never gets used and can be removed.

llllllllll:
Sorry, I see that it is never used. I am proposing that we should start using that argument to set the benchmark return source instead of adding it as an argument to `TradingAlgorithm`

npezolano:
They have a test here that uses it in a different way, https://github.com/quantopian/zipline/blob/3b769812706e525365f3c332fd69e2be79d612ed/tests/test_events_through_risk.py#L138. I can't get either of them to work.

Issue #926, opened 2015-12-21
Using pd.Panel as a data source does not work as expected.


elmq0022:
I am on zipline version 0.8.3 working on windows with python 2.7.  The items in Requirements.txt were installed via conda install and zipline was installed via pip.  

I have a pd.Panel of ~500 assets that are not loading correctly into the algorithm.run(pd.Panel) routine. 

Basically I was expecting the asset name (string) on the major_index to be converted to an appropriate object and mapped to a sid. This seems to only work about halfway.  

The **sid** and **symbol** functions from zipline.api do seem to work after the data is loaded, but when I grab prices from the history function the resulting data frame has column names that are a list of integers (not assets?).  So obviously the following does not work:

```python
prices = history(100, '1d', 'price').dropna()

for stock, wt in zip(prices.columnames, wts):
    order_target_percent(stock, wt)
```

I tried to work around this by casting the integer to its asset using **sid**(stock).  This pushed me past the initial error, but this still fails as cached price data only has integers as keys (instead of assets?).  So when the order method tires to look up the price it throws a KeyError (all the keys are ints and I'm passing equity assets).  I believe stack trace crashed in protocol.py

So, is  this just *user error* on my part?  Or, is something else off?

Any help is greatly appreciated.  Thanks!

EDIT:
I went back and tried the routine with data created from the  load_bars_from_yahoo function and I get the same error.  I feel like there is something odd is happening here.

llllllllll:
Hi, thanks for the bug report. I think this might be related to an issue I found in `TradingAlgorithm.run` where sids were getting added to the asset finder twice. I will try to repro this and get back to you with what I find.

elmq0022:
Thank you for following up.

elmq0022:
I did some additional debugging.  Basically I see bugs based on the transtion from the sid to generic assets design.  Is this correct?

The ```order``` function validates that it received an asset object (which is incorrect?) in the ```validate_order_params``` function.  

So, I was able to backtest with a monkey patched TradingAlgorithm by setting ```validate_order_params``` to ```pass```. Ignoring this asset validation, the order goes though based on just the provided sid.

I'm sure this is not the best way to fix this, but it works (for  now).  I guess there are more general design issues to resolve which is complicated by several potential data sources.  

Perhaps the construction of the data sources is the correct place to address this?   Would passing an list of assets as the ```pd.Panel.index``` resolve this?  I havent' tried it yet.  

Please let me know if I'm on the right track.

Thanks!

[...]

Issue #650, opened 2015-07-14
Tutorial buyapple.py fails in python 2.7.10


beebeed:
running buyapple.py fails in python 2.7.10 but runs in python 3.4.3 for zipline 8.0rc. The part of code that fails is `data[symbol('AAPL')]` found in
```
record(AAPL=data[symbol('AAPL')].price)
```
the error is
```
Traceback (most recent call last):
  File "C:\python2710_64\Scripts\run_algo.py", line 4, in <module>
    __import__('pkg_resources').run_script('zipline==0.8.0rc1', 'run_algo.py')
  File "C:\Python2710_64\lib\site-packages\pkg_resources\__init__.py", line 729, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "C:\Python2710_64\lib\site-packages\pkg_resources\__init__.py", line 1642, in run_script
    exec(code, namespace, namespace)
  File "c:\python2710_64\lib\site-packages\zipline-0.8.0rc1-py2.7-win-amd64.egg\EGG-INFO\scripts\run_algo.py", line 23, in <module>
    run_pipeline(print_algo=True, **parsed)
  File "C:\Python2710_64\lib\site-packages\zipline-0.8.0rc1-py2.7-win-amd64.egg\zipline\utils\cli.py", line 246, in run_pipeline
    perf = algo.run(source, overwrite_sim_params=overwrite_sim_params)
  File "C:\Python2710_64\lib\site-packages\zipline-0.8.0rc1-py2.7-win-amd64.egg\zipline\algorithm.py", line 529, in run
    for perf in self.gen:
  File "C:\Python2710_64\lib\site-packages\zipline-0.8.0rc1-py2.7-win-amd64.egg\zipline\gens\tradesimulation.py", line 120, in transform
    self.algo.instant_fill,
  File "C:\Python2710_64\lib\site-packages\zipline-0.8.0rc1-py2.7-win-amd64.egg\zipline\gens\tradesimulation.py", line 314, in _process_snapshot
    new_orders = self._call_handle_data()
  File "C:\Python2710_64\lib\site-packages\zipline-0.8.0rc1-py2.7-win-amd64.egg\zipline\gens\tradesimulation.py", line 343, in _call_handle_data
    self.simulation_dt,
  File "C:\Python2710_64\lib\site-packages\zipline-0.8.0rc1-py2.7-win-amd64.egg\zipline\utils\events.py", line 194, in handle_data
    event.handle_data(context, data, dt)
[...]

PR #959, opened 2016-01-20
Asset KeyError

richafrank:
Enable comparison of an Asset to an int64 on 32-bit python

by supporting any Integral. We use a number of mappings keyed by
int64, which otherwise raise KeyErrors for Assets.

Also ensures `algo.current_sids()` returns a list of Assets, instead of identifiers.

For #650 and #926

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -29,7 +29,7 @@ from pandas.tseries.tools import normalize_date
 from pandas.util.testing import assert_frame_equal
 
 from nose_parameterized import parameterized
-from numpy import full
+from numpy import full, int32, int64
 import sqlalchemy as sa
 
 from zipline.assets import (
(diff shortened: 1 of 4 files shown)


tests/test_assets.py
```diff
@@ -212,6 +212,7 @@ def test_asset_comparisons(self):
         self.assertEqual(s_23, s_23)
         self.assertEqual(s_23, 23)
         self.assertEqual(23, s_23)
+        self.assertEqual(int64(23), s_23)
```
llllllllll: let's also check np.int32 to be safe

zipline/assets/_assets.pyx
```diff
@@ -86,14 +88,20 @@ cdef class Asset:
...
+            x_as_int = int(x)
         else:
             return NotImplemented
 
         if isinstance(y, Asset):
             y_as_int = y.sid
         elif isinstance(y, int):
+            # ints are Integral, but much faster to special case
             y_as_int = y
+        elif isinstance(y, Integral):
```
llllllllll: as we discussed, maybe make this `isinstance(y, (np.int64, np.int32, Integral))` for performance.

richafrank: Added `long` as well.

PR #960, opened 2016-01-20
ENH: Add writer for equity minute bcolz format.

ehebert:
Implement a writer for minute data into a format comprised of multiple
ctables, one for each individual equity, with a common 'index' shared by
all ctables where a given a dt maps to the same array index for all
equities and fields.

This format is pulled from the lazy-mainline/Q2.0 branch, with some
changes to the interface.

Add basic retrieval of values at a given dt to reader. Not yet used by
Zipline simulations, but added to support unit tests.

# Other Notes

Questions:
- Currently the main entry point into the writer is `write(sid, days, df)`, where the df is a DataFrame for just the given sid. 
-- Providing `days` does feel redundant, when that information could be re-constructed from the `df` and the writer's `self._trading_days`.
-- `sid` could be also be redundant, if the df were formatted with a `MultiIndex` of (sid, dt), then the sids could be iterated over using `get_level_values`.

@ssanderson I took in a lot of notes from our in person conversation the other day.
Including taking in market_opens as a parameter, and then writing out the resulting index to free the writer/reader combo from requiring a `TradingEnvironment` or `tradingcalendar` module.

Where I did diverge from our discussion was:
- The writer class no longer requires subclassing to implement different sources, instead the `write` method takes a DataFrame with the OHLCV data. This change resulted from using the writer with our existing and incoming internal dataloader. I found having a writer which was agnostic of its input source more flexible to work with when wiring up different sources, and I believe is a more adoptable public API.

For some background on the choice of having the Writer agnostic to the input, I looked to the following:
- https://docs.python.org/2/library/csv.html
- https://golang.org/pkg/bufio/#example_Writer
- http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.to_excel.html

- The responsibility of the writer is just writing individual sid data, without adding the 'bring data up to date' logic that our internal loader provides. This change allowed dropping the writer into our existing internal loader. If/when we do make that logic a public API, I believe we should make a separate `Loader` (or some other name, `DataSet`/`DataSetManager`, I am not yet sure), which takes the writer as one of its components.

@jbredeche this implementation differs from our implementation in lazy-mainline in the following ways.
(For reference, https://github.com/quantopian/zipline/blob/lazy-mainline/zipline/data/us_equity_minutes.py)
- The aforementioned change to no longer using subclasses to get the data, instead passing in the DataFrame independent of the data source.
- Making the directory as part of the constructor. This change was 1) based on the above list of other Reader/Writers, 2) was more natural when using the writer with our existing internal loader 3) Allows more flexibility if we ever need to implement another writer that does not write to a directory, e.g. if a Postgres or SQLite writer is implemented then `write(sid, days, df)` would also work with a writer of that type.

- Allowing multiple passes on writing data, which is a requirement of our internal writers which read the source data in subgroups of days and equities. (Instead of only supporting reading all data for an equity at once.)

### DIFF
--- a/zipline/data/us_equity_minutes.py
+++ b/zipline/data/us_equity_minutes.py
@@ -251,10 +251,6 @@ class BcolzMinuteBarWriter(object):
     def first_trading_day(self):
         return self._first_trading_day
 
-    @property
-    def metadata_path(self):
-        return os.path.join(self._rootdir, METADATA_FILENAME)
-
     def sidpath(self, sid):
         """
         Parameters:
@@ -362,6 +358,11 @@ class BcolzMinuteBarWriter(object):
         df : pd.DataFrame
             DataFrame of market data with the following characteristics.
             columns : ('open', 'high', 'low', 'close', 'volume')
+                open : float64
+                high : float64
+                low  : float64
+                close : float64
+                volume : float64|int64
             index : DatetimeIndex of market minutes.
         """
         table = self._ensure_ctable(sid)
--- a/zipline/data/us_equity_minutes.py
+++ b/zipline/data/us_equity_minutes.py
@@ -54,7 +54,7 @@ def _sid_subdir_path(sid):
 
     Parameters:
     -----------
-    sid - int
+    sid : int
         Asset identifier.
 
     Returns:
--- a/zipline/data/us_equity_minutes.py
+++ b/zipline/data/us_equity_minutes.py
@@ -25,7 +25,6 @@ from pandas.core.datetools import normalize_date
 MINUTES_PER_DAY = 390
 DEFAULT_EXPECTEDLEN = 390 * 252 * 15
 OHLC_RATIO = 1000
-METADATA_FILENAME = 'metadata.json'
 
 
 class BcolzMinuteOverlappingData(Exception):
@@ -74,6 +73,52 @@ def _sid_subdir_path(sid):
         padded_sid[2:4],
         "{0}.bcolz".format(str(padded_sid))
     )
+
+
+class BcolzMinuteBarMetadata(object):
+
+    METADATA_FILENAME = 'metadata.json'
+
+    @classmethod
+    def metadata_path(cls, rootdir):
+        return os.path.join(rootdir, cls.METADATA_FILENAME)
+
+    @classmethod
+    def read(cls, rootdir):
+        path = cls.metadata_path(rootdir)
+        with open(path) as fp:
+            raw_data = json.load(fp)
+            first_trading_day = pd.Timestamp(
+                raw_data['first_trading_day'], tz='UTC')
+            minute_index = pd.to_datetime(raw_data['minute_index'],
+                                          utc=True)
+            ohlc_ratio = raw_data['ohlc_ratio']
+            return cls(first_trading_day, minute_index, ohlc_ratio)
+
+    def __init__(self, first_trading_day, minute_index, ohlc_ratio):
+        self.first_trading_day = first_trading_day
+        self.minute_index = minute_index
+        self.ohlc_ratio = ohlc_ratio
+
+    def write(self, rootdir):
+        """
+        Write the metadata to a JSON file in the rootdir.
+
+        Values contained in the metadata are:
+        first_trading_day : string
+            'YYYY-MM-DD' formatted representation of the first trading day
+             available in the dataset.
+        minute_index : list of integers
+             nanosecond integer representation of the minutes, the enumeration
+             of which corresponds to the values in each bcolz carray.
+        """
+        metadata = {
+            'first_trading_day': str(self.first_trading_day.date()),
+            'minute_index': self.minute_index.asi8.tolist(),
+            'ohlc_ratio': self.ohlc_ratio,
+        }
+        with open(self.metadata_path(rootdir), 'w+') as fp:
+            json.dump(metadata, fp)
 
 
 class BcolzMinuteBarWriter(object):
@@ -131,7 +176,7 @@ class BcolzMinuteBarWriter(object):
         -----------
         first_trading_day : datetime-like
             The first trading day in the data set.
-        
+
         rootdir : string
             Path to the root directory into which to write the metadata and
             bcolz subdirectories.
@@ -151,8 +196,8 @@ class BcolzMinuteBarWriter(object):
             of minutes in NYSE trading days.
 
         ohlc_ratio : int
-            The ratio by which to multiply the pricing data to convert the floats
-            from floats to an integer to fit within the np.uint32.
+            The ratio by which to multiply the pricing data to convert the
+            floats from floats to an integer to fit within the np.uint32.
 
             The default is 1000 to support pricing data which comes in to the
             thousands place.
@@ -166,7 +211,7 @@ class BcolzMinuteBarWriter(object):
 
             Defaults to supporting 15 years of NYSE equity market data.
 
-            see: http://bcolz.blosc.org/opt-tips.html#informing-about-the-length-of-your-carrays
+            see: http://bcolz.blosc.org/opt-tips.html#informing-about-the-length-of-your-carrays # noqa
         """
         self._rootdir = rootdir
         self._first_trading_day = first_trading_day
@@ -180,7 +225,12 @@ class BcolzMinuteBarWriter(object):
         self._minute_index = _calc_minute_index(
             self._market_opens, self._minutes_per_day)
 
-        self._write_metadata()
+        metadata = BcolzMinuteBarMetadata(
+            self._first_trading_day,
+            self._minute_index,
+            self._ohlc_ratio,
+        )
+        metadata.write(self._rootdir)
 
     @property
     def first_trading_day(self):
@@ -189,26 +239,6 @@ class BcolzMinuteBarWriter(object):
     @property
     def metadata_path(self):
         return os.path.join(self._rootdir, METADATA_FILENAME)
-
-    def _write_metadata(self):
-        """
-        Write the metadata to a JSON file in the rootdir.
-
-        Values contained in the metadata are:
-        first_trading_day : string
-            'YYYY-MM-DD' formatted representation of the first trading day
-             available in the dataset.
-        minute_index : list of integers
-             nanosecond integer representation of the minutes, the enumeration of which
-             corresponds to the values in each bcolz carray.
-        """
-        metadata = {
-            'first_trading_day': str(self.first_trading_day.date()),
-            'minute_index': self._minute_index.asi8.tolist(),
-            'ohlc_ratio': self._ohlc_ratio,
-        }
-        with open(self.metadata_path, 'w+') as fp:
-            json.dump(metadata, fp)
 
     def sidpath(self, sid):
         """
@@ -406,9 +436,9 @@ class BcolzMinuteBarReader(object):
 
         metadata = self._get_metadata()
 
-        self._first_trading_day = metadata['first_trading_day']
-        self._minute_index = metadata['minute_index']
-        self._ohlc_inverse = metadata['ohlc_inverse']
+        self._first_trading_day = metadata.first_trading_day
+        self._minute_index = metadata.minute_index
+        self._ohlc_inverse = 1.0 / metadata.ohlc_ratio
 
         self._carrays = {
             'open': {},
@@ -419,16 +449,7 @@ class BcolzMinuteBarReader(object):
         }
 
     def _get_metadata(self):
-        with open(os.path.join(self.rootdir, METADATA_FILENAME)) as fp:
-            raw_data = json.load(fp)
-            metadata = {
-                'first_trading_day': pd.Timestamp(
-                    raw_data['first_trading_day'], tz='UTC'),
-                'minute_index': pd.to_datetime(raw_data['minute_index'],
-                                               utc=True),
-                'ohlc_inverse': 1.0 / raw_data['ohlc_ratio']
-            }
-            return metadata
+        return BcolzMinuteBarMetadata.read(self.rootdir)
 
     def _get_carray_path(self, sid, field):
         sid_subdir = _sid_subdir_path(sid)
@@ -458,7 +479,7 @@ class BcolzMinuteBarReader(object):
         dt : datetime-like
             The datetime at which the trade occurred.
         field : string
-            The type of pricing data to retrieve. 
+            The type of pricing data to retrieve.
             ('open', 'high', 'low', 'close', 'volume')
 
         Returns:
--- a/zipline/data/us_equity_minutes.py
+++ b/zipline/data/us_equity_minutes.py
@@ -413,20 +413,16 @@ class BcolzMinuteBarWriter(object):
         close_col = np.zeros(minutes_count, dtype=np.uint32)
         vol_col = np.zeros(minutes_count, dtype=np.uint32)
 
-        if df is not None:
-            dt_ixs = np.searchsorted(all_minutes_in_window.values,
-                                     df.index.values)
+        dt_ixs = np.searchsorted(all_minutes_in_window.values,
+                                 df.index.values)
 
-            ohlc_ratio = self._ohlc_ratio
-            open_col[dt_ixs] = (df.open.values * ohlc_ratio).astype(np.uint32)
-            high_col[dt_ixs] = (df.high.values * ohlc_ratio).astype(np.uint32)
-            low_col[dt_ixs] = (df.low.values * ohlc_ratio).astype(np.uint32)
-            close_col[dt_ixs] = (df.close.values * ohlc_ratio).astype(
-                np.uint32)
-            vol_col[dt_ixs] = df.volume.values.astype(np.uint32)
-        else:
-            # If df is None, add zero rows for the given days.
-            pass
+        ohlc_ratio = self._ohlc_ratio
+        open_col[dt_ixs] = (df.open.values * ohlc_ratio).astype(np.uint32)
+        high_col[dt_ixs] = (df.high.values * ohlc_ratio).astype(np.uint32)
+        low_col[dt_ixs] = (df.low.values * ohlc_ratio).astype(np.uint32)
+        close_col[dt_ixs] = (df.close.values * ohlc_ratio).astype(
+            np.uint32)
+        vol_col[dt_ixs] = df.volume.values.astype(np.uint32)
 
         table.append([
             open_col,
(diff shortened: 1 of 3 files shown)


zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+    """
+    Format subdir path to limit the number directories in any given
+    subdirectory to 100.
+
+    The number in each directory is designed to support at least 100000
+    equities.
+
+    Parameters:
+    -----------
+    sid - int
```
ssanderson: The general format for this is usually `name : type` rather than `name - type`

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+        Asset identifier.
+
+    Returns:
+    --------
+    string
```
ssanderson: The format for this is usually the same as the input.  For example. `np.array` does:

```
Returns
-------
out : ndarray
    An array object satisfying the specified requirements.
```

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+            thousands place.
+
+        expectedlen : int
+            The expected length of the dataset, used when creating the initial
+            bcolz ctable.
+
+            If the expectedlen is not used, the chunksize and corresponding
+            compression ratios are not ideal.
+
+            Defaults to supporting 15 years of market data.
```
ssanderson: This is 15 years of market data for NYSE equities, right?  Might want to specify that here.

zipline/data/us_equity_minutes.py
ehebert: Will fix, I believe I used a ':' most everywhere else in this module.

zipline/data/us_equity_minutes.py
ehebert: Will do.

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+        
+        rootdir : string
+            Path to the root directory into which to write the metadata and
+            bcolz subdirectories.
+
+        market_opens : pd.Series
+            The market opens used as a starting point for each periodic span of
+            minutes in the index.
+
+        minutes_per_day : int
```
ssanderson: I don't think this should be optional.  `expectedlen` and `ohlc_ratio` don't have any bearing on correctness.  `minutes_per_day` does.

ssanderson: Also, the general format for optional values should be:
```
name : type, optional
    Sentence describing parameter.  Default value is <default>.
```
c.f. https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections

zipline/data/us_equity_minutes.py
ehebert: Fixed this and other `Returns` in docstrings.

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+        """
+        Parameters:
+        -----------
+        sid : int
+            Asset identifier.
+
+        Returns:
+        --------
+        pd.Timestamp
+            The midnight of the last date written in to the output for the
```
ssanderson: typo `into` should be one word.

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+            given sid.
+        """
+        try:
+            sizes_path = "{0}/close/meta/sizes".format(self.sidpath(sid))
```
ssanderson: This feels like it's relying a lot on bcolz implementation details.  I'd expect that a safer way to do this would be to make a `ctable` or `carray` and call `len` on it.  Does that not work and/or is it too slow?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+        except IOError:
+            return None
+
+    def _init_ctable(self, path):
+        """
+        Create empty ctable for given path.
+
+        Parameters:
+        -----------
+
```
ssanderson: extra whitespace here

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+        path : string
+            The path to rootdir of the new ctable.
+        """
+        # Only create the subdir on container creation.
+        sid_dirname = os.path.dirname(path)
+        if not os.path.exists(sid_dirname):
```
ssanderson: It looks like `_init_ctable` is only ever called from `write`, which already ensures that the appropriate directory exists.  Is this necessary?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+                 first_trading_day,
+                 rootdir,
+                 market_opens,
+                 minutes_per_day=MINUTES_PER_DAY,
+                 ohlc_ratio=OHLC_RATIO,
+                 expectedlen=DEFAULT_EXPECTEDLEN):
+        """
+        Parameters:
+        -----------
+
```
ssanderson: extra whitespace

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+                raise BcolzMinuteOverlappingData(dedent("""
+                window start={0} is before expected write date={1} for
+                sid={2}""".strip()).format(days[0], input_first_day, sid))
+            else:
+                days_to_zerofill = None
+
+        if days_to_zerofill is not None and len(days_to_zerofill):
+            self._zerofill(table, len(days_to_zerofill))
+
+        days_to_write_indexer = self._trading_days.slice_indexer(
```
ssanderson: Up above you've already aliased this as `tds`.

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+        indexer = all_minutes.slice_indexer(start=days_to_write[0])
+        all_minutes_in_window = all_minutes[indexer]
+
+        open_col = np.zeros(minutes_count, dtype=np.uint32)
+        high_col = np.zeros(minutes_count, dtype=np.uint32)
+        low_col = np.zeros(minutes_count, dtype=np.uint32)
+        close_col = np.zeros(minutes_count, dtype=np.uint32)
+        vol_col = np.zeros(minutes_count, dtype=np.uint32)
+
+        if df is not None:
```
ssanderson: When would I want to call this with `df = None`?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+        -----------
+
+        first_trading_day : datetime-like
+            The first trading day in the data set.
+        
+        rootdir : string
+            Path to the root directory into which to write the metadata and
+            bcolz subdirectories.
+
+        market_opens : pd.Series
```
ssanderson: What's the expected structure of this Series?  Mapping from midnight UTC -> start_date UTC?  Might be worth documenting that here.

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+            market_opens.index.slice_indexer(start=self._first_trading_day)]
+        self._trading_days = market_opens.index
+        self._minutes_per_day = minutes_per_day
+        self._expectedlen = expectedlen
+        self._ohlc_ratio = ohlc_ratio
+
+        self._minute_index = _calc_minute_index(
+            self._market_opens, self._minutes_per_day)
+
+        self._write_metadata()
```
ssanderson: Seems like we shouldn't be eagerly writing metadata if there's a separate `write` call being made later?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,402 @@
...
+
+    @property
+    def first_trading_day(self):
+        return self._first_trading_day
+
+    @property
+    def metadata_path(self):
+        return os.path.join(self._rootdir, METADATA_FILENAME)
```
ssanderson: Seems like this path construction logic should be broken out into a function that's shared between the reader and the writer, ala:

```
def bcolz_metadata_path(rootdir):
    return os.path.join(rootdir, METADATA_FILENAME)
```

zipline/data/us_equity_minutes.py
ehebert: I think it would benefit usage of BcolzEquityMinuteWriter to not need to pass the minutes per day on every instantiation.

What do you think of turning this class into `BcolzMinuteWriter` and then subclassing a `BcolzUSEquitiesMinuteWriter` which provides the minutes per day as a default?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,401 @@
...
+
+        Returns:
+        --------
+        out : string
+            Full path to the bcolz rootdir for the given sid.
+        """
+        sid_subdir = _sid_subdir_path(sid)
+        return join(self._rootdir, sid_subdir)
+
+    def last_date_in_output_for_sid(self, sid):
```
ssanderson: Style nit: I'm not sure we need `for_sid` in the name here.  Seems like that's clear enough from the signature?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,401 @@
...
+        If the length of the bcolz ctable is not exactly to the date before
+        the first day provided, fill the ctable with 0s up to that date.
+
+        Writes in blocks of the size of the days times minutes per day.
+
+        Parameters:
+        -----------
+        sid : int
+            The asset identifer for the data being written.
+        days : pd.DatetimeIndex
```
ssanderson: When would `days` be different from the index of the DataFrame itself.  Do we need this as a separate argument?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,401 @@
...
+            The days for which to write data from the given df.
+        df : pd.DataFrame
+            DataFrame of market data with the following characteristics.
+            columns : ('open', 'high', 'low', 'close', 'volume')
+            index : DatetimeIndex of market minutes.
+        """
+        path = self.sidpath(sid)
+        if not os.path.exists(path):
+            self._init_ctable(path)
```
ssanderson: Seems like `init_ctable` should just returned the newly `init`ed ctable?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,401 @@
...
+            sizes_path = "{0}/close/meta/sizes".format(self.sidpath(sid))
+            with open(sizes_path, mode='r') as f:
+                sizes = f.read()
+            data = json.loads(sizes)
+            num_days = data['shape'][0] / self._minutes_per_day
+            if num_days == 0:
+                # empty container
+                return None
+            return self._trading_days[num_days - 1]
+        except IOError:
```
ssanderson: When do we expect to hit this case?  Can we document that this will return None in the `Returns` section?

ssanderson: When do we expect to hit this case?  Can we document that this will return None in the `Returns` section?

zipline/data/us_equity_minutes.py
ehebert: It does work, but is too slow. Especially when the number of individual ctables is around ~10000.

zipline/data/us_equity_minutes.py
ehebert: It is not necessary, but a force of habit when calling `makedirs`. Will remove.

zipline/data/us_equity_minutes.py
ehebert: Removed, as well as other extra lines Parameters..

zipline/data/us_equity_minutes.py
ehebert: Good catch, changing over to use alias.

zipline/data/us_equity_minutes.py
ehebert: In a case where an asset was active over the date range, but had no trades.

zipline/data/us_equity_minutes.py
ehebert: Good call, updating.

zipline/data/us_equity_minutes.py
ssanderson: > I think it would benefit usage of BcolzEquityMinuteWriter to not need to pass the minutes per day on every instantiation.

How often are you realistically constructing one of these?  I buy the argument to ergonomics for things like DataFrame where you expect to use the constructor all over the place.  But any given codebase is going to have very few invocations of this class, and the implicit behavior here is pretty deeply wedded to Quantopian's particular usage.

zipline/data/us_equity_minutes.py
ssanderson: Or better yet, could everything up until line 313 be refactored to something like:

```
def ensure_ctable(self, sid):
    """Ensure that a ctable exists for ``sid``."""
    sidpath = self.sidpath(sid)
    if not os.path.exists(sidpath):
        return self._init_ctable(path)
    return bcolz.open(path)
```

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,395 @@
...
+
+        Parameters:
+        -----------
+        sid : int
+            The asset identifer for the data being written.
+        days : pd.DatetimeIndex
+            The days for which to write data from the given df.
+        df : pd.DataFrame
+            DataFrame of market data with the following characteristics.
+            columns : ('open', 'high', 'low', 'close', 'volume')
```
ssanderson: What are the dtypes of these columns?  Should they be `uint32`s?  `float64`s?

zipline/data/us_equity_minutes.py
```diff
@@ -11,27 +11,395 @@
...
+        table.flush()
+
+    def _zerofill(self, table, numdays):
+        num_to_prepend = numdays * self._minutes_per_day
+        prepend_array = np.zeros(num_to_prepend, np.uint32)
+        # Fill all OHLCV with zeros.
+        table.append([prepend_array] * 5)
+        table.flush()
+
+    def write(self, sid, days, df):
```
ssanderson: The expected usage pattern here feels weird to me.  If I understand correctly the expectation here is that I do something like:

```
writer = BcolzMinuteWriter(...)
days = all_my_days()
for sid in all_my_sids():
    writer.write(sid, days, data_for_sid(sid))
```
I can't imagine a world where I'd want `days` to be different from call to call here, which suggests to me that it should be a constructor parameter rather than a parameter to `write`.  Note that we don't have the same problem for the daily writers, because the expectation is that in normal usage there's only one call made to `write`.

ehebert: We talked about this in another channel, but repeating for posterity.

The writer needs to support writing a sid's data in multiple spans of days through its history because one of our internal data sources is stored with a daily zipfile containing all equities. To balance CPU and memory utilization, our internal loader gets the data for a space which is the product of a subset of sids and days and then flushes that data to disk.

Because of the nature of the bcolz format, writing a day at a time is punitive because of the constant compression of the last block.

zipline/data/us_equity_minutes.py
ehebert: This was one of my questions in the PR notes, we could read this from the DataFrame, and probably should to reduce redundancy of data.

Also, we could require the DataFrame to have a `MultiIndex` of `(sid, dt)` and then walk through the DataFrame over sids using `get_level_values('sid')`.

Do you think using the `MultiIndex` would be too much of a hoop for users of the API to jump through?

zipline/data/us_equity_minutes.py
ehebert: Our internal usage currently passes `float64`s.

zipline/data/us_equity_minutes.py
ehebert: I'd be happy with that.

zipline/data/us_equity_minutes.py
ehebert: Since the `write` is being called multiple times, I think writing the metadata eagerly works better.
However, admittedly it does blur the dataset directory's initial creation and successive updates.

It should have no functional difference to pass an ever expanding list of market_opens, however passing a list that is different otherwise would certainly cause issues.

This is a piece that could be made more robust, by checking the length and values of the `minute_index` when creating a writer for a pre-existing directory and failing if the values differ, but way be able to defer adding that robustness.

I will attempt adding at least a basic check now.

zipline/data/us_equity_minutes.py
ehebert: Agreed.

zipline/data/us_equity_minutes.py
ehebert: Yes, your EDIT is why I included `for_sid` in the method name, since it was ported from our internal usage which check all equities.

I'm 50/50 too.

zipline/data/us_equity_minutes.py
ehebert: This is hit in the case when the ctable has not yet been created.
However, before calling this we can just check for ctable existence.

zipline/data/us_equity_minutes.py
ehebert: Went with encapsulating the metadata serialization into its own class.

```
class BcolzMinuteBarMetadata(object):

    METADATA_FILENAME = 'metadata.json'

    @classmethod
    def metadata_path(cls, rootdir):
        return os.path.join(rootdir, cls.METADATA_FILENAME)

    @classmethod
    def read(cls, rootdir):
        path = cls.metadata_path(rootdir)
        with open(path) as fp:
            raw_data = json.load(fp)
            first_trading_day = pd.Timestamp(
                raw_data['first_trading_day'], tz='UTC')
            minute_index = pd.to_datetime(raw_data['minute_index'],
                                          utc=True)
            ohlc_ratio = raw_data['ohlc_ratio']
            return cls(first_trading_day, minute_index, ohlc_ratio)

    def __init__(self, first_trading_day, minute_index, ohlc_ratio):
        self.first_trading_day = first_trading_day
        self.minute_index = minute_index
        self.ohlc_ratio = ohlc_ratio

    def write(self, rootdir):
        """
        Write the metadata to a JSON file in the rootdir.

        Values contained in the metadata are:
        first_trading_day : string
            'YYYY-MM-DD' formatted representation of the first trading day
             available in the dataset.
        minute_index : list of integers
             nanosecond integer representation of the minutes, the enumeration
             of which corresponds to the values in each bcolz carray.
        """
        metadata = {
            'first_trading_day': str(self.first_trading_day.date()),
            'minute_index': self.minute_index.asi8.tolist(),
            'ohlc_ratio': self.ohlc_ratio,
        }
        with open(self.metadata_path(rootdir), 'w+') as fp:
            json.dump(metadata, fp)
```

ehebert:
Closing in favor of https://github.com/quantopian/zipline/pull/963

PR #962, opened 2016-01-21
Pipeline missing values

ssanderson:
Adds support for specifying `missing_values` to pipeline `Term`s, and makes `BoundColumn.latest` return a Filter instead of a Factor for expressions of dtype `bool`.

This also removes all implicit coercion behavior from `AdjustedArray`.  PipelineLoaders are now responsible for ensuring that loaded values match the dtypes of requested columns.

@llllllllll tagging you on this since most of the changes here are actually in the `BlazeLoader` and its tests.

### DIFF
--- a/tests/pipeline/test_adjusted_array.py
+++ b/tests/pipeline/test_adjusted_array.py
@@ -10,6 +10,7 @@ from numpy import (
     arange,
     array,
     full,
+    where,
 )
 from numpy.testing import assert_array_equal
 from six.moves import zip_longest
@@ -23,9 +24,21 @@ from zipline.lib.adjustment import (
 from zipline.lib.adjusted_array import AdjustedArray, NOMASK
 from zipline.utils.numpy_utils import (
     datetime64ns_dtype,
+    default_missing_value_for_dtype,
     float64_dtype,
+    int64_dtype,
     make_datetime64ns,
 )
+from zipline.utils.test_utils import check_arrays, parameter_space
+
+
+def moving_window(array, nrows):
+    """
+    Simple moving window generator over a 2D numpy array.
+    """
+    count = num_windows_of_length_M_on_buffers_of_length_N(nrows, len(array))
+    for i in range(count):
+        yield array[i:i + nrows]
 
 
 def num_windows_of_length_M_on_buffers_of_length_N(M, N):
@@ -66,6 +79,7 @@ def _gen_unadjusted_cases(dtype):
     nrows = 6
     ncols = 3
     data = arange(nrows * ncols).astype(dtype).reshape(nrows, ncols)
+    missing_value = default_missing_value_for_dtype(dtype)
 
     for windowlen in valid_window_lengths(nrows):
 
@@ -78,6 +92,7 @@ def _gen_unadjusted_cases(dtype):
             data,
             windowlen,
             {},
+            missing_value,
             [
                 data[offset:offset + windowlen]
                 for offset in range(num_legal_windows)
@@ -230,6 +245,7 @@ def _gen_overwrite_adjustment_cases(dtype):
 
 def _gen_expectations(baseline, adjustments, buffer_as_of, nrows):
 
+    missing_value = default_missing_value_for_dtype(baseline.dtype)
     for windowlen in valid_window_lengths(nrows):
 
         num_legal_windows = num_windows_of_length_M_on_buffers_of_length_N(
@@ -241,6 +257,7 @@ def _gen_expectations(baseline, adjustments, buffer_as_of, nrows):
             baseline,
             windowlen,
             adjustments,
+            missing_value,
             [
                 # This is a nasty expression...
                 #
@@ -267,9 +284,10 @@ class AdjustedArrayTestCase(TestCase):
                             data,
                             lookback,
                             adjustments,
+                            missing_value,
                             expected):
 
-        array = AdjustedArray(data, NOMASK, adjustments)
+        array = AdjustedArray(data, NOMASK, adjustments, missing_value)
         for _ in range(2):  # Iterate 2x ensure adjusted_arrays are re-usable.
             window_iter = array.traverse(lookback)
             for yielded, expected_yield in zip_longest(window_iter, expected):
@@ -282,9 +300,10 @@ class AdjustedArrayTestCase(TestCase):
                                         data,
                                         lookback,
                                         adjustments,
+                                        missing_value,
                                         expected):
 
-        array = AdjustedArray(data, NOMASK, adjustments)
+        array = AdjustedArray(data, NOMASK, adjustments, missing_value)
         for _ in range(2):  # Iterate 2x ensure adjusted_arrays are re-usable.
             window_iter = array.traverse(lookback)
             for yielded, expected_yield in zip_longest(window_iter, expected):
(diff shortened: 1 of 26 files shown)


tests/pipeline/test_term.py
```diff
@@ -330,6 +333,29 @@ class SomeFactorNoDType(SomeFactor):
         with self.assertRaises(InvalidDType):
             SomeFactor(dtype=1)
 
+    def test_latest_on_different_dtypes(self):
+        self.assertIsInstance(TestingDataSet.bool_col.latest, Filter)
+        self.assertIsInstance(TestingDataSet.float_col.latest, Factor)
+        self.assertIsInstance(TestingDataSet.datetime_col.latest, Factor)
+        self.assertIsInstance(TestingDataSet.int_col.latest, Factor)
+
+    def test_failure_timing_on_bad_missing_values(self):
```
ssanderson: @llllllllll I think this addresses the concern you raised in person today.

I originally had made `Column` fail eagerly, but that was annoying because the error couldn't tell you the name of the column that failed.  The compromise here is to fail in `_BoundColumnDescr`, which means we still barf in class construction, but we can give you an error that includes the name of the bad column.

tests/pipeline/test_term.py
```diff
@@ -330,6 +334,38 @@ class SomeFactorNoDType(SomeFactor):
...
+                self.assertIsInstance(column.latest, Factor)
+            else:
+                self.fail(
+                    "Unknown dtype %s for column %s" % (column.dtype, column)
+                )
+            # These should be the same value, plus this has the convenient
+            # property of correctly handling `NaN`.
+            self.assertIs(column.missing_value, column.latest.missing_value)
+
+    def test_failure_timing_on_bad_missing_values(self):
```
ssanderson: @llllllllll I think this addresses the concern you raised in person today.

I originally had made Column fail eagerly, but that was annoying because the error couldn't tell you the name of the column that failed. The compromise here is to fail in _BoundColumnDescr, which means we still barf in class construction, but we can give you an error that includes the name of the bad column.

tests/pipeline/test_adjusted_array.py
```diff
@@ -23,9 +24,21 @@
...
     datetime64ns_dtype,
+    default_missing_value_for_dtype,
     float64_dtype,
+    int64_dtype,
     make_datetime64ns,
 )
+from zipline.utils.test_utils import check_arrays, parameter_space
+
+
+def moving_window(array, nrows):
```
llllllllll: should we move this to `numpy_utils`?

ssanderson: I don't think we need to.  Anything that needs to do a moving window efficiently probably wants to use AdjustedArray.  This was just a simple utility for testing.

zipline/utils/test_utils.py
```diff
@@ -296,7 +324,7 @@ def make_simple_equity_info(sids, start_date, end_date, symbols=None):
     sids : array-like of int
     start_date : pd.Timestamp
     end_date : pd.Timestamp
-    symbols : list, optional
+    symbols : list, optionaln
```
llllllllll: typo

zipline/pipeline/loaders/equity_pricing_loader.py
```diff
@@ -81,12 +81,16 @@ def load_adjusted_array(self, columns, dates, assets, mask):
...
-        return dict(zip(columns, adjusted_arrays))
+        out = {}
+        for c, c_raw, c_adjs in zip(columns, raw_arrays, adjustments):
+            out[c] = AdjustedArray(
+                c_raw.astype(c.dtype),
+                mask,
+                c_adjs,
+                c.missing_value,
+            )
+        return out
```
llllllllll: I think this is probably nicer as a comprehension, but up to you.

zipline/pipeline/data/dataset.py
```diff
@@ -92,8 +132,15 @@ def qualname(self):
...
-        from zipline.pipeline.factors import Latest
-        return Latest(inputs=(self,), dtype=self.dtype)
+        if self.dtype == bool_dtype:
+            from zipline.pipeline.filters import Latest
+        else:
+            from zipline.pipeline.factors import Latest
+        return Latest(
+            inputs=(self,),
+            dtype=self.dtype,
+            missing_value=self.missing_value,
```
llllllllll: what does a missing value mean in a filter?

tests/test_test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+        assert cls.xy_invocations == list(product(cls.x_args, cls.y_args))
+
+    @parameter_space(x=x_args, y=y_args)
+    def test_xy(self, x, y):
+        self.xy_invocations.append((x, y))
+
+    def test_nothing(self):
+        # Ensure that there's at least one "real" test in the class, or else
+        # our {setUp,tearDown}Class won't be called if, for example,
+        # `parameter_space` returns None.
```
llllllllll: this is a great test class

tests/test_test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    y_args = [3, 4]
+
+    @classmethod
+    def setUpClass(cls):
+        cls.xy_invocations = []
+
+    @classmethod
+    def tearDownClass(cls):
+        # This is the only actual test here.
+        assert cls.xy_invocations == list(product(cls.x_args, cls.y_args))
```
llllllllll: `s/py.test/nose/`

llllllllll: also, is the order part of the contract of this function? if not should we make this a set?

tests/pipeline/test_term.py
llllllllll: Class construction time seems fine, it is worth it to know which column is incorrect and I imagine that very few (if any) people will be creating columns outside of a dataset.

zipline/pipeline/data/dataset.py
```diff
@@ -18,28 +23,62 @@ class Column(object):
...
     """
-    def __init__(self, dtype, name):
+    def __init__(self, dtype, missing_value, name):
         self.dtype = dtype
+
+        # Calculating missing values here guarantees that we fail quickly if
+        # the user fails to provide a missing value for a dtype that requires
+        # one (e.g. int64), but still enables us to provide an error message
+        # that points to the name of the failing column.
+        if missing_value is NotSpecified:
```
llllllllll: an we also check that the missing value is coercable to `dtype` here? For example:

```python
In [17]: class Ds(DataSet):
    c = Column('int64', np.nan)
   ....:     

In [18]: Ds.c
Out[18]: Ds.c::int64
```

zipline/utils/test_utils.py
```diff
@@ -664,3 +692,84 @@ def gen_calendars(start, stop, critical_dates):
...
+    """
+    Wrapper around subtest that allows passing keywords mapping names to
+    iterables of values.
+
+    The decorated test function will be called with the cross-product of all
+    possible inputs
+
+    Usage
+    -----
+    >>> class SomeTestCase(TestCase):
```
llllllllll: ERROR: test_test_utils_docs (tests.test_doctests.DoctestTestCase)

  File "/home/travis/build/quantopian/zipline/tests/test_doctests.py", line 68, in test_test_utils_docs

    self._check_docs(test_utils)

  File "/home/travis/build/quantopian/zipline/tests/test_doctests.py", line 49, in _check_docs

    raise e.exc_info[1]

NameError: name 'TestCase' is not defined

zipline/utils/test_utils.py
```diff
@@ -664,3 +692,84 @@ def gen_calendars(start, stop, critical_dates):
...
+    )
+
+    loader = make_seeded_random_loader(random_seed, calendar, sids)
+    get_loader = lambda column: loader
+
+    with tmp_asset_finder(equities=equity_info) as finder:
+        yield SimplePipelineEngine(get_loader, calendar, finder)
+
+
+def parameter_space(**params):
```
llllllllll: can you provide a `see also` to `subtest` and have `subtest` have a see also to `parameter_space`? I would also put a recomendation to use this function in the docstring of subtest

zipline/pipeline/loaders/equity_pricing_loader.py
ssanderson: I wrote this as a comprehension and then rewrote it this way because I thought it was easier to follow.  Going to leave it as is.

tests/test_test_utils.py
ssanderson: The way it's written right now you're guaranteed to get the result of calling `itertools.product` with the values in the order that they appear in the function's parameter list.  I don't see a reason to change that.

llllllllll: Sure, but that is an implementation detail, no?

ssanderson: Eh, I think it's fine to make it part of the contract.  It's nice to have defined semantics for the ordering.  I just pushed a test asserting that we flip the order if we flip the arglist.

llllllllll: kk

PR #965, opened 2016-01-22
PERF: make Asset comparison faster

llllllllll:


richafrank:
LGTM :+1:

zipline/assets/_assets.pyx
```diff
@@ -75,58 +87,37 @@ cdef class Asset:
...
-        if isinstance(x, Asset):
-            x_as_int = x.sid
-        elif isinstance(x, int):
-            # ints are Integral, but much faster to special case
-            x_as_int = x
-        elif isinstance(x, (long, np.int64, np.int32, Integral)):
-            x_as_int = int(x)
-        else:
+        try:
+            x_as_int = PyNumber_Index(x)
```
richafrank: Are we ok with:
```python
>>> 1 == 9999999999L
False
>>> Asset(1) == 9999999999L
  File "zipline/assets/_assets.pyx", line 99, in zipline.assets._assets.Asset.__richcmp__ (zipline/assets/_assets.c:2290)
    y_as_int = y.sid
OverflowError: value too large to convert to int
```

richafrank: FWIW That doesn't seem to be new to this PR:
```python
  File "zipline/assets/_assets.pyx", line 104, in zipline.assets._assets.Asset.__richcmp__ (zipline/assets/_assets.c:2348)
    y_as_int = int(y)
OverflowError: value too large to convert to int
```

llllllllll: That seems reasonable. The alternative would be to make this a bigger type like uint64 but idk if that is worth it and only moves the boundry up. The alternative is to catch this and return `NotImplemented` which will propagate as a type error. I think that error is less informative.

PR #974, opened 2016-01-26
Replace print with logger.info

jlowin:
For years I watched the tickers print, helpless to stop them from appearing every. single. time.

One day I did something about it.

### DIFF
--- a/zipline/data/loader.py
+++ b/zipline/data/loader.py
@@ -12,7 +12,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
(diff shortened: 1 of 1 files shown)


llllllllll:
Thanks! I have wanted to remove this line entirely. Do you think the log line should be there at all?

ssanderson:
@llllllllll I think it's nice to give some indication of progress during the load, since a user could easily mistake a blank terminal that's loading data as just being hung.  (By that logic we should probably have a better terminal UI for indicating backtest progress as well...).  Using a logger here also now means that you can filter out these logs if you don't want them by setting a handler with a suitable filter.

:+1: from me on this.

llllllllll:
kk, merging. Thanks for this.

PR #976, opened 2016-01-29
MAINT: fix pandas warning: Use "df.empty" instead of "not df"

jikamens:


jikamens:
@llllllllll if this looks ok to you please merge.

jikamens:
To fix:

```
...
  File ".../zipline/history/history_container.py", line 721, in update
  File ".../zipline/history/history_container.py", line 703, in frame_from_bardata
  File ".../local/lib/python2.7/site-packages/pandas/core/generic.py", line 714, in __nonzero__
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

llllllllll:
I think this should be:

```python
if sid_data is None:
    ...
```

The `df.get` returns `None` when not found.

jikamens:
>I think this should be:
>
>```python
>if sid_data is None:
>    ...
>```
>
>The `data.get` returns `None` when not found.

I'm pretty sure if that were the case then we wouldn't be getting a warning from Pandas about the truth value being ambiguous, right?

llllllllll:
It returns a dataframe is `sid in data` otherwise `None`. When we succeed, we get do `bool(DataFrame)` which raises a warning. With your patch, when we fail we will do `None.empty` which will also fail. An explicit check against `None` should handle both cases.

llllllllll:
Just saw the test failures. The `_data` here should be a mapping from sid to `SIDData`. Do we know why we are getting df here?

jikamens:
Well, I thought this was an easy fix, but apparently it's more complicated. I don't know why we're getting a DataFrame instead of a SIDData.

richafrank:
I believe this history code has been replaced now.

PR #977, opened 2016-01-30
BUG: Handle OverflowError by comparing False

richafrank:
instead of raising


richafrank:
@llllllllll as we discussed

llllllllll:
looks good

PR #978, opened 2016-02-01
ENH: Add padding method to minute bars writer.

ehebert:
So that consumers can write empty days worth of data, without needing
to construct a DataFrame with zero data force a write.

The internal loader uses `last_date_in_output_for_sid` to signify that
data has been attempted to be retrieved for all dates up until that, so
that when resuming a job those retrieval of data for those dates are not
re-attempted.

Also, used to make the write logic cleaneer, by making it only
necessary to create an array large enough for the given df.

### DIFF
--- a/tests/data/test_minute_bars.py
+++ b/tests/data/test_minute_bars.py
@@ -23,6 +23,7 @@ from pandas import (
     DataFrame,
     DatetimeIndex,
     Timestamp,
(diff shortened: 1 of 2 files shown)


ehebert:
cc: @jbredeche

jbredeche:
lgtm

PR #979, opened 2016-02-02
ENH: Allow passing of numpy arrays to writer.

ehebert:
For faster parsing and writing workflows, do not require a DataFrame.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -20,7 +20,6 @@ from os.path import join
 import json
 import os
 import pandas as pd
-from pandas.core.datetools import normalize_date
 
 US_EQUITIES_MINUTES_PER_DAY = 390
 
@@ -403,8 +402,6 @@ class BcolzMinuteBarWriter(object):
         -----------
         sid : int
             The asset identifer for the data being written.
-        days : pd.DatetimeIndex
-            The days for which to write data from the given df.
         df : pd.DataFrame
             DataFrame of market data with the following characteristics.
             columns : ('open', 'high', 'low', 'close', 'volume')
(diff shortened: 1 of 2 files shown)


zipline/data/minute_bars.py
```diff
@@ -415,11 +413,49 @@ def write(self, sid, df):
...
             index : DatetimeIndex of market minutes.
         """
+        cols = {
+            'open': df.open.values,
+            'high': df.high.values,
+            'low': df.low.values,
+            'close': df.close.values,
+            'volume': df.volume.values,
+        }
+        dts = df.index.values
```
jbredeche: @richafrank is this one of the perf regressions you mentioned in pandas v.recent?

richafrank: Yes, I think so.  Is this call to `values` a hot spot?

ehebert: In our internal code, this function would no longer be called, in favor of `write_cols`, so that we can avoid the overhead of `DataFrames` altogether when loading data.

This function is left here for case where convenience is valued over performance.

Open to changing this name to something like `write_dataframe`

richafrank: When you say "internal code", do you mean internal to a loop or code in an internal repo?  Wondering if zipline algos will need to call this function in a tight loop, or if it's just a one-time setup.

ehebert: (Summarizing conversation in another channel.)

'internal' meaning an 'internal repo'.
Algos would not be calling this function, it would part of the one-time setup, and it would most likely be the case that this would be passed `DataFrame`s that cover large spans of data.

PR #980, opened 2016-02-03
ENH: allow users to specify odo kwargs in from_blaze

llllllllll:


zipline/pipeline/loaders/blaze/core.py
```diff
@@ -221,20 +203,25 @@ class ExprData(namedtuple('ExprData', 'expr deltas resources')):
...
     def __repr__(self):
         # If the expressions have _resources() then the repr will
-        # drive computation so we box them.
+        # drive computation so we take the str here.
         cls = type(self)
         return super(ExprData, cls).__repr__(cls(
-            _ExprRepr(self.expr),
-            _ExprRepr(self.deltas),
-            self.resources,
+            str(self.expr),
```
richafrank: Why the `str` change?

llllllllll: the old box was the same as this but more complicated.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -595,9 +585,11 @@ def from_blaze(expr,
     ds = new_dataset(dataset_expr, deltas)
     # Register our new dataset with the loader.
     (loader if loader is not None else global_loader)[ds] = ExprData(
-        dataset_expr,
-        deltas,
-        resources,
```
richafrank: Why doesn't ExprData need the resources anymore?

llllllllll: we are binding the resources into the expression like we do with the earnings loader

richafrank: Ok, I follow now.

PR #982, opened 2016-02-04
Add auto_close_date support for equities

dmichalowicz:
- Adds an `auto_close_date` field to the Asset object, which is inherited by both equities and futures.
- Creates an `auto_close_date` column for equities in the schema of the assets database.
- Now closes equities and futures immediately prior to `before_trading_start`, as opposed to at market close.


dmichalowicz:
@jfkirk Assigning this to you for now, though still a work in progress. Also ping @ssanderson @richafrank

dmichalowicz:
@jfkirk @ssanderson This is ready for review.

tests/test_algorithm.py
```diff
@@ -1818,20 +1821,21 @@ class TestClosePosAlgo(TestCase):
...
         self.panel = pd.Panel({1: pd.DataFrame({
             'price': [1, 1, 2, 4], 'volume': [1e9, 1e9, 1e9, 0],
             'type': [DATASOURCE_TYPE.TRADE,
                      DATASOURCE_TYPE.TRADE,
                      DATASOURCE_TYPE.TRADE,
                      DATASOURCE_TYPE.CLOSE_POSITION]},
-            index=self.days)
+            index=self.days[:4])
         })
         self.no_close_panel = pd.Panel({1: pd.DataFrame({
```
ssanderson: Why are these now different lengths?

tests/test_algorithm.py
```diff
@@ -2018,3 +2022,99 @@ def test_remove_data(self):
...
+        dt = pd.Timestamp('2015-01-05', tz='UTC')
+        self.env = TradingEnvironment()
+        ix = self.env.trading_days.get_loc(dt)
+
+        self.metadata = make_rotating_equity_info(
+            3, dt, trading_day, 2, 2,
+        )
+
+        # Make all start dates equal.
+        self.metadata['start_date'] = [self.metadata['start_date'][0]]*3
```
ssanderson: You can just do `df[column] = scalar` instead of writing a list:

```
In [1]: df
Out[1]:
           a         b   c
0  -0.232151 -0.601906   5
1   1.795220 -0.535771   5
2   0.752929 -1.103029   5
3  -0.026765 -1.379774   5

In [2]: df.c = 1

In [3]: df
Out[3]:
           a         b  c
0  -0.232151 -0.601906  1
1   1.795220 -0.535771  1
2   0.752929 -1.103029  1
3  -0.026765 -1.379774  1
4   0.064691  1.046385  1
```

tests/test_algorithm.py
```diff
@@ -2018,3 +2022,99 @@ def test_remove_data(self):
...
+
+    def tearDown(self):
+        del self.env
+
+    def test_delisted_securities(self):
+        """
+        Make sure that after a security gets delisted, our portfolio holds the
+        correct number of securities and correct amount of cash.
+        """
+        self.env.write_data(equities_df=self.metadata)
```
ssanderson: You're calling `write_data` in all of these tests with `self.metadata`.  Can we move that logic into the shared setup (it should probably be in a `setUpClass` for performance)?

tests/test_algorithm.py
```diff
@@ -2018,3 +2022,99 @@ def test_remove_data(self):
...
+        algo = TestLiquidationAlgorithm(env=self.env)
+        algo.run(self.source)
+
+        expected_cash = [100000, 99830, 99830, 99860, 99860, 99940, 99940]
```
ssanderson: Can we annotate these changes to make it clearer what's being tested here?

tests/test_algorithm.py
```diff
@@ -2018,3 +2022,99 @@ def test_remove_data(self):
...
+        # Add an auto close date column, where liquidation occurs 2 days after
+        # the end date.
+        metadata = self.metadata.copy()
+        metadata['auto_close_date'] = metadata['end_date'] + (2 * trading_day)
+
+        self.env.write_data(equities_df=metadata)
+        algo = TestLiquidationAlgorithm(env=self.env)
+        algo.run(self.source)
+
+        expected_cash = [100000, 99830, 99830, 99830, 99860, 99860, 99940]
```
ssanderson: same comment as above.

zipline/assets/asset_db_migrations.py
```diff
@@ -101,9 +101,38 @@ def _downgrade_v1_to_v0(op, version_info_table):
...
+
+def _downgrade_v2_to_v1(op, version_info_table):
+    """
+    Downgrade assets db by removing the 'auto_close_date' column.
+    """
+    version_info_table.delete().execute()
+
+    # Drop indices before batch
+    # This is to prevent index collision when creating the temp table
+    op.drop_index('ix_equities_fuzzy_symbol')
```
ssanderson: It looks like this is copied from the migration above. @jfkirk do we need to add this to every downgrade?  I'd be a bit surprised if that were the case.

tests/test_assets.py
```diff
@@ -172,6 +172,27 @@ def build_lookup_generic_cases(asset_finder_type):
 
 class AssetTestCase(TestCase):
 
+    asset_attrs = ['sid',
```
jfkirk: It seems like we may want to add this list to assets.py, rather than here in the tests.

tests/test_algorithm.py
```diff
@@ -2018,3 +2022,99 @@ def test_remove_data(self):
...
+    """
+    Tests if delisted securities are properly removed from a portfolio holding
+    positions in said securities.
+    """
+    def setUp(self):
+        dt = pd.Timestamp('2015-01-05', tz='UTC')
+        self.env = TradingEnvironment()
+        ix = self.env.trading_days.get_loc(dt)
+
+        self.metadata = make_rotating_equity_info(
```
ssanderson: I'm inclined to say we should move the metadata construction here into a new test_utils function called `make_jagged_equity_info` or somesuch.

tests/test_algorithm.py
```diff
@@ -2018,3 +2022,99 @@ def test_remove_data(self):
         # initially only data for X should be sent and on the last day only
         # data for Y should be sent since X is expired
         np.testing.assert_array_equal(self.algo.data, expected_lengths)
+
+
+class TestAssetAutoClose(TestCase):
```
ssanderson: This is just testing auto closes of equities, right?

tests/test_algorithm.py
```diff
@@ -2018,3 +2022,99 @@ def test_remove_data(self):
...
+            columns=['price', 'volume'],
+        )
+
+        pan = pd.Panel({0: data_x, 1: data_y, 2: data_z})
+        self.source = DataPanelSource(pan)
+
+    def tearDown(self):
+        del self.env
+
+    def test_delisted_securities(self):
```
ssanderson: Since this is specifically testing equities, I'd call this `test_delisted_equities` (with similar modifications below).

zipline/assets/asset_db_migrations.py
jfkirk: I believe we will need it for every downgrade that uses a batch_alter_table. David, did you try this downgrade without the index drops?

tests/test_assets.py
ssanderson: Better yet, we should be dynamically generating this list from the types themselves:

```
In [1]: from zipline.assets import Asset
f
In [2]: from types import GetSetDescriptorType

In [3]: [name for name, value in vars(Asset).items() if isinstance(value, GetSetDescriptorType)]
Out[3]:
['auto_close_date',
 'sid',
 'start_date',
 'end_date',
 'exchange',
 'symbol',
 'first_traded',
 'asset_name']
```

zipline/finance/performance/position_tracker.py
```diff
@@ -56,6 +57,16 @@
                             'net_value'])
 
 
+def generate_close_event(sid, date):
+        # Return a CLOSE_POSITION event
```
ssanderson: Indentation here is off.

zipline/finance/performance/position_tracker.py
```diff
@@ -204,13 +216,7 @@ def auto_close_position_events(self, next_trading_day):
...
 
             for sid in sids:
-                # Yield a CLOSE_POSITION event
-                event = Event({
-                    'dt': date,
-                    'type': DATASOURCE_TYPE.CLOSE_POSITION,
-                    'sid': sid,
-                })
-                yield event
+                yield generate_close_event(sid, date)
```
ssanderson: The verb `generate` here makes me think that `generate_close_event` is a [generator function](https://docs.python.org/2/tutorial/classes.html#generators).

zipline/finance/performance/position_tracker.py
```diff
@@ -362,7 +368,7 @@ def maybe_create_close_position_transaction(self, event):
             dt=event.dt,
             price=price,
             commission=0,
-            order_id=0
+            order_id=uuid.uuid4().hex,
```
ssanderson: Why did this change?

tests/test_algorithm.py
dmichalowicz: As a result of auto-closing in `before_trading_start` as opposed to in `handle_market_close`, futures with an `auto_close_date` of the day after the last day of the simulation will no longer be auto-closed. With that said, I should work on a better test for that than this.

tests/test_algorithm.py
dmichalowicz: Yes, will change name.

zipline/assets/asset_db_migrations.py
dmichalowicz: Yes I tried it without the index drops but it failed

zipline/test_algorithms.py
```diff
@@ -947,6 +949,28 @@ def handle_data(self, data):
         self.i += 1
 
 
+class TestLiquidationAlgorithm(TradingAlgorithm):
```
ssanderson: We're generally moving away from the model of subclassing `TradingAlgorithm` for specific algos.  I'd rather write this as:

```
def initialize(context):
    ...

    ...

algo = TradingAlgorithm(initialize=initialize, handle_data=handle_data, ...)
```

zipline/test_algorithms.py
```diff
@@ -947,6 +949,28 @@ def handle_data(self, data):
...
+    def initialize(self, *args, **kwargs):
+        self.ordered = False
+        self.set_commission(PerTrade(0))
+        self.set_slippage(FixedSlippage(spread=0))
+        self.num_positions = []
+        self.cash = []
+
+    def handle_data(self, data):
+        if not self.ordered:
+            self.ordered = True
```
ssanderson: General coding style note: when I have a flag signalling that some operation has been performed, I generally try to set it only after that operation has been performed.  Here it's harmless, but it makes you more likely to do the right/expected thing in more complicated setups.

tests/test_algorithm.py
```diff
@@ -2018,3 +2022,99 @@ def test_remove_data(self):
...
+        self.env.write_data(equities_df=metadata)
+        algo = TestLiquidationAlgorithm(env=self.env)
+        algo.run(self.source)
+
+        expected_cash = [100000, 99830, 99830, 99830, 99860, 99860, 99940]
+        expected_num_positions = [0, 3, 3, 3, 2, 2, 1]
+        self.assertEqual(algo.cash, expected_cash)
+        self.assertEqual(algo.num_positions, expected_num_positions)
+
+    def test_cancel_open_orders(self):
```
ssanderson: I'm not sure this actually tests what we want.  If the orders are just filled normally, there won't be any open orders by the time the algo finishes.  At the very least, we should confirm that there **are** open orders at the end of the day before liquidation, and that they're removed on the subsequent call.

dmichalowicz: My thinking here was that if you place an order for a stock the day before it delists, it will never get filled and under the current circumstances would remain open indefinitely, meaning at the end of the backtest you'd still have an open order for it. This tests that at the end of the backtest this is not the case. I will fix it though to make it clearer.

tests/test_algorithm.py
```diff
@@ -1818,27 +1823,29 @@ class TestClosePosAlgo(TestCase):
 
     def setUp(self):
         self.env = TradingEnvironment()
-        self.days = self.env.trading_days[:4]
+        self.days = self.env.trading_days[:5]
         self.panel = pd.Panel({1: pd.DataFrame({
-            'price': [1, 1, 2, 4], 'volume': [1e9, 1e9, 1e9, 0],
+            'price': [1, 1, 2, 4, 8], 'volume': [1e9, 1e9, 1e9, 1e9, 0],
```
dmichalowicz: I just made this the same length as below for consistency.

tests/test_algorithm.py
```diff
@@ -2018,3 +2025,256 @@ def test_remove_data(self):
...
+            if not context.ordered:
+                for s in data:
+                    context.order(context.sid(s), 10)
+                context.ordered = True
+
+            context.cash.append(context.portfolio.cash)
+            context.num_positions.append(len(context.portfolio.positions))
+
+        algo = TradingAlgorithm(
+            initialize=self.initialize.im_func,
```
dmichalowicz: Not sure if `im_func` is the correct way to do this.

tests/test_algorithm.py
```diff
@@ -2018,3 +2025,256 @@ def test_remove_data(self):
...
+    @classmethod
+    def tearDownClass(cls):
+        cls.stack.close()
+
+    def test_delisted_equities(self):
+        """
+        Make sure that after an equity gets delisted, our portfolio holds the
+        correct number of equities and correct amount of cash.
+        """
+        def handle_data(context, data):
```
dmichalowicz: Each test's `handle_data` is slightly different; is it worth defining it every test, or should I try to make a shared `handle_data` like `initialize`?

zipline/assets/asset_db_migrations.py
```diff
@@ -101,9 +101,38 @@ def _downgrade_v1_to_v0(op, version_info_table):
...
+    op.drop_index('ix_equities_company_symbol')
+
+    # Execute batch op to allow column modification in SQLite
+    with op.batch_alter_table('equities') as batch_op:
+
+        # Delete 'auto_close_date'
```
ssanderson: This comment isn't adding any useful information here.

PR #983, opened 2016-02-04
Expiration perf

richafrank:
Look up expired futures from in-memory Futures instead of queries to the db.

### DIFF
--- a/zipline/utils/data.py
+++ b/zipline/utils/data.py
@@ -419,16 +419,25 @@ class SortedDict(MutableMapping):
     >>> d[-1] = 'negative one'
     >>> d[0] = 'zero'
     >>> d[2] = 'two'
-    >>> d
-    SortedDict[<built-in function abs>]([(0, 'zero'), (-1, 'negative one'), (1, 'one')])  # noqa
+    >>> d  # doctest: +NORMALIZE_WHITESPACE
+    SortedDict(<built-in function abs>,
+               [(0, 'zero'), (-1, 'negative one'), (2, 'two')])
     >>> d[1] = 'one'  # Mutating the mapping maintains the sort order.
+    >>> d  # doctest: +NORMALIZE_WHITESPACE
+    SortedDict(<built-in function abs>,
+               [(0, 'zero'), (-1, 'negative one'), (1, 'one'), (2, 'two')])
+    >>> del d[0]
+    >>> d  # doctest: +NORMALIZE_WHITESPACE
+    SortedDict(<built-in function abs>,
+               [(-1, 'negative one'), (1, 'one'), (2, 'two')])
+    >>> del d[2]
     >>> d
-    SortedDict[<built-in function abs>]([(0, 'zero'), (-1, 'negative one'), (1, 'one')])  # noqa
+    SortedDict(<built-in function abs>, [(-1, 'negative one'), (1, 'one')])
     """
-    def __init__(self, sort_key, mapping=None, **kwargs):
+    def __init__(self, key, mapping=None, **kwargs):
         self._map = {}
         self._sorted_key_names = []
-        self._sort_key = sort_key
+        self._sort_key = key
 
         self.update(merge(mapping or {}, kwargs))
 
@@ -464,8 +473,8 @@ class SortedDict(MutableMapping):
         _repr_running[call_key] = 1
         try:
             if not self:
-                return '%s[%r]()' % (self.__class__.__name__, self._sort_key)
-            return '%s[%r](%r)' % (self.__class__.__name__, self._sort_key,
-                                   self.items())
+                return '%s(%r)' % (self.__class__.__name__, self._sort_key)
+            return '%s(%r, %r)' % (self.__class__.__name__, self._sort_key,
+                                   list(self.items()))
         finally:
             del _repr_running[call_key]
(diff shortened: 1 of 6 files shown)


richafrank:
cc @jfkirk @dmichalowicz

zipline/utils/data.py
```diff
@@ -391,3 +395,85 @@ def _update_buffer(self, frame):
...
+    """A mapping of key-value pairs sorted by key according to the sort_key
+    function provided to the mapping.  Ties from the sort_key are broken by
+    comparing the original keys.  `iter` traverses the keys in sort order.
+
+    Parameters
+    ----------
+    mapping : mapping, optional
+    **kwargs
+        The initial mapping. This will double as the starting mapping so
+        the values must also be hashable.
```
llllllllll: The values do not need to be hashable here.

llllllllll: also you should add the `key` arg.

zipline/utils/data.py
```diff
@@ -391,3 +395,85 @@ def _update_buffer(self, frame):
...
+    >>> d  # doctest: +NORMALIZE_WHITESPACE
+    SortedDict[<built-in function abs>]([(-1, 'negative one'),
+                                         (1, 'one'),
+                                         (2, 'two')])
+    >>> del d[2]
+    >>> d  # doctest: +NORMALIZE_WHITESPACE
+    SortedDict[<built-in function abs>]([(-1, 'negative one'),
+                                         (1, 'one')])
+    """
+    def __init__(self, sort_key, mapping=None, **kwargs):
```
llllllllll: I would just call this `key` to match `sorted`

zipline/utils/data.py
```diff
@@ -391,3 +395,85 @@ def _update_buffer(self, frame):
...
+
+    def __len__(self):
+        return len(self._map)
+
+    def __repr__(self, _repr_running={}):
+        # Based on OrderedDict/defaultdict
+        call_key = id(self), get_ident()
+        if call_key in _repr_running:
+            return '...'
+        _repr_running[call_key] = 1
```
llllllllll: too bad there isn't reprlib in py2

zipline/utils/data.py
```diff
@@ -391,3 +395,85 @@ def _update_buffer(self, frame):
...
+        try:
+            if not self:
+                return '%s[%r]()' % (self.__class__.__name__, self._sort_key)
+            return '%s[%r](%r)' % (self.__class__.__name__, self._sort_key,
```
llllllllll: I would make the format `%s(%r, %r)` so you could reevaluate it.

zipline/utils/data.py
richafrank: Agreed, had mis-copied.

zipline/utils/data.py
richafrank: Sounds good to me.

llllllllll: do you think it makes sense to default `key` to `operator.le`?

zipline/utils/data.py
richafrank: Good call.

zipline/utils/data.py
```diff
@@ -391,3 +399,84 @@ def _update_buffer(self, frame):
...
+    def __iter__(self):
+        for key, name in self._sorted_key_names:
+            yield name
+
+    def __len__(self):
+        return len(self._map)
+
+    def __repr__(self, _repr_running={}):
+        # Based on OrderedDict/defaultdict
+        call_key = id(self), get_ident()
```
llllllllll: `_get_ident`

zipline/utils/data.py
llllllllll: also in py3 this only shows: `SortedDict[<function <lambda> at 0x7fae9a31cf28>](ItemsView(...))`, we might want `list(self.items())`

zipline/utils/data.py
```diff
@@ -391,3 +399,84 @@ def _update_buffer(self, frame):
...
+        self._map[name] = value
+        if len(self._map) > len(self._sorted_key_names):
+            key = self._sort_key(name)
+            pair = (key, name)
+            idx = _bisect_right(self._sorted_key_names, pair)
+            self._sorted_key_names.insert(idx, pair)
+
+    def __delitem__(self, name, _bisect_right=bisect.bisect_right):
+        del self._map[name]
+        idx = _bisect_right(self._sorted_key_names,
```
llllllllll: is this just a `bisect_left`?

zipline/utils/data.py
```diff
@@ -391,3 +399,84 @@ def _update_buffer(self, frame):
...
+        del self._sorted_key_names[idx - 1]
+
+    def __iter__(self):
+        for key, name in self._sorted_key_names:
+            yield name
+
+    def __len__(self):
+        return len(self._map)
+
+    def __repr__(self, _repr_running={}):
```
llllllllll: I think we are just using `_repr_running` as a `set`. We only check the keys and put in a single 1 constant. I think the intent might be more clear that way.

zipline/utils/data.py
richafrank: Agreed.

zipline/utils/data.py
richafrank: No strong opinion...

PR #985, opened 2016-02-09
BUG: Don't accumulate 0-sized positions.

jbredeche:
We properly remove empty positions before the user calls `context.portfolio.positions`.

But until the user does that, the empty positions sit around.  And if other internal code needs a list of the positions, they get empty positions.

### DIFF
--- a/zipline/finance/performance/position_tracker.py
+++ b/zipline/finance/performance/position_tracker.py
@@ -258,6 +258,19 @@ class PositionTracker(object):
             position = self.positions[sid]
 
         position.update(txn)
+
+        if position.amount == 0:
+            # if this position now has 0 shares, remove it from our internal
+            # bookkeeping.
+            del self.positions[sid]
+
+            try:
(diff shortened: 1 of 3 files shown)


zipline/finance/performance/position_tracker.py
```diff
@@ -258,6 +258,19 @@ def execute_transaction(self, txn):
...
+
+        if position.amount == 0:
+            # if this position now has 0 shares, remove it from our internal
+            # bookkeeping.
+            del self.positions[sid]
+
+            try:
+                # if this position exists in our user-facing dictionary,
+                # remove it as well.  
+                del self._positions_store[sid]
```
ehebert: Are these the cases where it wouldn't exist?

1) `context.positions` has not been called so positions_store has not been populated.
2) user code running `del self._positions_store[sid]`

jbredeche: yup, that's what I was thinking.

PR #986, opened 2016-02-09
BUG: Allow more than 998 equities for dividends.

ehebert:
Chunk queries into groups of 998, because of a SQLite limitation on
query size.

Also, point assets module at factored out value and function.


jbredeche:
lgtm!

PR #987, opened 2016-02-09
Tests for create_transaction using current_dt

ehebert:
With the changes to create_transaction using the current_dt, the `test_perf_tracking` suite became incompatible, because of the change to using the `SidView` `current_dt` instead of the `Event`, the tests now need `BarData`.

Also, add test to `test_slippage` which checks for the case where a transaction would have been created on a bar with no trades (which was also fixed in a previous patch.)


jbredeche:
lgtm, thanks

PR #988, opened 2016-02-10
more history tests

jbredeche:
History tests are not complete yet (still missing split/merger/dividend tests for daily history, and futures tests) but holding off for now.


ehebert:
Besides one comment, lgtm.

PR #990, opened 2016-02-10
Buyback auth in pipeline

mtydykov:
This adds buyback authorizations to Pipeline and refactors out much of the logic from earnings, since most of it can be reused for buyback authorizations and other events. 

Note: This PR won't be completely done until https://github.com/quantopian/zipline/pull/962 is completed, since we need missing values (see line 121 in `zipline/pipeline/loaders/events.py`).


mtydykov:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;B2BARcEJ&#x2F;2-eventvestor-buyback-auth"> EventVestor buyback_auth</a>

zipline/pipeline/factors/events.py
```diff
@@ -55,37 +84,54 @@ def _compute(self, arrays, dates, assets, mask):
...
+class BusinessDaysSincePreviousCashBuybackAuth(
+    BusinessDaysSincePreviousEvents
+):
+    """
+    Factor returning the number of **business days** (not trading days!) since
+    the most recent cash buyback authorization for each asset.
 
     See Also
     --------
     zipline.pipeline.factors.BusinessDaysUntilNextEarnings
```
llllllllll: this should be cashbuyback

zipline/pipeline/factors/events.py
```diff
@@ -55,37 +84,54 @@ def _compute(self, arrays, dates, assets, mask):
...
 
-        # Set masked values to NaT.
-        announce_dates[~mask] = np_NaT
 
-        # Convert row labels into a column vector for broadcasted comparison.
-        reference_dates = dates.values.astype(datetime64D_dtype)[:, newaxis]
-        return busday_count_mask_NaT(announce_dates, reference_dates)
+    See Also
+    --------
+    zipline.pipeline.factors.BusinessDaysUntilNextEarnings
```
llllllllll: same here

zipline/pipeline/loaders/blaze/events.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @preprocess(data_query_tz=optionally(ensure_timezone))
+    def __init__(self,
+                 expr,
+                 resources=None,
+                 odo_kwargs=None,
+                 data_query_time=None,
+                 data_query_tz=None,
+                 dataset=None,
+                 loader=None):
```
llllllllll: Can we call this `concrete_loader` or pick a name that shows how this differs from the loader being constructed?
Also, this argument and `dataset` need to go into the docstring.

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,126 @@
...
+        knew about the event on all prior dates.  This mode is only supported
+        if ``infer_timestamp`` is explicitly passed as a truthy value.
+
+    infer_timestamps : bool, optional
+        Whether to allow omitting the "timestamp" column.
+    """
+
+    def __init__(self,
+                 all_dates,
+                 events_by_sid,
```
llllllllll: noting what we talked about irl: add a frozenset which says what the expected columns are. The constructor will validate the shape of the dataframes passed to fail early with a good error message. If the order of this set is one, then we can coerce input from a datetimeindex or series.

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # Do not modify the original in place, since it may be used for other
+        #  purposes.
+        self.events_by_sid = (
+            events_by_sid.copy()
+        )
+        dates = self.all_dates.values
+
+        for k, v in iteritems(events_by_sid):
+            if "timestamp" not in v.columns:
```
llllllllll: We should use the `TS_FIELD_NAME` to make this easier to rename later.

zipline/pipeline/loaders/buyback_auth.py
```diff
@@ -0,0 +1,114 @@
...
+                 infer_timestamps=False,
+                 dataset=CashBuybackAuthorizations):
+        super(CashBuybackAuthorizationsLoader, self).__init__(
+            all_dates,
+            events_by_sid,
+            infer_timestamps=infer_timestamps,
+            dataset=dataset
+        )
+
+    def get_loader(self, column):
```
llllllllll: do you think we can generalize this function? Maybe provide the superclass a dict from column to loader attribute name?

We might try using a naming scheme where we first check that the column is in our dataset and then do: `getattr(`%s_loader` % column.name)`. Just trying to think of ways to share code between these subclasses.

tests/pipeline/test_earnings.py
```diff
@@ -372,14 +375,20 @@ def test_infer_timestamp(self):
...
         )
         self.assertEqual(
-            loader.announcement_dates.keys(),
+            loader.events_by_sid.keys(),
             announcement_dates.keys(),
         )
         assert_series_equal(
-            loader.announcement_dates[0],
-            pd.Series(index=[dtx[0]] * 10, data=dtx),
+            pd.Series(loader.events_by_sid[0][ANNOUNCEMENT_FIELD_NAME]),
```
llllllllll: I don't think we need to wrap this in `pd.Series`. If events_by_sid is a dataframe, then indexing by a column should return a series. I also recomend using `.loc` acessing here like `df.loc[:, ANNOUNCEMENT_FIELD_NAME]`. This makes it clear that we are looking up by column label instead of index label.

tests/pipeline/test_earnings.py
```diff
@@ -118,7 +120,8 @@ def num_days_between(start_date, end_date):
 
         def zip_with_dates(dts):
             return pd.Series(pd.to_datetime(dts), index=dates)
-
+        # TODO: tests will break because I now need mappings of sid ->
```
llllllllll: looks like you covered this

zipline/pipeline/loaders/buyback_auth.py
mtydykov: I'm trying the dictionary version; but is there a way to reference a method decorated with `@lazyval` from within my dictionary without actually evaluating it?

mtydykov: Ah, never mind. I guess that's what the second part of your comment addresses. Thanks!

mtydykov:
@llllllllll I think the PR's ready for a second pass. 

Might would be a good idea to add another set of tests just for events (i.e., `test_events.py`). The test would be for a generic loader, and we could test things like proper initialization of an `EventsLoader`, particularly since the logic has now gotten more complex (allowing DatetimeIndexes and Series as well as DataFrames). It'd help simplify current tests for the loaders a bit, since there is some redundancy. Think it's worth it?

tests/pipeline/test_buyback_auth.py
```diff
@@ -0,0 +1,511 @@
...
+
+def num_days_between(dates, start_date, end_date):
+    return num_days_in_range(dates, start_date, end_date)
+
+
+def zip_with_dates(index_dates, dts):
+    return pd.Series(pd.to_datetime(dts), index=index_dates)
+
+
+class BuybackAuthLoaderCommonTest:
```
llllllllll: no old style classes in py2

mtydykov: Could you clarify what you mean?

mtydykov: I read up on it, so I think I know what you mean, but could you confirm: I need to subclass from object, right?

abhijeetkalyan: Classes should inherit from object, ie `class Blah(object):`

tests/pipeline/test_earnings.py
```diff
@@ -378,14 +382,22 @@ def test_infer_timestamp(self):
...
         )
         self.assertEqual(
-            loader.announcement_dates.keys(),
+            loader.events_by_sid.keys(),
             announcement_dates.keys(),
         )
         assert_series_equal(
-            loader.announcement_dates[0],
-            pd.Series(index=[dtx[0]] * 10, data=dtx),
+            loader.events_by_sid[0].loc[:, ANNOUNCEMENT_FIELD_NAME],
```
llllllllll: this should just be standard getitem. I profiled and `df[colname]` is much faster than the loc. Sorry about the false information

tests/pipeline/test_events.py
```diff
@@ -0,0 +1 @@
+__author__ = 'mtydykov'
```
llllllllll: cruft

tests/pipeline/test_buyback_auth.py
```diff
@@ -0,0 +1,511 @@
+"""
+Tests for the reference loader for EarningsCalendar.
```
llllllllll: this is for buyback auth

llllllllll: this test feels like it shares a lot with the earnings cal, do you think we can look at generalizing this for event sets?

tests/pipeline/test_earnings.py
```diff
@@ -57,30 +56,33 @@ def setUpClass(cls):
 
         cls.earnings_dates = {
             # K1--K2--E1--E2.
-            A: to_series(
-                knowledge_dates=['2014-01-05', '2014-01-10'],
-                earning_dates=['2014-01-15', '2014-01-20'],
-            ),
+            A: pd.DataFrame({
```
llllllllll: if we revert this we can test the promotion from series logic

zipline/pipeline/loaders/blaze/buyback_auth.py
```diff
@@ -0,0 +1,162 @@
...
+                 **kwargs):
+        super(
+            BlazeCashBuybackAuthorizationsLoader, self
+        ).__init__(expr,
+                   resources=resources,
+                   odo_kwargs=odo_kwargs,
+                   data_query_time=data_query_time,
+                   data_query_tz=data_query_tz,
+                   dataset=dataset,
+                   concrete_loader=concrete_loader,
```
llllllllll: The concrete loader feels like a class attribute because you wouldn't want instances of the same type with different concrete loaders. What do you think about this?

zipline/pipeline/loaders/buyback_auth.py
```diff
@@ -0,0 +1,102 @@
...
+
+from ..data.buyback_auth import (
+    CashBuybackAuthorizations,
+    ShareBuybackAuthorizations
+)
+from .events import EventsLoader
+from zipline.utils.memoize import lazyval
+
+
+BUYBACK_ANNOUNCEMENT_FIELD_NAME = 'buyback_dates'
```
llllllllll: field names should be singular

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+    expected_cols : frozenset
+        Set of expected columns for the dataset, without timestamp.
+    """
+
+    def __init__(self,
+                 all_dates,
+                 events_by_sid,
+                 infer_timestamps=False,
+                 dataset=None,
+                 expected_cols=frozenset()):
```
llllllllll: I think this would make more sense as a class attribute. Users should not make two instances of the same class that expect different columns

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+        #  purposes.
+        self.events_by_sid = (
+            events_by_sid.copy()
+        )
+        dates = self.all_dates.values
+
+        for k, v in iteritems(events_by_sid):
+            # First, must convert to DataFrame.
+            if isinstance(v, pd.Series):
+                # If Series was passed, DateTime index is assumed.
```
llllllllll: what do you mean that a datetime index is assumed?

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+                self.events_by_sid[k] = pd.DataFrame(v)
```
llllllllll: I think we need to name the column here

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+            elif isinstance(v, pd.DatetimeIndex):
+                if not infer_timestamps:
+                    raise ValueError(
+                        "Got DatetimeIndex for sid %d.\n"
+                        "Pass `infer_timestamps=True` to use the first date in"
+                        " `all_dates` as implicit timestamp." % k
+                    )
+                self.events_by_sid[k] = pd.DataFrame(v)
+                v.index = [dates[0]] * len(v)
```
llllllllll: I think we need to name the column here. We can also pass the index as a keyword to the dataframe ctor.

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+            # Already a DataFrame
+            elif isinstance(v, pd.DataFrame):
```
llllllllll: I think this should not be an elif. If we have hit either of the two branches above then we will have promoted to a dataframe but we still need to make these checks. Also, as written now we will silently ignore anythin that is not a datetimeindex, series, or dataframe.

Can we add some tests for these conversion cases?

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+                if TS_FIELD_NAME not in v.columns:
+                    if not infer_timestamps:
+                        raise ValueError(
+                            "Got DataFrame without a '%s' column for sid %d.\n"
```
llllllllll: instead of `'%s'` you can use `%r`. This will correctly use double quotes if the string contains a single quote or will escape quotes in the case that both double and single quotes appear in the string.

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+                        )
+                    self.events_by_sid[k] = v = v.copy()
+                    v.index = [dates[0]] * len(v)
+                else:
+                    self.events_by_sid[k] = v.set_index(TS_FIELD_NAME)
+            else:
+                raise ValueError("Data for sid %s must be in DataFrame, "
+                                 "Series, or DatetimeIndex." % k)
+            # Once data is in a DF, make sure columns are correct.
+            cols_except_ts = (set(v.columns.values) -
```
llllllllll: we shouldn't need to thake the values of the columns here, this will coerce to an ndarray of strings. Also, why are we stripping out a sid column?

Do you think we should union the `TS_FIELD_NAME` with the expected fields earlier so that at runtime a user can inspect what the expected fields are and see that they need to iinclude `TS_FIELD_NAME`

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+        return merge(
+            self.get_loader(column).load_adjusted_array(
+                [column], dates, assets, mask
+            )
+            for column in columns
+        )
+
+    def mk_date_series(self, date_field_name):
+        return {sid: pd.Series(index=event.index,
+                               data=np.array(event[date_field_name]))
```
llllllllll: this is going to be making a lot of copies of the data. instead we can just take slices with:

```python
{sid: event_df[date_field_name] for sid, event_df in iteritems(self.events_by_sid)}`
```

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,154 @@
...
+            adjustments=None,
+        )
+
+    def _previous_event_value_loader(self,
+                                     previous_value_field,
+                                     event_date_field_name,
+                                     value_field_name):
+        return DataFrameLoader(
+            previous_value_field,
+            previous_value(
```
llllllllll: There is now asymmetry between `previous_date_frame` and `previous_value_field`. How do you feel about making `previous_date_field` accept the column names and do the slicing internally to be more consistent. The old API made more sense when we were storing the data as series but this is no longer true.

tests/pipeline/test_events.py
mtydykov: Whoops. Didn't mean to commit this.

tests/pipeline/test_buyback_auth.py
mtydykov: Yup, hence my comment about creating a `test_events.py`. Sounds good.

tests/pipeline/test_earnings.py
mtydykov: Yeah, but maybe instead of reverting, we can test that promotion logic separately from events/buyback auth in a common test for events (`test_events.py`).

zipline/pipeline/loaders/blaze/buyback_auth.py
mtydykov: Yeah, I agree

zipline/pipeline/loaders/events.py
mtydykov: Sorry, I just realized that comment makes no sense out of context. I should probably have just copied over the comment from the version of `loaders/earnings.py` that's on master:
`If a dict value is a Series, it's interpreted as a mapping from the date on which we learned an announcement was coming to the date on which the announcement was made.`

zipline/pipeline/loaders/events.py
mtydykov: What do you mean? Shouldn't the caller (whoever passes in `events_by_sid`) be responsible for naming the Series? How would we know the name otherwise at this point?

zipline/pipeline/loaders/events.py
mtydykov: If we hit either of the two previous branches, we won't need to make this check, because there will never be a `TS_FIELD_NAME` column at this point, though the index will already have been set. So I don't think it would make sense for the DataFrames generated in the other two cases to go through these checks, hence the `elif`.

What do you mean we'll silently ignore anything that's not any of those 3? That's why I raise a `ValueError` if we don't enter any of those 3 previous cases.

Yeah, I think I'll need to add tests in `test_events.py`.

zipline/pipeline/loaders/events.py
mtydykov: I'm stripping out a sid column because in some cases (e.g., see the tests for the blaze loaders), we get passed a DataFrame that contains a `sid` column, but not in all cases, and this is not one of the columns that needs to be checked.

We could do that, but I thought that we were making `TS_FIELD_NAME` optional on purpose, so that if it's not included, we could infer the timestamp?

zipline/pipeline/loaders/events.py
mtydykov: Makes sense

mtydykov:
@llllllllll comments addressed - except some that I wasn't clear on. 

Most notably, `test_events.py` has been filled with tests for setting up an `EventsLoader` and a `BlazeEventsLoader`. This helped simplify tests for buyback_auth and earnings a little bit. Let me know if you think there's something missing from those tests.

zipline/pipeline/loaders/blaze/earnings.py
```diff
@@ -1,29 +1,14 @@
...
-from zipline.utils.input_validation import ensure_timezone, optionally
-from zipline.utils.preprocess import preprocess
-
+from .events import BlazeEventsLoader
 
-ANNOUNCEMENT_FIELD_NAME = 'announcement_date'
 
-
-class BlazeEarningsCalendarLoader(PipelineLoader):
+class BlazeEarningsLoader(BlazeEventsLoader):
```
llllllllll: why did we rename this?

mtydykov: Because the concrete loader is called `EventsLoader`, so it seems like these 2 loaders should follow a pattern.

mtydykov: Whoops, that is not what I meant to rename.

mtydykov: Blaming PyCharm refactoring... and my occasional inability to read

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,183 @@
...
+        Whether to allow omitting the "timestamp" column.
+    dataset : DataSet
+        The DataSet object for which this loader loads data.
+    expected_cols : frozenset
+        Set of expected columns for the dataset, without timestamp.
+    """
+
+    @abc.abstractproperty
+    def expected_cols(self):
+        pass
```
llllllllll: `raise NotImplemented('expected_cols')`

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,183 @@
...
+        on which we learned about the event. If the DataFrames do not contain a
+         "timestamp" column, we assume we knew about the event on all prior
+         dates.  This mode is only supported if ``infer_timestamp`` is
+         explicitly passed as a truthy value.
+
+    infer_timestamps : bool, optional
+        Whether to allow omitting the "timestamp" column.
+    dataset : DataSet
+        The DataSet object for which this loader loads data.
+    expected_cols : frozenset
```
llllllllll: no longer an arg

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,183 @@
...
+    Abstract loader.
+
+    Does not currently support adjustments to the dates of known events.
+
+    Parameters
+    ----------
+    all_dates : pd.DatetimeIndex
+        Index of dates for which we can serve queries.
+    events_by_sid : dict[int -> pd.DataFrame], dict[int -> pd.Series],
+    or dict[int -> pd.DatetimeIndex]
```
llllllllll: this is more `dict[int -> pd.DataFrame or pd.Series or pd.DatetimeIndex]` because it does not need to be homogenous.

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,183 @@
...
+                              "Pass `infer_timestamps=True` to use the first "
+                              "date in `all_dates` as implicit timestamp.")
+
+DF_NO_TS_NOT_INFER_TS_ERROR = ("Got DataFrame without a '%r' column for sid "
+                               "%d.\nPass `infer_timestamps=True` to use the "
+                               "first date in `all_dates` as implicit "
+                               "timestamp.")
+
+TS_FIELD_NAME = "timestamp"
+SID_FIELD_NAME = "sid"
```
llllllllll: these constants are defined in a couple modules, could we try to centralize them?

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,183 @@
...
+                           "DatetimeIndex.")
+
+DTINDEX_NOT_INFER_TS_ERROR = ("Got DatetimeIndex for sid %d.\n"
+                              "Pass `infer_timestamps=True` to use the first "
+                              "date in `all_dates` as implicit timestamp.")
+
+DF_NO_TS_NOT_INFER_TS_ERROR = ("Got DataFrame without a '%r' column for sid "
+                               "%d.\nPass `infer_timestamps=True` to use the "
+                               "first date in `all_dates` as implicit "
+                               "timestamp.")
```
llllllllll: Why are we pulling these into module level constants?

zipline/pipeline/loaders/buyback_auth.py
```diff
@@ -0,0 +1,103 @@
...
+                 dataset=CashBuybackAuthorizations):
+        super(CashBuybackAuthorizationsLoader, self).__init__(
+            all_dates,
+            events_by_sid,
+            infer_timestamps=infer_timestamps,
+            dataset=dataset,
+        )
+
+    @property
+    def expected_cols(self):
```
llllllllll: also here: why not just make this a normal class level attribute?

zipline/pipeline/loaders/buyback_auth.py
```diff
@@ -0,0 +1,103 @@
+"""
+Reference implementation for EarningsCalendar loaders.
```
llllllllll: buyback auth

zipline/pipeline/loaders/blaze/events.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+        self._odo_kwargs = odo_kwargs if odo_kwargs is not None else {}
+        self._dataset = dataset
+        check_data_query_args(data_query_time, data_query_tz)
+        self._data_query_time = data_query_time
+        self._data_query_tz = data_query_tz
+
+    @abc.abstractproperty
+    def concrete_loader(self):
+        pass
```
llllllllll: this should raise a notimplementederror on access like: `NotImplementedError('concrete_loader')`

zipline/pipeline/loaders/blaze/events.py
```diff
@@ -0,0 +1,124 @@
...
+    resources : dict, optional
+        Mapping from the atomic terms of ``expr`` to actual data resources.
+    odo_kwargs : dict, optional
+        Extra keyword arguments to pass to odo when executing the expression.
+    data_query_time : time, optional
+        The time to use for the data query cutoff.
+    data_query_tz : tzinfo or str
+        The timezeone to use for the data query cutoff.
+    dataset : DataSet
+        The DataSet object for which this loader loads data.
```
llllllllll: need a newline before `notes`

zipline/pipeline/loaders/blaze/earnings.py
```diff
@@ -69,75 +59,21 @@ class BlazeEarningsCalendarLoader(PipelineLoader):
...
+                 **kwargs):
+        super(
+            BlazeEarningsLoader, self
+        ).__init__(expr, dataset=dataset,
+                   resources=resources, odo_kwargs=odo_kwargs,
+                   data_query_time=data_query_time,
+                   data_query_tz=data_query_tz, **kwargs)
+
+    @property
+    def concrete_loader(self):
```
llllllllll: why is this a property instead of a normal attribute?

zipline/pipeline/loaders/blaze/earnings.py
```diff
@@ -39,6 +24,10 @@ class BlazeEarningsCalendarLoader(PipelineLoader):
         The time to use for the data query cutoff.
     data_query_tz : tzinfo or str
         The timezeone to use for the data query cutoff.
+    dataset: DataSet
+        The DataSet object for which this loader loads data.
+    concrete_loader: EventsLoader
```
llllllllll: this is no longer an argument

zipline/pipeline/loaders/buyback_auth.py
```diff
@@ -0,0 +1,103 @@
...
+CASH_FIELD_NAME = 'cash'
+
+
+class CashBuybackAuthorizationsLoader(EventsLoader):
+    """
+    Reference loader for
+    :class:`zipline.pipeline.data.earnings.CashBuybackAuthorizations`.
+
+    events_by_sid: dict[sid -> pd.DataFrame(knowledge date,
+    event date, cash value)]
```
llllllllll: I would indent this so it doesn't look like two arguments.

zipline/pipeline/loaders/events.py
llllllllll: I think in that case we should just strip the `SID_FIELD_NAME` out at the call site instead of silently allowing it.

zipline/pipeline/loaders/events.py
llllllllll: We know what the name for the column should be because we can only promote from series in the case where we have one expected column

zipline/pipeline/loaders/events.py
llllllllll: sorry, I didn't see the final `else` clause.

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,183 @@
...
+        # Do not modify the original in place, since it may be used for other
+        #  purposes.
+        self.events_by_sid = (
+            events_by_sid.copy()
+        )
+        dates = self.all_dates.values
+
+        for k, v in iteritems(events_by_sid):
+            # First, must convert to DataFrame.
+            if isinstance(v, pd.Series):
```
llllllllll: the series and datetime index case can only happen when we only expect one col

zipline/pipeline/loaders/blaze/__init__.py
```diff
@@ -15,11 +25,16 @@
 __all__ = (
```
llllllllll: can we sort this to make it easier to read

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,191 @@
...
+
+        if infer_timestamps:
+            expected = pd.Series(index=[dtx[0]] * 10, data=dtx, )
+        else:
+            expected = pd.Series(index=dtx, data=dtx,)
+        # Check that index by first given date has been added
+        assert_series_equal(
+            loader.events_by_sid[0][ANNOUNCEMENT_FIELD_NAME],
+            expected,
+            check_names=False
```
llllllllll: I think we need to check the names of the columns because we use that to do slicing later.

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,191 @@
+"""
+Tests for setting up an EventsLoader and a BlazeEventsLoader.
+"""
+from nose_parameterized import parameterized
+
+import blaze as bz
+import pandas as pd
+from pandas.util.testing import assert_series_equal, TestCase, assertRaises
```
llllllllll: we can just use the `assertRaises` method of `TestCase`

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,191 @@
...
+from zipline.utils.memoize import lazyval
+from zipline.utils.numpy_utils import datetime64ns_dtype
+
+ABSTRACT_METHODS_ERROR = 'abstract methods concrete_loader'
+
+DAYS_SINCE_PREV = 'days_since_prev'
+
+PREVIOUS_ANNOUNCEMENT = 'previous_announcement'
+
+ANNOUNCEMENT_FIELD_NAME = 'announcement_date'
```
llllllllll: We should import these from wherever they are defined

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,191 @@
...
+            events_by_sid,
+            infer_timestamps=infer_timestamps,
+            dataset=dataset,
+        )
+
+
+dtx = pd.date_range('2014-01-01', '2014-01-10')
+
+
+def assert_loader_error(events_by_sid, error, msg, infer_timestamps=True):
```
llllllllll: idk if this should have a default arg since we should be explicit about this in the tests.

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,191 @@
...
+    with assertRaises(error) as context:
+        EventDataSetLoader(
+            dtx, events_by_sid, infer_timestamps=infer_timestamps,
+        )
+        assert msg in context.exception
```
llllllllll: we should use the assertion methods of TestCase instead of assert statements.

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,191 @@
...
+        wrong_col_name = 'some_other_col'
+        # Test wrong cols (cols != expected)
+        events_by_sid = {0: pd.DataFrame({wrong_col_name: dtx})}
+        assert_loader_error(
+            events_by_sid, ValueError, WRONG_COLS_ERROR % (
+                EventDataSetLoader.expected_cols, 0, wrong_col_name
+            )
+        )
+
+    @parameterized.expand([
```
llllllllll: good cases and comments to explain, thanks!

tests/pipeline/test_earnings.py
llllllllll: kk, I agree that it is probably best to test this in isolation in test_events

tests/pipeline/test_earnings.py
```diff
@@ -57,30 +56,33 @@ def setUpClass(cls):
...
-                knowledge_dates=['2014-01-05'] * 2,
-                earning_dates=['2014-01-10', '2014-01-15'],
-            ),
-            E: pd.Series(
-                data=[],
-                index=pd.DatetimeIndex([]),
-                dtype='datetime64[ns]',
-            ),
+            D: pd.DataFrame({
+                "timestamp": pd.to_datetime(['2014-01-05'] * 2),
```
llllllllll: should we be using `TS_FIELD_NAME` here?

tests/pipeline/test_buyback_auth.py
```diff
@@ -0,0 +1,464 @@
...
+sids = A, B, C, D, E = range(5)
+
+equity_info = make_simple_equity_info(
+    sids,
+    start_date=pd.Timestamp('2013-01-01', tz='UTC'),
+    end_date=pd.Timestamp('2015-01-01', tz='UTC'),
+)
+
+buyback_authorizations = {
+    # K1--K2--A1--A2--SC1--SC2--V1--V2.
```
llllllllll: I am not sure what the sc and v fields are here. I think that this case is till just `k1--k2--a1--a2`

zipline/pipeline/loaders/events.py
mtydykov: So, it makes sense to me why we'd have this, and I had a `NotImplemented` exception here originally, but it never gets there because it raises a `TypeError` first. Still necessary?

zipline/pipeline/loaders/buyback_auth.py
mtydykov: I think it's a neater/more elegant way of enforcing that each loader to adds `expected_cols`, since this way, the `EventsLoader` can define an abstract property and break immediately if the property isn't added.

zipline/pipeline/loaders/blaze/events.py
mtydykov: Same comment as for `expected_cols` - since I'm using `abc.abstractproperty`, won't the `NotImplementedError` never be reached? Maybe it's best practice to have it there anyway? Or maybe I shouldn't be using `abc.abstractproperty`?

zipline/pipeline/loaders/blaze/earnings.py
mtydykov: Same as for `expected_cols` - I think it's a more elegant way of enforcing that subclasses implement this.

zipline/pipeline/loaders/events.py
mtydykov: Could you clarify where you want the `SID_FIELD_NAME` to be removed?

zipline/pipeline/loaders/events.py
mtydykov: So you're saying that we should name the column based on what's in `expected_cols`?

zipline/pipeline/loaders/events.py
mtydykov: Could you clarify what you mean by that/what you want me to change?

tests/pipeline/test_buyback_auth.py
mtydykov: Forgot to consistently update the comments for this section; but the SC field is there (share count) and the V is there, but it should be modified since it refers to the cash field.

mtydykov: Ah, I see what you mean. I guess the logic is based around the knowledge/announcement dates, and the original comments from `test_earnings.py` reflected that. So to outline the logic of these tests, SC/V are not needed.

zipline/pipeline/loaders/events.py
mtydykov: I'd like to test that these error messages are correctly raised, and also better style. Did you mean why are they specifically module-level rather than class-level?

zipline/pipeline/loaders/buyback_auth.py
llllllllll: `abstractpropery` and `abstractmethod` only check that the value is in the class dict at all. You should make the abc have an abstract property to get the checks, but the subclasses can just put a frozen set in the class dict to satisfy the check.

zipline/pipeline/loaders/events.py
llllllllll: We cannot always adapt from a series. The only case where this promotion is valid is when we expect a single column other than timestamp

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,420 @@
...
+
+dtx = pd.date_range('2014-01-01', '2014-01-10')
+
+
+def assert_loader_error(events_by_sid, error, msg, infer_timestamps):
+    with TestCase.assertRaises(error) as context:
+        EventDataSetLoader(
+            dtx, events_by_sid, infer_timestamps=infer_timestamps,
+        )
+        TestCase.assertTrue(msg in context.exception)
```
llllllllll: assertContains(a, b)

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,420 @@
...
+    def test_no_concrete_loader_defined(self):
+        with TestCase.assertRaises(TypeError) as context:
+            BlazeEventDataSetLoaderNoConcreteLoader(
+                bz.Data(
+                    pd.DataFrame({ANNOUNCEMENT_FIELD_NAME: dtx,
+                                  SID_FIELD_NAME: 0
+                                  })
+                )
+            )
+            TestCase.assertTrue(ABSTRACT_METHODS_ERROR in context.exception)
```
llllllllll: self.assertContains

tests/pipeline/test_earnings.py
```diff
@@ -8,170 +8,101 @@
...
     make_simple_equity_info,
     tmp_asset_finder,
-    gen_calendars,
-    to_series,
-    num_days_in_range,
 )
 
-
-class EarningsCalendarLoaderTestCase(TestCase):
+earnings_dates = [
```
llllllllll: lift this into the test base class

tests/pipeline/test_buyback_auth.py
```diff
@@ -0,0 +1,297 @@
...
+        columns=[TS_FIELD_NAME,
+                 BUYBACK_ANNOUNCEMENT_FIELD_NAME,
+                 SHARE_COUNT_FIELD_NAME,
+                 CASH_FIELD_NAME],
+        dtype='datetime64[ns]'
+    ),
+]
+
+
+class CashBuybackAuthLoaderTestCase(TestCase, EventLoaderCommonTest):
```
llllllllll: rename EventLoaderCommonTest to EventLoaderCommonTestMixin

tests/pipeline/test_buyback_auth.py
```diff
@@ -0,0 +1,297 @@
...
+        ] = _expected_previous_buyback_share_count
+        self.cols[PREVIOUS_BUYBACK_ANNOUNCEMENT] = \
+            self.get_expected_previous_event_dates(dates)
+        self.cols[DAYS_SINCE_PREV] = self._compute_busday_offsets(
+            self.cols[PREVIOUS_BUYBACK_ANNOUNCEMENT]
+        )
+
+    @parameterized.expand(param_dates)
+    def test_compute_share_buyback_auth(self, dates):
+        self._test_compute(dates)
```
llllllllll: rename `_test_compute` to `test_compute` on the base and then classes will inherit this test when they use the mixin

zipline/pipeline/loaders/blaze/__init__.py
mtydykov: Done

tests/pipeline/test_events.py
mtydykov: Unfortunately, that doesn't work (`has no attribute assertContains`)

richafrank: I think he meant `self.assertIn`.

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,420 @@
...
+            BlazeEventDataSetLoaderNoConcreteLoader, self
+        ).__init__(expr,
+                   dataset=dataset,
+                   **kwargs)
+
+
+class BlazeEventLoaderTestCase(TestCase):
+    # Blaze loader: need to test failure if no concrete loader
+    def test_no_concrete_loader_defined(self):
+        with TestCase.assertRaises(TypeError) as context:
```
richafrank: Any reason not to use `self.assertRaises`?

mtydykov: Nope, that works too

tests/pipeline/test_events.py
mtydykov: Yeah, I'm using that one

zipline/pipeline/loaders/blaze/events.py
mtydykov: Done

tests/pipeline/test_events.py
mtydykov: Done - not sure why the diff is still showing the old version

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,203 @@
...
+        self.all_dates = all_dates
+        # Do not modify the original in place, since it may be used for other
+        #  purposes.
+        self.events_by_sid = (
+            events_by_sid.copy()
+        )
+        dates = self.all_dates.values
+
+        for k, v in iteritems(events_by_sid):
+            # Already a DataFrame
```
mtydykov: @llllllllll could you look over this logic again? I had to change it around a bit to incorporate taking expected_cols into account.

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,203 @@
...
+                if cols_except_ts != self.expected_cols:
+                    raise ValueError(
+                        WRONG_COLS_ERROR .format(
+                            expected_columns=self.expected_cols,
+                            sid=k,
+                            resulting_columns=v.columns.values
+                        )
+                    )
+            # Not a DataFrame and we only expect 1 column
+            elif len(self.expected_cols) == 1:
```
llllllllll: We should invert this check to flatten the logic here. This can be: `if len(self.expected_cols) != 1: raise ...` and then the rest of the code does not need to be indented.

zipline/pipeline/loaders/events.py
```diff
@@ -0,0 +1,203 @@
...
+                else:
+                    # We expect 1 column, but we got something other than a
+                    # Series, DatetimeIndex, or DataFrame.
+                    raise ValueError(WRONG_SINGLE_COL_DATA_FORMAT_ERROR
+                                     .format(sid=k))
+            else:
+                # We expected multiple columns, but we got something other
+                # than a DataFrame.
+                raise ValueError(WRONG_MANY_COL_DATA_FORMAT_ERROR
+                                 .format(sid=k))
```
llllllllll: Could we try to not break the line on an attribute access. I find this style really hard to read

tests/pipeline/test_buyback_auth.py
```diff
@@ -0,0 +1,287 @@
...
+        }, index=dates)
+        self.cols[PREVIOUS_BUYBACK_ANNOUNCEMENT] = \
+            self.get_expected_previous_event_dates(dates)
+        self.cols[PREVIOUS_BUYBACK_CASH] = _expected_previous_cash
+        self.cols[DAYS_SINCE_PREV] = self._compute_busday_offsets(
+            self.cols[PREVIOUS_BUYBACK_ANNOUNCEMENT]
+        )
+
+
+class ShareBuybackAuthLoaderTestCase(EventLoaderCommonMixin, TestCase):
```
llllllllll: Mixins normally come after the concrete super class

zipline/pipeline/loaders/events.py
mtydykov: Doesn't work (anything in a DataFrame would error)

zipline/pipeline/loaders/events.py
mtydykov: Done - not showing up for some reason

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,481 @@
...
+    DF_NO_TS_NOT_INFER_TS_ERROR,
+    DTINDEX_NOT_INFER_TS_ERROR,
+    EventsLoader,
+    SERIES_NO_DTINDEX_ERROR,
+    WRONG_COLS_ERROR,
+    WRONG_MANY_COL_DATA_FORMAT_ERROR,
+    WRONG_SINGLE_COL_DATA_FORMAT_ERROR
+)
+from zipline.utils.memoize import lazyval
+from zipline.utils.numpy_utils import (datetime64ns_dtype,
```
llllllllll: please indent this on a newline for consistency

tests/pipeline/test_events.py
```diff
@@ -0,0 +1,481 @@
...
+                                       NaTD,
+                                       make_datetime64D)
+from zipline.utils.test_utils import gen_calendars, num_days_in_range, \
```
llllllllll: same here

zipline/pipeline/loaders/events.py
llllllllll: talked about this in person, I am wrong

zipline/pipeline/loaders/events.py
llllllllll: are you sure it was committed?

zipline/pipeline/loaders/events.py
mtydykov: Yup, just making a note

llllllllll:
One last comment, can we squash some of these commits? After that, merge on pass!

mtydykov:
Yeah, for sure, that is a LOT of commits. Sounds good, yay!

PR #991, opened 2016-02-10
BUG: Add right-binding operators to NumExprFilter.

ssanderson:
Fixes a bug where doing a boolean comparison between a non-numexpr
Filter and a NumExprFilter would fail because we don't implement
`__rand__` and `__ror__`.

### DIFF
--- a/tests/pipeline/test_numerical_expression.py
+++ b/tests/pipeline/test_numerical_expression.py
@@ -1,3 +1,4 @@
+from itertools import permutations
 from operator import (
     add,
     ge,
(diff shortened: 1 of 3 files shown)


tests/pipeline/test_numerical_expression.py
```diff
@@ -460,26 +465,57 @@ def test_comparisons(self):
...
         self.fake_raw_data = {
             f: arange(25).reshape(5, 5),
             g: arange(25).reshape(5, 5) - eye(5),
             h: full((5, 5), 5),
+            custom_filter: custom_filter_mask,
         }
 
         # Should be True on the diagonal.
-        eye_filter = f > g
+        eye_filter = (f > g)
```
llllllllll: why the parens?

ssanderson: I think it's slightly more readable.

llllllllll: kk

ssanderson: It's easy to misread this as a chained comparison.

llllllllll:
Test was really clear, thanks! lgtm

PR #992, opened 2016-02-11
Fix error like data[symbol('AAPL')]

waveyeung:
also fix data[sid(0)]

### DIFF
--- a/zipline/protocol.py
+++ b/zipline/protocol.py
@@ -25,6 +25,7 @@ from zipline.utils.algo_instance import get_algo_instance
 from zipline.utils.serialization_utils import (
     VERSION_LABEL
 )
+from zipline.assets import Equity
 
 # Datasource type should completely determine the other fields of a
 # message with its type.
@@ -525,7 +526,10 @@ class BarData(object):
         self._data[name] = value
 
     def __getitem__(self, name):
-        return self._data[name]
+        if isinstance(name,Equity):
+            return self._data[name.sid]
+        else:
+            return self._data[name]
 
     def __delitem__(self, name):
         del self._data[name]


richafrank:
Hi @waveyeung ,  Thanks for the PR.  Do you have an example of code that was failing without this change?  

Also, have you tried using the latest of zipline master from github?  I ask because this was likely already fixed by #959.  On latest master, I see:
```python
In [1]: from zipline.assets import Equity

In [2]: d = {1: 'found'}

In [3]: d[Equity(sid=1)]
Out[3]: 'found'
```
even with 32-bit python, so the explicit use of the `sid` attribute should be unnecessary.

waveyeung:
Hi @richafrank 
My env is zipline-0.8.3 from pip, py2.7

```
from zipline.api import record, symbol, symbols, sid

def initialize(context):
    context.stocks = symbols('AAPL', 'MFST')

    for stock in encontext.stocks:
        price = data[stock].price # or price = data[sid(0)].price
```

code raise excetption:

>   File "C:\Python27\lib\site-packages\zipline\protocol.py", line 521, in __getitem__
>     return self._data[name]
> KeyError: Equity(0, symbol=u'AAPL', asset_name=None, exchange=None, start_date=Timestamp('1970-01-01 00:00:00+0000', tz='UTC'), end_date=Timestamp('2116-02-20 23:53:38.427387903+0000', tz='UTC'), first_traded=None

checked source code at zipline master from github, it doesn't seems to solve the problem.

waveyeung:
Hi @richafrank 
I found master fix the problem, good job!

richafrank:
Great - glad to hear it!

PR #996, opened 2016-02-15
Added Dockerfile

jxstanford:
Builds docker image with  TA-Lib, Jupyter, and checked out version of zipline and
starts Jupyter server on port https://localhost:8888 when container runs.  The
default password for the Jupyter server is 'jupyter'.

Instructions on usage found in the Dockerfile comments.  Example image at
docker.io/jxstanford/zipline.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,81 @@
+#
+# Dockerfile for an image with the currently checked out version of zipline installed. To build:
+#
+#    docker build -t quantopian/zipline .
+# 
+# To run:
+#
+#    docker run -v=/path/to/your/notebooks:/project -p 8888:8888/tcp --name zipline -it quantopian/zipline
+#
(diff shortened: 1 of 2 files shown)


jxstanford:
Hope you find this useful.  I think it should help people get a more consistent build experience with zipline.  I've been using it with the latest versions of pycharm (which can use a docker container as a remote execution environment).

jxstanford:
I'm going to withdraw this PR for the moment.  After tinkering with pycharm and the container, I realized that the pycharm/ipython integration has some trouble with SSL and passwords on the jupyter server.  I'll make docker default to no password/no SSL, but allow them to be set easily via the --env flag (or a compose file).

I would also like to find a more elegant way to manage custom requirements such as matplotlib.  Currently, you would need to log in, pip install them, and then commit the changes back to your local image.  Would like to defer to a local requirements file specified in an env var.

Will resubmit when those things are cleaned up.

PR #998, opened 2016-02-16
MAINT: Upgrade Logbook version

jikamens:
Upgrade Logbook to 0.12.5. This required changing a usage of
`logbook.NullHandler()` which passed `bubble=True`, since
`NullHandler` no longer supports the `bubble` argument.

### DIFF
--- a/conda/logbook/meta.yaml
+++ b/conda/logbook/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: logbook
-  version: "0.10.0"
+  version: "0.12.5"
 
 source:
-  fn: Logbook-0.10.0.tar.gz
-  url: https://pypi.python.org/packages/source/L/Logbook/Logbook-0.10.0.tar.gz
-  md5: 92439ce6f71f3120d65d84c2a3ab5047
+  fn: Logbook-0.12.5.tar.gz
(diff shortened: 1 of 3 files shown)


ssanderson:
:+1: from me assuming we're happy with any downstream implications of this.

richafrank:
I think we need to update our conda recipe for logbook. https://github.com/quantopian/zipline/blob/master/conda/logbook/meta.yaml was generated with `conda skeleton`

jikamens:
>I think we need to update our conda recipe for logbook.

Any idea why the conda recipe has a dependency on six? I don't see six referenced anywhere in the source code of Logbook 0.10.0 or 0.12.5.

richafrank:
> Any idea why the conda recipe has a dependency on six?

At least in version 0.10, they specify it as [install_requires] (https://github.com/getlogbook/logbook/blob/2694230b0ce21a8e369b4b6663da78f26289e920/setup.py#L122).

jikamens:
Oh, I see it now. Silly me trying to use "git grep" in a directory that isn't a git repository. That dependency is gone in 0.12.5.

I can easily update the conda recipe, but how do I test that update?

PR #1000, opened 2016-02-17
Introduces new API for accessing price data:

jbredeche:
data.get_spot_value
data.can_trade
data.is_stale
data.history


jbredeche:
The `test_history.py` diff is huge because I completely rewrote the history tests.

tests/test_bar_data.py
```diff
@@ -0,0 +1,418 @@
+from unittest import TestCase
```
ehebert: Copyright.

tests/test_bar_data.py
```diff
@@ -0,0 +1,418 @@
...
+from zipline.utils.test_utils import write_minute_data_for_asset, \
+    create_daily_df_for_asset, DailyBarWriterFromDataFrames, assert_same
+
+OHLC = ["open", "high", "low", "close"]
+OHLCP = OHLC + ["price"]
+ALL_FIELDS = OHLCP + ["volume", "last_traded"]
+
+# offsets used in test data
+field_info = {
+    "open": 1,
```
ehebert: Could we make these mantissas instead?
So that same value is not in different slots for OHLC. e.g. low of i, and close of i+1 would have the same value.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+                    self.simulation_dt_func(),
+                    self.data_frequency
+                )
+            else:
+                # assume fields is iterable
+                # return a Series indexed by field
+                return pd.Series(data={
+                    field: self.data_portal.get_spot_value(
+                        assets,
+                        field,
```
ehebert: We should make (but not required for this PR) `get_spot_value` accept the iterable as well.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+                        self.simulation_dt_func(),
+                        self.data_frequency)
+                    for field in fields
+                }, index=fields, name=assets.symbol)
+        else:
+            if isinstance(fields, str):
+                # assume assets is iterable
+                # return a Series indexed by asset
+                return pd.Series(data={
+                    asset: self.data_portal.get_spot_value(
```
ehebert: If `get_spot_value` returns the values in the same order as they are passed in, the returned array from `get_spot_value` can be passed to the Series without the extra step of the dict.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+                        assets,
+                        field,
+                        self.simulation_dt_func(),
+                        self.data_frequency)
+                    for field in fields
+                }, index=fields, name=assets.symbol)
+        else:
+            if isinstance(fields, str):
+                # assume assets is iterable
+                # return a Series indexed by asset
```
ehebert: What do you think of doing a `field = fields` here, just so that scalar/iterable is clearer.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+                        self.simulation_dt_func(),
+                        self.data_frequency)
+                    for asset in assets
+                    }, index=assets, name=fields)
+            else:
+                # both assets and fields are iterable
+                data = {}
+
+                for field in fields:
+                    series = pd.Series(data={
```
ehebert: This works, but I think having `get_spot_value` return a ndarray may help here.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+            })
+
+    cdef _is_stale_for_asset(self, asset, dt, data_portal):
+        if asset.start_date > dt:
+            return False
+
+        if asset.end_date <= dt:
+            return False
+
+        current_val = data_portal.get_spot_value(asset, "close", dt,
```
ehebert: I'd recommend using `volume` since a check against 0 is cheaper.

zipline/data/data_portal.py
```diff
@@ -286,41 +286,45 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
-            field,
-            dt,
-        )
-
-        if extra_source_val is not None:
-            return extra_source_val
-
-        if field not in BASE_FIELDS:
-            raise KeyError("Invalid column: " + str(field))
+        # extra_source_val = self._check_extra_sources(
```
ehebert: Remove?

zipline/data/data_portal.py
```diff
@@ -35,16 +35,17 @@
...
-    'open_price': 'open',
-    'high': 'high',
-    'low': 'low',
-    'close': 'close',
-    'close_price': 'close',
-    'volume': 'volume',
-    'price': 'close'
-}
+BASE_FIELDS = frozenset([
+    "open", "high", "low", "close", "volume", "price", "last_traded"
```
ehebert: Is `open_price` intentionally no longer supported with this change?

tests/test_algorithm.py
```diff
@@ -706,7 +706,7 @@ def setUpClass(cls):
         for sid in cls.sids:
             equities_metadata[sid] = {
                 'start_date': cls.sim_params.period_start,
-                'end_date': cls.sim_params.period_end
+                'end_date': cls.env.next_trading_day(cls.sim_params.period_end)
```
ehebert: Why does the equity need to end beyond the trading range?

tests/test_bar_data.py
```diff
@@ -0,0 +1,429 @@
...
+        self.assertEqual(4, bar_data.spot_value(self.ASSET1, "high"))
+        self.assertEqual(1, bar_data.spot_value(self.ASSET1, "low"))
+        self.assertEqual(2, bar_data.spot_value(self.ASSET1, "close"))
+        self.assertEqual(200, bar_data.spot_value(self.ASSET1, "volume"))
+        self.assertEqual(2, bar_data.spot_value(self.ASSET1, "price"))
+        self.assertEqual(self.days[0],
+                         bar_data.spot_value(self.ASSET1, "last_traded"))
+
+        for field in OHLCP:
+            self.assertTrue(np.isnan(bar_data.spot_value(self.ASSET2, field)))
```
ehebert: So that it can be seen which if any values fail. It would be helpful here to either include the field as the 3rd argument to the assert, or `np.testing.assert_array_equal` does support checking `np.nan`s against eachother, so this could be.

```
np.testing.assert_array_equal(
   [np.nan] * 5,
   [bar_data.spot_value(self.ASSET2, field) for field in OHLCP]
)
```

 

```

File: tests/test_blotter.py [hunk 12]
```diff
@@ -58,7 +58,9 @@ def setUpClass(cls):
         cls.env.write_data(equities_data={
             24: {
                 'start_date': cls.sim_params.trading_days[0],
-                'end_date': cls.sim_params.trading_days[-1]
+                'end_date': cls.env.next_trading_day(
```
ehebert: Why is this the only change in this file?

zipline/_protocol.pxd
```diff
@@ -18,6 +18,10 @@ cdef class BarData:
     cdef object simulation_dt_func
     cdef object data_frequency
     cdef dict _views
+    cdef object _get_equity_price_view(BarData, object)
+    cdef object _create_sid_view(BarData, object)
+    cdef object _is_stale_for_asset(BarData, object, object, object)
```
ehebert: Could type these as `bool`.

(`from cpython import `bool`)

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+        -------
+        boolean or Series of booleans, indexed by asset.
+        """
+        dt = self.simulation_dt_func()
+        data_portal = self.data_portal
+
+        if isinstance(assets, Asset):
+            return self._can_trade_for_asset(assets, dt, data_portal)
+        else:
+            return pd.Series(data={
```
ehebert: Have we done some profiling of whether a dict versus Series would be more performant here?

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+                asset: self._can_trade_for_asset(asset, dt, data_portal)
+                for asset in assets
+            })
+
+    cdef _can_trade_for_asset(self, asset, dt, data_portal):
+        if asset.start_date > dt:
```
ehebert: I'd recommend evaluating to the positive first.

```
if asset.start_date <= dt <= self.asset.end_date:
    return data_portal.get_spot_value(asset, 'volume', dt, self._data_frequency) != 0
else:
    return False
```

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+            return False
+
+        if asset.end_date < dt:
+            return False
+
+        # is there a last price?
+        return not np.isnan(
+            data_portal.get_spot_value(asset, "price", dt, self.data_frequency)
```
ehebert: Would recommend checking volume here since a comp against 0 is cheaper than `nan`.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+        if isinstance(assets, Asset):
+            return self._is_stale_for_asset(assets, dt, data_portal)
+        else:
+            return pd.Series(data={
+                asset: self._is_stale_for_asset(asset, dt, data_portal)
+                for asset in assets
+            })
+
+    cdef _is_stale_for_asset(self, asset, dt, data_portal):
+        if asset.start_date > dt:
```
ehebert: Can you put a docstring here about the semantics of `stale`? `stale` means no trade for  data for the current bar, correct?

Do we need the `asset.start_date` check here? Returning `False` when there is no data seems counter intuitive, and this function is only needed if and only if the stock is 'active' correct?

That is, we are incurring extra dt comps that may be unneeded.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+                asset_list,
+                self.simulation_dt_func(),
+                bar_count,
+                frequency,
+                fields
+            )
+
+            if single_asset:
+                # single asset, single field, return a series.
+                return df[assets]
```
ehebert: Possible TODO: Make `get_history_window` return just the one asset if called with only one value.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+                    field: self.data_portal.get_history_window(
+                        [assets],
+                        self.simulation_dt_func(),
+                        bar_count,
+                        frequency,
+                        field
+                    )[assets] for field in fields
+                })
+            else:
+                df_dict = {
```
ehebert: Have not profiled, but I the extra DataFrame creation per sid at a large number will be expensive. We should look at either having `get_history_window` return a `Panel` or 3-dim ndarray.

zipline/algorithm.py
```diff
@@ -786,8 +786,18 @@ def _calculate_order_value_amount(self, asset, value):
...
+        if self.datetime >= asset.end_date:
+            raise CannotOrderDelistedAsset(
+                msg="Cannot order {0}, as it stopped trading on {1}.".format(
+                    asset.symbol, asset.end_date
+                )
+            )
         last_price = self.trading_client.current_data[asset].price
 
+        if np.isnan(last_price):
+            raise
```
ehebert: Should name an exception here.

zipline/data/data_portal.py
```diff
@@ -429,24 +429,25 @@ def _get_minute_spot_value_future(self, asset, column, dt):
...
-            asset.sid,
-            dt,
-            column
+            asset.sid, dt, column
         )
 
-        if result > 0 or column == "volume":
-            return result
+        if result == 0 or np.isnan(result):
+            if column == "volume":
```
ehebert: I'd recommend restructuring so that volume and OHLC handling are not comingled.
With the implementation above when `volume` is non-zero `np.isnan(result)` is also called.

That extra call would be removed with:

```
if column == "volume":
     if result == 0:
        return 0
elif np.isnan(result) and not ffill:
       return  np.nan
```

zipline/data/data_portal.py
```diff
@@ -717,27 +742,28 @@ def get_history_window(self, assets, end_dt, bar_count, frequency, field,
...
-        assets = np.array([
-            (self.env.asset_finder.retrieve_asset(asset) if
-             isinstance(asset, int) else asset) for asset in assets])
+        if field not in OHLCVP_FIELDS:
+            raise ValueError("Invalid field: {0}").format(field)
 
         if frequency == "1d":
-            df = self._get_history_daily_window(assets, end_dt, bar_count,
-                                                field_to_use)
+            if field == "price":
```
ehebert: To remove some duplication at the beginning of the function could do:
```
if field == 'price':
    field = 'close'
```

zipline/data/data_portal.py
```diff
@@ -1017,7 +1050,8 @@ def _get_daily_window_for_sid(self, asset, field, days_in_window,
         else:
             return_array = np.zeros((bar_count,))
 
-        return_array[:] = np.NAN
+        if field != "volume":
+            return_array[:] = np.NAN
```
ehebert: We do nothing with volume because the array is already zeros, correct? Maybe a comment to that effect?

zipline/data/us_equity_pricing.py
```diff
@@ -528,6 +528,16 @@ def _spot_col(self, colname):
     def get_last_traded_dt(self, asset, day):
         volumes = self._spot_col('volume')
         search_day = day
+
+        if search_day >= asset.end_date:
+            # move to the last day before the asset end date
+            try:
+                search_day = self._calendar[
+                    self._calendar.get_loc(asset.end_date) - 1
```
ehebert: I believe this can also be accomplished by `asset.end_date - self._calendar.freq`.
(Though not 100% if this calendar object has a `freq`).

zipline/data/us_equity_pricing.py
```diff
@@ -528,6 +528,16 @@ def _spot_col(self, colname):
     def get_last_traded_dt(self, asset, day):
         volumes = self._spot_col('volume')
         search_day = day
+
+        if search_day >= asset.end_date:
+            # move to the last day before the asset end date
+            try:
```
ehebert: Would `searchsorted(asset.end_date, 'left')` encapsulate this entire logic?

zipline/sources/benchmark_source.py
```diff
@@ -35,16 +35,38 @@ def __init__(self, benchmark_sid, env, trading_days, data_portal,
...
-                self.data_portal
-            )
+            self._precalculated_series = \
+                self._initialize_precalculated_series(
+                    self.benchmark_asset,
+                    self.env,
+                    self.trading_days,
+                    self.data_portal
+                )
+        else:
```
ehebert: Why was this change needed with the new API?

zipline/utils/test_utils.py
```diff
@@ -678,6 +678,58 @@ def write_bcolz_minute_data(env, days, path, df_dict):
...
 
+def write_minute_data_for_asset(env, writer, start_dt, end_dt, sid,
+                                interval=1, start_val=1):
+
+    asset_minutes = env.minutes_for_days_in_range(start_dt, end_dt)
+    minutes_count = len(asset_minutes)
+    minutes_arr = np.array(range(start_val, start_val + minutes_count))
+
+    df = pd.DataFrame({
+        "open": minutes_arr + 1,
```
ehebert: Same comment earlier about using `0.1, 0.2` instead.

zipline/utils/test_utils.py
```diff
@@ -678,6 +678,58 @@ def write_bcolz_minute_data(env, days, path, df_dict):
...
+        "high": days_arr + 2,
+        "low": days_arr - 1,
+        "close": days_arr,
+        "volume": days_arr * 100,
+        "day": [day.value for day in days]
+    })
+
+    if interval > 1:
+        # only keep every 'interval' rows
+        for idx, minute in enumerate(days_arr):
```
ehebert: I think this should be `idx, daily`.
Also, can do this all at once like this:

```
q, r = divmod(len(days_arr), interval)
where_zero = np.tile([False] + [True] * (interval - 1), q + (1 if r else 0))
where_zero = where_zero[:len(days_arr)]
df[where] = 0
```

zipline/utils/test_utils.py
```diff
@@ -678,6 +678,58 @@ def write_bcolz_minute_data(env, days, path, df_dict):
...
+    df = pd.DataFrame({
+        "open": minutes_arr + 1,
+        "high": minutes_arr + 2,
+        "low": minutes_arr - 1,
+        "close": minutes_arr,
+        "volume": 100 * minutes_arr,
+        "dt": asset_minutes
+    }).set_index("dt")
+
+    if interval > 1:
```
ehebert: See below for a technique to do the indexing only once.

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+        fields: string or iterable of string.  Valid values are "open", "high",
+            "low", "close", "volume", "price", and "last_traded".
+
+        bar_count: integer number of bars of trade data
+
+        frequency: string. "1m" for minutely data or "1d" for daily date
+
+        Returns
+        -------
+        series or DataFrame or Panel, depending on the dimensionality of
```
ehebert: `Series`

zipline/_protocol.pyx
```diff
@@ -76,6 +79,289 @@ cdef class BarData:
...
+
+            If multiple assets and single field are passed in, the returned
+            DataFrame is indexed by dt, and has assets as columns.
+
+            If a single asset and multiple fields are passed in, the returned
+            DataFrame is indexed by dt, and has fields as columns.
+
+            If multiple assets and multiple fields are passed in, the returned
+            Panel is indexed by field, has dt as the major axis, and assets
+            as the minor axis.
```
ehebert: Does this table summarize this correctly?


|                  | scalar Asset                         | iterable Assets                                         |
|----------------|---------------------------------------|--------------------------------------------------------|
| scalar field | scalar (float or int)                   | `DataFrame(index=dts, columns=assets)                      |
| iterable fields | `DataFrame(index=dts, columns=fields)` | `Panel(index=fields, major_axis=dt, minor_axis=assets) `    |

zipline/data/us_equity_pricing.py
ssanderson: Do we want the last entry if `get_loc` raises a KeyError (which would happen if the search day isn't in the calendar)?  Wondering if the try/except block is too broad.

zipline/finance/blotter.py
```diff
@@ -227,10 +232,9 @@ def get_transactions(self, bar_data):
         closed_orders = []
         transactions = []
 
-        for asset, asset_orders in iteritems(self.open_orders):
-            trade_bar = bar_data[asset]
-
-            for order, txn in self.slippage_func(trade_bar, asset_orders):
+        for sid, asset_orders in iteritems(self.open_orders):
+            asset = self.asset_finder.retrieve_asset(sid)
```
ssanderson: This is making a SQL Query per order.  You probably want `retrieve_all` to do this in one query.

zipline/_protocol.pyx
ehebert: @ssanderson do you have an opinion on the appropriate values for index and columns, and what would be more natural usage in an algorithm?

I think we should change the order to:

|                  | scalar Asset                         | iterable Assets                                         |
|----------------|---------------------------------------|--------------------------------------------------------|
| scalar field | scalar (float or int)                   | `DataFrame(index=assets, columns=dts)                      |
| iterable fields | `DataFrame(index=fields, columns=dts)` | `Panel(index=assets, major_axis=fields, minor_axis=dts) `    |

For a usage pattern of:
```
window = data.history(symbol['FOO'], ['open', 'close'], 90, '1d')
a = window['close']

window = data.history([symbol['FOO', 'BAR'], 'close', 90, '1d')
b = window['FOO']

window = data.history([symbol['FOO', 'BAR'], ['open', 'close'], 90, '1d')
c = window['FOO', 'close']

a == b == c
```

I can be convinced that `window['close', 'FOO']` is a more intuitive ordering, but recommended doing `[asset, field]` to preserver the ordering of the parameters to the history method.

zipline/finance/slippage.py
```diff
@@ -49,27 +49,29 @@ def volume_for_bar(self):
         return self._volume_for_bar
 
     @abc.abstractproperty
-    def process_order(self, trade_bar, order):
+    def process_order(self, bar_data, price, volume, order):
```
ssanderson: Unrelated to this PR, but it looks like this whole file is indented one space?  That's not the case on master.

zipline/_protocol.pyx
ssanderson: @ehebert I would transpose most of your suggestions.

The almost-universal convention for DataFrames is to use dt as the index, because the expectation is that you're generally accessing time-series values sequentially, and pandas uses columnar storage.  The reprs of dates are also much longer than most column names, which makes them inconvenient to display as column labels.  This is what I'd expect for the labels here:

|                  | scalar Asset                         | iterable Assets                                         |
|----------------|---------------------------------------|--------------------------------------------------------|
| scalar field | scalar (float or int)                   | `DataFrame(index=dts, columns=assets)                      |
| iterable fields | `DataFrame(columns=fields, index=dts)` | `Panel(items=fields, major_axis=dts, minor_axis=assets) `    |

The reason you want `fields` as the first index of a panel is that you guarantee that slices along the first axis yield blocks with homogenous storage. (All entries for a given field are guaranteed to have the same dtype.)  You then want `dt, asset` as major and minor axes so that slices by field yield `DataFrame`s with dates as the index.

ehebert: @ssanderson you've convinced, we should be mindful of how the contiguous and have that take priority on the API.

zipline/_protocol.pyx
jbredeche: `data_portal.get_spot_value` doesn't return an array, it returns scalar

zipline/_protocol.pyx
jbredeche: good idea, done

zipline/_protocol.pyx
jbredeche: I'm open to it, but not optimizing for the multi-asset, multi-field case yet.  can do that later.

zipline/data/data_portal.py
jbredeche: yeah.  the new API uses "open" and "close".

BarData will do the translation from `open_price` to `open` and `close_price` to `close` to support the old API.

tests/test_algorithm.py
jbredeche: because the mock data is written up to, and including, the last day of the trading range.  the asset's end date is the first day for which we don't have data.

the change to `price` to return `NaN` after the asset end date exposed this mismatch.

tests/test_bar_data.py
jbredeche: good idea, done

tests/test_blotter.py
jbredeche: the price/NaN issue.  we need the asset to live to the day after the last trading day, so that we can get prices on the last trading day.

zipline/_protocol.pxd
jbredeche: thanks, done

zipline/_protocol.pyx
jbredeche: no.  can do it later.

zipline/_protocol.pyx
jbredeche: done, but your logic there isn't right (we need to check if there is a previous price, not if volume != 0)

zipline/_protocol.pyx
jbredeche: see note above - checking volume wouldn't be enough.

zipline/_protocol.pyx
jbredeche: the docstring is up in the main user-facing method, `is_stale`.

zipline/_protocol.pyx
jbredeche: agreed, can do later.

zipline/algorithm.py
jbredeche: oops, good catch.

zipline/data/data_portal.py
jbredeche: done

zipline/data/data_portal.py
jbredeche: that would break the logic later in the method that decides whether to ffill

zipline/data/data_portal.py
jbredeche: yes.  will comment.

zipline/finance/slippage.py
```diff
@@ -145,8 +146,8 @@ def process_order(self, trade_bar, order):
                 return
 
         return create_transaction(
-            trade_bar,
             order,
+            bar_data.current_dt,
```
ssanderson: It looks like this is the only place where this needs the `bar_data` object now.  Could we just pass the dt so that this function just operates on plain old data?

zipline/finance/slippage.py
```diff
@@ -181,11 +182,11 @@ def __init__(self, spread=0.0):
         """
         self.spread = spread
 
-    def process_order(self, trade_bar, order):
+    def process_order(self, bar_data, price, volume, order):
         return create_transaction(
-            trade_bar,
             order,
-            trade_bar.price + (self.spread / 2.0 * order.direction),
+            bar_data.current_dt,
```
ssanderson: Same note as above.

zipline/utils/test_utils.py
```diff
@@ -678,6 +678,58 @@ def write_bcolz_minute_data(env, days, path, df_dict):
...
+    days_count = len(days)
+    days_arr = np.array(range(2, days_count + 2))
+
+    df = pd.DataFrame({
+        "open": days_arr + 1,
+        "high": days_arr + 2,
+        "low": days_arr - 1,
+        "close": days_arr,
+        "volume": days_arr * 100,
+        "day": [day.value for day in days]
```
ssanderson: I think this just wants to be `days.values`?

zipline/_protocol.pyx
```diff
@@ -76,6 +81,289 @@ cdef class BarData:
...
+            "last_traded", "open", "high", "low", "close", "volume"
+
+        Returns
+        -------
+        Float, pandas Series, or pandas DataFrame.  See notes
+        below.
+
+        Notes
+        -----
+        If a single asset and a single field are passed in, a scalar float
```
ssanderson: This might be a case where putting a table in here like the one @ehebert commented with would be the most helpful way to describe this behavior.  (Sphinx can also auto-generate nice HTML renderings for tables embedded in docstrings.)

zipline/_protocol.pyx
```diff
@@ -76,6 +81,289 @@ cdef class BarData:
...
+        If a list of assets and a single field are passed in, a pandas Series
+        is returned whose indices are the assets, and whose values are scalar
+        values for each asset for the given field.
+
+        If a list of assets and a list of fields are passed in, a pandas
+        DataFrame is returned, indexed by asset.  The columns are the requested
+        fields, filled with the scalar values for each asset for each field.
+
+        "price" returns the last known close price of the asset.  If there is
+        no last known value, NaN is returned.  If a value is found, and we
```
ssanderson: Might be worth clarifying that `NaN` is only returned if we've never seen a trade.

zipline/_protocol.pyx
```diff
@@ -76,6 +81,289 @@ cdef class BarData:
...
+        had to cross an adjustment boundary (split, dividend, etc) to get it,
```
ssanderson: :+1: for clarifying the edge case here.

zipline/_protocol.pyx
```diff
@@ -76,6 +81,289 @@ cdef class BarData:
...
+        the value is adjusted before being returned.
+
+        "last_traded" returns the date of the last trade event of the asset,
+        even if the asset has stopped trading. If there is no last known value,
+        NaT is returned.
```
ssanderson: Is this numpy.NaT or pd.NaT?  (They actually have pretty different behavior.)

zipline/_protocol.pyx
```diff
@@ -76,6 +81,289 @@ cdef class BarData:
             self.data_frequency
         )
 
+    def spot_value(self, assets, fields):
+        """
+        Returns the spot value of the given assets for the given fields
+        at the current simulation time.  Spot values are the as-traded price
```
ssanderson: Based on the comment below on `price`, it seems like this isn't quite true anymore?

zipline/_protocol.pyx
```diff
@@ -76,6 +81,289 @@ cdef class BarData:
...
+        Parameters
+        ----------
+        assets : Asset or iterable of Assets
+
+        fields : string or iterable of strings.  Valid values are: "price",
+            "last_traded", "open", "high", "low", "close", "volume"
+
+        Returns
+        -------
+        Float, pandas Series, or pandas DataFrame.  See notes
```
ssanderson: This is an Int if we request a scalar volume, right?

zipline/data/us_equity_pricing.py
jbredeche: I rewrote that chunk of code, agreed with @ssanderson that the try/except was too broad.  Now, we simply go back to the last trading day before the asset end date, if needed.

```
        if day >= asset.end_date:
            # go back to one day before the asset ended
            search_day = self._calendar[
                self._calendar.get_loc(asset.end_date) - 1
            ]
        else:
            search_day = day
```

zipline/sources/benchmark_source.py
jbredeche: we were passing an `int` into `data_portal.get_history_window`, and I wanted to avoid that.  Also cleaned up a bit of the initialization logic.

zipline/utils/test_utils.py
jbredeche: I think my version is more readable, and it's fast enough

zipline/_protocol.pyx
jbredeche: fixed

zipline/finance/blotter.py
jbredeche: good catch, fixed

zipline/finance/slippage.py
jbredeche: I don't see that here.

zipline/finance/slippage.py
jbredeche: we could, but I think it adds a lot of power for the user to have access to `bar_data` when building a custom slippage function.  They can do `history` calls, check volatility, etc.

tests/test_bar_data.py
jbredeche: thanks, done

zipline/_protocol.pyx
jbredeche: good idea, done

zipline/utils/test_utils.py
jbredeche: agreed, fixed

zipline/_protocol.pyx
jbredeche: I'm clarifying, but `NaN` is returned if we've never seen a trade, OR if the asset is dead.

zipline/_protocol.pyx
jbredeche: `pd.NaT`

zipline/utils/test_utils.py
jbredeche: that change triggered some warnings:
```
A value is trying to be set on a copy of a slice from a DataFrame

See the the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  self._setitem_with_indexer(indexer, value)

```

leaving it alone for now.

zipline/_protocol.pyx
jbredeche: yes, will clarify

zipline/_protocol.pyx
jbredeche: yeah.  but the vast majority of the time, spot values aren't adjusted, and I want to reinforce that to users.  I'll fiddle with the wording.

Issue #999, opened 2016-02-16
Travis builds failing when build cache enabled


jikamens:
Failures visible in https://travis-ci.org/quantopian/zipline/builds/109687842

Build cache disabled in https://github.com/quantopian/zipline/commit/d8bbaec5a8f2cdcb1c938aa3f3885adff3f3cddd until we figure out how to fix this.

jikamens:
Failures are intermittent.

ssanderson:
From slack discussion:
```
ssanderson [15:27] 
from reading https://docs.travis-ci.com/user/caching/#How-does-the-caching-work%3F, it looks like travis just tars up any listed directories and sticks them in s3

[15:28] 
so I imagine that depending on the order that the builds in matrix run, we'd get stale versions of dependencies that would **appear** to be correct just from looking at the filesystem

[15:29] 
I think the long term fix if we want to allow caching is to supply separate cache directories for each entry in the build matrix

[15:29] 
and explicitly build packages into those directories

[15:29] 
and then install from those locations

rich [15:29] 
that makes sense to me

ssanderson [15:29] 
we only had one entry in the build matrix per python version until pretty recently, right?

rich [15:30] 
correct

[15:30] 
now we have multiple numpy's

[15:30] 
(2)

ssanderson [15:30] 
ok, I imagine that's the culprit then
```

richafrank:
Re-enabled by #1002.  Can re-open if there are problems again.

PR #1002, opened 2016-02-17
BLD: Use separate travis pip cache subdirs per numpy version

richafrank:
Round 1: np 1.9 and np 1.10 jobs will all run with empty caches, each will populate a subdirectory and update the travis cache, last one wins - let's say 1.9.

Round 2: All jobs will run, but the cache will only have 1.9, which will speed up only those jobs. 1.10 will be slower, but will now add its data to the cache and become the last to upload it, now with both 1.9 and 1.10.

Round 3: All jobs can use the cache for a performance gain during installation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,16 @@ matrix:
   fast_finish: true
   include:
   - python: 2.7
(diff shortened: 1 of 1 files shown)


richafrank:
For #999

PR #1006, opened 2016-02-23
PERF: Batch and cache '1d' history calls.

ehebert:
Instead of fetching the '1d' portion which holds data for previous days
390 times a day, cache the last array retrieved for each array.

Also, call `load_raw_array` with all assets at once instead of one at a
time, since the daily bcolz file is tuned for reading many equities at
once.

A further optimization on top of this would be to prefetch data from the
daily bcolz file.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -25,6 +25,7 @@ from six.moves import reduce
 
 from zipline.assets import Asset, Future, Equity
 from zipline.data.us_equity_pricing import NoDataOnDate
+from zipline.pipeline.data.equity_pricing import USEquityPricing
 
 from zipline.utils import tradingcalendar
 from zipline.utils.memoize import remember_last
@@ -106,6 +107,26 @@ class DataPortal(object):
         elif self._equity_minute_reader is not None:
             self._first_trading_day = \
                 self._equity_minute_reader.first_trading_day
+
+        # The `equity_daily_reader_array` lookups provide lru cache of 1 for
+        # daily history reads from the daily_reader.
+        # `last_remembered` or lru_cache can not be used, because the inputs
+        # to the function are not hashable types, and the order of the assets
+        # iterable needs to be preserved.
+        # The function that implements the cache will insert a value for the
+        # given field of (frozenset(assets), start_dt, end_dt).
+        #
+        # This is  optimized for algorithms that call history once per field
+        # in handle_data or a scheduled function.
+        self._equity_daily_reader_array_keys = {
+            'open': None,
+            'high': None,
+            'low': None,
+            'close': None,
+            'volume': None,
+            'price': None
+        }
+        self._equity_daily_reader_array_data = {}
 
     def handle_extra_source(self, source_df, sim_params):
         """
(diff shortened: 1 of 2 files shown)


ehebert:
On my local development machine, this patch takes a minute mode algorithm which does 3 bars of '1d' history for 6 equities from 1.4 seconds per day to 0.5 seconds per day.

zipline/data/data_portal.py
```diff
@@ -107,6 +108,16 @@ def __init__(self,
...
                 self._equity_minute_reader.first_trading_day
 
+        self._equity_daily_reader_array_keys = {
+            'open': None,
+            'high': None,
+            'low': None,
+            'close': None,
+            'volume': None,
+            'price': None
+            }
```
jbredeche: tabbing

zipline/data/data_portal.py
```diff
@@ -527,20 +538,26 @@ def _get_history_daily_window(self, assets, end_dt, bar_count,
...
             else:
-                data.append(self._get_history_daily_window_equity(
-                    asset, days_for_window, end_dt, field_to_use
-                ))
-
+                eq_assets.append(asset)
+        eq_data = self._get_history_daily_window_equities(
+            eq_assets, days_for_window, end_dt, field_to_use
+        )
+        if future_data:
```
jbredeche: Not really part of your PR, but this made me realize we need to do a check to make sure all the assets passed to a `history` call are of the same type (this will become required in 24/5 futures backtesting).  Right now, if a future and an equity are passed in here, we'll only return the futures information.

zipline/data/data_portal.py
```diff
@@ -606,20 +623,17 @@ def _get_history_daily_window_future(self, asset, days_for_window,
...
-    def _get_history_daily_window_equity(self, asset, days_for_window,
-                                         end_dt, field_to_use):
+    def _get_history_daily_window_equities(
+            self, assets, days_for_window, end_dt, field_to_use):
         ends_at_midnight = end_dt.hour == 0 and end_dt.minute == 0
 
-        # get the start and end dates for this sid
-        end_date = self._get_asset_end_date(asset)
-
-        if ends_at_midnight or (days_for_window[-1] > end_date):
```
jbredeche: what are the implications of getting rid of the `end_date` check?

zipline/data/data_portal.py
```diff
@@ -843,8 +857,10 @@ def _get_minute_window_for_asset(self, asset, field, minutes_for_window):
             return self._get_minute_window_for_future(asset, field,
                                                       minutes_for_window)
         else:
-            return self._get_minute_window_for_equity(asset, field,
-                                                      minutes_for_window)
+            # TODO: Make caller accept assets.
```
jbredeche: by caller, do you mean `_get_minute_window_for_equities`?

zipline/data/data_portal.py
```diff
@@ -1020,8 +1042,33 @@ def _apply_all_adjustments(self, data, asset, dts, field,
 
         np.around(data, 3, out=data)
 
-    def _get_daily_window_for_sid(self, asset, field, days_in_window,
-                                  extra_slot=True):
+    def _equity_daily_reader_arrays(self, field, dts, assets):
+        # Temporary cache shim before loader is pulled in.
+        # Custom memoization, because of unhashable types.
```
jbredeche: can you clarify these comments?  not sure what they mean

zipline/data/data_portal.py
```diff
@@ -107,6 +108,16 @@ def __init__(self,
...
 
+        self._equity_daily_reader_array_keys = {
+            'open': None,
+            'high': None,
+            'low': None,
+            'close': None,
+            'volume': None,
+            'price': None
+            }
+        self._equity_daily_reader_array_data = {}
```
jbredeche: can you add a comment explaining what these two hashes are for?

zipline/data/data_portal.py
ehebert: Will fix.

zipline/data/data_portal.py
ehebert: Yes.

zipline/data/data_portal.py
jbredeche: I'll add an issue for this.

ehebert: We concat the data, so both would be returned; though this behavior is not covered by testing yet. (Figured out this out when the implementation was not attempting to concatenate at all.)

zipline/data/data_portal.py
ehebert: So far have not run into implications, the `load_raw_array` handles the case by providing `nans` before and after the window already.

zipline/data/data_portal.py
ehebert: No, I mean that `_get_history_minute_window` can and should pass an iterable of assets to `_get_minute_window_for_asset(s)` instead of calling multiple times.

With that change the start and end locations would only need to be calculated once per bar, the multiplication for the thousands place can be done only once, etc.

PR #1007, opened 2016-02-23
PERF: If the window is one day, no adjustments.

ehebert:
Noticed in the case where the window function is used to fetch the
minutes for the current day to update the last slot in a '1d' history
window.

Further improvement can be made on this by prefetching the day's data
into a Float64Window instead and then iteratively getting the latest
value.

On local machine this cuts the time per day for an algo which does 3
bars of '1d' history for 6 algos by about 33%. (0.5 seconds -> 0.35 seconds)

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -916,6 +916,8 @@ class DataPortal(object):
             return np.full((len(minutes_for_window), len(assets), np.nan))
 
         num_minutes = len(minutes_for_window)
+        start_date = normalize_date(minutes_for_window[0])
+        end_date = normalize_date(minutes_for_window[-1])
 
         return_data = np.zeros((len(minutes_for_window), len(assets)),
                                dtype=np.float64)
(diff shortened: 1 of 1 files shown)


zipline/data/data_portal.py
```diff
@@ -916,6 +916,8 @@ def _get_minute_window_for_equities(
             return np.full((len(minutes_for_window), len(assets), np.nan))
 
         num_minutes = len(minutes_for_window)
+        start_date = normalize_date(minutes_for_window[0])
```
jbredeche: Two thoughts

1) We could have the tradesim loop set the current_day on the dataportal, like we used to, to avoid this call to `normalize_date`, but that probably isn't worth it.
2) Anytime we use daily history, we know that it's going to call this method with the minute from the last day - is it worth passing in an extra param to signal that we know it's all on the same day, so no need to `normalize_date` twice and then do a date comparison?

ehebert: I heartily agree with 2). up to even considering a separate function entirely.

ehebert: i.e. we would benefit from a function where the interface is `intraday_value_up_to_dt(assets, field, dt)`, which can use a cached value of `intraday_value_up_to_dt(assets, field, dt - 1)` to answer the question of 'what is the lowest low to this point' etc.

e.g.

```
def intraday_value_up_to_dt(self, asset, field, dt):
    try:
        prev = self.intraday_cache[(asset, field, dt - 1)]
        value_at_dt = # read from prefetched window
        if field == 'low': 
            value = min(value_at_dt, prev)
    except KeyError:
        # get values to dt, except for open and close, which just need the first and last
        if field == 'low':
           value = values_to_dt.min()
    self._intraday_cache[(asset, field, dt)] = value
    return value   
```

PR #1009, opened 2016-02-24
Deprecate batch transform

abhijeetkalyan:
Deprecating for now, before we rewrite the sample pairtrade algo to use history and then kill batch altogether.

### DIFF
--- a/zipline/transforms/batch_transform.py
+++ b/zipline/transforms/batch_transform.py
@@ -34,6 +34,7 @@ from six import (
 
 from zipline.utils.algo_instance import get_algo_instance
 from zipline.utils.data import MutableIndexRollingPanel
+from zipline.utils.deprecate import deprecated
 from zipline.protocol import Event
 
 log = logbook.Logger('BatchTransform')
@@ -151,6 +152,8 @@ class BatchTransform(object):
 
     """
 
+    @deprecated(msg="Batch transforms are deprecated and will be removed"
+                    " in a future release. Please use 'history' instead.")
     def __init__(self,
                  func=None,
                  refresh_period=0,
@@ -161,7 +164,6 @@ class BatchTransform(object):
                  compute_only_full=True,
                  bars='daily',
                  downsample=False):
-
         """Instantiate new batch_transform object.
 
         :Arguments:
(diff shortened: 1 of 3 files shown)


zipline/transforms/batch_transform.py
```diff
@@ -486,11 +487,13 @@ def _get_field_names(self, event):
             return set.union(self.field_names, self.latest_names)
 
 
+@deprecated(msg="batch_transform is deprecated and will be removed"
```
richafrank: Should this live on `BatchTransform` instead?  Not sure if there are other pathways to it.

zipline/utils/deprecate.py
```diff
@@ -0,0 +1,46 @@
...
+    Parameters
+    ----------
+    msg : str
+        The message to display in the deprecation warning.
+    stacklevel : int
+        How far up the stack the warning needs to go, before
+        showing the relevant calling lines.
+    Usage
+    -----
+    @deprecated(msg='sample_fn', stacklevel=2)
```
richafrank: I think 'sample_fn' is probably not a message that someone would want to use. Maybe this example should use default values?

zipline/utils/deprecate.py
```diff
@@ -0,0 +1,46 @@
...
+    def sample_fn(*args, **kwargs):
+    """
+    def deprecated_dec(fn):
+        @wraps(fn)
+        def wrapper(*args, **kwargs):
+            warnings.warn(
+                msg or "Function %s is deprecated." % fn.__name__,
```
richafrank: I *think* this decorator is generic enough that we could use it with classes too, so maybe we don't need to specify "Function " in the default message.

zipline/utils/deprecate.py
abhijeetkalyan: Totally agree - this was previously `name='sample_fn'`, when it made a little more sense. I'll provide a better message

zipline/utils/deprecate.py
llllllllll: If we used this as a class decorator then the name bound in the scope will not be the class object but the wrapper function. This will make py2 style super and isinstance fail. We could check if this is a class and transform the `__new__` instead.

richafrank: Maybe not worth the complexity of supporting that then?

abhijeetkalyan: Feels like it's out of the scope of this PR, not sure if it's worth adding to the utility going forward. I guess we could add it when we need it

llllllllll: Sure, just saying that this will not work with classes seamlessly

PR #1010, opened 2016-02-24
benchmark sources

llllllllll:
Related: https://groups.google.com/forum/#!topic/zipline/pmE4-SMKg20

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -524,34 +524,43 @@ class TradingAlgorithm(object):
     # TODO: make a new subclass, e.g. BatchAlgorithm, and move
     # the run method to the subclass, and refactor to put the
     # generator creation logic into get_generator.
-    def run(self, source, overwrite_sim_params=True,
+    def run(self,
+            source,
+            overwrite_sim_params=True,
             benchmark_return_source=None):
         """Run the algorithm.
 
-        :Arguments:
-            source : can be either:
-                     - pandas.DataFrame
-                     - zipline source
-                     - list of sources
+        Parameters
+        ----------
+        source : DataSource-like or list[DataSource]
+            The source or sources to use as the input data for the algorithm
+            and benchmark.
+        overwrite_sim_params : bool, optional
+            Update the internal sim_params based on the source data.
+        benchmark_return_source : DataSource, optional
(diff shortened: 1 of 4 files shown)


zipline/algorithm.py
```diff
@@ -138,44 +138,74 @@ def handle_data(context, data):
...
+        algo_filename : str, optional
+            The filename for the algoscript. This will be used in exception
+            tracebacks. default: '<string>'.
+        data_frequency : {'daily', 'minute'}, optional
+            The duration of the bars.
+        capital_base : float, optional
+            How much capital to start with. default: 1.0e5
+        instant_fill : bool, optional
+            Whether to fill orders immediately or on next bar. default: False
+        asset_finder : AssetFinder, optional
```
ssanderson: I don't think you can actually pass an AssetFinder right now:

https://github.com/quantopian/zipline/pull/1010/files#diff-710b8bff6f1b5d930511d643b73c6987R269

llllllllll: Thanks, I didn't audit the existing entries. I will make sure there are no other invalid params there.

llllllllll:
I have a simple case where I am passing a minutely benchmark and getting a `benchmark_period_(return|volitility)` of all `NaN`. This seems incorrect to me:

```python
import numpy as np
import pandas as pd

from zipline.algorithm import TradingAlgorithm
from zipline.api import get_datetime
from zipline.sources import BenchmarkSource
from zipline.finance.trading import TradingEnvironment

dates = TradingEnvironment().minutes_for_days_in_range(
    '2014-01-06',
    '2014-01-10',
)
data = pd.Panel({
    0: pd.DataFrame(
        {'price': np.arange(len(dates)), 'volume': np.arange(len(dates))},
        index=dates,
    ),
})
bench = BenchmarkSource(pd.Series(np.arange(len(dates)), index=dates))


def initialize(context):
    pass


    print(get_datetime())


results = TradingAlgorithm(
    start=dates[0],
    end=dates[-1],
    data_frequency='minute',
    emission_rate='minute',
    initialize=initialize,
    handle_data=handle_data,
).run(data, benchmark_return_source=bench)
```

llllllllll:
we have changed the source api

PR #1012, opened 2016-02-25
DEV: allow partial uploads in conda_build_matrix

llllllllll:
I cannot get the builds to run in parallel without locking up. We can invest more time if this becomes an annoyance

### DIFF
--- a/etc/conda_build_matrix.py
+++ b/etc/conda_build_matrix.py
@@ -43,11 +43,18 @@ def mkargs(py_version, npy_version, output=False):
     help='Upload the last built packages without rebuilding.',
 )
 @click.option(
+    '--allow-partial-uploads',
+    is_flag=True,
+    default=False,
+    help='Upload any packages that were built even if some of the builds'
+    ' failed.',
+)
+@click.option(
     '--user',
     default='quantopian',
     help='The anaconda account to upload to.',
 )
-def main(upload, upload_only, user):
+def main(upload, upload_only, allow_partial_uploads, user):
     if upload_only:
         # if you are only uploading you shouldn't need to specify both flags
         upload = True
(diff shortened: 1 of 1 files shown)


richafrank:
> I cannot get the builds to run in parallel without locking up. 

Since this is dev-only, maybe it would be easy to get it working on a single python version.

etc/conda_build_matrix.py
```diff
@@ -74,11 +80,21 @@ def main(upload, upload_only, user):
...
+            out, err = p.communicate()
+            if p.returncode:
+                # only change the status to failure if we are not allowing
+                # partial uploads
+                status |= not allow_partial_uploads
+                print(
+                    'failed to get the output name for python=%s numpy=%s\n'
+                    '%s' % (py_version, npy_version, err.decode('utf-8')),
+                )
+            files.append(out.decode('utf-8').strip())
```
richafrank: Should this be inside the `p.returncode` conditional?

etc/conda_build_matrix.py
```diff
@@ -74,11 +80,21 @@ def main(upload, upload_only, user):
                     err.decode('utf-8'),
                 ))
         elif upload:
-            files.append(subprocess.Popen(
+            p = subprocess.Popen(
                 **mkargs(py_version, npy_version, output=True)
```
richafrank: How does this differ from the functionality of `proc`?  Are we `conda build`ing twice?

etc/conda_build_matrix.py
llllllllll: yup

etc/conda_build_matrix.py
llllllllll: nope, this just prints the filename that the last command built

richafrank: Ahhh `--output`

etc/conda_build_matrix.py
```diff
@@ -43,11 +43,17 @@ def mkargs(py_version, npy_version, output=False):
     help='Upload the last built packages without rebuilding.',
 )
 @click.option(
+    '--allow-partial-uploads',
+    is_flag=True,
+    default=False,
+    help='Allow uploads',
```
richafrank: Could you include that this only works with `--update-only`?

etc/conda_build_matrix.py
```diff
@@ -74,11 +80,21 @@ def main(upload, upload_only, user):
...
+            out, err = p.communicate()
+            if p.returncode:
+                # only change the status to failure if we are not allowing
+                # partial uploads
+                status |= not allow_partial_uploads
+                print(
+                    'failed to get the output name for python=%s numpy=%s\n'
+                    '%s' % (py_version, npy_version, err.decode('utf-8')),
+                )
+                files.append(out.decode('utf-8').strip())
```
richafrank: Actually, this should be in an `else`, right?

llllllllll: I am struggling ;_;, thanks

etc/conda_build_matrix.py
llllllllll: this should work without upload only. I will fix this because it should be more descriptive

richafrank: Ah, I see - that's the second loop.

PR #1017, opened 2016-02-29
ENH: Add capabilities to cancel open orders at market close

lianga888:
Currently, hardcode to cancel at EOD

### DIFF
--- a/tests/finance/test_cancel_policy.py
+++ b/tests/finance/test_cancel_policy.py
@@ -1,11 +1,20 @@
-import datetime
-
-import pytz
-
+#
+# Copyright 2016 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 from unittest import TestCase
 
 from zipline.finance.cancel_policy import NeverCancel, EODCancel
-from zipline.finance.order import Order
 from zipline.gens.sim_engine import (
     BAR,
     DAY_END
@@ -15,43 +24,11 @@ from zipline.gens.sim_engine import (
 class CancelPolicyTestCase(TestCase):
 
     def test_eod_cancel(self):
-
-        order = Order(
-            dt=datetime.datetime(2006, 1, 5, 14, 30, tzinfo=pytz.utc),
-            sid=133,
-            amount=1,
-            cancel_policy=EODCancel()
-        )
-
-        current_dt = datetime.datetime(2006, 1, 5, 20, 59, tzinfo=pytz.utc)
-        current_event = BAR
-
-        should_cancel = order.should_cancel(current_dt, current_event)
-        self.assertEqual(should_cancel, False)
-
-        current_dt = datetime.datetime(2006, 1, 5, 21, 0, tzinfo=pytz.utc)
-        current_event = DAY_END
-
-        should_cancel = order.should_cancel(current_dt, current_event)
-        self.assertEqual(should_cancel, True)
+        cancel_policy = EODCancel()
+        self.assertTrue(cancel_policy.should_cancel(DAY_END))
+        self.assertFalse(cancel_policy.should_cancel(BAR))
 
     def test_never_cancel(self):
-
-        order = Order(
-            dt=datetime.datetime(2006, 1, 5, 14, 30, tzinfo=pytz.utc),
-            sid=133,
-            amount=1,
-            cancel_policy=NeverCancel()
-        )
-
-        current_dt = datetime.datetime(2006, 1, 5, 20, 59, tzinfo=pytz.utc)
-        current_event = BAR
-
-        should_cancel = order.should_cancel(current_dt, current_event)
-        self.assertEqual(should_cancel, False)
-
-        current_dt = datetime.datetime(2006, 1, 5, 21, 0, tzinfo=pytz.utc)
-        current_event = DAY_END
-
-        should_cancel = order.should_cancel(current_dt, current_event)
-        self.assertEqual(should_cancel, False)
+        cancel_policy = NeverCancel()
+        self.assertFalse(cancel_policy.should_cancel(DAY_END))
+        self.assertFalse(cancel_policy.should_cancel(BAR))
--- a/tests/finance/test_cancel_policy.py
+++ b/tests/finance/test_cancel_policy.py
@@ -1,0 +1,57 @@
+import datetime
+
+import pytz
+
+from unittest import TestCase
+
+from zipline.finance.cancel_policy import NeverCancel, EODCancel
+from zipline.finance.order import Order
+from zipline.gens.sim_engine import (
+    BAR,
+    DAY_END
+)
+
+
+class CancelPolicyTestCase(TestCase):
+
+    def test_eod_cancel(self):
+
+        order = Order(
+            dt=datetime.datetime(2006, 1, 5, 14, 30, tzinfo=pytz.utc),
+            sid=133,
+            amount=1,
+            cancel_policy=EODCancel()
+        )
+
+        current_dt = datetime.datetime(2006, 1, 5, 20, 59, tzinfo=pytz.utc)
+        current_event = BAR
+
+        should_cancel = order.should_cancel(current_dt, current_event)
+        self.assertEqual(should_cancel, False)
+
+        current_dt = datetime.datetime(2006, 1, 5, 21, 0, tzinfo=pytz.utc)
+        current_event = DAY_END
+
+        should_cancel = order.should_cancel(current_dt, current_event)
+        self.assertEqual(should_cancel, True)
+
+    def test_never_cancel(self):
+
+        order = Order(
+            dt=datetime.datetime(2006, 1, 5, 14, 30, tzinfo=pytz.utc),
+            sid=133,
+            amount=1,
+            cancel_policy=NeverCancel()
+        )
+
+        current_dt = datetime.datetime(2006, 1, 5, 20, 59, tzinfo=pytz.utc)
+        current_event = BAR
+
+        should_cancel = order.should_cancel(current_dt, current_event)
+        self.assertEqual(should_cancel, False)
+
+        current_dt = datetime.datetime(2006, 1, 5, 21, 0, tzinfo=pytz.utc)
+        current_event = DAY_END
+
+        should_cancel = order.should_cancel(current_dt, current_event)
+        self.assertEqual(should_cancel, False)
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -15,6 +15,8 @@
 from collections import namedtuple
 import datetime
 from datetime import timedelta
+
+from logbook import TestHandler, WARNING
 from mock import MagicMock
 from nose_parameterized import parameterized
 from six import iteritems, itervalues
@@ -57,7 +59,7 @@ from zipline.errors import (
     SymbolNotFound,
     RootSymbolNotFound,
     UnsupportedDatetimeFormat,
-    CannotOrderDelistedAsset)
+    CannotOrderDelistedAsset, SetCancelPolicyPostInit, UnsupportedCancelPolicy)
 from zipline.test_algorithms import (
     access_account_in_init,
     access_portfolio_in_init,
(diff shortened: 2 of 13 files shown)


zipline/finance/cancel_policy.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class EODCancel(CancelPolicy):
+    def __init__(self, warn_on_cancel=True):
+        self.warn_on_cancel = warn_on_cancel
+
+    def should_cancel(self, dt, event, order):
+        if event == DAY_END:
+            if self.warn_on_cancel:
+                log_warning(order.amount, order.sid, order.filled,
```
jbredeche: Feels a bit strange to log this warning when we're simply checking if we should cancel, and not actually doing the cancelling here.  (I think it's correct that we're not doing the cancelling here).

zipline/gens/tradesimulation.py
```diff
@@ -180,6 +184,10 @@ def handle_benchmark(date):
             stack.enter_context(self.processor)
             stack.enter_context(ZiplineAPI(self.algo))
             for dt, action in self.clock:
+
+                if algo.data_frequency == 'minute':
```
jbredeche: can you move this to the `DAY_END` handler?  worried about the perf impact of checking every minute.  if we eventually have intra-day cancellation policies, we can revisit this.

tests/finance/test_cancel_policy.py
```diff
@@ -0,0 +1,60 @@
+import datetime
```
ehebert: License/Copyright.

tests/finance/test_cancel_policy.py
```diff
@@ -0,0 +1,60 @@
...
+
+class CancelPolicyTestCase(TestCase):
+
+    def cancel_order(self, order):
+        order.cancel()
+
+    def test_eod_cancel(self):
+
+        order = Order(
+            dt=datetime.datetime(2006, 1, 5, 14, 30, tzinfo=pytz.utc),
```
ehebert: I think we now more commonly use `Timestamp` for setting arbitrary dates, e.g.  `Timestamp('2006-01-05 14:31', tz='UTC')`, but not sure if there is an official policy/standard.

tests/test_algorithm.py
```diff
@@ -878,6 +879,7 @@ def test_minute_data(self, algo_class):
             )
 
             algo = algo_class(sim_params=sim_params, env=env)
+            algo.cancel_policy = NeverCancel()
```
ehebert: Should this be a parameter to Algorithm?

zipline/finance/blotter.py
```diff
@@ -77,7 +78,8 @@ def __repr__(self):
     def set_date(self, dt):
         self.current_dt = dt
 
-    def order(self, sid, amount, style, order_id=None):
+    def order(self, sid, amount, style, cancel_policy=NeverCancel(),
```
ehebert: Is part of the design that in the same simulation/algorithm, that individual orders can have different cancellation policies?

zipline/finance/cancel_policy.py
```diff
@@ -0,0 +1,70 @@
+import abc
```
ehebert: Licence/Copyright.

zipline/finance/cancel_policy.py
```diff
@@ -0,0 +1,70 @@
...
+        'Your order for %s shares of %s has been partially filled. %s '
+        'shares were successfully purchased. The remaining %s shares '
+        'are being canceled based on the policy %s.' %
+        (amount, sid, filled, amount - filled, policy_name)
+    )
+
+
+class CancelPolicy(with_metaclass(abc.ABCMeta)):
+
+    def should_cancel(self, dt, event, order):
```
ehebert: This looks like it should be an `abstractmethod`?

zipline/gens/tradesimulation.py
ehebert: Agreed.

richafrank:
My understanding is that the "DEV" prefix is for tools/changes that devs use to maintain a project, not for features in the project itself.  What about "ENH"?

zipline/finance/blotter.py
richafrank: When is this default used?

zipline/finance/order.py
```diff
@@ -63,6 +67,7 @@ def __init__(self, dt, sid, amount, stop=None, limit=None, filled=0,
         self.limit_reached = False
         self.direction = math.copysign(1, self.amount)
         self.type = zp.DATASOURCE_TYPE.ORDER
+        self.cancel_policy = cancel_policy if cancel_policy else NeverCancel()
```
richafrank: When is this default used?

zipline/finance/cancel_policy.py
```diff
@@ -0,0 +1,70 @@
...
+from zipline.gens.sim_engine import DAY_END
+from zipline.utils.serialization_utils import (
+    VERSION_LABEL
+)
+
+warning_logger = logbook.Logger('AlgoWarning')
+
+
+def log_warning(amount, sid, filled, policy_name):
+    warning_logger.warn(
```
richafrank: When do we want to use a logger for warnings versus the stdlib `warnings` module?

zipline/finance/cancel_policy.py
```diff
@@ -0,0 +1,70 @@
...
+            if self.warn_on_cancel:
+                log_warning(order.amount, order.sid, order.filled,
+                            self.__class__.__name__)
+            return True
+
+        return False
+
+
+class NeverCancel(CancelPolicy):
+    def __init__(self, warn_on_cancel=False):
```
ssanderson: Why does this take a `warn` parameter if it will never be used?

zipline/finance/cancel_policy.py
ssanderson: @richafrank one distinction is that `warnings` are easier to silence, but also only fire once by default.  I think logging here makes sense, since each trigger of this is a distinct event that shouldn't be amalgamated with other cancellation warnings.

zipline/finance/cancel_policy.py
lianga888: I can take it out, but does it make sense to have different cancel policies take different params?

zipline/finance/cancel_policy.py
```diff
@@ -0,0 +1,70 @@
...
+class CancelPolicy(with_metaclass(abc.ABCMeta)):
+
+    def should_cancel(self, dt, event, order):
+        pass
+
+    def __getstate__(self):
+
+        state_dict = \
+            {k: v for k, v in iteritems(self.__dict__)
+                if not k.startswith('_')}
```
ssanderson: Why are we skipping private members here?  How would we expect these to be repopulated if we actually had to use this code?

zipline/finance/cancel_policy.py
ssanderson: @jbredeche I recommended moving this here.  Since it's the cancel policy that owns the logic for whether or we should log, it feels reasonable to me that it should own the logic for warning as well.  The alternative, I think, would be to have another interface method (e.g. `should_warn`), which the blotter would consult before logging itself, but then you lose the ability in the future to have different warning messages for different cancellation policies, which seems like a desirable behavior.

zipline/finance/cancel_policy.py
ssanderson: Yes.  There's no guarantee or expectation that different subclasses of a type will have the same constructor signature.

Imagine if we added `CancelAtTimeIfNotFilled`.  That would presumably have to take a `time` as a parameter, but you wouldn't retroactively add that as an unused parameter to every other CancelPolicy.

zipline/finance/order.py
```diff
@@ -37,8 +41,8 @@
 
 
 class Order(object):
-    def __init__(self, dt, sid, amount, stop=None, limit=None, filled=0,
-                 commission=None, id=None):
+    def __init__(self, dt, sid, amount, cancel_policy=None, stop=None,
```
ssanderson: IMO this shouldn't be a default if the expectation is that every `Order` instance will have this attribute.  c.f. @richafrank's comment below.

zipline/algorithm.py
```diff
@@ -240,6 +241,9 @@ def __init__(self, *args, **kwargs):
                 commission=PerShare()
             )
 
+        # Cancel open orders at end of day by default
+        self.cancel_policy = EODCancel(warn_on_cancel=True)
```
ssanderson: I don't think this default belongs in Zipline.  If we want to enforce this on Quantopian, we should move it into a downstream subclass.

zipline/finance/blotter.py
ssanderson: @ehebert yes.  The long-term plan for this in my mind is that you can call `set_cancel_policy` to set the default, but you can override it on a per-order basis in `TradingAlgorithm.order`.

w/r/t @richafrank's comment here, I don't think this should be a default.  We should be explicit when we're passing a policy here, since this is only ever called from `TradingAlgorithm.order`.

In general, if we want to provide default behaviors for user-facing APIs, we should inject the default as early as possible in the call stack, and then pass values explicitly from there on out.  Adding defaults at every layer makes it much harder to change the default later, since we have to go and find all the places where we ever passed a cancel policy.  It also makes it much harder to figure out where values are actually created.

tests/test_blotter.py
```diff
@@ -112,6 +114,20 @@ def test_blotter_order_types(self, style_obj, expected_lmt, expected_stp):
...
+        blotter = Blotter('minute', self.env.asset_finder)
+
+        blotter.order(24, 100, MarketOrder(), cancel_policy=EODCancel())
+        blotter.order(25, 100, MarketOrder(), cancel_policy=EODCancel())
+
+        self.assertEqual(len(blotter.new_orders), 2)
+        for order in blotter.new_orders:
+            self.assertEqual(order.status, ORDER_STATUS.OPEN)
+
+        blotter.execute_cancel_policy(None, DAY_END)
```
ssanderson: This should also asssert that if we execute the cancel policy **not** on DAY_END, that we don't cancel the order.

tests/test_algorithm.py
ssanderson: :+1: please don't mutate objects after they've been created.

tests/finance/test_cancel_policy.py
```diff
@@ -0,0 +1,37 @@
...
+from zipline.finance.cancel_policy import NeverCancel, EODCancel
+from zipline.gens.sim_engine import (
+    BAR,
+    DAY_END
+)
+
+
+class CancelPolicyTestCase(TestCase):
+
+    def cancel_order(self, order):
```
jbredeche: what's this method for?

tests/test_blotter.py
```diff
@@ -112,6 +114,25 @@ def test_blotter_order_types(self, style_obj, expected_lmt, expected_stp):
         self.assertEqual(result.limit, expected_lmt)
         self.assertEqual(result.stop, expected_stp)
 
+    def test_blotter_eod_cancellation(self):
```
jbredeche: do you also need a test for the blotter not canceling orders if its policy is `NeverCancel`?

zipline/finance/blotter.py
```diff
@@ -136,6 +140,22 @@ def cancel(self, order_id):
             # along with newly placed orders.
             self.new_orders.append(cur_order)
 
+    def execute_cancel_policy(self, event):
+        if self.cancel_policy.should_cancel(event):
+            for orders_by_sid in itervalues(self.open_orders):
+                for order in orders_by_sid:
+                    self.cancel(order.id)
+                    if self.cancel_policy.warn_on_cancel:
```
jbredeche: `self.cancel_policy.warn_on_cancel` can't change during these loops. can you pull the check out of the loops, maybe right after the `should_cancel` check?

zipline/gens/tradesimulation.py
```diff
@@ -185,6 +189,9 @@ def handle_benchmark(date):
                 elif action == DAY_START:
                     once_a_day(dt)
                 elif action == DAY_END:
+                    if algo.data_frequency == 'minute':
```
jbredeche: the `data_frequency` isn't changing for the lifetime of this simulation, so we shouldn't pay the cost of checking this (even though it's daily and not minutely, so the perf impact isn't huge).  also, it's a bit misleading to future readers of this code.

one way to get around this is something like this, above the main loop:
```
if algo.data_frequency == "minute":
    def execute_cancelation_policy():
        algo.blotter.execute_cancel_policy(DAY_END)
else:
    def execute_cancelation_policy():
        pass
```

and then call `execute_cancelation_policy()` inside the `DAY_END` handler.

zipline/finance/cancel_policy.py
ssanderson: I don't think we need to be updating every file with this:

http://www.apache.org/dev/apply-license.html#copy-per-file

> DO I HAVE TO HAVE A COPY OF THE LICENSE IN EACH SOURCE FILE?
Only one full copy of the license is needed per distribution. See the policy.

zipline/finance/cancel_policy.py
```diff
@@ -0,0 +1,67 @@
...
+
+        self.__dict__.update(state)
+
+
+class EODCancel(CancelPolicy):
+    def __init__(self, warn_on_cancel=True):
+        self.warn_on_cancel = warn_on_cancel
+
+    def should_cancel(self, event):
+        return True if event == DAY_END else False
```
ssanderson: This is just `event == DAY_END`.

zipline/algorithm.py
```diff
@@ -237,7 +238,9 @@ def __init__(self, *args, **kwargs):
                 data_frequency=self.data_frequency,
                 asset_finder=self.asset_finder,
                 slippage_func=VolumeShareSlippage(),
-                commission=PerShare()
+                commission=PerShare(),
+                # Default to NeverCancel in zipline
+                cancel_policy=kwargs.pop('cancel_policy', NeverCancel())
```
jbredeche: what happened to the API method you had introduced in a previous version of this code?

zipline/algorithm.py
```diff
@@ -1026,6 +1029,10 @@ def set_commission(self, commission):
         self.blotter.commission = commission
 
     @api_method
+    def set_cancel_policy(self, cancel_policy):
```
ehebert: @jbredeche re: https://github.com/quantopian/zipline/pull/1017/files#r54516398 it looks like it is here.
I believe the other `cancel_policy` in kwargs setting a default.

zipline/finance/cancel_policy.py
ehebert: Fair enough, however each file should still have copyright?

zipline/finance/blotter.py
```diff
@@ -136,6 +140,23 @@ def cancel(self, order_id):
             # along with newly placed orders.
             self.new_orders.append(cur_order)
 
+    def execute_cancel_policy(self, event):
```
ehebert: Since we currently only call the cancel policy at day end, we could cut the extra parameter of `event`.

PR #1018, opened 2016-02-29
BUG: Ensure consistent ordering of amounts, prices & multipliers

StewartDouglas:
To calculate our position values / exposures, we multiply together the corresponding `amounts`, `last_sale_prices` and `multipliers`. This assumes that these three iterables are in the same order. In fact this is not necessarily the case, because while `multipliers` are stored as an `OrderedDict`, the `amounts` and `last_sale_prices` are lists which are populated by iterating over a regular `dict` (actually a subclass, called `positiondict`), and we cannot be sure of the order in which keys/values are stored within in this `dict`. In particular, the order of this `dict` may change in arbitrary ways when we call `.update()` on it, whereas the ordering of existing elements in the `OrderedDict`s will not change.

We have not previously seen this issue because in the case equities-only algorithms, all the `multipliers` are the same (== 1) and so the order in which they are stored does not matter. This *is* an issue when dealing with mixed equities and futures algorithms, since in the case of futures we would like to [set the positions values to zero](https://github.com/quantopian/zipline/blob/master/zipline/finance/performance/position_tracker.py#L153) (since, unlike equities, you can't sell your positions for cash.) It is also an issue with pure-futures algorithms, since different futures can have different multipliers leading to different exposures.

To resolve this issue, we change `positiondict` so that it subclasses `OrderedDict` rather than just `dict`. Since `self.positions` and `self_position_value_multipliers / self._position_exposure_multipliers` are populated in the same order, we ensure that values from `amounts`, `last_sale_prices` and `_position_value_multipliers` will all now be returned in the same order.

I still need to update our tests the cover the above.

### DIFF
--- a/tests/test_perf_tracking.py
+++ b/tests/test_perf_tracking.py
@@ -2552,7 +2552,9 @@ class TestPositionTracker(unittest.TestCase):
     def setUpClass(cls):
         cls.env = TradingEnvironment()
         futures_metadata = {3: {'multiplier': 1000},
-                            4: {'multiplier': 1000}}
+                            4: {'multiplier': 1000},
+                            1032201401: {'multiplier': 50},
+                            }
         cls.env.write_data(equities_identifiers=[1, 2],
                            futures_data=futures_metadata)
 
@@ -2601,7 +2603,6 @@ class TestPositionTracker(unittest.TestCase):
         pt.update_positions({1: pos1, 2: pos2, 3: pos3, 4: pos4})
 
         # Test long-only methods
-
         pos_stats = pt.stats()
         self.assertEqual(100, pos_stats.long_value)
         self.assertEqual(100 + 300000, pos_stats.long_exposure)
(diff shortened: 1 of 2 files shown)


ehebert:
Reading through `position_tracker` it looks like the values will be added to positions and multipliers in the same order, but the guarantee that elements are not added out of order is not high.

It may be worth looking again at how this data is structured, and whether we switch back to just using dictionaries without relying on the order of insertion.

StewartDouglas:
I agree that values are **added** to positions and multipliers in the same order, the problem arises when we then **read** from positions and multipliers: since positions is a `dict` the read-order won't necessarily correspond to the write order. Hence my proposal to change positions to subclass `OrderedDict`. 

@ehebert what do you think about merging this fix (after adding tests), and then refactoring the code at a later date so that it doesn't rely on the order of the data (for example, by changing `amounts` and `last_sale_prices` are `sid -> value` dicts).

PR #1019, opened 2016-02-29
PERF: Prefetch daily history windows.

ehebert:
Use AdjustedArrayWindow to prefetch daily history windows, advancing the
window forward on each call for a new end date of the window.

Locally, on an algorithm with 6 daily windows this speeds up the daily
1d history from 300ms -> 200ms per day, saving about 33%.

The main bottlenecks this addresses is:
- Re-reading the data from the daily bar bcolz files every simulation
day.
- Remove need to do all adjustments and rounding every day.

Also, changed the `load_raw_arrays` in the Panel daily bar reader to
provide nans for missing days, to match the behavior of the bcolz
loader. (Discovered when running test_examples.)

# Other Notes

- The repeated calculation of `size` based on the passed `dts` to history makes me think the interface is not 100% correct here. Perhaps we should be using last_dt and size as the parameter instead? Which would match the actual history calls parameters of `(assets, bar_count, '1d', fields)` more closely.

- This is on the path of speeding up history calls on the lazy-mainline branch, after this the next steps are most likely:
-- addressing `get_minute_window_for_equities` when called in minute mode and with `1d` 
-- Doing a prefetch on minute data.

### DIFF
--- a/zipline/data/us_equity_loader.py
+++ b/zipline/data/us_equity_loader.py
@@ -1,0 +1,235 @@
+# Copyright 2016 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from numpy import dtype, around
+
+from six import iteritems
+
+from zipline.pipeline.data.equity_pricing import USEquityPricing
+from zipline.lib._float64window import AdjustedArrayWindow as Float64Window
+from zipline.lib.adjustment import Float64Multiply
+from zipline.utils.cache import CachedObject, Expired
+
+
+class SlidingWindow(object):
+    """
+    Wrapper around an AdjustedArrayWindow which supports monotonically
+    increasing (by datetime) requests for a sized window of data.
+
+    Parameters
+    ----------
+    window : AdjustedArrayWindow
+       Window of pricing data with prefetched values beyond the current
+       simulation dt.
+    cal_start : int
+       Index in the overall calendar at which the window starts.
+    """
+
+    def __init__(self, window, cal_start):
+        self.window = window
+        self.cal_start = cal_start
+        self.current = around(next(window), 3)
+        self.most_recent_ix = self.cal_start
+
+    def get(self, end_ix):
+        """
+        Returns
+        -------
+        out : A np.ndarray of the equity pricing up to end_ix after adjustments
+              and rounding have been applied.
+        """
+        if self.most_recent_ix == end_ix:
+            return self.current
+
+        self.current = around(self.window.seek(end_ix - self.cal_start + 1), 3)
+
+        self.most_recent_ix = end_ix
+        return self.current
+
+
+class USEquityHistoryLoader(object):
+    """
+    Loader for sliding history windows of adjusted US Equity Pricing data.
+
+    Parameters
+    ----------
+    daily_reader : DailyBarReader
+        Reader for daily bars.
+    adjustment_reader : SQLiteAdjustmentReader
+        Reader for adjustment data.
+    """
+
+    def __init__(self, daily_reader, adjustment_reader):
+        self._daily_reader = daily_reader
+        self._calendar = daily_reader._calendar
+        self._adjustments_reader = adjustment_reader
+        self._daily_window_blocks = {}
+
+        self._prefetch_length = 40
+
+    def _get_adjustments_in_range(self, assets, days, field):
+        """
+        Get the Float64Multiply objects to pass to an AdjustedArrayWindow.
+
+        For the use of AdjustedArrayWindow in the loader, which looks back
+        from current simulation time back to a window of data the dictionary is
+        structured with:
+        - the key into the dictionary for adjustments is the location of the
+        day from which the window is being viewed.
+        - the start of all multiply objects is always 0 (in each window all
+          adjustments are overlapping)
+        - the end of the multiply object is the location before the calendar
+          location of the adjustment action, making all days before the event
+          adjusted.
+
+        Parameters
+        ----------
+        assets : iterable of Asset
+            The assets for which to get adjustments.
+
+        days : iterable of datetime64-like
+            The days for which adjustment data is needed.
+        field : str
+            OHLCV field for which to get the adjustments.
+
+        Returns
+        -------
+        out : The adjustments as a dict of loc -> Float64Multiply
+        """
+        sids = {int(asset): i for i, asset in enumerate(assets)}
+        start = days[0]
+        end = days[-1]
+        adjs = {}
+        for sid, i in iteritems(sids):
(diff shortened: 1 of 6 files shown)


zipline/data/us_equity_loader.py
```diff
@@ -0,0 +1,230 @@
...
+    Parameters
+    ----------
+    window : AdjustedArrayWindow
+       Window of pricing data with prefetched values beyond the current
+       simulation dt.
+    cal_start : int
+       Index in the overall calendar at which the window starts.
+    cal_end : int
+       Index in the overall calendar at which the window
+    adjustement_reader : SQLiteAdjustmentReader
```
ssanderson: This looks like it's out of date?

ehebert: Typo, artifact of basing docstring on loader. Removing.

ehebert: Also, fixing type of `adjustement` reader in other docstring while at it.

zipline/data/us_equity_loader.py
```diff
@@ -0,0 +1,230 @@
...
+                                                         i,
+                                                         i,
+                                                         m[1])]
+                divs = self._adjustments_reader.get_adjustments_for_sid(
+                    'dividends', sid)
+                for d in divs:
+                    dt = d[0]
+                    if start < dt <= end:
+                        end_loc = days.get_loc(dt)
+                        adjs[end_loc] = [Float64Multiply(0,
```
ssanderson: I think this is broken if we have a merger and a dividend or a split on the same date? (We're just overwriting it seems like?)

zipline/data/us_equity_loader.py
```diff
@@ -0,0 +1,230 @@
...
+                                                         end_loc - 1,
+                                                         i,
+                                                         i,
+                                                         d[1])]
+            splits = self._adjustments_reader.get_adjustments_for_sid(
+                'splits', sid)
+            for s in splits:
+                dt = s[0]
+                if field == 'volume':
+                    ratio = s[1] / 1.0
```
ssanderson: I think you want the inverse of this?  Is this currently passing tests?

zipline/data/us_equity_loader.py
ehebert: Good catch. Had made it append at one point, but that must have been lost in a stash/side-branch.
Adding back in append behavior now.

zipline/data/us_equity_loader.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def __init__(self, daily_reader, adjustment_reader):
+        self._daily_reader = daily_reader
+        self._calendar = daily_reader._calendar
+        self._adjustments_reader = adjustment_reader
+        self._daily_window_blocks = {}
+
+        self._prefetch_length = 40
+
+    def _get_adjustments_in_range(self, assets, days, field):
```
ssanderson: I think this is just re-implementing `load_adjustments_from_sqlite` in `_adjustments.pyx`.  Am I missing something here?

zipline/data/us_equity_loader.py
ehebert: You are right, that was meant to be `1.0 / s[1]`.
It does pass tests, which means the `split + volume` test case is probably not covered.
Fixing now, and checking tests.

zipline/data/us_equity_loader.py
ehebert: Initial implementation did use `load_adjustments_from_sqlite`, but was incurring a cost (though I do not heavy the number in front of me, and can get it again) of querying for all splits/mergers/dividends, even when a handful are desired.

Specifically,
```
    cdef set split_sids = _get_split_sids(
        adjustments_db,
        start_date,
        end_date,
    )
    cdef set merger_sids = _get_merger_sids(
        adjustments_db,
        start_date,
        end_date,
    )
    cdef set dividend_sids = _get_dividend_sids(
        adjustments_db,
        start_date,
        end_date,
    )
```

If it does end up being too costly, we could alter `_adjustments.pyx` to only do the extra filtering query iff the number of sids is greater than a certain threshold.

ehebert: Also, the `date_loc` needs to be `effective_date` - 1, instead of the `effective_date`.

Though, whether or not to do that shift on the date, could be parameterized.

zipline/data/us_equity_loader.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        Get the Float64Multiply objects to pass to an AdjustedArrayWindow.
+
+        For the use of AdjustedAraryWindow in the loader, which looks back
```
jbredeche: typo - AdjustedArrayWindow

zipline/data/us_equity_loader.py
```diff
@@ -0,0 +1,240 @@
...
+        days : iterable of datetime64-like
+            The days for which adjustment data is needed.
+        field : str
+            OHLCV field for which to get the adjustments.
+
+        Returns
+        -------
+        out : The adjustments as a dict of loc -> Float64Multiply
+        """
+        sids = {int(asset): i for i, asset in enumerate(assets)}
```
jbredeche: faster to do `asset.sid`?  not sure.

zipline/data/us_equity_loader.py
```diff
@@ -0,0 +1,240 @@
...
+        start = days[0]
+        end = days[-1]
+        adjs = {}
+        for sid, i in iteritems(sids):
+            if field != 'volume':
+                mergers = self._adjustments_reader.get_adjustments_for_sid(
```
jbredeche: for all these queries to sqlite, is it worth specifying the date range in the query instead of pulling them all into memory every time?

zipline/data/us_equity_loader.py
```diff
@@ -0,0 +1,236 @@
...
+                                           i,
+                                           ratio)
+                    try:
+                        adjs[end_loc].append(mult)
+                    except KeyError:
+                        adjs[end_loc] = [mult]
+        return adjs
+
+    def _ensure_sliding_window(
+            self, assets, start, end, size, start_ix, end_ix, field):
```
jbredeche: can you add some comments explaining what this method is doing?

zipline/lib/_windowtemplate.pxi
```diff
@@ -53,10 +53,16 @@ cdef class AdjustedArrayWindow:
         self.adjustment_indices = sorted(adjustments, reverse=True)
         self.window_length = window_length
         self.anchor = window_length + offset
+        self.next_anchor = self.anchor
         self.max_anchor = data.shape[0]
 
         self.next_adj = self.pop_next_adj()
 
+    property anchor:
+
```
jbredeche: don't need this empty line

zipline/data/us_equity_loader.py
ehebert: Potentially, yes. And that may be part of moving over to using `load_adjustments`.
However, this is not a hot bottleneck.

zipline/data/us_equity_loader.py
ehebert: Might be more semantically correct to use `asset.sid`. This function is only called about once a quarter in a simulation, so any gain would be marginal.

zipline/lib/_windowtemplate.pxi
ehebert: Ended up removing the property, in favor of a `seek` method.

jbredeche:
lgtm, looks like it needs rebasing.

ehebert:
Merged with 7e70b000e97be92194d5e7a413764ed876568f77. Mea culpa for not doing via UI.

PR #1023, opened 2016-03-02
MAINT: Build shims to maintain some Q1 behavior, with some warnings, in Q2.

jbredeche:


jbredeche:
I have to do a bit more work here, hold off on review.

tests/test_algorithm.py
```diff
@@ -231,22 +231,6 @@ def fake_method(*args, **kwargs):
             with ZiplineAPI(algo):
                 self.assertIs(sentinel, getattr(zipline.api, name)())
 
-    def test_cannot_iterate_over_data(self):
```
ehebert: Should we just invert this test instead of deleting? i.e. ensure that data can be iterated over, (and possibly check for warning?)

tests/test_api_shim.py
```diff
@@ -0,0 +1,120 @@
...
+    @classmethod
+    def create_algo(cls, code, filename=None):
+        return TradingAlgorithm(
+            script=code,
+            sim_params=cls.sim_params,
+            env=cls.env,
+            algo_filename=filename
+        )
+
+    def test_iterate_data(self):
```
ehebert: Ah, I see the aforementioned test was moved here.

zipline/_protocol.pyx
```diff
@@ -391,8 +399,74 @@ cdef class BarData:
             normalize_date(self.simulation_dt_func())
         )
 
+    #################
+    # OLD API SUPPORT
+    #################
+    cdef _calculate_universe(self):
```
ehebert: This could be memoized using `remember_last` with the addition of `dt` as parameter.

zipline/algorithm.py
```diff
@@ -440,17 +440,28 @@ def _create_generator(self, sim_params):
...
+    def _calculate_universe(self):
+        # this exists to provide backwards compatibility for older,
+        # deprecated APIs, particularly around the iterability of
+        # BarData (ie, 'for sid in data).
+
+        # our universe is the intersection of:
+        # 1) all assets in our positions
+        # 2) all assets in open orders
+        return set(
+            list(self.perf_tracker.position_tracker.positions.keys()) +
```
ehebert: `keys` should already be a list.

jbredeche: not in python3.

zipline/data/data_portal.py
```diff
@@ -1362,3 +1366,38 @@ def get_fetcher_assets(self, day):
...
+                [asset], dt, 2, "1d", "price", ffill=True
+            )[asset]
+
+            return (hst.iloc[-1] - hst.iloc[0]) / hst.iloc[0]
+
+        if bars is None:
+            raise ValueError("bars cannot be None!")
+
+        price_arr = self.get_history_window(
+            [asset], dt, bars, "1m", "price", ffill=True
```
ehebert: Is this always `1m`? i.e. is this only to support Q minute mode, or should we provide this backwards compatibility for public daily mode as well?

jbredeche: great catch. fixing.

zipline/finance/blotter.py
```diff
@@ -229,34 +229,37 @@ def get_transactions(self, bar_data):
             example, Zipline models the commission cost into the fill price
             of the transaction), then this is None.
         """
+
         closed_orders = []
         transactions = []
 
-        assets = self.asset_finder.retrieve_all(self.open_orders.keys())
-        asset_dict = {asset.sid: asset for asset in assets}
+        if len(self.open_orders) > 0:
```
ehebert: Why is this required for shimming?

zipline/gens/tradesimulation.py
```diff
@@ -35,7 +35,8 @@ class AlgorithmSimulator(object):
         'daily': 'daily_perf'
     }
 
-    def __init__(self, algo, sim_params, data_portal, clock, benchmark_source):
+    def __init__(self, algo, sim_params, data_portal, clock, benchmark_source,
+                 universe_func):
```
ehebert: `universe_func` should probably have a default of `None`?

zipline/zipline_deprecation_warning.py
```diff
@@ -13,20 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from cpython cimport bool
```
ehebert: Looks like `git` thought this was a rename, when it is not.

Why remove the `pxd`?

zipline/algorithm.py
```diff
@@ -440,17 +440,28 @@ def _create_generator(self, sim_params):
...
+        )
+
+        return self.trading_client.transform()
+
+    def _calculate_universe(self):
+        # this exists to provide backwards compatibility for older,
+        # deprecated APIs, particularly around the iterability of
+        # BarData (ie, 'for sid in data).
+
+        # our universe is the intersection of:
```
richafrank: intersection -> union

zipline/algorithm.py
ssanderson: If you're treating the keys of a dictionary as a set-like object, you probably want `viewkeys`: https://docs.python.org/2/library/stdtypes.html#dictionary-view-objects

`self.perf_tracker.position_tracker.positions.viewkeys() | self.blotter.open_orders.viewkeys()` is equivalent to this, but without the repeated traversals.

zipline/zipline_deprecation_warning.py
jbredeche: we don't need it anymore.  it was there to support downstream subclassing of some of our Cython classes, but we don't need to do that anymore.

zipline/_protocol.pyx
```diff
@@ -391,8 +399,74 @@ cdef class BarData:
...
+        iter(self.keys())
+
+    def __getitem__(self, name):
+        return self._get_equity_price_view(name)
+
+    cdef _warn_user(self, msg):
+        warnings.warn(
+            msg,
+            category=ZiplineDeprecationWarning,
+            stacklevel=1
```
ssanderson: `stacklevel=1` probably isn't what you want here, given that this is almost always being called from our code.

jbredeche: I have a test for this, it is what I want, because this is Cython code.  the first python stackframe is the algorithm code.

zipline/finance/blotter.py
jbredeche: it's not.  I just noticed it as a small optimization and wanted to take care of it before I forgot.

zipline/gens/tradesimulation.py
jbredeche: not sure, what do we gain from that?

zipline/_protocol.pyx
```diff
@@ -20,8 +21,17 @@ import numpy as np
 from cpython cimport bool
 
 from zipline.assets import Asset
+from zipline.zipline_deprecation_warning import ZiplineDeprecationWarning
 
 cdef class BarData:
+    cdef object data_portal
+    cdef object simulation_dt_func
+    cdef object data_frequency
+    cdef object _views
```
richafrank: Does the forward declaration change the generated C?  Should any of these be more explicitly typed, e.g. `cdef dict _views`?

zipline/_protocol.pyx
```diff
@@ -29,7 +39,8 @@ cdef class BarData:
 
     This is what is passed as `data` to the `handle_data` function.
     """
-    def __init__(self, data_portal, simulation_dt_func, data_frequency):
+    def __init__(self, data_portal, simulation_dt_func, data_frequency,
+                 universe_func=None):
```
richafrank: When is the default used?

zipline/_protocol.pyx
richafrank: I have the same question.

zipline/_protocol.pyx
```diff
@@ -391,8 +399,74 @@ cdef class BarData:
...
+        return len(self._calculate_universe())
+
+    def keys(self):
+        self._warn_user("Iterating over the assets in `data` is "
+                        "deprecated.")
+
+        return list(self._calculate_universe())
+
+    def iterkeys(self):
+        iter(self.keys())
```
richafrank: missing `return` - must be untested.

zipline/_protocol.pyx
```diff
@@ -422,12 +496,17 @@ cdef class SidView:
             column = "close"
         elif column == "open_price":
             column = "open"
+        elif column == "dt":
+            return self.dt
```
richafrank: These seem unnecessary, given that `__getattr__` is called when no instance or class attribute was found.

zipline/_protocol.pyx
```diff
@@ -452,4 +531,39 @@ cdef class SidView:
...
+        )
+
+    def returns(self):
+        self._warn_user("The `returns` method is deprecated.")
+        return self.data_portal.get_simple_transform(
+            self.asset, "returns", self.simulation_dt_func(),
+            self.data_frequency
+        )
+
+    cdef _warn_user(self, msg):
```
richafrank: Maybe this should be named something like `_warn_deprecated`.

zipline/zipline_deprecation_warning.py
```diff
@@ -13,20 +13,6 @@
 # See the License for the specific language governing permissions and
```
richafrank: Instead of a having a module dedicated to only a single warning class, what do you think about naming this `warnings.py`, like we have `errors.py`?  Then we can add more later without making the namespace confusing.

zipline/algorithm.py
```diff
@@ -439,19 +441,25 @@ def _create_generator(self, sim_params):
...
+            self._create_benchmark_source(),
+            universe_func=self._calculate_universe
         )
 
+        return self.trading_client.transform()
+
+    def _calculate_universe(self):
+        # this exists to provide backwards compatibility for older,
+        # deprecated APIs, particularly around the iterability of
+        # BarData (ie, 'for sid in data).
```
richafrank: missing close quote

zipline/algorithm.py
```diff
@@ -485,16 +506,15 @@ def run(self, data=None, overwrite_sim_params=True):
...
                 equities = []
-                futures = []
-                for asset in assets:
+                for asset in self._assets_from_source:
                     if isinstance(asset, Equity):
                         equities.append(asset)
-                    elif isinstance(asset, Future):
-                        futures.append(asset)
-                if equities:
+                if self._equities:
```
richafrank: Where is this defined?  Do we no longer have any tests that run with a DataFrame?

zipline/algorithm.py
```diff
@@ -1145,11 +1165,14 @@ def cancel_order(self, order_param):
...
-    def history(self, assets, bar_count, frequency, field, ffill=True):
-        try:
-            assets = list(assets)
-        except TypeError:
-            assets = [assets]
+    def history(self, bar_count, frequency, field, ffill=True):
+        warnings.warn(
+            "The `history` method is deprecated.  Use `data.history` instead.",
+            category=ZiplineDeprecationWarning,
+            stacklevel=4
```
richafrank: Why 4?

zipline/finance/blotter.py
richafrank: `if self.open_orders:`

zipline/finance/blotter.py
```diff
@@ -246,34 +246,37 @@ def get_transactions(self, bar_data):
...
+            assets = self.asset_finder.retrieve_all(self.open_orders.keys())
```
richafrank: `iter` over a dict will yield the keys, so I don't think we need to make a new list of them.  Or are we mutating them here?

zipline/zipline_deprecation_warning.py
jbredeche: love it.

zipline/zipline_deprecation_warning.py
ehebert: Just confirming this means `live_bar_data.pyx` is removed in the corresponding internal patch?

jbredeche: yup

zipline/algorithm.py
```diff
@@ -485,16 +506,15 @@ def run(self, data=None, overwrite_sim_params=True):
                 copy_panel.items = self._write_and_map_id_index_to_sids(
                     copy_panel.items, copy_panel.major_axis[0],
                 )
-                assets = self.trading_environment.asset_finder.retrieve_all(
-                    copy_panel.items)
+                self._assets_from_source = \
+                    self.trading_environment.asset_finder.retrieve_all(
```
richafrank: Do we want to `set` this, like on master?  Not sure if there can be dupes.

zipline/_protocol.pyx
jbredeche: Can't add an 'arbitrary' decorator to a `cdef` function.

zipline/algorithm.py
jbredeche: That's what I needed to get back into the algocode.

zipline/_protocol.pyx
richafrank: Is there a perf hit then?  Should we define one in cython?

zipline/algorithm.py
richafrank: Sorry, I should have been clearer.  If an algo in zipline has:

```python
def handle_data(self, data):
    self.history(...)
```
then probably it should be `stacklevel=1`.  Trying to understand which usage this is for and if we care about any others.

jbredeche: in your example, I think that goes through a bunch of the `ZiplineAPI` decorator machinery, necessitating the higher stacklevel number.

richafrank: Hmm, I believe [`api_method`] (https://github.com/quantopian/zipline/blob/8e59d12daf0e59f23e8eeb48ab27d0cdf1fe19b5/zipline/utils/api_support.py#L56-L56) returns its input function.

zipline/_protocol.pyx
jbredeche: by a bunch of tests.  I prefer not requiring a universe func because in the future, we won't pass one when the old API is deprecated.

zipline/algorithm.py
jbredeche: well, the test currently verifies that we are getting to the proper stackframe to generate the warning, so I'm inclined to leave it alone for now.

PR #1024, opened 2016-03-02
BUG: Fix history crash at data start.

ehebert:
Fix a bug when history is called in a backtest with a start date near
the first trading day of the daily bars file, where load_raw_arrays
would hit a key error because data was requested which preceded dates
stored in the daily bar reader's calendar.

In the case where the first day is before the daily bar start, truncate
the query to days in the daily bar readers range, and then prepend nans.

### DIFF
--- a/zipline/data/us_equity_loader.py
+++ b/zipline/data/us_equity_loader.py
@@ -188,11 +188,12 @@ class USEquityHistoryLoader(object):
             td = self.env.trading_days
             offset = td.get_loc(start) - td.get_loc(
                 self._daily_reader.first_trading_day)
-            if end < self._daily_reader.first_trading_day:
-                array = full((size, 1), nan)
+            pre_slice = self._calendar.slice_indexer(start, end)
+            fill_size = pre_slice.stop - pre_slice.start
+            if field != 'volume':
+                pre_array = full((fill_size, 1), nan)
             else:
-                pre_slice = self._calendar.slice_indexer(start, end)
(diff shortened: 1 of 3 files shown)


zipline/data/us_equity_loader.py
```diff
@@ -52,7 +53,8 @@ def get(self, end_ix):
         if self.most_recent_ix == end_ix:
             return self.current
 
-        self.current = around(self.window.seek(end_ix - self.cal_start + 1), 3)
+        target = end_ix - self.cal_start - self.offset + 1
+        self.current = around(self.window.seek(target), 3)
```
jbredeche: what does this `3` mean?

zipline/data/us_equity_loader.py
```diff
@@ -36,10 +36,11 @@ class SlidingWindow(object):
        Index in the overall calendar at which the window starts.
     """
 
-    def __init__(self, window, cal_start):
+    def __init__(self, window, cal_start, offset):
```
jbredeche: can you add `offset` to the docstring?

zipline/data/us_equity_loader.py
```diff
@@ -176,16 +179,39 @@ def _ensure_sliding_window(
         except KeyError:
             pass
 
-        start_ix = self._calendar.get_loc(start)
-        end_ix = self._calendar.get_loc(end)
+        # Handle case where data is request before the start of data available
```
jbredeche: typo 'requested'

PR #1027, opened 2016-03-04
MAINT: add buyback auth to factors init.

mtydykov:
Forgot to add buyback auth factors to `__init__`. Also, small formatting fixes.


llllllllll:
lgtm, @richafrank any idea about the gfortan stuff?

mtydykov:
There's also an import error? Doesn't seem related to any of my changes.

richafrank:
> lgtm, @richafrank any idea about the gfortan stuff?

I think we need to pin the version of libgfortran that we're conda installing to whatever last passed (I think 1 instead of 3?).

mtydykov:
Is that something I should do in this PR? Or elsewhere?

richafrank:
Seems to be failing for other PRs too, so could fix it separately.

richafrank:
Should be fixed on maser now.

PR #1030, opened 2016-03-07
WIP (DO NOT MERGE): Introduce new ExchangeCalendar and TradingSchedule classes

StewartDouglas:
Refactor our calendar and timing logic so that it is independent of the `TradingEnvironment`, and can be generalized to futures exchanges which trade 24/5.

The ExchangeCalendar's API looks like the following public attributes and methods:
- `name`: The name of this exchange, e.g. `NYSE`.
- `tz`: The timezone in which this exchange is located.
- `is_open_on_minute`: Is the exchange open* at the given minute?
- `is_open_on_date`: Is the exchange open* at any point during the specified calendar day.
- `trading_days`: Return all of the exchange sessions between the given start and end, inclusive.
- `opens_and_closes`: Return the opens and closes of the exchange during the specified date.
- `session_date`: Return the `session` to which a given minute belongs. For example, a futures contract trading at 7pm EST on Sunday is considered to be part of Monday's session. 
- `minutes_for_date`: For a given datetime, return the minutes in the corresponding session
- `minute_window`: Returns a `DatetimeIndex` containing a window of minutes.

The ExchangeCalendar API still has a number of unanswered questions, including:
- Should the methods such as `trading_days` only accept canonicalized dates, or should they map arbitrary datetimes to session boundaries before returning the date range.
- Should `is_open_on_date` return `True` when given Sunday datetimes for relevant futures exchanges.
These questions will be answered in party by what the current consumers of this information expect.

The main lookup data structure, `self.schedule`, is a pandas `DataFrame`, but we should review this and decide whether to instead use a numpy array of minutes or pandas `DatetimeIndex`. There will be space vs. time trade-offs associated with this decision.

This work borrows significantly from the work done by @ssanderson here: https://github.com/quantopian/zipline/pull/556

* We define `open` to mean: is the exchange accepting orders.


jfkirk:
Note from our conversation: We may want two separate `is_open_on_date` and `is_session_date` methods to break up the two separate answers for different consumers.

zipline/finance/trading.py
```diff
@@ -69,7 +70,7 @@ def __init__(
         bm_symbol='^GSPC',
         exchange_tz="US/Eastern",
         max_date=None,
-        env_trading_calendar=tradingcalendar,
+        env_trading_calendar=NYSEExchangeCalendar,
```
jfkirk: Why pass in a calendar class, rather than an instantiated Calendar object? Doing the construction internally precludes us from giving the calendar arguments, if that is something that we ever want to do.

That may not be worth worrying about, though, if the eventual goal is to have all of the Calendar work removed from this class anyway.

zipline/finance/trading.py
```diff
@@ -504,53 +507,49 @@ def __init__(self, period_start, period_end,
...
+        assert self.period_end >= cal.first_trading_day, \
             "Period end falls before the first known trading day."
 
-        self.first_open = self._calculate_first_open(env)
-        self.last_close = self._calculate_last_close(env)
-
-        start_index = env.get_index(self.first_open)
-        end_index = env.get_index(self.last_close)
+        self.first_open = self._calculate_first_open(cal)
+        self.last_close = self._calculate_last_close(cal)
```
jfkirk: Note to self: these fields seem more like they should be SimulationParameters, not attributes of the environment

zipline/finance/trading.py
StewartDouglas: Good point about the calendar arguments. Were you thinking instead of instantiating the `NYSEExchangeCalendar` outside of the `TradingEnvironment`, and then passing it in? That doesn't seem ideal to me, since it implies that the `NYSEExchangeCalendar` would be instantiated every time `trading.py` was imported. How about we add `min_date` and `max_date` kwargs to the `TradingEnvironment`, as Scotty did here: https://github.com/quantopian/zipline/pull/556/files#diff-84720364fb47256a081315e8449ee0d2R101

zipline/utils/exchange_calendar.py
```diff
@@ -0,0 +1,386 @@
...
+from pandas.tseries.offsets import CustomBusinessDay
+
+start_default = pd.Timestamp('1990-01-01', tz='UTC')
+end_base = pd.Timestamp('today', tz='UTC')
+# Give an aggressive buffer for logic that needs to use the next trading
+# day or minute.
+end_default = end_base + pd.Timedelta(days=365)
+
+
+def delta_from_time(t):
```
jfkirk: @StewartDouglas what is the purpose of this function? is it cruft/placeholder?

zipline/utils/exchange_calendar.py
```diff
@@ -0,0 +1,386 @@
...
+    days : DatetimeIndex
+        The "base" time which we want to change.
+    t : datetime.time
+        The time we want to offset @days by
+    tz : pytz.timezone
+        The timezone which these times represent
+    day_offset : int
+        The number of days we want to offset @days by
+    """
+    days = DatetimeIndex(days).tz_localize(None).tz_localize(tz)
```
jfkirk: Why localize to none, then back to tz, rather than just straight to tz?

zipline/utils/exchange_calendar.py
StewartDouglas: I think Scotty had planned on using this on his branch but it's not necessary for ours: it can be deleted.

zipline/utils/exchange_calendar.py
StewartDouglas: This is so we can easily flip from midnight UTC to midnight EST. We index our schedule in UTC, but in order to offset correctly given that `NYSE_OPEN` and `NYSE_CLOSE` are timezone naive, we must my convert our dates into EST: UTC -> Naive -> EST.

zipline/utils/exchange_calendar.py
```diff
@@ -0,0 +1,386 @@
...
+    opens_or_closes.values[indexer] = special_opens_or_closes.values
+
+
+class ExchangeCalendar(with_metaclass(ABCMeta)):
+    """
+    An ExchangeCalendar represents the timing information of a single market
+    exchange.
+
+    Properties
+    ----------
```
StewartDouglas: This should be renamed to `Parameters`. We should then have an `Attributes` section below it, documenting `start` and `end`: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#documenting-classes

jfkirk:
Closing in deference to new branch.

PR #1031, opened 2016-03-07
Ignore missing deltas in pipeline

mtydykov:
This branch addresses an [issue](https://github.com/quantopian/qexec/issues/9196) in which running Pipeline for certain premium datasets for certain date ranges results in a KeyError because there are sids that exist in the databazaar deltas table but that are not returned by the pipeline engine. This fix filters out those sids and logs a message that keeps track of which sids are not found.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -37,8 +37,11 @@ from zipline.utils.numpy_utils import (
     int64_dtype,
     repeat_last_axis,
 )
-from zipline.utils.test_utils import tmp_asset_finder, make_simple_equity_info
-
+from zipline.utils.test_utils import (
+    tmp_asset_finder,
+    make_simple_equity_info,
+    make_test_handler
+)
 
 nameof = op.attrgetter('name')
 dtypeof = op.attrgetter('dtype')
@@ -1128,3 +1131,82 @@ class BlazeToPipelineTestCase(TestCase):
                 window_length=3,
                 compute_fn=op.itemgetter(-1),
             )
+
+    def test_ignore_nonexistent_delta(self):
+        missing_sid = ord('E')
+        handler = make_test_handler(self)
+        asset_info = asset_infos[0][0]
+        base_dates = pd.DatetimeIndex([
+            pd.Timestamp('2014-01-01'),
+            pd.Timestamp('2014-01-04')
+        ])
+        repeated_dates = base_dates.repeat(4)
+        baseline = pd.DataFrame({
+            'sid': (self.sids + (missing_sid,)) * 2,
+            'value': (0., 1., 2., 3., 1., 2., 3., 4.),
+            'int_value': (0, 1, 2, 3, 1, 2, 3, 4.),
+            'asof_date': repeated_dates,
+            'timestamp': repeated_dates
+        })
+        expr = bz.Data(baseline, name='expr', dshape=self.dshape)
+        deltas = bz.Data(
+            odo(
+                bz.transform(
+                    expr,
+                    value=expr.value + 10,
+                    timestamp=expr.timestamp + timedelta(days=1),
+                ),
+                pd.DataFrame,
+            ),
+            name='delta',
+            dshape=self.dshape,
+        )
+        expected_views = keymap(pd.Timestamp, {
+            '2014-01-03': np.array([[10.0, 11.0, 12.0],
+                                    [10.0, 11.0, 12.0],
+                                    [10.0, 11.0, 12.0]]),
+            '2014-01-06': np.array([[10.0, 11.0, 12.0],
+                                    [10.0, 11.0, 12.0],
+                                    [11.0, 12.0, 13.0]]),
+        })
+        if len(asset_info) == 4:
+            expected_views = valmap(
+                lambda view: np.c_[view, [np.nan, np.nan, np.nan]],
+                expected_views,
+            )
+            expected_output_buffer = [10, 11, 12, np.nan, 11, 12, 13, np.nan]
+        else:
+            expected_output_buffer = [10, 11, 12, 11, 12, 13]
+
+        cal = pd.DatetimeIndex([
+            pd.Timestamp('2014-01-01'),
+            pd.Timestamp('2014-01-02'),
+            pd.Timestamp('2014-01-03'),
+            # omitting the 4th and 5th to simulate a weekend
+            pd.Timestamp('2014-01-06'),
+        ])
+        with handler.applicationbound():
+            with tmp_asset_finder(equities=asset_info) as finder:
+                expected_output = pd.DataFrame(
+                    expected_output_buffer,
+                    index=pd.MultiIndex.from_product((
+                        sorted(expected_views.keys()),
+                        finder.retrieve_all(asset_info.index),
+                    )),
+                    columns=('value',),
+                )
+                self._run_pipeline(
+                    expr,
+                    deltas,
+                    expected_views,
+                    expected_output,
+                    finder,
+                    calendar=cal,
+                    start=cal[2],
+                    end=cal[-1],
+                    window_length=3,
+                    compute_fn=op.itemgetter(-1),
+                )
+            message = handler.records[0].message
+            exp_msg = "Didn't find the following sids in asset index: {}"
+            self.assertIn(exp_msg.format(set([missing_sid])), message)
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -5,6 +5,7 @@ from __future__ import division
 
 from collections import OrderedDict
 from datetime import timedelta, time
+from itertools import product, chain
 from unittest import TestCase
 import warnings
 
@@ -40,7 +41,6 @@ from zipline.utils.numpy_utils import (
 from zipline.utils.test_utils import (
     tmp_asset_finder,
     make_simple_equity_info,
-    make_test_handler
 )
 
 nameof = op.attrgetter('name')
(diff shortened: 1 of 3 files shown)


tests/pipeline/test_blaze.py
```diff
@@ -1128,3 +1131,82 @@ def test_novel_deltas_macro(self):
                 window_length=3,
                 compute_fn=op.itemgetter(-1),
             )
+
+    def test_ignore_nonexistent_delta(self):
```
mtydykov: This test has a structure very similar to `test_novel_deltas`. We could simplify it a bit, figure out a way to parameterize (even though there's an extra context manager here, so that might get tricky), or just leave it as is.

zipline/utils/test_utils.py
```diff
@@ -881,3 +881,9 @@ def decorator(f):
         param_sets = product(*(params[name] for name in argnames))
         return subtest(param_sets, *argnames)(f)
     return decorator
+
+
+def make_test_handler(testcase, *args, **kwargs):
+    handler = TestHandler(*args, **kwargs)
+    testcase.addCleanup(handler.close)
```
llllllllll: is addCleanup called in teardown or teardownClass?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -952,6 +954,17 @@ def collect_expr(e):
                 columns=added_query_fields + list(map(getname, columns)),
             )
         )
+        if not materialized_deltas.empty and have_sids:
+            missing_sids = materialized_deltas[
+                ~materialized_deltas.sid.isin(assets)
+            ]
+            log.debug(
+                "Didn't find the following sids in asset index: {}".format(
```
llllllllll: dont call format in the log.debug call, just pass the formatters as seperate arguments. This prevents the formatting from happening if debug logs are being dropped.

Also, I am still not sure we need this logging. it is 100% valid for there to be assets here that are not in the baseline and that is an expected case. If we really want this logging to happen downstream we might want to look at adding this in the downstream consumer of zipline.

Also noting that I am pretty sure we do not show debug logs by default so this might need to be an info wherever we agree this should go.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -194,6 +195,7 @@
 )
 is_invalid_deltas_node = complement(flip(isinstance, valid_deltas_node_types))
 get__name__ = op.attrgetter('__name__')
+log = logbook.Logger('BlazeLoader')
```
llllllllll: see comment below about the log; however, can we name loggers with `__name__` to make it easier to find the source of the logging?

tests/pipeline/test_blaze.py
```diff
@@ -1128,3 +1131,82 @@ def test_novel_deltas_macro(self):
...
+            expected_output_buffer = [10, 11, 12, 11, 12, 13]
+
+        cal = pd.DatetimeIndex([
+            pd.Timestamp('2014-01-01'),
+            pd.Timestamp('2014-01-02'),
+            pd.Timestamp('2014-01-03'),
+            # omitting the 4th and 5th to simulate a weekend
+            pd.Timestamp('2014-01-06'),
+        ])
+        with handler.applicationbound():
```
llllllllll:
```python
with handler.applicationbound(), tmp_asset_finder(...):
```

This also might be easier to follow if you just do with `make_test_handler(self)` or move the `handler = ...` to right before the ctx manager

tests/pipeline/test_blaze.py
llllllllll: we can always parameterize on the context manager, and in one case use `zipline.utils.context_tricks.nop_context`

zipline/utils/test_utils.py
mtydykov: According to the docs, it's called after tearDown, so I assume in teardownClass?

tests/pipeline/test_blaze.py
llllllllll: actually, in general I am not sure how I feel about testing the log messages. are these really static parts of the api of the function?

tests/pipeline/test_blaze.py
```diff
@@ -1128,3 +1131,82 @@ def test_novel_deltas_macro(self):
...
+                    expected_views,
+                    expected_output,
+                    finder,
+                    calendar=cal,
+                    start=cal[2],
+                    end=cal[-1],
+                    window_length=3,
+                    compute_fn=op.itemgetter(-1),
+                )
+            message = handler.records[0].message
```
llllllllll: we should scan the records to see if exactly one matches the pattern because this might not be the first thing logged.

zipline/utils/test_utils.py
llllllllll: err, I think that just means it is after tearDown and not in tearDownClass. This is the behavior we want though so it is okay

tests/pipeline/test_blaze.py
richafrank: Does it need to be applicationbound?

zipline/utils/test_utils.py
```diff
@@ -881,3 +881,9 @@ def decorator(f):
         param_sets = product(*(params[name] for name in argnames))
         return subtest(param_sets, *argnames)(f)
     return decorator
+
+
+def make_test_handler(testcase, *args, **kwargs):
```
llllllllll: Could we put a docstring on this utility so others know how to use this?

zipline/pipeline/loaders/blaze/core.py
mtydykov: Right now, the only cases where this happens is with invalid sids in the dbz database. E.g., there's a symbol that we symbol mapped incorrectly because we used a stale assets db. There's another symbol that a vendor is sending us updates for that expired years ago. Basically, all the times that this has been failing currently give us info that we probably want to know about. Maybe there's a better way/place to do this, though.

llllllllll: The databazaar is only one consumer of this, users may come with their own datasets that have some intersection with the assets that exist in their asset finder but also contains others. Also, this is relying on the implementation detail that the assets requested is currently the full universe in existance. There are plans to make this filter to only sids that have matched some other filters so I am not totally sure this logging at this particular spot is desired for all users of zipline

zipline/utils/test_utils.py
mtydykov: Sure. I just moved this here as is from qexec's `test_utils` (with the intention of getting rid of it there, since qexec depends on zipline). Of course, if we don't end up using it to log missing sids, probably doesn't need to be here.

tests/pipeline/test_blaze.py
mtydykov: I'm not really sure; there might be better ways of doing this. I'm using it the same way it's being used in the qexec tests.

zipline/utils/test_utils.py
llllllllll: We can still leave this here, but in either case it should be documented

zipline/pipeline/loaders/blaze/core.py
mtydykov: Ok. Maybe there should be another place (or just separate analysis entirely) to find the missing sids. I can get rid of the logging; it'll simplify things quite a bit.

zipline/utils/test_utils.py
mtydykov: Done

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -953,6 +953,13 @@ def collect_expr(e):
             )
         )
 
+        # It's not guaranteed that assets returned by the engine will contain
+        # all sids from the deltas table; filter out such mismatches here.
+        if not materialized_deltas.empty and have_sids:
+            materialized_deltas = materialized_deltas[
+                materialized_deltas.sid.isin(assets)
```
llllllllll: we should use the `SID_FIELD_NAME` constant here

zipline/utils/test_utils.py
```diff
@@ -881,3 +881,16 @@ def decorator(f):
...
+    """
+    Returns a TestHandler which will be used by the given testcase. This
+    handler can be used to test log messages.
+
+    testcase: unittest.TestCase
```
llllllllll:
```
Parameters
----------------
```

Also missing the return value

tests/pipeline/test_blaze.py
```diff
@@ -843,10 +849,21 @@ def compute(self, today, assets, out, data):
             check_dtype=False,
         )
 
-    @with_extra_sid
-    def test_deltas(self, asset_info):
-        expr = bz.Data(self.df, name='expr', dshape=self.dshape)
-        deltas = bz.Data(self.df, dshape=self.dshape)
+    @with_ignore_sid
+    def test_deltas(self, asset_info, extra_sid_to_ignore):
```
llllllllll: `extra_sid_to_ignore` sounds like it should the `sid` itself, maybe we could rename to something like `add_extra_sid` to show that it is actually a boolean. What do you think?

tests/pipeline/test_blaze.py
```diff
@@ -843,10 +849,21 @@ def compute(self, today, assets, out, data):
...
+        if not extra_sid_to_ignore:
+            df = self.df
+        else:
+            dates = self.dates.repeat(4)
+            df = pd.DataFrame({
```
llllllllll: we shouldn't duplicate `self.df` here because it could change. We could just add the data for `E` to a copy of `self.df`

tests/pipeline/test_blaze.py
mtydykov: sounds good

llllllllll:
lgtm. can you rebase the wip commits. other than that merge on pass

PR #1032, opened 2016-03-08
FIX: Make unexpected keyword arguments return meaningful errors

lianga888:
Cython returns unhelpful errors when you pass unexpected keywords.
Wrap relevant BarData methods with a decorator that checks the
keywords and returns a better error if needed.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -95,6 +95,10 @@ from zipline.test_algorithms import (
     noop_algo,
     record_float_magic,
     record_variables,
+    call_with_kwargs,
+    call_without_kwargs,
+    call_with_bad_kwargs_current,
+    call_with_bad_kwargs_history
 )
 from zipline.utils.context_tricks import CallbackManager
 import zipline.utils.events
@@ -1008,7 +1012,7 @@ class TestAlgoScript(TestCase):
 
         days = 251
 
-        trades_by_sid = {
+        cls.trades_by_sid = {
             0: factory.create_trade_history(
                 0,
                 [10.0] * days,
@@ -1025,14 +1029,9 @@ class TestAlgoScript(TestCase):
                 cls.env)
         }
 
-        cls.data_portal = create_data_portal_from_trade_history(cls.env,
-                                                                cls.tempdir,
-                                                                cls.sim_params,
-                                                                trades_by_sid)
-
-        cls.zipline_test_config = {
-            'sid': 0,
-        }
+        cls.data_portal = create_data_portal_from_trade_history(
+            cls.env, cls.tempdir, cls.sim_params, cls.trades_by_sid
+        )
 
     @classmethod
     def tearDownClass(cls):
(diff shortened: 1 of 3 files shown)


jbredeche:
@ehebert can you take a look?

zipline/_protocol.pyx
```diff
@@ -21,6 +21,28 @@ from cpython cimport bool
 
 from zipline.assets import Asset
 
+
+class assertKeywords(object):
```
richafrank: This feels like it belongs with the other pre-processors in [input_validation.py] (https://github.com/quantopian/zipline/blob/a3dbf75/zipline/utils/input_validation.py).  Should we have a cython decorator that wraps something there?

richafrank: I don't think this falls into any of the python naming conventions, either `assert_keywords` or `AssertKeywords`.

tests/test_algorithm.py
```diff
@@ -1294,6 +1297,38 @@ def test_account_in_init(self):
...
+            with self.assertRaises(TypeError) as cm:
+                test_algo = TradingAlgorithm(
+                    script=script,
+                    sim_params=self.sim_params,
+                    env=self.env,
+                )
+                test_algo.run(self.data_portal)
+
+            self.assertIn("got an unexpected keyword argument 'blahblah'",
+                          cm.exception.message)
```
ssanderson: `BaseException.message` is deprecated in Python 3.  You want `cm.exception.args[0]` here.

tests/test_algorithm.py
```diff
@@ -1294,6 +1297,38 @@ def test_account_in_init(self):
...
+        )
+        test_algo.run(self.data_portal)
+
+    def test_bad_kwargs(self):
+        """
+        Test that api methods on the data object called with bad kwargs return
+        a meaningful TypeError that we create, rather than an unhelpful cython
+        error
+        """
+        for script in [call_with_bad_kwargs_history,
```
ehebert: This seems like a good case for `nose.parameterized`.

e.g.:
```
@paremeterized.expand([('history', call_with_bad_kwargs_history),
                                         'current', call_with_bad_kwargs_current)])
def test_bad_kwargs(self, test_name, algo_text):
    ....
```

zipline/test_algorithms.py
```diff
@@ -894,3 +894,36 @@ def handle_data(context, data):
     context.incr += 1
     record(data=float('%s'))
 """
+
+call_with_kwargs = """
```
ehebert: For completion, we could also test `call_without_kwargs`.
e.g. `data.history(symbol('TEST'), 'price', 5, '1d')`

zipline/_protocol.pyx
```diff
@@ -21,6 +21,28 @@ from cpython cimport bool
...
+    """
+    Asserts that the keywords passed into the wrapped function are included
+    in those passed into this decorator. If not, raise a TypeError with a
+    meaningful message, unlike the one cython returns by default.
+    """
+
+    def __init__(self, *args):
+        self.names = args
+
+    def __call__(self, func):
```
ehebert: Double checked by comparing this branch against `lazy-mainline` with an algo that uses `data.history(context.stocks, 'high', 3, '1d')` and at least in that case, this wrapper seems to be in the clear as far as adding overhead.

zipline/_protocol.pyx
```diff
@@ -21,6 +21,28 @@ from cpython cimport bool
...
+
+class assert_keywords(object):
+    """
+    Asserts that the keywords passed into the wrapped function are included
+    in those passed into this decorator. If not, raise a TypeError with a
+    meaningful message, unlike the one cython returns by default.
+    """
+
+    def __init__(self, *args):
+        self.names = args
```
ehebert: If this is only every checking membership, could use a `set` here too. But the number of elements is so small, that the tradeoff between checking every element is probably marginal.

zipline/_protocol.pyx
```diff
@@ -21,6 +21,28 @@ from cpython cimport bool
 
 from zipline.assets import Asset
 
+
+class assert_keywords(object):
+    """
+    Asserts that the keywords passed into the wrapped function are included
+    in those passed into this decorator. If not, raise a TypeError with a
+    meaningful message, unlike the one cython returns by default.
```
ehebert: What does the Cython error that we are covering over look like?

Not a blocker for this patch, but is 'fixing' the Cython error something we should upstream?

zipline/_protocol.pyx
```diff
@@ -21,6 +21,28 @@ from cpython cimport bool
...
+    """
+
+    def __init__(self, *args):
+        self.names = args
+
+    def __call__(self, func):
+        def assert_keywords_and_call(*args, **kwargs):
+            for field in kwargs.keys():
```
ehebert: Calling `keys()` is redundant, iterating over a dict `for field in kwargs` has the same behavior.

zipline/_protocol.pyx
lianga888: It's something like "func() takes exactly x positional arguments (1 given)" when you misspell one of the keywords

PR #1033, opened 2016-03-08
BUG: Fixes incorrect behaviour in market open and close

abhijeetkalyan:
Fixes schedule_function's handling of after_open and before_close dates. 
We now look at *this* period start and period end, instead of the next period start and period end. The next period start and period end are implicitly creating state(when we call a rule for the first time, when its `_next_period_start` is None, and then again when its `_next_period_start `has a value). 

As a result we were seeing the following impure behaviour:
```python
In [33]: bc.should_trigger(test_close_time2, env)
Out[33]: True

In [34]: bc.should_trigger(test_close_time2, env)
Out[34]: False

In [35]: bc.should_trigger(test_close_time2, env)
Out[35]: False
```
because of the messed up state, we weren't correctly keeping track of where we were, which led to messed up behaviour with `schedule_function`. We avoid that here by simply keeping track of this period start and end, and updating the period if the date we're on is not the same as the cached date. In tests, we now only create one before_close and one after_open object, so that we can catch any regressions that affect their internal state, instead of creating fresh objects in each subtest that have a good internal state.

There is some performance impact, but we're still doing OK. Here's the code I used to test perf:
**this branch**
```python
In [11]: tdays = env.trading_days[2000:3000]

In [12]: ao = time_rules.market_open(minutes=30)

In [13]: def test_trigger(tdays, rule):
   ....:     for tday in tdays:
   ....:         mins = env.market_minutes_for_day(tday)
   ....:         for min in mins:
   ....:             rule.should_trigger(min, env)
   ....:             

In [14]: %timeit test_trigger(tdays, ao)
1 loops, best of 3: 1.13 s per loop

In [15]: bc = time_rules.market_close(minutes=30)

In [16]: %timeit test_trigger(tdays, ao)
1 loops, best of 3: 1.13 s per loop

In [17]: %timeit test_trigger(tdays, bc)
1 loops, best of 3: 1.07 s per loop
```

**lazy-mainline**
```python
In [1]: def test_trigger(tdays, rule):
    for tday in tdays:
        mins = env.market_minutes_for_day(tday)
        for min in mins:
            rule.should_trigger(min, env)
   ...:             

In [2]: ao = time_rules.market_open(minutes=30)

In [3]: tdays = env.trading_days[2000:3000]

In [4]: %timeit test_trigger(tdays, ao)
1 loops, best of 3: 838 ms per loop

In [5]: bc = time_rules.market_close(minutes=30)

In [6]: %timeit test_trigger(tdays, bc)
1 loops, best of 3: 730 ms per loop
```

**master**
```python
In [1]: def test_trigger(tdays, rule):
    for tday in tdays:
        mins = env.market_minutes_for_day(tday)
        for min in mins:
            rule.should_trigger(min, env)
   ...:             

In [2]: ao = time_rules.market_open(minutes=30)

In [4]: tdays = env.trading_days[2000:3000]

In [5]: %timeit test_trigger(tdays, ao)
1 loops, best of 3: 2.55 s per loop

In [6]: bc = time_rules.market_close(minutes=30)

In [7]: %timeit test_trigger(tdays, bc)
1 loops, best of 3: 2.51 s per loop
```

### DIFF
--- a/zipline/utils/events.py
+++ b/zipline/utils/events.py
@@ -240,7 +240,7 @@ class EventRule(six.with_metaclass(ABCMeta)):
 
 class StatelessRule(EventRule):
     """
-    A stateless rule has no state.
+    A stateless rule has no observable side effects.
     This is reentrant and will always give the same result for the
     same datetime.
     Because these are pure, they can be composed to create new rules.
@@ -336,28 +336,38 @@ class AfterOpen(StatelessRule):
             datetime.timedelta(minutes=1),  # Defaults to the first minute.
         )
 
-        self._next_period_start = None
-        self._next_period_end = None
+        self._period_start = None
+        self._period_end = None
+        self._period_close = None
 
         self._one_minute = datetime.timedelta(minutes=1)
 
     def calculate_dates(self, dt, env):
         # given a dt, find that day's open and period end (open + offset)
-        self._next_period_start = env.get_open_and_close(dt)[0]
-        self._next_period_end = \
-            self._next_period_start + self.offset - self._one_minute
+        self._period_start, self._period_close = env.get_open_and_close(dt)
+        self._period_end = \
+            self._period_start + self.offset - self._one_minute
 
     def should_trigger(self, dt, env):
-        if self._next_period_start is None:
+        # There are two reasons why we might want to recalculate the dates.
+        # One is the first time we ever call should_trigger, when
+        # self._period_start is none. The second is when we're on a new day,
+        # and need to recalculate the dates. For performance reasons, we rely
+        # on the fact that our clock only ever ticks forward, since it's
+        # cheaper to do dt1 <= dt2 than dt1.date() != dt2.date(). This means
+        # that we will NOT correctly recognize a new date if we go backwards
+        # in time(which should never happen in a simulation, or in a live
+        # trading environment)
+        if (
+            self._period_start is None or
+            self._period_close <= dt
+        ):
             self.calculate_dates(dt, env)
 
-        if self._next_period_start <= dt < self._next_period_end:
+        if self._period_start <= dt < self._period_end:
             # haven't made it past the offset yet
             return False
         else:
-            if dt >= self._next_period_end:
-                self.calculate_dates(env.next_trading_day(dt), env)
-
             return True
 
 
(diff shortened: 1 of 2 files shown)


zipline/utils/events.py
```diff
@@ -375,28 +379,33 @@ def __init__(self, offset=None, **kwargs):
...
     def should_trigger(self, dt, env):
-        if self._next_period_start is None:
+        # There are two reasons why we might want to recalculate the dates.
+        # One is the first time we ever call should_trigger, when
+        # self._period_start is none. The second is when we're on a new day,
+        # and need to recalculate the dates.
+        if (
+            self._period_start is None or
+            self._period_start.date() != dt.date()
+        ):
```
llllllllll: I think we can simplify this if statement

zipline/utils/events.py
```diff
@@ -375,28 +379,33 @@ def __init__(self, offset=None, **kwargs):
...
             self.calculate_dates(dt, env)
 
-        if dt <= self._next_period_start:
-            return False
-        else:
-            if dt > self._next_period_end:
-                self.calculate_dates(env.next_trading_day(dt), env)
-
+        if self._period_start < dt <= self._period_end:
```
llllllllll: I think we can simplify this if statement

abhijeetkalyan:
@jbredeche updated this PR with some performance improvements. We now do a less-than-equals check to see whether we need to calculate a new date(if `dt` is gte the `period_close`, we calculate a new date). This puts us back to lazy-mainline level perf numbers. The updated perf numbers for this branch are:
```python
In [1]: tdays = env.trading_days[2000:3000]

In [2]: def test_trigger(tdays, rule):
    for tday in tdays:
        mins = env.market_minutes_for_day(tday)
        for min in mins:
            rule.should_trigger(min, env)
   ...:             

In [3]: ao = time_rules.market_open(minutes=30)

In [4]: %timeit test_trigger(tdays, ao)
1 loops, best of 3: 851 ms per loop

In [5]: bc = time_rules.market_close(minutes=30)

In [6]: %timeit test_trigger(tdays, bc)
1 loops, best of 3: 858 ms per loop

```
**However**, we're now relying explicitly on the fact that the clock only moves forward. This logic will break if we ever pass a date prior to what we've passed before, which means that these rules are no longer truly stateless - but in our world of time only marching forward, they will behave just like any other stateless rules.

Since this now depends on time only moving forward, I've also updated the tests to pass in a sorted list of random dates, when testing before close and after open.

zipline/utils/events.py
```diff
@@ -336,28 +336,39 @@ def __init__(self, offset=None, **kwargs):
...
+        self._period_close = None
 
         self._one_minute = datetime.timedelta(minutes=1)
 
     def calculate_dates(self, dt, env):
         # given a dt, find that day's open and period end (open + offset)
-        self._next_period_start = env.get_open_and_close(dt)[0]
-        self._next_period_end = \
-            self._next_period_start + self.offset - self._one_minute
+        self._period_start = env.get_open_and_close(dt)[0]
```
jbredeche: this is doing an extra `get_open_and_close` unnecessarily.

you can probably do:
```
self._period_start, self._period_close = env.get_open_and_close(dt)
```

zipline/utils/events.py
```diff
@@ -336,28 +336,39 @@ def __init__(self, offset=None, **kwargs):
...
-        if self._next_period_start is None:
+        # There are two reasons why we might want to recalculate the dates.
+        # One is the first time we ever call should_trigger, when
+        # self._period_start is none. The second is when we're on a new day,
+        # and need to recalculate the dates. For performance reasons, we rely
+        # on the fact that our clock only ever ticks forward, since it's
+        # cheaper to do dt1 <= dt2 than dt1.date() != dt2.date(). This means
+        # that we will NOT correctly recognize a new date if we go backwards
+        # in time(which should never happen in a simulation, or in a live
+        # trading environment)
```
jbredeche: great comment

zipline/utils/events.py
abhijeetkalyan: Yeah, both this and the one in `AfterOpen`. Fixing in the morning

PR #1034, opened 2016-03-08
ENH: Add min date to TradingEnvironment.

ehebert:
Allow creation of TradingEnvironment to specify a minimum date, so that
trading days, market opens, etc. can trimmed to a range more relevant to
the backtest.

This changes is with an eye towards storing all market minutes in the
trading environment, where storing values for much more than the
simulation range starts to become more costly.

### DIFF
--- a/tests/test_finance.py
+++ b/tests/test_finance.py
@@ -25,6 +25,7 @@ from unittest import TestCase
 from datetime import datetime, timedelta
 
 import numpy as np
+import pandas as pd
 
 from nose.tools import timed
 
@@ -546,8 +547,16 @@ class TradingEnvironmentTestCase(TestCase):
         self.assertTrue(all(friday == minutes[31:421]))
         self.assertTrue(all(thursday == minutes[421:]))
 
+    def test_min_date(self):
+        min_date = pd.Timestamp('2016-03-04', tz='UTC')
+        env = TradingEnvironment(min_date=min_date)
+
+        self.assertGreaterEqual(env.first_trading_day, min_date)
+        self.assertGreaterEqual(env.treasury_curves.index[0],
+                                min_date)
+
     def test_max_date(self):
-        max_date = datetime(2008, 8, 1, tzinfo=pytz.utc)
+        max_date = pd.Timestamp('2008-08-01', tz='UTC')
         env = TradingEnvironment(max_date=max_date)
 
         self.assertLessEqual(env.last_trading_day, max_date)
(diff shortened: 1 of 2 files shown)


tests/test_finance.py
```diff
@@ -546,8 +547,16 @@ def test_market_minute_window(self):
         self.assertTrue(all(friday == minutes[31:421]))
         self.assertTrue(all(thursday == minutes[421:]))
 
+    def test_min_date(self):
+        min_date = pd.Timestamp('2016-03-04', tz='UTC')
+        env = TradingEnvironment(min_date=min_date)
+
+        self.assertLessEqual(env.first_trading_day, min_date)
```
richafrank: If we're specifying a minimum, shouldn't the `first_trading_day` be greater than or equal to it?

zipline/finance/trading.py
```diff
@@ -82,10 +83,7 @@ def __init__(
         # `tc_td` is short for "trading calendar trading days"
         tc_td = env_trading_calendar.trading_days
 
-        if max_date:
-            self.trading_days = tc_td[tc_td <= max_date].copy()
-        else:
-            self.trading_days = tc_td.copy()
```
richafrank: Are we ok to remove the `copy`?

tests/test_finance.py
ehebert: You are right, copy/paste error.

zipline/finance/trading.py
```diff
@@ -82,10 +83,7 @@ def __init__(
...
+        self.trading_days = tc_td[tc_td.slice_indexer(min_date, max_date)]
```
richafrank: Does this produce an inclusive range?

zipline/finance/trading.py
ehebert: I think so, the DatetimeIndex is immutable.

zipline/finance/trading.py
ehebert: Yes, e.g.:

```
(Pdb++) tc_td[tc_td.slice_indexer(pd.Timestamp('2016-03-04', tz='UTC'), pd.Time\
stamp('2016-03-08', tz='UTC'))]
DatetimeIndex(['2016-03-04', '2016-03-07', '2016-03-08'], dtype='datetime64[ns]', freq='C', tz='UTC')
```

richafrank: Great.

zipline/finance/trading.py
richafrank: Perfect.

PR #1037, opened 2016-03-09
BUG: proper accounting of leftover_cash for multiple concurrent splits

jbredeche:
If there were multiple splits that affected our holdings at the same
time, the leftover cash wasn’t properly being summed.

for https://github.com/quantopian/qexec/issues/9353


jbredeche:
small and easy, just merging.

PR #1039, opened 2016-03-09
TST: Update tests to use new data API

abhijeetkalyan:
Updated the tests that were using the old data API. After discussion with @jbredeche a couple of the tests that have use cases we don't really support anymore(`test_get_sid`, for example), have been killed. 

The PR itself is still WIP because I'm adding another commit that tests that the old and new data APIs return identical results for OHLCVP, but this commit itself is ready to be looked at.

UPDATE: Added a test that verifies old and new data APIs both hit the same code path.

### DIFF
--- a/tests/test_api_shim.py
+++ b/tests/test_api_shim.py
@@ -1,5 +1,6 @@
 import warnings
 from unittest import TestCase
+from mock import patch
 import pandas as pd
 from testfixtures import TempDirectory
 
@@ -9,6 +10,7 @@ from zipline.data.minute_bars import BcolzMinuteBarWriter, \
     US_EQUITIES_MINUTES_PER_DAY, BcolzMinuteBarReader
 from zipline.data.us_equity_pricing import BcolzDailyBarReader
 from zipline.finance.trading import TradingEnvironment, SimulationParameters
+from zipline.protocol import BarData
 from zipline.utils.test_utils import write_minute_data_for_asset, \
     create_daily_df_for_asset, DailyBarWriterFromDataFrames
 from zipline.zipline_warnings import ZiplineDeprecationWarning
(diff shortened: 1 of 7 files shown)


tests/test_fetcher.py
```diff
@@ -223,47 +223,13 @@ def initialize(context):
...
-
-    assert "signal" not in data[sid(24)]
+    record(ibm_signal=data.current(sid(3766), "signal"))
+    record(dell_signal=data.current(sid(25317), "signal"))
     """)
 
         self.assertEqual(5, results["ibm_signal"].iloc[-1])
         self.assertEqual(5, results["dell_signal"].iloc[-1])
 
-    def test_fetch_csv_with_nomatch_symbol(self):
```
abhijeetkalyan: I killed this test because with our ffilling behaviour, the test isn't accurate anymore.

tests/test_fetcher.py
```diff
@@ -364,9 +330,15 @@ def initialize(context):
 
 def handle_data(context, data):
     if {should_have_data}:
-        assert 'cpi' in data['urban']
+        try:
```
abhijeetkalyan: Although this isn't behaviour we want to really support, I added this workaround here because it seems like we still want to test this underlying behaviour of fetcher here, despite the changes in the data API.

tests/test_api_shim.py
```diff
@@ -149,6 +151,55 @@ def create_algo(cls, code, filename=None, sim_params=None):
...
+            with patch(patch_meth) as gsv:
+                fun()
+                gsv.assert_called_with(
+                    self.asset1,
+                    field,
+                    test_minute,
+                    'minute'
+                )
+
+        for field in ohlcvp_fields:
```
jbredeche: this is great. can you do the same for `history`?

abhijeetkalyan:
Updated to also test `history` vs `data.history`, in addition to the ohlcvp fields.

PR #1041, opened 2016-03-09
Fix buyback auth tests in pipeline

mtydykov:
This PR fixes the way buyback auth previous dates/values are computed and the way it is tested. Previously, the computations and tests were the same as for earnings calendar, but it doesn't make sense for them to be the same because while earnings calendar announcement dates are for the future, buyback auth announcement dates are always for some number ( > 0) of days before the current day. This means that for calculating previously known dates/values, the timestamp (knowledge date) should be used, rather than the event date, because we won't actually know about the event on the same day that it happens. This also means that most of the test cases for earnings calendar don't apply to buyback auth.

### DIFF
--- a/zipline/pipeline/loaders/utils.py
+++ b/zipline/pipeline/loaders/utils.py
@@ -5,7 +5,7 @@ import pandas as pd
 from six import iteritems
 from six.moves import zip
 
-from zipline.utils.numpy_utils import NaTns, NaTD
+from zipline.utils.numpy_utils import NaTns
 
 
 def next_date_frame(dates, events_by_sid, event_date_field_name):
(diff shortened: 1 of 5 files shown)


zipline/pipeline/loaders/utils.py
```diff
@@ -133,18 +102,25 @@ def previous_value(date_index, events_by_sid, event_date_field, value_field,
...
         df = events_by_sid[sid]
         df = df[df[event_date_field] <= d_n]
+        event_date_vals = df[event_date_field].values
+        # Get knowledge dates corresponding to the values in which we are
+        # interested
+        kd_vals = df[df[event_date_field] <= d_n].index.values
+        # The date at which a previous event is first known is the max of the
+        #  kd and the event date.
+        index_dates = np.array([max([kd, val])
+                                for kd, val in zip(kd_vals, event_date_vals)])
```
llllllllll: This can be done vectorized in numpy with:
```python
index_dates = np.maximum(kd_vals, event_date_vals)
```

I don't think this is quite correct though, if we have a timeline like:

```
<---|---|---|--->
    t0  t1  t3
```
O `t0` there is a buyback authorization which we learn about on `t1`. On `t3` we ask "what is the date of the most recent buyback?". This should still be `t0`; however, on `t1` when we ask this we should get `NaT`. The previous event loader was not built with adjustments in mind because this is not a problem with earnings / dividends where you know about them in advance.

zipline/pipeline/loaders/utils.py
```diff
@@ -60,70 +60,39 @@ def next_date_frame(dates, events_by_sid, event_date_field_name):
     return pd.DataFrame(index=dates, data=cols)
 
 
-def previous_date_frame(date_index, events_by_sid, event_date_field_name):
+def get_previous_date_or_value(events_by_sid, date_index, missing_value,
```
mtydykov: Then we should probably also change `next_date_frame` to be `next_event_frame`, no? And change the functionality of that method to allow handling values that are not dates.

zipline/pipeline/loaders/utils.py
mtydykov: What's not correct? The logic that I have now? Not seeing why it's wrong.

mtydykov: What you described sounds to me like what we want to have happen, right?

llllllllll: I described what we want, but that is not what will happen, on t3, the date we report will be `t1`, also, just realized that I failed to properly count 3 elements and did 0, 1, 3 for no reason.

PR #1042, opened 2016-03-09
TST: move test_utils and adds test fixture classes

llllllllll:
Adds zipline.testing.fixtures.ZiplineTestCase to manage setup and
teardown and adds mixins to define fixtures like an asset finder or
trading calendar.


ssanderson:
LGTM

PR #1043, opened 2016-03-10
MAINT: Adds get_ctable method to BcolzMinuteBarReader

yankees714:
For convenience in accessing the raw ctables.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -552,6 +552,23 @@ class BcolzMinuteBarReader(object):
 
         return carray
 
+    def get_ctable(self, sid):
+        """Get the ctable of minute bars for a given sid.
+
+        Returns
+        -------
+        ctable or None
+            The ctable of minute bars for this sid, or None if no ctable
+            exists.
+        """
+
+        ctable_path = os.path.join(self._rootdir, _sid_subdir_path(sid))
+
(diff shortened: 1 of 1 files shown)


zipline/data/minute_bars.py
```diff
@@ -552,6 +552,10 @@ def _open_minute_file(self, field, sid):
 
         return carray
 
+    def get_ctable(self, sid):
+        ctable_path = os.path.join(self._rootdir, _sid_subdir_path(sid))
```
ssanderson: I think this is `self.sidpath`?

ssanderson: This looks like it's replicating the behavior of `_ensure_ctable`, except without the guaranteed initialization.  When would we want to call this without guaranteeing that the table already exists?

yankees714: Ahh yeah missed `self.sidpath` somehow, that should be enough for my use case.

yankees714: @ssanderson Realized the confusion here, the methods you mentioned above of the BcolzMinuteBarWriter, whereas I was adding `get_ctable` to BcolzMinuteBarReader, for consumers of the minute bars. But they probably should share this path creation logic.

yankees714:
This is superseded by `BcolzMinuteBarReader.unadjusted_window`  (#1050) .

PR #1044, opened 2016-03-10
FIX: Rollback granularity on fetcher sources to daily

lianga888:
For every fetcher source, take the last signal on a given day for
a given asset or data source as the signal for the entire day

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -141,6 +141,9 @@ class DataPortal(object):
 
         self._in_bts = False
 
+    def _reindex_extra_source(self, df, source_date_index):
+        return df.reindex(index=source_date_index, method='ffill')
+
     def handle_extra_source(self, source_df, sim_params):
         """
         Extra sources always have a sid column.
@@ -151,7 +154,8 @@ class DataPortal(object):
         if source_df is None:
             return
 
-        self._extra_source_df = source_df
+        # Normalize all the dates in the df
+        source_df.index = source_df.index.normalize()
 
         # source_df's sid column can either consist of assets we know about
         # (such as sid(24)) or of assets we don't know about (such as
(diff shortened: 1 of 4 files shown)


zipline/data/data_portal.py
```diff
@@ -190,18 +187,21 @@ def handle_extra_source(self, source_df, sim_params):
         for group_name in group_names:
             group_dict[group_name] = grouped_by_sid.get_group(group_name)
 
+        extra_source_df = None
```
jbredeche: can you just directly initialize `extra_source_df = pd.DataFrame()` here, then you don't need the None check below?

zipline/data/data_portal.py
```diff
@@ -1360,24 +1363,13 @@ def get_fetcher_assets(self, day):
...
-        asset_list = self._extra_source_df.loc[date_to_use]["sid"]
+        day = normalize_date(dt)
 
-        # make sure they're actually assets
-        asset_list = [asset for asset in asset_list
-                      if isinstance(asset, Asset)]
-
-        return asset_list
+        assets = self._extra_source_df.loc[day]['sid']
+        if isinstance(assets, pd.Series):
```
jbredeche: when would they not be `pd.Series`?

lianga888: I'm not that familiar with pandas, but I noticed that if `self._extra_source_df` only has 1 row with the indexed `day`, then `self._extra_source_df.loc[day]` returns a Series and then `self._extra_source_df.loc[day]['sid']` returns a scalar. Only if there are multiple rows with that `day` does `self._extra_source_df.loc[day]['sid']` return a Series

zipline/data/data_portal.py
lianga888: yeah, that makes sense. Also, added a check in `get_fetcher_assets` to return `[]` when the df is empty

zipline/data/data_portal.py
```diff
@@ -139,6 +139,9 @@ def __init__(self,
         }
         self._equity_daily_reader_array_data = {}
 
+    def _reindex_extra_source(self, df, source_date_index):
```
jbredeche: why did you move this to be a separate method?  (it's not wrong, I'm just curious).  it took me a few reads to realize that it's not a functional change.

zipline/data/data_portal.py
```diff
@@ -190,18 +187,21 @@ def handle_extra_source(self, source_df, sim_params):
         for group_name in group_names:
             group_dict[group_name] = grouped_by_sid.get_group(group_name)
 
+        extra_source_df = pd.DataFrame()
```
jbredeche: can you add a comment explaining what `extra_source_df` is here?  looks like a dataframe to which we're appending other dataframes below?

zipline/data/data_portal.py
lianga888: It gets overridden in live_data_portal, which does not forward fill https://github.com/quantopian/qexec/pull/9440

PR #1045, opened 2016-03-10
ENH: Better warning message on partial fills

abhijeetkalyan:
Even went with 'canceled' instead of 'cancelled' :us: 

@lianga888 can you take a look?

### DIFF
--- a/zipline/finance/blotter.py
+++ b/zipline/finance/blotter.py
@@ -166,15 +166,31 @@ class Blotter(object):
                     self.cancel(order.id, relay_status=False)
 
                     if warn:
-                        warning_logger.warn(
-                            'Your order for %s shares of %s has been '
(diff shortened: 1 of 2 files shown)


lianga888:
lgtm :)

zipline/finance/blotter.py
```diff
@@ -166,15 +166,31 @@ def execute_cancel_policy(self, event):
...
+                                    order_sym=order.sid.symbol,
+                                    order_filled=order.filled,
+                                    order_failed=order.amount - order.filled,
+                                )
+                            )
+                        else:
+                            warning_logger.warn(
+                                'Your order for {order_amt} shares of '
+                                '{order_sym} failed to fill by the end of day '
+                                'and was canceled'.format(
```
jbredeche: missing period after `canceled`

PR #1046, opened 2016-03-10
Classifiers

ssanderson:


zipline/pipeline/factors/factor.py
```diff
@@ -5,20 +5,22 @@
 from operator import attrgetter
 from numbers import Number
 
-from numpy import inf
+from numpy import inf, where, nanmean, nanstd
```
richafrank: Should we use `nanmean` from our math_utils instead?  It uses bottleneck if it's available.

ssanderson: yeah, will do

zipline/pipeline/classifiers/latest.py
```diff
@@ -0,0 +1,29 @@
...
+    def _validate(self):
+        if self.inputs[0].dtype != int64_dtype:
+            raise TypeError(
+                "{name} expected an input of dtype int64, "
+                "but got {not_bool} instead.".format(
+                    name=type(self).__name__,
+                    not_bool=self.inputs[0].dtype,
+                )
+            )
+        super(Latest, self)._validate()
```
richafrank: I noticed that `Classifier._validate` makes sure to return `super()._validate()`.  Do we need to here too?

zipline/lib/normalize.py
```diff
@@ -0,0 +1,44 @@
...
+    >>> naive_grouped_rowwise_apply(data, labels, lambda row: row - row.min())
+    array([[0., 1., 0.],
+           [0., 0., 2.],
+           [0., 0., 0.]])
+    >>> naive_grouped_rowwise_apply(data, labels, lambda row: row / row.sum())
+    array([[ 0.33333333,  0.66666667,  1.        ],
+           [ 0.33333333,  1.        ,  0.66666667],
+           [ 1.        ,  1.        ,  1.        ]])
+    """
+    out = np.empty_like(data)
```
dmichalowicz: Should this be

```python
if out is not None:
    out = np.empty_like(data)
```

tests/pipeline/test_factor.py
```diff
@@ -43,6 +48,20 @@ class F(Factor):
...
+    missing_value = -1
+    inputs = ()
+    window_length = 0
+
+
+class OtherC(Classifier):
+    dtype = int64_dtype
+    missing_value = -1
+    inputs = ()
+    window_length = 0
```
dmichalowicz: What's the purpose of having this copy?

zipline/pipeline/classifiers/latest.py
ssanderson: Nothing uses these values right now, and all the existing classifiers just return None, but we should probably be consistent.

zipline/lib/normalize.py
ssanderson: Yep.

tests/pipeline/test_factor.py
ssanderson: It's used in `test_normalizations` below for a case where I needed two different classifiers.

dmichalowicz:
@ssanderson Looks good by me. I worked through testing `demean` and `zscore` with different masks and groupby's and they worked well.

ssanderson:
@dmichalowicz thanks!  I'll merge this once I finish making tweaks to the docstrings.

PR #1047, opened 2016-03-10
BUG: Don't mutate open orders list as we cancel

lianga888:
In execute_cancel_policy, make a copy of the orders list, so it
doesn't get mutated as we cancel orders

### DIFF
--- a/zipline/finance/blotter.py
+++ b/zipline/finance/blotter.py
@@ -16,9 +16,10 @@ import math
 
 from logbook import Logger
 from collections import defaultdict
+from copy import copy
 
 import pandas as pd
-from six import iteritems, itervalues
+from six import iteritems
 
 from zipline.finance.order import Order
 
@@ -140,7 +141,8 @@ class Blotter(object):
                 # along with newly placed orders.
                 self.new_orders.append(cur_order)
 
-    def cancel_all_orders_for_asset(self, asset):
+    def cancel_all_orders_for_asset(self, asset, warn=False,
+                                    relay_status=True):
         """
         Cancel all open orders for a given asset.
         """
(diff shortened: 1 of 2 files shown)


zipline/finance/blotter.py
```diff
@@ -162,7 +162,8 @@ def execute_cancel_policy(self, event):
             warn = self.cancel_policy.warn_on_cancel
 
             for orders_by_sid in itervalues(self.open_orders):
-                for order in orders_by_sid:
+                # So we don't mutate the list as we cancel orders
+                for order in orders_by_sid[:]:
```
ssanderson: The function directly above this is `cancel_all_orders_for_asset`, which handles this correctly (and documents that this is an issue.)  Is there a reason we're not just using that here?

lianga888: Doesn't directly fit because of the warning messages, but if you think it makes more sense, I could refactor the code to make it use that method

ssanderson: I would make `cancel_all` take a `warn` parameter and then move the warning logic here to that method.

ssanderson: So the updated logic here would look something like:

```
warn = self.cancel_policy.warn_on_cancel
for sid in self.open_orders:
    self.cancel_all(sid, warn=warn)
```

ssanderson: Also, is there a reason we renamed that method to `cancel_all_orders_for_asset`?  That name makes it look like the method should accent an Asset object, when it actually accepts a `sid`.

lianga888: The open orders dict is now keyed by Asset in Q2 I believe

ssanderson: https://github.com/quantopian/zipline/pull/1047/files#diff-9dc7cb37547c5feca2c0771447ce1a63R147 needs to be updated then.

lianga888: yep, a lot of those tests

PR #1048, opened 2016-03-11
Lazy but tidy

richafrank:
Cleaned up some code no longer needed by lazy-mainline.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -1262,7 +1262,7 @@ class DataPortal(object):
 
         Parameters
         ----------
-        sids : list
+        sids : container
             Sids for which we want splits.
 
         dt: pd.Timestamp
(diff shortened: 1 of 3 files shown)


zipline/data/data_portal.py
```diff
@@ -1284,8 +1284,7 @@ def get_splits(self, sids, dt):
             "SELECT sid, ratio FROM SPLITS WHERE effective_date = ?",
             (seconds,)).fetchall()
 
-        sids_set = set(sids)
-        splits = [split for split in splits if split[0] in sids_set]
+        splits = [split for split in splits if split[0] in sids]
```
jbredeche: the `sids` container can be on the order of 500+ sids, so that's why I used a set.  Think we should go back to that?

richafrank: Looked like `sids` is already a set?

jbredeche: indeed.  I'll see myself out.

PR #1050, opened 2016-03-14
ENH: Unadjusted window data for minute bars.

ehebert:
Add a method to minute bar reader which returns the OHLCV for all
requested fields for a list assets over the specified start and end
minutes.

Initial usage is intended for use by a loader which consumes minute bar
data to resample into daily bars, but may also be used when aggregating
minute data during '1d' history calls in Q2.0.

This iteration does not include including of early closes.

### DIFF
--- a/tests/data/test_minute_bars.py
+++ b/tests/data/test_minute_bars.py
@@ -23,7 +23,8 @@ from pandas import (
     DataFrame,
     DatetimeIndex,
     Timestamp,
-    NaT
+    Timedelta,
+    NaT,
 )
 from testfixtures import TempDirectory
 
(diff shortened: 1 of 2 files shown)


ehebert:
Just realized that this does not handle `nan`s properly, going to add a test case and fix that ASAP.

ehebert:
With https://github.com/quantopian/zipline/commit/56f22c34a4363db000f2ba55f6720165c23c7ad8 the aforementioned `nan` case should now be handled.

ehebert:
Changed name to `unadjusted_window`, and on closer inspection that new line was fine. Merging.

PR #1051, opened 2016-03-14
DEV: Make all the data.* methods work for non-market-minutes by using the last market close

jbredeche:
Give responsibility of altering datetimes on the fly to `BarData`, so that `DataPortal` is unsullied.


freddiev4:
Uh... why is coveralls commenting here after 2 years?

PR #1052, opened 2016-03-14
BUG: Empty pipeline failed to tz_localize

dmichalowicz:
`tz_localize` was failing on a dataframe with an empty MultiIndex.

### DIFF
--- a/tests/pipeline/test_pipeline_algo.py
+++ b/tests/pipeline/test_pipeline_algo.py
@@ -566,3 +566,43 @@ class PipelineAlgorithmTestCase(TestCase):
             # TradingAlgorithm.
             overwrite_sim_params=False,
         )
+
+    def test_empty_pipeline(self):
+
+        # For ensuring we call before_trading_start.
+        count = [0]
+
+        def initialize(context):
+            pipeline = attach_pipeline(Pipeline(), 'test')
+
+            vwap = VWAP(window_length=10)
+            pipeline.add(vwap, 'vwap')
+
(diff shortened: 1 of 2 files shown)


zipline/pipeline/engine.py
```diff
@@ -384,6 +384,20 @@ def _to_narrow(self, data, mask, dates, assets):
         If mask[date, asset] is True, then result.loc[(date, asset), colname]
         will contain the value of data[colname][date, asset].
         """
+        if not mask.any():
+            # Manually handle the empty DataFrame case. This is a workaround
```
ssanderson: Might be worth noting that this also saves us the work of applying a known-empty mask to the array.

tests/pipeline/test_pipeline_algo.py
```diff
@@ -566,3 +566,36 @@ def handle_data(context, data):
...
+            pipeline.add(vwap, 'vwap')
+
+            # Nothing should have prices less than 0.
+            pipeline.set_screen(vwap < 0)
+
+        def handle_data(context, data):
+            pass
+
+        def before_trading_start(context, data):
+            context.results = pipeline_output('test')
```
ssanderson: Can we assert that we get an empty DataFrame here?

ssanderson: Should probably also assert that BTS actually called here.

dmichalowicz:
Got the okay from @ssanderson

PR #1053, opened 2016-03-14
FIX: Remove serialization of zipline objects

lianga888:
- Removes the `__getstate__` and `__setstate__` methods, used for pickle serialization, from zipline objects
- Removes relevant tests


richafrank:
Do we still need https://github.com/quantopian/zipline/blob/73592a8b310faa81044234c29fac8cecf93fdfe9/zipline/utils/serialization_utils.py ?

lianga888:
Yea, it's still used in qexec in various places. Something to do with the asset finder across the simulation/deluge pipe according to @jdricklefs

jdricklefs:
We pickle the AssetFinder in some instances. That code could move out of this repository, though.

PR #1054, opened 2016-03-14
WithTradingEnvironmnet and WithSimParams

llllllllll:


zipline/testing/fixtures.py
```diff
@@ -108,30 +115,136 @@ def enter_instance_context(self, context_manager):
...
-    with an AssetFinder. The default finder just a `tmp_asset_finder`
+    with an AssetFinder. The default finder just a `tmp_asset_finder` whose
+    arguments come from:
+
+      equities=cls.make_equities_info(),
+      futures=cls.make_futures_info(),
+      exchanges=cls.make_exchanges_info(),
+      root_symbols=cls.make_root_symbols_info(),
+
+    Each of these fields may be overridden with a dataframe to write.
```
ssanderson: "with a function that returns a DataFrame to write"?

zipline/testing/fixtures.py
```diff
@@ -108,30 +115,136 @@ def enter_instance_context(self, context_manager):
...
+    """ZiplineTestCase mixing providing cls.log_handler as an instance-level
+    fixture.
+
+    after init_instance_fixtures has been called `self.log_handler` will be a
+    new `FileHandler` which has been pushed with the value of
+    `self.LOGGER_PATH`
+    """
+    LOGGER_PATH = 'test.log'
+
+    def init_instance_fixtures(self):
```
ssanderson: Is there a reason we're making a new handler and entering it for every test method?

zipline/testing/fixtures.py
```diff
@@ -108,30 +115,136 @@ def enter_instance_context(self, context_manager):
...
 
+class WithLogger(object):
+    """ZiplineTestCase mixing providing cls.log_handler as an instance-level
+    fixture.
+
+    after init_instance_fixtures has been called `self.log_handler` will be a
+    new `FileHandler` which has been pushed with the value of
+    `self.LOGGER_PATH`
+    """
+    LOGGER_PATH = 'test.log'
```
ssanderson: This should probably be a function that returns something in a temp directory.

zipline/testing/fixtures.py
```diff
@@ -108,30 +115,136 @@ def enter_instance_context(self, context_manager):
...
+
+    def init_instance_fixtures(self):
+        super(WithLogger, self).init_instance_fixtures()
+
+        self.log_handler = log_handler = FileHandler(self.LOGGER_PATH)
```
ssanderson: handler.applicationbound is a context manager.

zipline/testing/fixtures.py
```diff
@@ -108,30 +115,136 @@ def enter_instance_context(self, context_manager):
...
+    SIM_PARAMS_END = None
+    SIM_PARAMS_CAPITAL_BASE = float("1.0e5")
+    SIM_PARAMS_NUM_DAYS = None
+    SIM_PARAMS_DATA_FREQUENCY = 'daily'
+    SIM_PARAMS_EMISSION_RATE = 'daily'
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(WithSimParams, cls).init_class_fixtures()
+        cls.sim_params = factory.create_simulation_parameters(
```
ssanderson: `create_simulation_parameters` is just a wrapper around `SimulationParameters` with logic for defaults.  Given that we're already defining logic for default here, I'd say we should just directly invoke `SimulationParameters`.

llllllllll: Do you think I should duplicate the logic in this function here? The function provides different checks based on passing a num_days or start/end. I think it is probably best to defer to this function to make this easier to use.

ssanderson: I could see the argument to leaving create_simulation_parameters, but I'd put a concrete default here so it's clear what you're getting as a consumer of this.

Also, we don't pass `year` anywhere to create_simulation_parameters as far as I can tell, so I think we can just remove that.

ssanderson:
LGTM

zipline/testing/fixtures.py
llllllllll: updated to pass the start and end explicitly

PR #1055, opened 2016-03-15
ENH: Speedups by moving computation-heavy things to Cython

jbredeche:
Moving a bunch of arithmetic-heavy work to Cython takes this algorithm, on my machine (in qprof) from ~46 seconds to ~29 seconds:

```
def initialize(context):
    pass

	data.current(sid(24), "high")

	data.current(sid(24), "price")
```

(2012-01-04 to 2014-01-03, 1M, minutely)

the difference is much more pronounced with even more illiquid securities.  I'll have more data tomorrow.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -20,8 +20,13 @@ from os.path import join
 import json
 import os
 import pandas as pd
+from zipline.gens.sim_engine import NANOS_IN_MINUTE
 
-from zipline.utils.memoize import remember_last
+from zipline.data._minute_bar_internal import (
+    minute_value,
+    find_position_of_minute,
+    find_last_traded_position_internal
+)
 
 US_EQUITIES_MINUTES_PER_DAY = 390
 
@@ -519,7 +524,11 @@ class BcolzMinuteBarReader(object):
         metadata = self._get_metadata()
 
         self._first_trading_day = metadata.first_trading_day
-        self._minute_index = metadata.minute_index
+
+        # store a list of "minute epoch" values
+        self._minute_value_index = \
+            metadata.minute_index.astype('datetime64[m]').astype(int)
+
         self._ohlc_inverse = 1.0 / metadata.ohlc_ratio
 
         self._carrays = {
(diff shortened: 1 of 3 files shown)


zipline/data/_array_utilities.pyx
```diff
@@ -0,0 +1,52 @@
...
+
+from bcolz.carray_ext cimport carray
+
+cdef inline int int_min(int a, int b): return a if a <= b else b
+
+@cython.cdivision(True)
+def minute_value(np.ndarray[np.long_t, ndim=1] arr, Py_ssize_t pos):
+    cdef short q, r
+
+    q = cython.cdiv(pos, 390)
```
ehebert: With this being a function not associated with the `BcolzMinuteBarClass`, it may make sense to pass the `minutes_per_day` as a parameter, instead of hardcoding to 390.

zipline/data/_array_utilities.pyx
```diff
@@ -0,0 +1,52 @@
...
+def find_position_of_minute(np.ndarray[np.long_t, ndim=1] arr,
+                             np.long_t minute_val):
+    cdef Py_ssize_t market_open_loc, market_open, delta
+
+    market_open_loc = np.searchsorted(arr, minute_val, side='right') - 1
+    market_open = arr[market_open_loc]
+
+    delta = minute_val - market_open
+
+    return (market_open_loc * 390) + delta
```
ehebert: Likewise here with the 390 minutes_per_day.

zipline/data/minute_bars.py
```diff
@@ -520,6 +525,10 @@ def __init__(self, rootdir):
 
         self._first_trading_day = metadata.first_trading_day
         self._minute_index = metadata.minute_index
+
+        self._minute_value_index = \
```
ehebert: I would recommend this instead for clarity, `self._minute_index.values.astype('datetime64[m]).astype(int)`.

zipline/data/minute_bars.py
```diff
@@ -601,32 +610,23 @@ def get_last_traded_dt(self, asset, dt):
 
     def _find_last_traded_position(self, asset, dt):
         volumes = self._open_minute_file('volume', asset)
-        start_date = asset.start_date
+        start_date_minutes = asset.start_date.value / NANOS_IN_MINUTE
+        dt_minutes = dt.value / NANOS_IN_MINUTE
```
ehebert: Not for this PR, but this does make me ask the question, should our `dt` be a `datetime64[m]` when in minute mode, to minimize times that conversions are needed?

zipline/data/_array_utilities.pyx
```diff
@@ -0,0 +1,52 @@
...
+def minute_value(np.ndarray[np.long_t, ndim=1] arr, Py_ssize_t pos):
+    cdef short q, r
+
+    q = cython.cdiv(pos, 390)
+    r = cython.cmod(pos, 390)
+
+    return arr[q] + r
+
+def find_position_of_minute(np.ndarray[np.long_t, ndim=1] arr,
+                             np.long_t minute_val):
```
ehebert: I believe the convention we have been following is to import all numpy functions.
In part because that allows us to be specific about which are cimports or not.

PR #1061, opened 2016-03-17
BUG: Fix '1d' history for last day in slot.

ehebert:
Fix the following behavior for the last slot in history with a '1d'
freq.

- open 
 - old
    Always used the open on the first minute, (9:31)
 - fix
    Uses first non-nan open.
- high
 - old
    The existence of any nan from market open to dt would propagate and
    make the result nan.
 - fix
    nans no longer affect high, unless there are only nans between
    market open and given dt
- low
 - old
    The existence of any nan from market open to dt would propagate and
    make the result nan.
 - fix
    nans no longer affect low, unless there are only nans between
    market open and given dt
- close
 - old
    The close at the given dt was always used. If the close was a nan at
    that dt, a nan would be returned.
 - fix
    The close at the given dt, unless that a value is a nan; in that
    case the most recent non-nan value is returned.

This patch's main intent is to fix the bug affecting results, however with
this fix there is a decent sized performance gain from cutting out the
calls to `all_minutes_for_day`, as well as cutting the retrieval of the
minute window from market_window -> current dt, which is order ~n^2 with
respect to how many minute values are accessed.

This implementation leaves some performance considerations on the table.
Particularly:
- The minute window access in each of the aggregation methods could be
fetched and stored in memory instead of going back to the bcolz tape on
every query.
- The minute windows could be structured so that they are accessed in
groups of assets. This implementation incurs the cost of getting the
minute position from the cache (and/or calculating) the number of bars
times the number of assets, when that retrieval could only be done once
per minute.

Also, adds test fixtures for bcolz minute data and the tempdir
dependency for the bcolz data, for use by the new test suite.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -59,6 +59,362 @@ OHLCVP_FIELDS = frozenset([
 HISTORY_FREQUENCIES = set(["1m", "1d"])
 
 
+class DailyHistoryAggregator(object):
+    """
+    Converts minute pricing data into a daily summary, to be used for the
+    last slot in a call to history with a frequency of `1d`.
+
+    This summary is the same as a daily bar rollup of minute data, with the
+    distinction that the summary is truncated to the `dt` requested.
+    i.e. the aggregation slides forward during a the course of simulation day.
+
+    Provides aggregation for `open`, `high`, `low`, `close`, and `volume`.
+    The aggregation rules for each price type is documented in their respective
+
+    """
+
+    def __init__(self, market_opens, minute_reader):
+        self._market_opens = market_opens
+        self._minute_reader = minute_reader
+
+        # The caches are structured as (date, market_open, entries), where
+        # entries is a dict of asset -> (last_visited_dt, value)
+        #
+        # Whenever an aggregation method determines the current value,
+        # the entry for the respective asset should be overwritten with a new
+        # entry for the current dt.value (int) and aggregation value.
+        #
+        # When the requested dt's date is different from date the cache is
+        # flushed, so that the cache entries do not grow unbounded.
+        #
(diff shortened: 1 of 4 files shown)


zipline/data/data_portal.py
```diff
@@ -59,6 +59,362 @@
 HISTORY_FREQUENCIES = set(["1m", "1d"])
 
 
+class DailyHistoryAggregator(object):
```
jbredeche: stylistic thing, can you move this into its own file?  when I open `data_portal.py`, I prefer immediately seeing the `DataPortal` code.

zipline/data/data_portal.py
```diff
@@ -59,6 +59,362 @@
...
+        Once the first non-nan open is seen, that value remains constannt per
+        asset for the remainder of the day.
+
+        Returns
+        -------
+        np.array with dtype=float64, in order of assets parameter.
+        """
+        date = dt.date()
+        dt_value = dt.value
+        if self._opens_cache is None or self._opens_cache[0] != date:
```
jbredeche: these ~5 lines are repeated in each of the methods, should we generalize into a helper method?

zipline/data/data_portal.py
ehebert: I see your point. On the other hand the `DailyHistoryAggregator` is a support class for the `DataPortal`.

As we discussed in person, that could mean that the name of the module itself could renamed to not imply that it is just data_portal.

Either way, going to defer/punt on that refactoring.

zipline/data/data_portal.py
ehebert: Added with ac5ec75667424b7cba0fcbabda47440b31e54937

PR #1062, opened 2016-03-17
Lazy prevent history before trading start

ehebert:
Unreverts #1040 

Also, make env.market_minutes lazy so that the memory and CPU cost is not hit in contexts that do not use the minutes.

### DIFF
--- a/zipline/errors.py
+++ b/zipline/errors.py
@@ -593,3 +593,11 @@ class AssetDBImpossibleDowngrade(ZiplineError):
         "The existing Asset database is version: {db_version} which is lower "
         "than the desired downgrade version: {desired_version}."
     )
+
+
+class HistoryWindowStartsBeforeData(ZiplineError):
+    msg = (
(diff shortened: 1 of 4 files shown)


zipline/errors.py
```diff
@@ -593,3 +593,11 @@ class AssetDBImpossibleDowngrade(ZiplineError):
         "The existing Asset database is version: {db_version} which is lower "
         "than the desired downgrade version: {desired_version}."
     )
+
+
+class HistoryWindowStartsBeforeData(ZiplineError):
+    msg = (
+        "History window extends beyond environment first date, "
+        "{first_trading_day}. To use history with bar count, {bar_count}, "
+        "start simulation on or after, {suggested_start_day}."
```
jbredeche: Can you tweak to:

```
"History window extends before {first_trading_day}.  To use this history window, start the backtest on or after {suggested_start_day}."
```

ehebert: Done.

PR #1063, opened 2016-03-17
Dividends in pipeline

mtydykov:
Adds custom loader and tests for dividends to Pipeline.


tests/pipeline/test_dividends.py
```diff
@@ -0,0 +1,496 @@
...
+            # announcements that happened on the same day for the same sid.
+            3: get_values_for_date_ranges(zip_date_index_with_vals,
+                                          vals[3],
+                                          date_invervals[3],
+                                          dates),
+            4: zip_date_index_with_vals(dates, ['NaN'] * len(dates)),
+        }, index=dates)
+
+
+class DividendsByAnnouncementDateTestCase(TestCase, EventLoaderCommonMixin):
```
mtydykov: Should these tests now be using fixtures? e.g., for the `tmp_asset_finder`? Should we create a new fixture in zipline for databazaar (maybe it can replace much of `EventLoaderCommonMixin`, or replace it entirely)?

zipline/pipeline/loaders/utils.py
```diff
@@ -36,28 +44,35 @@ def next_date_frame(dates, events_by_sid, event_date_field_name):
...
+        for knowledge_date, event_date, value in iterkv:
             date_mask = (
                 (knowledge_date <= raw_dates) &
                 (raw_dates <= event_date)
             )
             value_mask = (event_date <= data) | (data == NaTns)
-            data[date_mask & value_mask] = event_date
-
-    return pd.DataFrame(index=dates, data=cols)
+            data_indeces = np.where(date_mask & value_mask)
```
llllllllll: indices

zipline/pipeline/loaders/utils.py
```diff
@@ -260,3 +275,39 @@ def check_data_query_args(data_query_time, data_query_tz):
...
+    return pd.Series(pd.to_datetime(dts), index=index_dates)
+
+
+def get_values_for_date_ranges(zip_date_index_with_vals,
+                               vals_for_date_intervals,
+                               date_intervals,
+                               date_index):
+    """
+    Returns a
+    :param zip_vals_for_date_index:
```
llllllllll: numpy style instead of default sphinx

zipline/pipeline/loaders/utils.py
```diff
@@ -260,3 +275,39 @@ def check_data_query_args(data_query_time, data_query_tz):
...
+    :param num_days_between_dates:
+    :param vals_for_date_intervals:
+    :param date_intervals:
+    :return:
+    """
+    # Fill in given values for given date ranges.
+    return zip_date_index_with_vals(
+        date_index,
+        list(
+            itertools.chain(*[
```
llllllllll: instead of chain* we can use chain.from_iterable or concat. No need to wrap the genexpr in `[]` which makes it a list comprehension

tests/pipeline/base.py
```diff
@@ -161,15 +160,6 @@ def get_equity_info(cls):
             end_date=pd.Timestamp('2015-01-01', tz='UTC'),
         )
 
-    def zip_with_floats(self, dates, flts):
```
llllllllll: was this just dead code? do we need to replace this with the stuff you added in utils.py?

tests/pipeline/test_buyback_auth.py
```diff
@@ -39,6 +37,8 @@
     BlazeCashBuybackAuthorizationsLoader,
     BlazeShareBuybackAuthorizationsLoader,
 )
+from zipline.pipeline.loaders.utils import get_values_for_date_ranges, \
```
llllllllll: please use parens to wrap imports that don't fit on a single line.

tests/pipeline/test_dividends.py
```diff
@@ -0,0 +1,496 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls._cleanup_stack = stack = ExitStack()
+        cls.cols = {}
+        cls.dataset = {sid:
+                       frame.drop([EX_DATE_FIELD_NAME,
+                                   PAY_DATE_FIELD_NAME], axis=1)
+                       for sid, frame
+                       in enumerate(dividends_cases)}
+        cls.finder = stack.enter_context(
```
llllllllll: WithTmpAssetFinder

zipline/pipeline/loaders/utils.py
mtydykov: Wow, I also did not finish that docstring. Good catch!

zipline/pipeline/loaders/utils.py
```diff
@@ -260,3 +275,39 @@ def check_data_query_args(data_query_time, data_query_tz):
                 data_query_tz,
             ),
         )
+
+
+def zip_with_floats(dates, flts):
```
ssanderson: You can pass `dtype` to Series directly.

zipline/pipeline/loaders/utils.py
ssanderson: All of this seems like it wants to be using array primitives rather than lazy streaming values.

mtydykov: @ssanderson I'm not sure; did you mean using `np.array`?

tests/pipeline/test_dividends.py
mtydykov: We also need to change `test_buyback_auth.py` and `test_earnings.py` to use that, right?

zipline/pipeline/loaders/utils.py
ssanderson: @mtydykov what I mean is that this is taking a bunch of numpy-ish datastructures (meaning, either numpy or pandas objects) and producing a numpy-ish datastructure, but all of the data processing here is being done on sequences of python objects.  That generally means we're losing most or all of the performance benefits of working with numpy objects in the first place.

zipline/pipeline/loaders/utils.py
mtydykov: Done (on diff line)

zipline/pipeline/loaders/utils.py
```diff
@@ -260,3 +274,53 @@ def check_data_query_args(data_query_time, data_query_tz):
...
+    Returns
+    -------
+    date_index_with_vals : pd.Series
+        A Series indexed by the given DatetimeIndex and with values assigned
+        to dates based on the given date intervals.
+    """
+    # Fill in given values for given date ranges.
+    return zip_date_index_with_vals(
+        date_index,
+        np.repeat(vals_for_date_intervals,
```
mtydykov: @ssanderson I've changed this a bit to make the code clearer, but I'm not sure what to do from here to get the performance benefits you mentioned.

mtydykov:
@llllllllll dividends PR ready for another pass

zipline/pipeline/loaders/utils.py
```diff
@@ -260,3 +274,53 @@ def check_data_query_args(data_query_time, data_query_tz):
                 data_query_tz,
             ),
         )
+
+
+def zip_with_floats(dates, flts):
+        return pd.Series(flts, index=dates, dtype='float')
+
+
+def num_days_between(dates, start_date, end_date):
```
ssanderson: I don't think this function does anything?

zipline/pipeline/loaders/utils.py
ssanderson: Big :+1: on the usage of `repeat` here.

If you look at the implementation of `num_days_in_range` (which is equivalent to `num_days_between`, see comment above), you'll see that all it's doing is a sorted lookup into the DatetimeIndex with the `start, end` pair you're providing, and then subtracting the difference.  So the this whole expression is essentially doing something like:
```
In [31]: dates = pd.date_range('1990', '1990-01-15', freq='2D')
In [32]: dates
Out[32]:
DatetimeIndex(['1990-01-01', '1990-01-03', '1990-01-05', '1990-01-07',
               '1990-01-09', '1990-01-11', '1990-01-13', '1990-01-15'],
              dtype='datetime64[ns]', freq='2D', tz=None)

In [33]: values = array([1, 2, 3])

In [34]: bounds = [(Timestamp('1990-01-02'), Timestamp('1990-01-06')),
                  (Timestamp('1990-01-03'), Timestamp('1990-01-08')),
                  (Timestamp('1990-01-11'), Timestamp('1990-01-14'))]

In [35]: deltas = [data.searchsorted(end) - data.searchsorted(start) for start, end in bounds]

In [36]: deltas
Out[36]: [2, 3, 2]

In [37]: np.repeat(values, deltas)
Out[37]: array([1, 1, 2, 2, 2, 3, 3])
```

The only thing here that's still happening iteratively in Python is the slicing to determine the number of repeats for each value you have.  You could make that vectorized as well by passing all the starts at once to `searchsorted`, then passing all the ends, and then doing a vectorized subtraction.  With the input types you have right now, this would look something like:

```
dates.values.searchsorted([b[1].asm8 for b in bounds]) - dates.values.searchsorted([b[0].asm8 for b in bounds])
```
which probably isn't much faster (in fact it would probably be slower until the inputs got very large), because we have to coerce the datetimeindex to an array and coerce all the Timestamps to arrays of `datetime64[ns]` to do the broadcasted lookups.  If this function turns out to be a bottleneck, we'd probably want to change the input formats for the bounds from a list of pairs to two parallel arrays of `(start_date, end_date)`.  Unless it's very straightforward to do so, that change probably isn't worth making without more profiling.

zipline/testing/fixtures.py
```diff
@@ -177,7 +177,14 @@ class WithAssetFinder(object):
     def _make_info(cls):
         return None
 
-    make_equities_info = _make_info
+    @classmethod
+    def make_equities_info(cls):
+        return make_simple_equity_info(
+            cls.get_sids(),
```
ssanderson: Where is `get_sids` coming from here?

mtydykov: It's coming from `EventLoaderCommonMixin` in `base.py`, which is probably not the best way to organize this. A better way to organize this might be to have `EventLoaderCommonMixin` be turned into a fixture which subclasses from `WithAssetFinder` and to have `make_equities_info` be defined there instead.

ssanderson: > A better way to organize this might be to have EventLoaderCommonMixin be turned into a fixture which subclasses from WithAssetFinder and to have make_equities_info be defined there instead.

:+1: on this re-org.  As-is this will break in any other consumers of this fixture.

mtydykov:
Ready for another pass. Of note: `EventLoaderCommonMixin` moved to `fixtures.py` and changed to fixture `WithZiplineDataBazaar`.

zipline/pipeline/data/dividends.py
```diff
@@ -0,0 +1,25 @@
+"""
+Dataset representing dates of upcoming dividends.
+"""
+from zipline.utils.numpy_utils import datetime64ns_dtype, float64_dtype
+
+from .dataset import Column, DataSet
+
+
+class DividendsByExDate(DataSet):
+    next_ex_date = Column(datetime64ns_dtype)
```
llllllllll: should this just be `(next|previous)_date`, we know it is ex_date from the dataset

zipline/pipeline/data/dividends.py
```diff
@@ -0,0 +1,25 @@
...
+
+class DividendsByPayDate(DataSet):
+    next_pay_date = Column(datetime64ns_dtype)
+    previous_pay_date = Column(datetime64ns_dtype)
+    next_amount = Column(float64_dtype)
+    previous_amount = Column(float64_dtype)
+
+
+class DividendsByAnnouncementDate(DataSet):
+    previous_announcement_date = Column(datetime64ns_dtype)
```
llllllllll: why no next here?

zipline/testing/fixtures.py
```diff
@@ -292,3 +299,126 @@ def init_class_fixtures(cls):
...
+    """
+    ZiplineTestCase mixin providing common test methods/behaviors for event
+    data loaders.
+
+    `get_sids` must return the sids being tested.
+    `get_dataset` must return {sid -> pd.DataFrame}
+    `loader_type` must return the loader class to use for loading the dataset
+    `make_asset_finder` returns a default asset finder which can be overridden.
+    """
+    @abc.abstractproperty
```
llllllllll: this isn't an abc

llllllllll: these should probably be class methods with reasonable defaults which can be overridden by the user of the fixture.

zipline/testing/fixtures.py
```diff
@@ -292,3 +299,126 @@ def init_class_fixtures(cls):
...
+    def get_dataset(self):
+        raise NotImplementedError('get_dataset')
+
+    @abc.abstractproperty
+    def loader_type(self):
+        raise NotImplementedError('loader_type')
+
+    @classmethod
+    def make_asset_finder(cls):
```
llllllllll: the way to implement this pattern would be to override the class method `make_equities_info` and have the return value of `make_simple_equity_info(...)`; however, I am not sure if it should be this fixture's job to setup the equities info.

zipline/testing/fixtures.py
```diff
@@ -292,3 +299,126 @@ def init_class_fixtures(cls):
...
+            The dates we can serve.
+
+        Returns
+        -------
+        args : tuple[any]
+            The arguments to forward to the loader positionally.
+        """
+        return dates, self.get_dataset()
+
+    def setup_engine(self, dates):
```
llllllllll: we may want to namespace these methods like: pipeline_event_loader_setup_engine` because we expect to have potentially many fixtures that all provide data. This make the consumer better understand which fixture they are hooking.

zipline/pipeline/loaders/utils.py
```diff
@@ -36,28 +42,35 @@ def next_date_frame(dates, events_by_sid, event_date_field_name):
...
+        values = df[return_field_name]
+        data = date_cols[equity]
         if not event_dates.index.is_monotonic_increasing:
             event_dates = event_dates.sort_index()
 
         # Iterate over the raw Series values, since we're comparing against
         # numpy arrays anyway.
-        iterkv = zip(event_dates.index.values, event_dates.values)
-        for knowledge_date, event_date in iterkv:
+        iterkv = zip(event_dates.index.values, event_dates.values, values)
```
llllllllll: idk if we want to keep calling this iterkv because it is an iterator of three-tuples.

zipline/pipeline/loaders/blaze/dividends.py
```diff
@@ -0,0 +1,229 @@
...
+
+    concrete_loader = DividendsByPayDateLoader
+
+    def __init__(self,
+                 expr,
+                 resources=None,
+                 odo_kwargs=None,
+                 data_query_time=None,
+                 data_query_tz=None,
+                 dataset=DividendsByPayDate,
```
llllllllll: each of the classes seems to have this `__init__` overridden, should the default dataset just be a class attribute so that we don't need to have these argument forwarding constructors?

zipline/pipeline/data/dividends.py
mtydykov: See https://github.com/quantopian/qexec/issues/8897

zipline/testing/fixtures.py
mtydykov: What would be a reasonable default for `loader_type`? `None`?

zipline/testing/fixtures.py
mtydykov: Whose job should it be?

llllllllll: The consumer's

zipline/pipeline/loaders/blaze/dividends.py
```diff
@@ -0,0 +1,187 @@
...
+
+    _expected_fields = frozenset({
+        TS_FIELD_NAME,
+        SID_FIELD_NAME,
+        CASH_AMOUNT_FIELD_NAME,
+        ANNOUNCEMENT_FIELD_NAME
+    })
+
+    concrete_loader = DividendsByAnnouncementDateLoader
+    dataset=DividendsByAnnouncementDate
```
llllllllll: space around assignment

zipline/pipeline/loaders/blaze/events.py
```diff
@@ -63,8 +64,7 @@ def __init__(self,
                  resources=None,
                  odo_kwargs=None,
                  data_query_time=None,
-                 data_query_tz=None,
-                 dataset=None):
```
llllllllll: this still needs to be an argument so that downstream consumers may change it. We just want to use the class level attribute as the default.

zipline/pipeline/loaders/blaze/events.py
```diff
@@ -56,6 +56,7 @@ class BlazeEventsLoader(PipelineLoader):
     If the '{TS_FIELD_NAME}' field is not included it is assumed that we
     start the backtest with knowledge of all announcements.
     """
+    dataset = None
```
llllllllll: We probably want to call this `default_dataset` because the constructor still needs to accept this as an argument.

zipline/pipeline/loaders/blaze/events.py
```diff
@@ -64,7 +64,11 @@ def __init__(self,
                  resources=None,
                  odo_kwargs=None,
                  data_query_time=None,
-                 data_query_tz=None):
+                 data_query_tz=None,
+                 dataset=default_dataset):
+        if not dataset:
```
llllllllll: We should compare this against None directly instead of checking truthiness

llllllllll:
One comment, other than that lgtm

PR #1064, opened 2016-03-18
DEV: Clean up slippage API.

jbredeche:
`process_order` now takes BarData and the Order object and
returns a tuple of (execution_price, execution_volume).  Our
internal code is now responsible for creating the transaction.

Also, enforce that Orders and Transactions use Assets, not ints.

### DIFF
--- a/zipline/finance/slippage.py
+++ b/zipline/finance/slippage.py
@@ -42,18 +42,20 @@ class SlippageModel(with_metaclass(abc.ABCMeta)):
         return self._volume_for_bar
 
     @abc.abstractproperty
-    def process_order(self, bar_data, price, volume, order):
+    def process_order(self, data, order):
         pass
 
-    def simulate(self, bar_data, asset, orders_for_asset):
+    def simulate(self, data, asset, orders_for_asset):
         self._volume_for_bar = 0
-        volume = bar_data.current(asset, "volume")
+        volume = data.current(asset, "volume")
 
         if volume == 0:
             return
 
-        price = bar_data.current(asset, "price")
-        dt = bar_data.current_dt
+        # can use the close price, since we verified there's volume in this
+        # bar.
+        price = data.current(asset, "close")
+        dt = data.current_dt
 
         for order in orders_for_asset:
             if order.open_amount == 0:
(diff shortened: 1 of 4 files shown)


jbredeche:
cc @ehebert @ssanderson

zipline/finance/slippage.py
```diff
@@ -63,8 +63,20 @@ def simulate(self, bar_data, asset, orders_for_asset):
...
+
             try:
-                txn = self.process_order(bar_data, price, volume, order)
+                execution_price, execution_volume = \
+                    self.process_order(data, order)
+
+                if execution_price is not None:
+                    txn = create_transaction(
+                        order,
+                        data.current_dt,
```
ehebert: Should this function just take data (or have data baked in) so that the real `current_dt` is always enforced?

zipline/finance/slippage.py
```diff
@@ -95,7 +107,9 @@ def __repr__(self):
                    volume_limit=self.volume_limit,
                    price_impact=self.price_impact)
 
-    def process_order(self, bar_data, price, volume, order):
+    def process_order(self, data, order):
+        price = data.current(order.asset, "price")
```
ehebert: I would advise only every accessing price after the volume check. (And we should add a comment that says 'do not move this before the `cur_volume`'.)

If there is no volume, i.e. when an asset is illiquid, the 'get last traded value' branch of 'price' will be triggered.

When processing orders, the value used should only be the current close.
I would also recommend using `close` instead of `price` since the ffill behavior of `price` should not be used in this context.

jbredeche: The volume check is done one level up in the call stack, so `process_order` will never be called.

https://github.com/quantopian/zipline/pull/1064/files#diff-7553035d28cf2d71ff96e833ebb225c3R52

zipline/finance/slippage.py
```diff
@@ -156,10 +168,10 @@ def __init__(self, spread=0.0):
         """
         self.spread = spread
 
-    def process_order(self, bar_data, price, volume, order):
-        return create_transaction(
-            order,
-            bar_data.current_dt,
+    def process_order(self, data, order):
+        price = data.current(order.asset, "price")
```
ehebert: Like above, I believe only `close` should only be used here, possibly with is `nan` check before returning the order.

PR #1066, opened 2016-03-18
MAINT: fix bug in issue template

llllllllll:
due to rounding error the bitness function gave the correct result on 64b but not on 32b


richafrank:
:+1:

PR #1067, opened 2016-03-20
Added a Dockerfile for repeatable runtimes

jxstanford:
See the comments in the Dockerfile on how to build the image and run the container, and how to execute an algo inside the container.


richafrank:
@ssanderson Would you take a look at this?

jxstanford:
Happy to answer any question.  Hopefully the comments in the Dockerfile are enough to cover basic usage.

ssanderson:
I'll take a look at this today.

Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+# 
+VOLUME ${PROJECT_DIR}
+EXPOSE ${NOTEBOOK_PORT}
+
+#
+# install zipline from source
+# 
+ADD . /zipline
+WORKDIR /zipline
+RUN rm -rf /zipline/dist || /bin/true
```
ssanderson: This might be more easily accomplished by adding a [`.dockerignore`](https://docs.docker.com/engine/reference/builder/#dockerignore-file) file to the Zipline root.

ssanderson: Other patterns probably worth excluding: `build` (103 MB on my machine) `.eggs` (136 MB).

Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+#
+#    docker build -t quantopian/zipline .
+# 
+# To run the container:
+#
+#    docker run -v=/path/to/your/notebooks:/projects -p 8888:8888/tcp --name zipline -it quantopian/zipline
+#
+# To access Jupyter when running docker locally (you may need to add NAT rules):
+#
+#    https://127.0.0.1:8888
```
ssanderson: We might want to warn here that this will trigger a warning in most browsers about having an unverified certificate.

Dockerfile
```diff
@@ -0,0 +1,87 @@
+#
+# Dockerfile for an image with the currently checked out version of zipline installed. To build:
+#
+#    docker build -t quantopian/zipline .
+# 
+# To run the container:
+#
+#    docker run -v=/path/to/your/notebooks:/projects -p 8888:8888/tcp --name zipline -it quantopian/zipline
```
ssanderson: One thing I noticed while testing this is that if you don't mount an external directory as a volume, you get an error if you try to create a new notebook.  Is that intentional and/or should we warn people that failure creating notebooks likely means failure to mount a local directory?

ssanderson:
@jxstanford this looks great!  I had a couple small notes and a few questions, but by and large I think this looks like a super useful addition.

Dockerfile
jxstanford: Interesting. Can you past the error?  I am able to do that without issue.  I start the container:

jxstanford:
Good catch on the `.dockerignore`.  I'll add it and push a commit when we figure out what to do with the notebook volume issue.

Dockerfile
ssanderson: I think the error I was seeing is actually more interesting than I thought.  If I run the container with

`docker run -p 8888:8888 --name zipline -it quantopian/zipline`, when I try to open a notebook I get this in the browser:

screenshot from 2016-03-29 12-13-57

In the javascript console, I have a bunch of errors that look like:

```
Failed to load resource: net::ERR_INSECURE_RESPONSE
main.min.js?v=cee9d5d…:13001Object
main.min.js?v=cee9d5d…:13003API request failed (404): error
https://localhost:8888/api/terminals?_=1459267991926 Failed to load resource: net::ERR_INSECURE_RESPONSE
main.min.js?v=cee9d5d…:13001Object
main.min.js?v=cee9d5d…:13003API request failed (404): error
https://localhost:8888/api/sessions?_=1459267991927 Failed to load resource: net::ERR_INSECURE_RESPONSE
main.min.js?v=cee9d5d…:13001Object
main.min.js?v=cee9d5d…:13003API request failed (404): error
https://localhost:8888/api/terminals?_=1459267991928 Failed to load resource: net::ERR_INSECURE_RESPONSE
main.min.js?v=cee9d5d…:13001Object
main.min.js?v=cee9d5d…:13003API request failed (404): error
```

ssanderson: The kicker here is that if I restart top the container and then restart it with `docker start zipline`, I can create notebooks fine!

ssanderson: Hrm...and now I can't reproduce it at all.  My best guess is there are some funky browser cache interactions happening here.  Probably not worth digging into unless you're really excited about that sort of thing...

jxstanford: This seems to be an interaction between the server with unsigned certificate and your browser.  What OS/browser are you using?  When you connect to the notebook server do you connect to https://127.0.0.1:8888, and do you get a warning about unsafe site or similar?  

Since the server is probably only going to accessed used locally, we could probably revert to an HTTP server with no certificate.  I was tinkering with SSL/TLS for another project and thought everything was OK, so I dropped it in this one.  It’s probably not required here.  Let me know if you agree/disagree.

ssanderson: I'm on Ubuntu 15.10, Chrome 49.0.2623.87.

I think using SSL by default makes sense, since it decreases the likelihood that someone runs this container publicly without securing it.  The flipside I suppose is that the person who's going to carelessly/unknowingly deploy this publicly probably isn't going to bother to change the password.  At the very least, it's nice to have the self-signed cert steps in code here, since they're a bit tricky.

One option might be to make it easier to run without a password via a flag like `--no-password`.  Doing that would probably require switching to using an ENTRYPOINT rather than a CMD though, which sounds like it breaks PyCharm in some way?  Another option might be to support setting a `NO_PASSWORD` envvar, and not pass `--certfile` and `--keyfile` if it's set.

My initial thought with all of the above though is that it's probably adding more complexity than its worth until someone actually runs into a problem trying to use this.

jxstanford:
@ssanderson I think I've got everything cleaned up.  I also updated the comments in the `Dockerfile`.  The short story is that I cleaned up the unnecessary volume statements, remove the contents of the zipline folder after building in the container, but leave the folder in case developers want to mount code from the host, and work on it in the container.  I do that quite a bit in another project. 

It's a big image, but most of it looks to be related to ta-lib, numpy, scipy, and pandas.  I suppose that's to be expected.  Happy to take another pass if you find something amiss.

.dockerignore
```diff
@@ -0,0 +1,6 @@
+MANIFEST.in
+**/*pyc
```
ssanderson: :+1:

ssanderson:
@jxstanford LGTM.  Thanks for the help!

jxstanford:
NP.  Feel free to kick related bugs/feature requests over to me.

ssanderson:
@jxstanford sure thing.  @llllllllll currently has a branch going that builds a much larger and more robust dataset than Yahoo's data using https://www.quandl.com/data/WIKI.  Once his branch lands, a nice addition to this might be adding another volume for persisting the wiki dataset on disk outside the container.

Issue #529, opened 2015-03-08
Rerunning an algo


npezolano:
Is it possible to rerun an algo and continue from a different time period. Take the following simple example, I want to run the algo from t1 to t2, and then run it again from t2 to t3, however the algo's portfolios gets overwritten.

```python
from zipline import TradingAlgorithm
from zipline.utils.factory import load_from_yahoo

from datetime import datetime
import pytz
import matplotlib.pyplot as plt

class BuyApple(TradingAlgorithm):
    def initialize(self):
        self.invested = False
        self.i = 0

    def handle_data(self, data):
        self.order('AAPL',1)
        self.i += 1

# Load data for t1 to t2
start = datetime(2001, 1, 1, 0, 0, 0, 0, pytz.utc)
end = datetime(2002, 1, 1, 0, 0, 0, 0, pytz.utc)
data = load_from_yahoo(stocks=['AAPL'], indexes={}, start=start,
                   end=end, adjusted=False)

# Run algorithm
dma = BuyApple(capital_base=10000000)
perf = dma.run(data)
perf.positions[-1] # 'amount': 247

# Load data for t2 to t3
start = datetime(2002, 1, 2, 0, 0, 0, 0, pytz.utc)
end = datetime(2002, 5, 1, 0, 0, 0, 0, pytz.utc)
data = load_from_yahoo(stocks=['AAPL'], indexes={}, start=start,
                   end=end, adjusted=False)

perf = dma.run(data)
perf.positions[-1] #  'amount': 83
```

twiecki:
That's an interesting use-case. Definitely makes sense to me that you'd want to do that.

One thing to try is to set `overwrite_sim_params=True` in `run()`: https://github.com/quantopian/zipline/blob/master/zipline/algorithm.py#L400

But note that you then would have to specify your own `sim_params` and pass them to `__init__` which set start and end date of the whole length.

npezolano:
Thanks for the response, I tried your approach, the problem with the above approach is that the `PerformanceTracker` gets set to `None` here, https://github.com/quantopian/zipline/blob/master/zipline/algorithm.py#L474

twiecki:
Yes, that's indeed the problem :-/. I don't think we currently allow for this then but there are some plans to rework this logic a bit so we'll keep this in mind to make sure to support it.

npezolano:
This can be closed by #852 .

PR #1068, opened 2016-03-20
Rerunning an algo #529

tomyitav:
Addressing issue Rerunning an algo #529. The class AlgoRunnerIterator iterates over times set in a configuration file, and re runs a trading algorithm that uses the given initialize and handle data functions.


llllllllll:
Hi @tomyitav, thanks for opening this pr! Sorry it has taken so long to get
back to you.

The first thing that I noticed about this pr is that it appears you are coming
from Java based on some of your naming choices and the use of `;`. In python we
normally use `snake_case` for variables, functions, and methods and use
`PascalCase` for class names. There are a few exceptions to this but this is
the norm. In python the ending semicolon is optional and very rarely used as it
does not change the semantics. PEP 8 is the standard python style guide which
zipline follows, it is available here:
https://www.python.org/dev/peps/pep-0008/. We also run a linter on our code to
ensure that these rules are followed. This has caused the tests to fail on the
pr. You can see all of the errors here:
https://travis-ci.org/quantopian/zipline/jobs/117313295#L2148.

In python the iteration protocol is somewhat different from Java. Instead of
`iterator`, `next`, and `hasNext` we have `__iter__` and `__next__`. `__iter__`
is used to return a new iterator over some container. The object that is
returned must support a `__next__` (`next` in python 2) method which returns
the next element or raises `StopIteration` when done.

While in java we can desugar loops like:

```java
for (Object ob : myCollection) {
    System.out.println(ob);
}
```
into
```java
for (Iterator<Object> iterator = myCollection.iterator(); iterator.hasNext();) {
    Object ob = iterator.next();
    System.out.println(ob);
}
```

In Python loops desugar like:

```python
for ob in my_collection:
    print(ob)
```
into
```python
it = iter(my_collection)  # calls `__iter__`
while True:
    try:
        ob = next(it)  # calls `__next__` in py3 or `next` in py2
    except StopIteration:
        break

    print(ob)
```

You can read more about the iteration protocol here:
https://docs.python.org/3.4/library/stdtypes.html#typeiter

Python also gives us some nice shorthand for converting loops into these
iterable objects with generators:
https://docs.python.org/3.4/glossary.html#term-generator. A generator looks
like a function but instead of `return`ing a single value, it may have multiple
`yield`s. For example, imagine we wanted to define our own `(x)range`, or an
iterator that yields integers from `[0, n)`. One option would be to define a
class that implements an `__iter__` and a `__next__` method but that is a lot
of boiler plate. Instead we could write:

```python
def range(n):
    value = 0
    while value < n:
        yield value
        value += 1
```
We can then use this like:
```python
>>> for n in range(5):
...     print(n)
0
1
2
3
4
```

Regarding the feature proposed, I don't think it quite implements the feature
request in #529. In that issue the author wanted the ability to resume
execution of an algo while still maintaining the positions. This would allow
the user to analyze the performance after some time and then continue running
the algo without clearing the state. The current iterator that you have defined
does provide a nice shorthand for running an algo over different time windows
though. I think we can make it more pythonic by using a generator here. For
example:

```python
def algo_iter(symbols, date_bounds, **kwargs):
    """Create an iterator over algorithm results using yahoo data for
    the given sybols and date ranges.

    Parameters
    ----------
    symbols : iterable[str]
        The symbols to load data for.
    date_bounds : iterable[(datetime, datetime)]
        The start and end dates to run the algorithm for.
    **kwargs
        Forwarded to ``TradingAlgorithm`` to construct the actual algorithm.

    Yields
    ------
    perf : pd.DataFrame
        The performance for the algorithm over the given interval.
    """
    for start, end in date_bounds:
        yield TradingAlgorithm(**kwargs).run(
            load_bars_from_yahoo(stocks=symbols, start=start, end=end),
        )
```

This generator will do basically the same thing as the iterator you defined but
accepts the `date_bounds` as an iterator of `(start_date, end_date)` tuples.
This allows users to provide dates in other ways that just a csv but does not
stop them from loading the csv and passing the dates themselves.

If we wanted to do that, the call would look like:

```python
dates = pd.read_csv(
    'date.csv',
    index_col='start',  # set the `start` column to the index
    squeeze=True,  # sqeeze our remaining single column df into a Series
    parse_dates=[0, 1],  # parse the dates out of the two columns
)

for result in algo_iter(['IBM', 'MSFT'], dates.items(), ...):
    ...
```

We can also make a potential optimization by scanning all of the dates ahead of
time to download one large block of data:

```python
def algo_iter(symbols, date_bounds, fetch_dense=False, **kwargs):
    """Create an iterator over algorithm results using yahoo data for
    the given sybols and date ranges.

    Parameters
    ----------
    symbols : iterable[str]
        The symbols to load data for.
    date_bounds : iterable[(datetime, datetime)]
        The start and end dates to run the algorithm for.
    fetch_dense : bool, optional
        Should a single dense block of data be loaded. If you will use most
        of the data between start and end this will make less requests against
        yahoo and will be less IO.
    **kwargs
        Forwarded to ``TradingAlgorithm`` to construct the actual algorithm.

    Yields
    ------
    perf : pd.DataFrame
        The performance for the algorithm over the given interval.
    """
    if fetch_dense:
        date_bounds = list(date_bounds)
        start, end = reduce(
            lambda a, b: (min(a[0], b[0]), max(a[1], b[1])),
            date_bounds,
        )
        _data = load_bars_from_yahoo(stocks=symbols, start=start, end=end)

        def get_data(start, end):
            return _data.loc[:, start:end]

    else:
        def get_data(start, end):
            return load_bars_from_yahoo(stocks=symbols, start=start, end=end)

    for start, end in date_bounds:
        yield TradingAlgorithm(**kwargs).run(get_data(start, end))
```

This implementation optionally makes a single request to yahoo for all of the
data at once and then slices into that to get the data for any given date
range.

Since this doesn't fully implement the feature in #529 I am not sure we want to
add this utility. Can you tell us a little more about how you are using this,
maybe there is a use case we are not seeing.

FreddieV4:
Hi @tomyitav I'm going to close this PR for now. If you'd like to rebase and address the suggestions that @llllllllll has made above then feel free to re-open or open a new PR 🙂

PR #1069, opened 2016-03-21
FIX: Make overnight adjustments for before_trading_start

lianga888:
- breaks the logic for getting adjustments into a new method `get_adjustments`, used by `get_adjusted_value` and by BarData.history when we need to handle non market minutes
- `get_adjusted_value` is now a public method used by BarData.current when we need to handle non market minutes

### DIFF
--- a/tests/test_bar_data.py
+++ b/tests/test_bar_data.py
@@ -476,6 +476,48 @@ class TestMinuteBarData(TestBarDataBase):
         with handle_non_market_minutes(bar_data):
             self.assertTrue(bar_data.is_stale(self.HILARIOUSLY_ILLIQUID_ASSET))
 
+    def test_overnight_adjustments(self):
+        # verify there is a split for SPLIT_ASSET
+        splits = self.adjustments_reader.get_adjustments_for_sid(
+            "splits",
+            self.SPLIT_ASSET.sid
+        )
+
+        self.assertEqual(1, len(splits))
+        split = splits[0]
+        self.assertEqual(
+            split[0],
+            pd.Timestamp("2016-01-06", tz='UTC')
+        )
+
+        # Current day is 1/06/16
+        day = self.days[1]
+        eight_fortyfive_am_eastern = \
+            pd.Timestamp("{0}-{1}-{2} 8:45".format(
+                day.year, day.month, day.day),
+                tz='US/Eastern'
+            )
+
+        bar_data = BarData(self.data_portal,
+                           lambda: eight_fortyfive_am_eastern,
+                           "minute")
+
+        expected = {
+            'open': 391 / 2.0,
+            'high': 392 / 2.0,
+            'low': 389 / 2.0,
+            'close': 390 / 2.0,
+            'volume': 39000 * 2.0,
+            'price': 390 / 2.0,
+        }
+
+        with handle_non_market_minutes(bar_data):
+            for field in OHLCP + ['volume']:
+                value = bar_data.current(self.SPLIT_ASSET, field)
+
+                # Assert the price is adjusted for the overnight split
+                self.assertEqual(value, expected[field])
+
 
 class TestDailyBarData(TestBarDataBase):
     @classmethod
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -849,6 +849,65 @@ class MinuteEquityHistoryTestCase(HistoryTestCaseBase):
         # last 5 minutes should not be adjusted
         np.testing.assert_array_equal(np.array(range(782, 787)), window3[395:])
 
+    def test_overnight_adjustments(self):
+        # Should incorporate adjustments on midnight 01/06
+        current_dt = pd.Timestamp("2015-01-06 8:45", tz='US/Eastern')
+        bar_data = BarData(self.data_portal, lambda: current_dt, "minute")
+
+        expected = {
+            'open': np.arange(383, 393) / 2.0,
+            'high': np.arange(384, 394) / 2.0,
+            'low': np.arange(381, 391) / 2.0,
+            'close': np.arange(382, 392) / 2.0,
+            'volume': np.arange(382, 392) * 100 * 2.0,
+            'price': np.arange(382, 392) / 2.0,
+        }
+
+        with handle_non_market_minutes(bar_data):
+            # Single field, single asset
+            for field in ALL_FIELDS:
+                values = bar_data.history(self.SPLIT_ASSET, field, 10, '1m')
+                np.testing.assert_array_equal(values.values, expected[field])
+
+            # Multi field, single asset
+            values = bar_data.history(
+                self.SPLIT_ASSET, ['open', 'volume'], 10, '1m'
+            )
+            np.testing.assert_array_equal(values.open.values,
+                                          expected['open'])
+            np.testing.assert_array_equal(values.volume.values,
+                                          expected['volume'])
+
+            # Single field, multi asset
+            values = bar_data.history(
+                [self.SPLIT_ASSET, self.ASSET2], 'open', 10, '1m'
+            )
+            np.testing.assert_array_equal(values[self.SPLIT_ASSET].values,
+                                          expected['open'])
+            np.testing.assert_array_equal(values[self.ASSET2].values,
+                                          expected['open'] * 2)
+
+            # Multi field, multi asset
+            values = bar_data.history(
+                [self.SPLIT_ASSET, self.ASSET2], ['open', 'volume'], 10, '1m'
+            )
+            np.testing.assert_array_equal(
+                values.open[self.SPLIT_ASSET].values,
+                expected['open']
+            )
+            np.testing.assert_array_equal(
+                values.volume[self.SPLIT_ASSET].values,
+                expected['volume']
+            )
+            np.testing.assert_array_equal(
+                values.open[self.ASSET2].values,
+                expected['open'] * 2
+            )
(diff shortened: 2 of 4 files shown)


tests/test_history.py
```diff
@@ -812,6 +812,23 @@ def test_minute_dividends(self):
...
+            [self.SPLIT_ASSET],
+            last_traded,
+            10,
+            "1m",
+            "close",
+            handle_bts=True
+        )[self.SPLIT_ASSET]
+
+        np.testing.assert_array_equal(window.values,
+                                      np.array(range(382, 392))/2.0)
```
ssanderson: `np.arange`

zipline/data/data_portal.py
```diff
@@ -637,7 +638,51 @@ def _check_extra_sources(self, asset, column, dt):
 
                 raise KeyError
 
-    def get_spot_value(self, asset, field, dt, data_frequency):
+    def _get_overnight_adjustments(self, assets, field, dt):
+
+        if not isinstance(assets, (list, tuple)):
```
ssanderson: As far as I can tell this is only ever called once, by our own internal code.  Why are we doing a type coercion here?

zipline/data/data_portal.py
```diff
@@ -637,7 +638,51 @@ def _check_extra_sources(self, asset, column, dt):
...
+            assets = [assets]
+
+        # dt at bts is forced back to 4 pm previous day for the purpose of
```
ssanderson: Is this the previous day, or the previous **trading day**?  If the latter, then this is going to be broken on mondays/holidays/etc.

zipline/data/data_portal.py
```diff
@@ -841,11 +901,17 @@ def _get_minute_spot_value(self, asset, column, dt, ffill=False):
         if dt == last_traded_dt or dt.date() == last_traded_dt.date():
             return result
 
+        # If we are in before_trading_start, dt is 4pm the previous day, and
```
ssanderson: See comment above.  This seems unlikely to be correct on the days after non-trading days.

zipline/data/data_portal.py
```diff
@@ -637,7 +638,51 @@ def _check_extra_sources(self, asset, column, dt):
...
+            else:
+                merger_adj = div_adj = []
+
+            ratio = reduce(mul, split_adj + merger_adj + div_adj, 1.0)
+            adjs.append(ratio)
+
+        return adjs
+
+    def get_spot_value(self, asset, field, dt, data_frequency,
+                       handle_bts=False):
```
ssanderson: I don't think this should have a default value.  Providing one is inviting us to create bugs by accidentally passing the wrong value.

As a general rule of thumb, business logic functions that aren't called by users should almost never use default values.

ssanderson:
> renamed the handle_non_market_minutes handler to handle_bts_minute. It's currently only used in the before_trading_start, so I felt the name change made sense. In the future, if we need a handler for all non market minutes, that can take on the former name

Pedagical Moment:

In general, we should prefer to name functions based on **what they represent** rather than **how they happen to be used**.  The fact that we're only using `handle_non_market_minutes` in before_trading_start doesn't necessarily mean that we should rename it to reflect that more narrow use, especially if we expect to expand that usage down the road.  The real question for making that name choice should be: "is there something essential about what this function does such that it **only** works in before_trading_start?"  If the answer to that question is "yes", then it *might* be reasonable to do the rename you've made here, but we also might want to ask ourselves whether we could implement a simpler function that naturally handles non-market minute correctly in a generic way.

tests/test_history.py
```diff
@@ -812,6 +812,23 @@ def test_minute_dividends(self):
         # last 5 minutes should not be adjusted
         np.testing.assert_array_equal(np.array(range(782, 787)), window3[395:])
 
+    def test_overnight_adjustments(self):
+
+        # Should incorporate adjustments on midnight 01/06
+        last_traded = pd.Timestamp("2015-01-05 21:00", tz='UTC')
```
ssanderson: Can we add a test asserting that this correctly handles the window boundary crossing a non-trading day?

zipline/data/data_portal.py
```diff
@@ -701,12 +701,50 @@ def get_spot_value(self, asset, field, dt, data_frequency):
                 else:
                     return self._get_minute_spot_value(asset, field, dt)
 
-    def _get_adjusted_value(self, asset, field, dt,
-                            perspective_dt,
-                            data_frequency,
-                            spot_value=None):
+    def get_adjustments(self, assets, field, dt, perspective_dt):
+
```
jbredeche: can you add a docstring?  purpose, parameters, what this returns

zipline/data/data_portal.py
```diff
@@ -701,12 +701,50 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
+        return adjs
+
+    def get_adjusted_value(self, asset, field, dt,
+                           perspective_dt,
+                           data_frequency,
+                           spot_value=None):
         """
-        Private method that returns a scalar value representing the value
+        Method that returns a scalar value representing the value
```
jbredeche: how about "Returns a scalar value..." (don't need "Method that")

zipline/_protocol.pyx
```diff
@@ -459,7 +466,25 @@ cdef class BarData:
...
+                }
+
+                if self._adjust_minutes:
+                    adjs = {
+                        field: self.data_portal.get_adjustments(
+                            assets,
+                            field,
+                            self._get_current_minute(),
+                            self.simulation_dt_func()
+                        )[0] for field in fields
```
jbredeche: why the `[0]`?

zipline/_protocol.pyx
```diff
@@ -459,7 +466,25 @@ cdef class BarData:
...
+                    }
+
+                    df_dict = {field: df * adjs[field]
```
jbredeche: don't some fields need to be multiplied by the adjustment ratio (open, high, etc) and some divided (volume)?

tests/test_bar_data.py
```diff
@@ -476,6 +476,38 @@ def test_is_stale_at_midnight(self):
...
+            pd.Timestamp("{0}-{1}-{2} 8:45".format(
+                day.year, day.month, day.day),
+                tz='US/Eastern'
+            )
+
+        bar_data = BarData(self.data_portal,
+                           lambda: eight_fortyfive_am_eastern,
+                           "minute")
+
+        with handle_non_market_minutes(bar_data):
```
jbredeche: I think you need to test with every possible field value (OHLCVP), and verify that for OHLCP, the values are multiplied by the split ratio, and for V, the value is divided by the split ratio.

tests/test_api_shim.py
```diff
@@ -211,22 +211,32 @@ def assert_get_spot_value_called(fun, field):
...
+                        test_end_minute,
+                        'minute'
+                    )
         # Ensure that data.current(sid(n), field) has the same behaviour as
         # data[sid(n)].field.
         for field in ohlcvp_fields:
             assert_get_spot_value_called(
                 lambda: getattr(bar_data[self.asset1], field),
                 field,
+                False
```
jbredeche: can you remind me what this new parameter is for?

jbredeche:
@lianga888 this is much cleaner than the last version, nicely done.

left a few comments, but the big hole is that you need to test for OHLCVP, not just P.  Volume has adjustments applied inversely (divided, instead of multiplied), and you haven't implemented or tested that.

zipline/_protocol.pyx
lianga888: data_portal.get_adjustments returns a list of adjustments for the list of assets that are passed in. In this case, since `assets` is just one asset, it gets coerced into a list, so the return value is a list of len 1

zipline/_protocol.pyx
lianga888: Yeah, that logic is in data_portal.get_adjustments's `split_adj_factor`, which returns the inverse for volume

tests/test_api_shim.py
lianga888: To avoid code repetition in BarData.current, I used a partial to store the function (data_portal.get_spot_value vs. data_portal.get_adjusted_value) that we want to use. These are called with keywords, so this assert method must be aware (through the is_new flag) that the new api calls it this way, while the old api just calls with positional args

tests/test_bar_data.py
jbredeche: ah got it, cool.  can you add a test for all the other fields?

jbredeche:
getting this via qprof on this branch:
```
  File "/Users/jean/repo/qexec/zipline_repo/zipline/data/data_portal.py", line 700, in get_spot_value
    True)
  File "/Users/jean/repo/qexec/zipline_repo/zipline/data/data_portal.py", line 854, in _get_minute_spot_value
    dt, "minute", spot_value=result
  File "/Users/jean/repo/qexec/zipline_repo/zipline/data/data_portal.py", line 781, in get_adjusted_value
    ratio = self.get_adjustments(asset, field, dt, perspective_dt)[0]
  File "/Users/jean/repo/qexec/zipline_repo/zipline/data/data_portal.py", line 733, in get_adjustments
    if dt <= adj_dt <= perspective_dt]
  File "pandas/tslib.pyx", line 832, in pandas.tslib._Timestamp.__richcmp__ (pandas/tslib.c:15596)
TypeError("Cannot compare type 'Timestamp' with type 'float'",)
```

investigating

PR #1071, opened 2016-03-22
FIX: Allow order_target_percent, order_target_value, etc on an asset's last day

jbredeche:
There was an off-by-one bug that was erroneously preventing you from
ordering an asset on its last trading day.

There’s a bunch we still need to do to solidy this - these date checks
don’t apply to the simple `order` method, etc.  I wanted to get this fix in first.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -730,7 +730,7 @@ class TestTransformAlgorithm(TestCase):
         for sid in cls.sids:
             equities_metadata[sid] = {
                 'start_date': cls.sim_params.period_start,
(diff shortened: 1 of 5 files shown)


tests/test_algorithm.py
```diff
@@ -730,7 +730,7 @@ def setUpClass(cls):
         for sid in cls.sids:
             equities_metadata[sid] = {
                 'start_date': cls.sim_params.period_start,
-                'end_date': cls.env.next_trading_day(cls.sim_params.period_end)
+                'end_date': cls.sim_params.period_end
```
jbredeche: we didn't need to go to the next trading day, so I removed it.

PR #1072, opened 2016-03-22
run_pipeline failing given too many factors

dmichalowicz:
NumericalExpressions fail to merge if one of them has >10 inputs.

### DIFF
--- a/tests/pipeline/test_numerical_expression.py
+++ b/tests/pipeline/test_numerical_expression.py
@@ -8,7 +8,9 @@ from operator import (
     methodcaller,
     mul,
     ne,
+    sub,
 )
+from string import ascii_uppercase
 from unittest import TestCase
 
 import numpy
@@ -146,6 +148,41 @@ class NumericalExpressionTestCase(TestCase):
             f + (f > 2)
         with self.assertRaises(TypeError):
             (f > f) > f
+
+    def test_many_inputs(self):
+        """
+        Test adding NumericalExpressions with >10 inputs.
+        """
+        # Create an initial NumericalExpression by adding two factors together.
+        f = self.f
+        expr = f + f
+
+        self.fake_raw_data = {f: full((5, 5), 0, float)}
+        expected = 0
+
+        # Alternate between adding and subtracting factors. Because subtraction
+        # is not commutative, this ensures that we are combining factors in the
+        # correct order.
+        ops = (add, sub)
+
+        for i, name in enumerate(ascii_uppercase):
+            op = ops[i % 2]
+            NewFactor = type(
+                name,
+                (Factor,),
+                dict(dtype=float64_dtype, inputs=(), window_length=0),
+            )
+            new_factor = NewFactor()
+
+            # Again we need a NumericalExpression, so add two factors together.
+            new_expr = new_factor + new_factor
+            self.fake_raw_data[new_factor] = full((5, 5), i + 1, float)
+            expr = op(expr, new_expr)
+
+            # Double the expected output since each factor is counted twice.
+            expected = op(expected, (i + 1) * 2)
+
+        self.check_output(expr, full((5, 5), expected, float))
 
     def test_combine_datetimes(self):
         with self.assertRaises(TypeError) as e:
--- a/zipline/pipeline/expression.py
+++ b/zipline/pipeline/expression.py
@@ -248,7 +248,18 @@ class NumericalExpression(ComputableTerm):
         new_inputs.
         """
         expr = self._expr
-        for idx, input_ in enumerate(self.inputs):
+
+        # If we have 11+ variables, some of our variable names may be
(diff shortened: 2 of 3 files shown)


zipline/pipeline/expression.py
```diff
@@ -248,7 +248,12 @@ def _rebind_variables(self, new_inputs):
         new_inputs.
         """
         expr = self._expr
-        for idx, input_ in enumerate(self.inputs):
+
+        # This call to `reversed` is required to handle the case in which the
+        # length of self.inputs is greater than 10. Otherwise, expressions
+        # containing an x_10 will be improperly overwritten at the same time
+        # that x_1 is overwritten.
```
ssanderson: The essence of the bug here is that some of our variable names are substrings of other variable names, so when we replace `x_1` before `x_10`, we inadvertently modify `x_10` because it contains `x_1` as a substring.  By enumerating in reverse order, we ensure that every variable name which is a substring of another variable name is processed **after** the variable of which it is a substring.  This guarantees that the substitution of any given variable index only every affects exactly its own index.

tests/pipeline/test_numerical_expression.py
```diff
@@ -147,6 +147,41 @@ def test_validate_bad(self):
...
+        for class_name in 'PQRSTUVWXYZ':
+            factors.append(
+                type(
+                    class_name,
+                    (Factor,),
+                    dict(dtype=float64_dtype, inputs=(), window_length=0),
+                )
+            )
+
+        short_num_expr = NumericalExpression(
```
ssanderson: I don't think there's a benefit to manually constructing these.  We can just use '+' or similar.

tests/pipeline/test_numerical_expression.py
```diff
@@ -147,6 +147,41 @@ def test_validate_bad(self):
         with self.assertRaises(TypeError):
             (f > f) > f
 
+    def test_many_inputs(self):
+        factors = []
+
+        # Create 11 factors to use as inputs. We need at least 11 to create a
+        # string expression containing an x_10. Then, ensure that this long
+        # expression can correctly merge with another.
+        for class_name in 'PQRSTUVWXYZ':
```
ssanderson: Might be worth going up a bit higher here:

```
from string import ascii_uppercase
[a + b for a, b in product(ascii_uppercase, ascii_uppercase)]
```
^ will give you all length-two pairs of uppercase letters

tests/pipeline/test_numerical_expression.py
```diff
@@ -147,6 +147,41 @@ def test_validate_bad(self):
...
+        # Merge our long NumericalExpression with another.
+        long_expr, short_expr, new_inputs = \
+            long_num_expr.build_binary_op('+', short_num_expr)
+
+        # Test that it passes validation.
+        NumericalExpression(
+            ' + '.join([long_expr, short_expr]),
+            new_inputs,
+            dtype=float64_dtype,
+        )
```
ssanderson: We should test actually computing these to make sure we don't scramble the variables.

tests/pipeline/test_numerical_expression.py
```diff
@@ -148,39 +150,40 @@ def test_validate_bad(self):
...
+        new_factor_names = ascii_uppercase.replace('F', '')
+        self.fake_raw_data = {f: full((5, 5), 0, float)}
+        expected = 0
+
+        # Alternate between adding and subtracting factors. Because subtraction
+        # is not commutative, this ensures that we are combining factors in the
+        # correct order.
+        ops = (add, sub)
+
+        for i, name in enumerate(new_factor_names):
```
ssanderson: Another common way to write something like this is:

```
from itertools import repeat
for op, name in zip(new_factor_names, repeat(ops)):
    ...
```
```

File: tests/pipeline/test_numerical_expression.py [hunk 6]
```diff
@@ -148,39 +150,40 @@ def test_validate_bad(self):
...
-                    dict(dtype=float64_dtype, inputs=(), window_length=0),
-                )
+        """
+        Test adding NumericalExpressions with >10 inputs.
+        """
+        # Create an initial NumericalExpression by adding two factors together.
+        f = self.f
+        expr = f + f
+
+        new_factor_names = ascii_uppercase.replace('F', '')
```
ssanderson: Is this actually necessary?  I don't think it's a big deal to have two factors named `F` in the expression.

dmichalowicz: No it's not necessary, will remove

tests/pipeline/test_numerical_expression.py
dmichalowicz: I still need the index here though, for the data.

ssanderson:
LGTM.  Can you add a bugfix entry to the whatsnew?

ssanderson:
LGTM

PR #1075, opened 2016-03-23
Quantiles

ssanderson:
Adds `Factor.quantiles` as a method for computing a Classifier from a Factor.


zipline/pipeline/factors/factor.py
```diff
@@ -685,6 +695,105 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
...
+
+        Parameters
+        ----------
+        mask : zipline.pipeline.Filter, optional
+            Mask of values to ignore when computing deciles.
+
+        Returns
+        -------
+        deciles : zipline.pipeline.classifiers.Quantiles
+            A Classifier producing integer labels ranging from 0 to 4.
```
dmichalowicz: 0 to 9

docs/source/whatsnew/0.8.5.txt
```diff
@@ -32,13 +32,20 @@ Enhancements
...
+  them as the ``groupby`` parameter to factor normalization methods.
+
 * Added factor normalization methods:
   :meth:`zipline.pipeline.Factor.demean` and
   :meth:`zipline.pipeline.Factor.zscore`. (:issue:`1046`)
 
-* Implemented :class:`zipline.pipeline.Classifier`, a new core pipeline API
-  term representing grouping keys.  Classifiers are primarily used by passing
-  them as the ``groupby`` parameter to factor normalization methods.
+* Added :meth:`zipline.pipeline.Factor.quantiles`, a method for computing a
```
dmichalowicz: Are there additions to help docs for quantiles as well?

tests/pipeline/test_factor.py
```diff
@@ -553,3 +649,191 @@ class DateFactor(Factor):
...
+                                 [0, 0, 1, 1, 2, 2],
+                                 [0, 0, 1, 1, 2, 2]]),
+            # In the limiting case, we just have every column different.
+            '6': permuted_array([[0, 1, 2, 3, 4, 5],
+                                 [0, 1, 2, 3, 4, 5],
+                                 [0, 1, 2, 3, 4, 5],
+                                 [0, 1, 2, 3, 4, 5],
+                                 [0, 1, 2, 3, 4, 5],
+                                 [0, 1, 2, 3, 4, 5]]),
+        }
```
dmichalowicz: Would it be worth testing the case where number of bins does not evenly divide the number of columns? i.e. if your factor had 4 columns and you wanted 3 bins, should that give this?
```python
[[0, 0, 1, 2],
 [0, 0, 1, 2],
...
```

Or, if you ask for more bins than there are columns, should that raise a certain error?

ssanderson: At the time when we get the bin count, we don't know how many columns we're going to compute with (this is true even if we're not doing any grouping, since the number of assets in existence at any given time fluctuates significantly.)

w/r/t uneven dividing, the exact right answer there depends on the interpolation method used by `pd.qcut`.  I'll take a look and see if I can figure out a sane way to test this reliably.

PR #1077, opened 2016-03-23
MAINT: Provide prefetch history loader for 1m.

ehebert:
Factor out USEquityHistoryLoader into a base class which can be
overridden to support daily or minute history.

In support of reading the minute bars for a range of market minute by
the loader from the reader, put the entire minute index for the minute
bar reader into memory so that the two indexes can be readily compared.
(Though we may want to consider a better way to store just the half day
 information, since the half day only accounts for ~1% of trading days.)

Change some test case sim params so that the tests are started far
enough from the dataset start so that history calls can be made.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -107,7 +107,9 @@ class HistoryTestCaseBase(TestCase):
         cls.adj_reader = cls.create_adjustments_reader()
 
         cls.create_data()
-        cls.create_data_portal()
+
+    def setUp(self):
+        self.create_data_portal()
 
     @classmethod
     def create_data_portal(cls):
@@ -655,6 +657,9 @@ class MinuteEquityHistoryTestCase(HistoryTestCaseBase):
             check_internal_consistency(
                 bar_data, self.SHORT_ASSET, ALL_FIELDS, 30, "1m"
             )
+
+        # Reset data portal because it has advanced past next test date.
+        self.create_data_portal()
 
         # choose a window that contains the last minute of the asset
         bar_data = BarData(self.data_portal, lambda: minutes[15], "minute")
@@ -1263,7 +1268,7 @@ class DailyEquityHistoryTestCase(HistoryTestCaseBase):
                 "close"
             )[asset]
 
-            # first value should be halved, second value unadjusted
+            # first value should be doubled, second value unadjusted
             np.testing.assert_array_equal([1, 3], window2)
 
             window2_volume = self.data_portal.get_history_window(
@@ -1274,7 +1279,10 @@ class DailyEquityHistoryTestCase(HistoryTestCaseBase):
                 "volume"
             )[asset]
 
-            np.testing.assert_array_equal(window2_volume, [100, 300])
+            if asset == self.SPLIT_ASSET:
+                np.testing.assert_array_equal(window2_volume, [400, 300])
+            elif asset == self.MERGER_ASSET:
+                np.testing.assert_array_equal(window2_volume, [200, 300])
 
             # straddling both events
             window3 = self.data_portal.get_history_window(
@@ -1295,7 +1303,10 @@ class DailyEquityHistoryTestCase(HistoryTestCaseBase):
                 "volume"
             )[asset]
 
-            np.testing.assert_array_equal(window3_volume, [50, 150, 400])
+            if asset == self.SPLIT_ASSET:
+                np.testing.assert_array_equal(window3_volume, [800, 600, 400])
+            elif asset == self.MERGER_ASSET:
+                np.testing.assert_array_equal(window3_volume, [200, 300, 400])
 
     def test_daily_dividends(self):
         # self.DIVIDEND_ASSET had dividends on 1/6 and 1/7
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -26,7 +26,8 @@ from six.moves import reduce
 from zipline.assets import Asset, Future, Equity
 from zipline.data.us_equity_pricing import NoDataOnDate
 from zipline.data.us_equity_loader import (
-    USEquityHistoryLoader,
+    USEquityDailyHistoryLoader,
+    USEquityMinuteHistoryLoader,
 )
 
 from zipline.utils import tradingcalendar
@@ -451,7 +452,7 @@ class DataPortal(object):
 
         self._equity_daily_reader = equity_daily_reader
         if self._equity_daily_reader is not None:
-            self._equity_history_loader = USEquityHistoryLoader(
+            self._equity_history_loader = USEquityDailyHistoryLoader(
                 self.env,
                 self._equity_daily_reader,
                 self._adjustment_reader
@@ -466,6 +467,11 @@ class DataPortal(object):
             self._equity_daily_aggregator = DailyHistoryAggregator(
                 self.env.open_and_closes.market_open,
                 self._equity_minute_reader)
+            self._equity_minute_history_loader = USEquityMinuteHistoryLoader(
+                self.env,
+                self._equity_minute_reader,
+                self._adjustment_reader
+            )
             self.MINUTE_PRICE_ADJUSTMENT_FACTOR = \
                 self._equity_minute_reader._ohlc_inverse
 
@@ -496,6 +502,15 @@ class DataPortal(object):
             'price': None
         }
         self._equity_daily_reader_array_data = {}
+        self._equity_minute_loader_array_keys = {
+            'open': None,
+            'high': None,
+            'low': None,
+            'close': None,
+            'volume': None,
+            'price': None
+        }
+        self._equity_minute_loader_array_data = {}
 
         self._in_bts = False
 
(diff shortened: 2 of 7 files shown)


zipline/data/_minute_bar_internal.pyx
```diff
@@ -59,15 +59,10 @@ def find_position_of_minute(ndarray[long_t, ndim=1] market_opens,
     """
     cdef Py_ssize_t market_open_loc, market_open, delta
 
-    market_open_loc = \
-        searchsorted(market_opens, minute_val, side='right') - 1
-    market_open = market_opens[market_open_loc]
-    delta = minute_val - market_open
-
-    return (market_open_loc * minutes_per_day) + delta
+    return searchsorted(minutes, minute_val, side='right') - 1
```
jbredeche: you don't need the variables defined in line 60 anymore

zipline/data/_minute_bar_internal.pyx
```diff
@@ -100,17 +95,15 @@ def find_last_traded_position_internal(
     """
     cdef Py_ssize_t minute_pos, current_minute
 
-    minute_pos = int_min(
-        find_position_of_minute(market_opens, end_minute, minutes_per_day),
+    start_pos = searchsorted(minutes, start_minute, side='left')
```
jbredeche: you should `cdef` `start_pos` and `end_pos` above, right? (and get rid of `current_minute`)

zipline/data/_minute_bar_internal.pyx
```diff
@@ -100,17 +95,15 @@ def find_last_traded_position_internal(
...
+
+    end_pos = minute_pos = int_min(
```
jbredeche: does `end_pos` still need to exist?

zipline/data/data_portal.py
```diff
@@ -495,6 +501,15 @@ def __init__(self,
             'price': None
         }
         self._equity_daily_reader_array_data = {}
+        self._equity_minute_loader_array_keys = {
```
jbredeche: can you add a comment explaining what's going to be stored in `self._equity_minute_loader_array_keys` and `self._equity_minute_loader_array_data`?  I assume it's cached data of some sort, but hard to tell.

zipline/data/data_portal.py
```diff
@@ -1441,17 +1365,11 @@ def _apply_all_adjustments(self, data, asset, dts, field,
...
 
-            data *= price_adj_factor
-
-            # if anything is zero, it's a missing bar, so replace it with NaN.
-            # we only want to do this for non-volume fields, because a missing
-            # volume should be 0.
-            data[data == 0] = np.NaN
-
-        np.around(data, 3, out=data)
+            if price_adj_factor is not None:
```
jbredeche: this is likely going to conflict with lazy-mainline, where this method has been renamed and tweaked a bit.

zipline/data/minute_bars.py
```diff
@@ -612,31 +610,26 @@ def get_last_traded_dt(self, asset, dt):
...
-            US_EQUITIES_MINUTES_PER_DAY
-        )
-
-        return pd.Timestamp(minute_epoch, tz='UTC', unit="m")
+        minute_epoch = self._minute_value_index[pos]
+        return pd.Timestamp(minute_epoch, tz='UTC')
 
     @remember_last
-    def _find_position_of_minute(self, minute_dt):
+    def _find_position_of_minute(self, minute_dt_val):
```
jbredeche: can you update the docstring,`minute_dt_val` is the nanosecond value right?

zipline/data/minute_bars.py
```diff
@@ -706,3 +698,16 @@ def unadjusted_window(self, fields, start_dt, end_dt, sids):
                 out *= self._ohlc_inverse
             results.append(out)
         return results
+
+    def unadjusted_window_for_minutes(self, fields, minutes, sids):
```
jbredeche: can you add a docstring here?  hard to decipher what this method does.

zipline/data/minute_bars.py
```diff
@@ -706,3 +698,16 @@ def unadjusted_window(self, fields, start_dt, end_dt, sids):
...
+        start_dt = minutes[0]
+        end_dt = minutes[-1]
+        results = self.unadjusted_window(fields, start_dt, end_dt, sids)
+        if len(results[0][0]) != len(minutes):
+            start_idx = self._find_position_of_minute(start_dt.value)
+            end_idx = self._find_position_of_minute(end_dt.value)
+            data_minutes = self._minute_value_index[start_idx:end_idx + 1]
+            indices = np.in1d(data_minutes, minutes.asi8)
+            # Should do this to ndarray.
+            results = [r[:, indices] for r in results]
```
jbredeche: just to make sure I understand, this entire block handles early closes?  can you add some comments to that effect?

zipline/data/minute_bars.py
```diff
@@ -706,3 +698,16 @@ def unadjusted_window(self, fields, start_dt, end_dt, sids):
...
+
+    def unadjusted_window_for_minutes(self, fields, minutes, sids):
+        start_dt = minutes[0]
+        end_dt = minutes[-1]
+        results = self.unadjusted_window(fields, start_dt, end_dt, sids)
+        if len(results[0][0]) != len(minutes):
+            start_idx = self._find_position_of_minute(start_dt.value)
+            end_idx = self._find_position_of_minute(end_dt.value)
+            data_minutes = self._minute_value_index[start_idx:end_idx + 1]
+            indices = np.in1d(data_minutes, minutes.asi8)
```
jbredeche: from the docs, it seems like `np.in1d` is O(len(data_minutes) * len(minutes)) - is that fast enough for our use here, or am I misreading?

jbredeche: nvm, https://github.com/numpy/numpy/blob/v1.10.1/numpy/lib/arraysetops.py#L305-L404 shows that it's sorting the arrays first, phew.

jbredeche: actually, there's an `assume_unique` flag that I think we can set to True, which would make this faster: https://github.com/numpy/numpy/blob/v1.10.1/numpy/lib/arraysetops.py#L318

zipline/data/us_equity_loader.py
```diff
@@ -121,7 +136,7 @@ def _get_adjustments_in_range(self, assets, days, field):
                 for m in mergers:
                     dt = m[0]
                     if start < dt <= end:
-                        end_loc = days.get_loc(dt)
+                        end_loc = dts.searchsorted(dt)
```
jbredeche: why this change?  because `dts` can be minutes now?

zipline/data/us_equity_loader.py
```diff
@@ -179,50 +195,25 @@ def _ensure_sliding_window(
...
-                end_ix = self._calendar.get_loc(end)
-            if field != 'volume':
-                pre_array = full((fill_size, 1), nan)
-            else:
-                pre_array = full((fill_size, 1), 0)
-        else:
-            offset = 0
-            start_ix = self._calendar.get_loc(start)
-            end_ix = self._calendar.get_loc(end)
+        start = dts[0]
```
jbredeche: style comment/question, why are `start` and `end` defined far away from each other?  easier to read if they're defined right next to each other, I think.

zipline/data/us_equity_loader.py
jbredeche: out of scope for this PR, but would love to see less duplicates of the intricate logic around each merger, dividend, and split (all the `Float64Multiply` construction, etc).  I'll submit a PR afterwards.

zipline/data/us_equity_loader.py
```diff
@@ -168,10 +183,11 @@ def _get_adjustments_in_range(self, assets, days, field):
         return adjs
 
     def _ensure_sliding_window(
-            self, assets, start, end, size, field):
+            self, assets, dts, size, field):
```
jbredeche: I have no idea what this method does.  can you add some comments explaining?

zipline/data/us_equity_loader.py
```diff
@@ -179,50 +195,25 @@ def _ensure_sliding_window(
...
-                pre_array = full((fill_size, 1), nan)
-            else:
-                pre_array = full((fill_size, 1), 0)
-        else:
-            offset = 0
-            start_ix = self._calendar.get_loc(start)
-            end_ix = self._calendar.get_loc(end)
+        start = dts[0]
+
+        offset = 0
```
jbredeche: `offset` is only ever assigned 0, then passed into `SlidingWindow` below - can we get rid of it?

tests/test_history.py
```diff
@@ -1255,7 +1260,7 @@ def test_daily_splits_and_mergers(self):
                 "close"
             )[asset]
 
-            # first value should be halved, second value unadjusted
+            # first value should be doubled, second value unadjusted
```
jbredeche: still looks like it's getting halved (2 -> 1), is this comment change intentional?

zipline/data/data_portal.py
ehebert: I do not see that rename on current `lazy-mainline`.

zipline/data/minute_bars.py
ehebert: Tuning this would be good, howevere it is also called relatively infrequently. Once per every 1490 minute bars.

zipline/data/minute_bars.py
ehebert: Yes, I am having some pause on how the half days are handled here.
I believe we can do it instead by just storing the minutes that *aren't* market minutes and filtering those out; since half days are only 2-3 times a year we should avoid some of the upfront costs that loading all minutes adds.

In that case, would revert the changes that change `_minute_bar_internal.pyx`.

tests/test_history.py
ehebert: You're right, that comment was meant to go with lines 1276 and 1279.

zipline/data/us_equity_loader.py
ehebert: Yes, because they can be `dts`.

I am planning on factoring this out as well.

zipline/data/us_equity_loader.py
ehebert: The reason that it is not, is if the previous window has not expired, `start` is not needed.

zipline/data/us_equity_loader.py
ehebert: Going the other direction, and passing `offset` to `Float64Window` as well.

ehebert:
Closing in favor of https://github.com/quantopian/zipline/pull/1091

PR #1078, opened 2016-03-23
FIX: Issue with DailyHistoryAggregator.close calling itself recursively

lianga888:
Calls its recursively and nests the array that it returns. Do not nest it
https://github.com/quantopian/qexec/issues/9543

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -12,7 +12,7 @@ from testfixtures import TempDirectory
 from zipline import TradingAlgorithm
 from zipline._protocol import handle_non_market_minutes
 from zipline.assets import Asset
-from zipline.data.data_portal import DataPortal
+from zipline.data.data_portal import DataPortal, DailyHistoryAggregator
 from zipline.data.minute_bars import (
     BcolzMinuteBarReader,
     BcolzMinuteBarWriter,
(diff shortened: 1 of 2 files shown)


zipline/data/data_portal.py
```diff
@@ -350,7 +350,7 @@ def closes(self, assets, dt):
                         asset, dt, 'close')
                     if pd.isnull(val):
                         val = self.closes([asset],
-                                          pd.Timestamp(prev_dt, tz='UTC'))
+                                          pd.Timestamp(prev_dt, tz='UTC'))[0]
```
jbredeche: should this fix also be applied to the case above (lines 342-344)?

lianga888: hmm yeah i think so

jbredeche: cool, can you add a test that exercises that case too

tests/test_history.py
```diff
@@ -1751,3 +1751,99 @@ def test_skip_minutes_multiple(self, field):
                                     self.expected_values[asset][field][i],
                                     err_msg="sid={0} field={1} dt={2}".format(
                                         asset, field, minute))
+
+
+class DailyHistoryAggregatorTestCase(TestCase):
```
ehebert: I agree with your assessment that we should test the `DailyHistoryAggregator` directly.

My recommendation would be to change the existing tests which use the fixtures paradigm to use:

```
    def init_instance_fixtures(self):
        super(MinuteToDailyAggregationTestCase, self).init_instance_fixtures()
        # Set up a fresh data portal for each test, since order of calling
        # needs to be tested.
        self.equity_daily_aggregator = DailyHistoryAggregator(
            self.env.open_and_closes.market_open,
            self.bcolz_minute_bar_reader,
```

PR #1079, opened 2016-03-23
MAINT: pull aggregator logic out to separate method

jbredeche:
for downstream subclassing


ehebert:
See comment on internal repo, I think we can do the needed logic that the sublcass was providing purely from the minute reader interface.

jbredeche:
no longer needed.

PR #1080, opened 2016-03-24
BUG: Fix accidental array during recursion.

ehebert:
The recursion on finding the last close was incorrectly using the array
returned by the close method instead of demoting it to scalar. When this
value was appended to the full close results return arrays with bizarre
shapes, e.g. `(2, 1, 1, 1, 1, 1)`, because a list built of scalars and
lists was passed to the array constructor.

Change daily aggregrator test suite to test the DailyHistoryAggregator
directly instead of routing through data portal, since the assignment of
the values when using a bar count of 1 was papering over the unexpected
shape.

Add a check on returned values in tests that the return is a `Real` to
prevent regression on the assignment of an array when a

Credit to @aliang for identifying root cause.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -1,6 +1,8 @@
 from textwrap import dedent
 from unittest import TestCase
 
+from numbers import Real
+
 import pandas as pd
 import numpy as np
 from numpy import nan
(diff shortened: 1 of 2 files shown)


ehebert:
@lianga888 here is the branch that is an alternative to https://github.com/quantopian/zipline/pull/1078 (porting existing tests, so that we have the multiple test cases covered.)

When testing, removed the `[0]` from each place where it is needed, and different tests failed, so I believe the addition of checking for `Real` plus existing test cases is sufficient.

lianga888:
lgtm

PR #1082, opened 2016-03-24
FIX: Shim for q1 history call inside before_trading_start

lianga888:
Make sure the old history API gets the window as of the last market minute, gets adjustments between the last market minute and current time, and applies the adjustments

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -308,6 +308,8 @@ class TradingAlgorithm(object):
         self._before_trading_start = None
         self._analyze = None
 
+        self._in_bts = False
+
         self.event_manager = EventManager(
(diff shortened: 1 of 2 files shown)


zipline/algorithm.py
```diff
@@ -1246,14 +1252,35 @@ def history(self, bar_count, frequency, field, ffill=True):
...
+            return self.data_portal.get_history_window(
+                assets,
+                self.datetime,
+                bar_count,
+                frequency,
+                field,
+                ffill,
+            )
+        else:
+            adjusted_dt = \
```
jbredeche: can you add comments here explaining what is going on? looks right to me.

PR #1083, opened 2016-03-24
BUG: correctly translate # of days into # of minutes for simple transforms

jbredeche:


zipline/data/data_portal.py
```diff
@@ -1739,6 +1740,31 @@ def get_fetcher_assets(self, dt):
         else:
             return [assets] if isinstance(assets, Asset) else []
 
+    @lru_cache(8)
```
llllllllll: why 8 exactly?

zipline/data/data_portal.py
```diff
@@ -15,6 +15,7 @@
 from operator import mul
 
 import bcolz
+from cachetools import lru_cache
```
llllllllll: this is available in functools32, if we add this we should also update `memoize_last` in zipline.utils.memoize to be implemented as `lru_cache(1)`

zipline/data/data_portal.py
```diff
@@ -1739,6 +1740,31 @@ def get_fetcher_assets(self, dt):
...
+        )
+
+        minutes_count = \
+            sum(210 if day in self.env.early_closes else 390 for day in days)
+
+        # add the minutes for today
+        today_open = self.env.get_open_and_close(ending_minute)[0]
+
+        minutes_count += int(
+            (ending_minute - today_open).total_seconds() / 60
```
llllllllll: instead of `int(a / b)` we can use `a // b`

zipline/data/data_portal.py
jbredeche: good call, switched and done.

jbredeche:
thanks @llllllllll for the py2/py3 compatibility work!

setup.py
```diff
@@ -189,15 +198,22 @@ def read_requirements(path, strict_bounds, conda_format=False):
...
 
         return list(reqs)
 
 
 def install_requires(strict_bounds=False, conda_format=False):
-    return read_requirements('etc/requirements.txt',
+    reqs = read_requirements('etc/requirements.txt',
                              strict_bounds=strict_bounds,
                              conda_format=conda_format)
+    if sys.version_info.major == 2 or conda_format:
```
richafrank: Why do we need the `or conda_format` portion?

llllllllll: I am pretty sure this is executed with the python from the conda root

PR #1084, opened 2016-03-24
TST: Parametrizes spot price test to test dividends and mergers also.

abhijeetkalyan:
@jbredeche I just used the format of the existing test, and parameterized for dividends and mergers. Is there anything else we'd like to test here?


abhijeetkalyan:
Spoke offline, merging

PR #1085, opened 2016-03-24
MAINT: Write open and closes for minute bar format

ehebert:
Write arrays representing corresponding market opens and market closes,
which will eventually replace the `minute_index` field.

The market closes are being added for incoming work on another branch
which will use the market closes to generate a list of non-market
minutes to filter out when returning data from `unadjusted_window`.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -94,9 +94,18 @@ class BcolzMinuteBarMetadata(object):
             minute_index = pd.to_datetime(raw_data['minute_index'],
                                           utc=True)
             ohlc_ratio = raw_data['ohlc_ratio']
-            return cls(first_trading_day, minute_index, ohlc_ratio)
+            return cls(first_trading_day,
+                       minute_index,
+                       None,  # currently only writing market_opens
+                       None,  # currently only writing market_closes
+                       ohlc_ratio)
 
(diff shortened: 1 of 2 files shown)


ehebert:
@jbredeche this branch is in support of changing `unadjusted_window` on the lazy-mainline branch to filter out the non-market minutes ala https://github.com/quantopian/zipline/compare/lazy-mainline...store-open-and-close-instead-of-all-minutes

This is instead of the changes to the minute bar reader in https://github.com/quantopian/zipline/pull/1077 which was relying on loading all minutes into memory and a separate `unadjusted_window_for_minutes_function`.

jbredeche:
Looks good. Are you writing the opens in case we need them in the future, or do we also need them now?

ehebert:
We will need them in the next patch on lazy-mainline. Long term the minute
index will be removed in favor of calculating off of the market opens.

zipline/data/minute_bars.py
```diff
@@ -184,6 +205,7 @@ def __init__(self,
                  first_trading_day,
                  rootdir,
                  market_opens,
+                 market_closes,
```
yankees714: Can you update the parameters in the docstring?

yankees714:
Looks good to me as well :+1:

zipline/data/minute_bars.py
ehebert: Done.

PR #1086, opened 2016-03-26
ENH: Filter unadjusted window to market minutes.

ehebert:
Use market closes in minute reader to filter out non-market minutes, by
creating a 'blacklist' of minutes between the early close and would be
market close if the day were regular.

Added because the incoming minute history loader needs an interface to
query for the exact market minutes in a range from the reader.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -15,10 +15,12 @@ from textwrap import dedent
 
 import bcolz
 from bcolz import ctable
-import numpy as np
+from intervaltree import IntervalTree
 from os.path import join
 import json
 import os
+import numpy as np
+from numpy import timedelta64
 import pandas as pd
 from zipline.gens.sim_engine import NANOS_IN_MINUTE
 
@@ -28,7 +30,7 @@ from zipline.data._minute_bar_internal import (
     find_last_traded_position_internal
 )
 
-from zipline.utils.memoize import remember_last
+from zipline.utils.memoize import remember_last, lazyval
 
 US_EQUITIES_MINUTES_PER_DAY = 390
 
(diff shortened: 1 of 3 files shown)


ehebert:
In support of removing `unadjusted_window_for_minutes` in https://github.com/quantopian/zipline/pull/1077 in favor of `unadjusted_window` filtering out early closes by default.

jbredeche:
Can you give an overview of the reason behind this change? What's the IntervalTree for, and why is this better than what we had before?

ehebert:
This change is so that the equity history loader in https://github.com/quantopian/zipline/pull/1077 can ask the reader for the minutes in the market minute range without needing to check for the rare case where there are market minutes.

The implementation of making `unadjusted_window` market minute aware by using the interval tree is better than the implementation in the current #1077, because here pulling a full index of the minutes available into memory is not required.

By only storing the exclusions, we take advantage of the fact that early market closes only occur 2 or 3 times a year.

Because the number of actual exclusion ranges is so small we may be able to use a more basic structure, but the semantics of the interval tree, especially in answering the question 'does this range overlap' is worth the slight upfront cost in complexity/pre-calculation.

Also, just update the PR so that the frame of reference for the values in the tree is the position value of the minutes, so that we do not have to skip back and forth between `Timestamp`, position integer, and int64 value of the `Timestamp`.

zipline/data/minute_bars.py
```diff
@@ -587,6 +585,78 @@ def _get_metadata(self):
...
+        occurs on days which had an early close, i.e. days where the close
+        based on the regular period of minutes per day and the market close
+        do not match.
+
+        Returns:
+        --------
+        List of DatetimeIndex representing the minutes to exclude because
+        of early closes.
+        """
+        market_opens = self._market_opens.values.astype('datetime64[m]')
```
jbredeche: can we calculate this once in `__init__`?

zipline/data/minute_bars.py
```diff
@@ -587,6 +585,78 @@ def _get_metadata(self):
...
+        market_closes = self._market_closes.values.astype('datetime64[m]')
+        minutes_per_day = (
+            self._market_closes.values.astype('datetime64[m]')
+            -
+            self._market_opens.values.astype('datetime64[m]')
```
jbredeche: you already calculated these values a few lines above

zipline/data/minute_bars.py
jbredeche: oh, this is only done once, nvm.

zipline/data/minute_bars.py
ehebert: Good catch, will change over to using those. (Had intended to switch over to those during a refactoring.)

zipline/data/minute_bars.py
```diff
@@ -566,9 +563,10 @@ def __init__(self, rootdir):
 
         self._first_trading_day = metadata.first_trading_day
 
-        # store a list of "minute epoch" values
-        self._minute_value_index = \
-            metadata.minute_index.astype('datetime64[m]').astype(int)
+        self._market_opens = metadata.market_opens
+        self._market_open_values = metadata.market_opens.values.\
+            astype('datetime64[m]').astype(int)
+        self._market_closes = metadata.market_closes
```
jbredeche: why are the opens being coerced to `datetime64[m]` but not the closes?

zipline/data/minute_bars.py
ehebert: Since we only call this function during tree init, no need to hold on to the values as `datetime64[m]`, at least at this time.

PR #1092, opened 2016-03-29
FIX: Refactor Q1 history API logic so that algoproxy can use it

lianga888:
Allow Qexec and Zipline to share `get_history_window`, which holds the logic that deals with midnight datetimes in before_trading_start


jbredeche:
nice work, LGTM

PR #1093, opened 2016-03-29
MAINT: add dividends datasets to inits

mtydykov:
Adds dividends data/loaders to inits.

### DIFF
--- a/zipline/pipeline/factors/events.py
+++ b/zipline/pipeline/factors/events.py
@@ -3,15 +3,13 @@ Factors describing information about event data (e.g. earnings
 announcements, acquisitions, dividends, etc.).
 """
 from numpy import newaxis
-from zipline.pipeline.data.buyback_auth import (
+from ..data import (
     CashBuybackAuthorizations,
-    ShareBuybackAuthorizations
-)
-from zipline.pipeline.data.dividends import (
+    ShareBuybackAuthorizations,
     DividendsByAnnouncementDate,
-    DividendsByExDate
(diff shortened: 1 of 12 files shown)


zipline/pipeline/factors/events.py
```diff
@@ -3,15 +3,15 @@
...
     CashBuybackAuthorizations,
     ShareBuybackAuthorizations
 )
-from zipline.pipeline.data.dividends import (
+from ..data import (
     DividendsByAnnouncementDate,
     DividendsByExDate
 )
-from zipline.pipeline.data.earnings import EarningsCalendar
+from zipline.pipeline.data import EarningsCalendar
```
llllllllll: All three of these groups should now be merged

zipline/pipeline/loaders/blaze/dividends.py
```diff
@@ -6,12 +6,12 @@
...
     TS_FIELD_NAME,
 )
-from zipline.pipeline.data.dividends import (
+from ...data import (
     DividendsByExDate,
     DividendsByAnnouncementDate,
     DividendsByPayDate
 )
-from zipline.pipeline.loaders.dividends import (
+from ...loaders import (
```
llllllllll: I think that two levels up should be the max for relative imports, after that I think it stops helping the readability.

mtydykov: Should we go back to the way it was before for these, then?

PR #1095, opened 2016-03-30
Pass a mask (filter) to custom factors

dmichalowicz:
When a custom factor is passed a mask, it computes only over the assets for which the mask returns `True`. This allows computationally expensive factors to not have to compute over an entire universe of stocks.

### DIFF
--- a/tests/pipeline/test_engine.py
+++ b/tests/pipeline/test_engine.py
@@ -62,6 +62,7 @@ from zipline.pipeline.factors import (
     MaxDrawdown,
     SimpleMovingAverage,
 )
+from zipline.pipeline.term import NotSpecified
 from zipline.testing import (
     make_rotating_equity_info,
     make_simple_equity_info,
@@ -93,6 +94,14 @@ class AssetID(CustomFactor):
 
     def compute(self, today, assets, out, close):
         out[:] = assets
+
+
+class OpenPrice(CustomFactor):
+    window_length = 1
+    inputs = [USEquityPricing.open]
+
+    def compute(self, today, assets, out, open):
+        out[:] = open
 
 
 def assert_multi_index_is_product(testcase, index, *levels):
@@ -353,6 +362,62 @@ class ConstantInputTestCase(TestCase):
             avg_result,
             DataFrame(expected_avg, index=dates, columns=self.assets),
         )
+
+    def test_masked_factor(self):
+        """
+        Test that a Custom Factor computes the correct values when passed a
+        mask. The mask/filter should be applied prior to computing any values,
+        as opposed to computing the factor across the entire universe of
+        assets. Any assets that are filtered out should be filled with missing
+        values.
+        """
+        loader = self.loader
+        dates = self.dates[5:10]
+        assets = self.assets
+        asset_ids = self.asset_ids
+        constants = self.constants
+        num_dates = len(dates)
+        open = USEquityPricing.open
+        engine = SimplePipelineEngine(
+            lambda column: loader, self.dates, self.asset_finder,
+        )
+
+        # These are the expected values for the OpenPrice factor. If we pass
+        # OpenPrice a mask, any assets that are filtered out should have all
+        # NaN values. Otherwise, we expect its computed values to be the
+        # asset's open price.
+        values = array([constants[open]] * num_dates, dtype=float)
+        missing_values = array([nan] * num_dates)
+
+        for asset_id in asset_ids:
+            mask = AssetID() <= asset_id
+            factor1 = OpenPrice(mask=mask)
+
+            # Test running our pipeline both with and without a second factor.
+            # We do not explicitly test the resulting values of the second
+            # factor; we just want to implicitly ensure that the addition of
+            # another factor to the pipeline term graph does not cause any
+            # unexpected exceptions when calling `run_pipeline`.
+            for factor2 in (None,
+                            RollingSumDifference(mask=NotSpecified),
+                            RollingSumDifference(mask=mask)):
+                if factor2 is None:
+                    columns = {'factor1': factor1}
+                else:
+                    columns = {'factor1': factor1, 'factor2': factor2}
+                pipeline = Pipeline(columns=columns)
+                results = engine.run_pipeline(pipeline, dates[0], dates[-1])
+                factor1_results = results['factor1'].unstack()
+
+                expected = {
+                    asset: values if asset.sid <= asset_id else missing_values
+                    for asset in assets
+                }
+
+                assert_frame_equal(
+                    factor1_results,
+                    DataFrame(expected, index=dates, columns=assets),
+                )
 
     def test_rolling_and_nonrolling(self):
         open_ = USEquityPricing.open
--- a/tests/pipeline/test_engine.py
+++ b/tests/pipeline/test_engine.py
@@ -397,30 +397,36 @@ class ConstantInputTestCase(TestCase):
             expected_values = where(mask, expected_value, nan)
             return DataFrame(expected_values, index=dates, columns=assets)
 
-        # Produce a mask that looks like:
-        #
-        #       Equity(0 [A])  Equity(1 [B])  Equity(2 [C])  Equity(3 [D])
-        # Day 1          True           True           True          False
-        # Day 2          True           True          False          False
-        # Day 3          True          False          False          False
-        #
         cascading_mask = AssetIDPlusDay() < (asset_ids[-1] + dates[0].day)
+        expected_cascading_mask_result = array(
+            [[True,  True,  True, False],
+             [True,  True, False, False],
+             [True, False, False, False]],
+            dtype=bool,
+        )
 
-        # And another one that looks like:
-        #
-        #       Equity(0 [A])  Equity(1 [B])  Equity(2 [C])  Equity(3 [D])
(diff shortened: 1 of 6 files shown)


dmichalowicz:
@ssanderson Ready for a look

tests/pipeline/test_engine.py
```diff
@@ -354,6 +363,62 @@ def test_numeric_factor(self):
...
+        open = USEquityPricing.open
+        engine = SimplePipelineEngine(
+            lambda column: loader, self.dates, self.asset_finder,
+        )
+
+        # These are the expected values for the OpenPrice factor. If we pass
+        # OpenPrice a mask, any assets that are filtered out should have all
+        # NaN values. Otherwise, we expect its computed values to be the
+        # asset's open price.
+        values = array([constants[open]] * num_dates, dtype=float)
```
ssanderson: you want to explicitly use `float64` here, otherwise you'll get float32 on 32-bit systems.

tests/pipeline/test_engine.py
```diff
@@ -354,6 +363,62 @@ def test_numeric_factor(self):
...
+        missing_values = array([nan] * num_dates)
```
ssanderson: `np.full(nan, num_dates)`

tests/pipeline/test_engine.py
```diff
@@ -354,6 +363,62 @@ def test_numeric_factor(self):
...
+
+        for asset_id in asset_ids:
```
ssanderson: rather than run separate pipelines for each asset, could we just build all the terms we want and run one pipeline here?

tests/pipeline/test_engine.py
```diff
@@ -354,6 +363,62 @@ def test_numeric_factor(self):
...
+            mask = AssetID() <= asset_id
+            factor1 = OpenPrice(mask=mask)
+
+            # Test running our pipeline both with and without a second factor.
+            # We do not explicitly test the resulting values of the second
```
ssanderson: Is there a reason not to test this?

tests/pipeline/test_engine.py
```diff
@@ -354,6 +363,62 @@ def test_numeric_factor(self):
...
+            # factor; we just want to implicitly ensure that the addition of
+            # another factor to the pipeline term graph does not cause any
+            # unexpected exceptions when calling `run_pipeline`.
+            for factor2 in (None,
```
ssanderson: This is only ever testing with a mask that's constant w/r/t each asset.  We should probably test with a filter that sometimes returns True and other times returns False.

zipline/pipeline/engine.py
```diff
@@ -242,8 +242,11 @@ def _mask_and_dates_for_term(self, term, workspace, graph, dates):
         Load mask and mask row labels for term.
         """
         mask = term.mask
-        offset = graph.extra_rows[mask] - graph.extra_rows[term]
-        return workspace[mask][offset:], dates[offset:]
+        mask_offset = graph.extra_rows[mask] - graph.extra_rows[term]
+        dates_offset = (
+            graph.extra_rows[self._root_mask_term] - graph.extra_rows[term]
```
ssanderson: Might be worth commenting here that we're comparing against `_root_mask_term` because that's what determines the shape of the top-level dates array.

ssanderson: Also might be worth renaming the `dates` parameter to something like `all_dates` to be clear that we're slicing of the portion that's actually relevant to the particular term we're computing.

zipline/pipeline/graph.py
```diff
@@ -190,8 +190,23 @@ def _add_to_graph(self, term, parents, extra_rows):
...
 
+        if isinstance(term, ComputableTerm):
+            # For computable terms, we want to manually add the term's mask to
+            # the graph with zero extra rows. A computable term does not
+            # directly require its mask to have any extra rows. Only loadable
+            # terms should dictate how many extra rows a mask should compute.
+            self._add_to_graph(
+                term.mask,
+                parents,
+                extra_rows=0,
```
ssanderson: I think this actually wants to be `extra_rows` here, not 0.  If we need 3 extra rows of a `ComputableTerm`, then we also need 3 extra rows of the mask.  What we **don't** need is an additional `term.extra_input_rows` of the mask.

tests/pipeline/test_engine.py
dmichalowicz: Not in particular, I've added this coverage

tests/pipeline/test_engine.py
dmichalowicz: I've simplified the cases a bit to now only call `run_pipeline` twice; once with a single masked factor and once with two masked factors

tests/pipeline/test_engine.py
dmichalowicz: Good point, I've added this in.

dmichalowicz:
@ssanderson I merged your changes and updated based on your comments.

tests/pipeline/test_engine.py
```diff
@@ -372,52 +381,59 @@ def test_masked_factor(self):
...
+        factor2_expected = create_expected_results(factor2_values)
+
+        # Produce a mask that looks like:
+        #
+        #       Equity(0 [A])  Equity(1 [B])  Equity(2 [C])  Equity(3 [D])
+        # Day 1          True           True           True          False
+        # Day 2          True           True          False          False
+        # Day 3          True          False          False          False
+        #
+        mask = AssetIDPlusDay() < asset_ids[-1] + dates[0].day
```
ssanderson: Might help readability here to put parens around the RHS of the `<` expression.

ssanderson: This isn't hitting the cases where an asset transitions from False -> True.

tests/pipeline/test_engine.py
```diff
@@ -372,52 +381,59 @@ def test_masked_factor(self):
...
-                assert_frame_equal(
-                    factor1_results,
-                    DataFrame(expected, index=dates, columns=assets),
-                )
+        missing_values = full(num_dates, nan)
+        factor1_values = full(num_dates, constants[open], dtype=float64)
+        factor2_values = full(num_dates,
+                              3.0 * (constants[open] - constants[close]))
+
+        def create_expected_results(values):
```
ssanderson: Clearer here might be to just compute the expected mask and then use `np.where(mask, factor_values, nan)`.

ssanderson: This would also make it easier to test with more complex masks.  See note below.

tests/pipeline/test_engine.py
```diff
@@ -372,52 +381,59 @@ def test_masked_factor(self):
...
+        pipeline = Pipeline(columns=columns)
+        results = engine.run_pipeline(pipeline, dates[0], dates[-1])
+        factor1_results = results['factor1'].unstack()
+        assert_frame_equal(factor1_results, factor1_expected)
+
+        # Test running a pipeline with a second factor. This ensures that
+        # adding another factor to the pipeline with a different window length
+        # does not cause any unexpected behavior, especially when both factors
+        # share the same mask.
+        columns['factor2'] = RollingSumDifference(mask=mask)
```
ssanderson: Can we also test that adding the second factor doesn't affect the result of the first factor?

zipline/pipeline/engine.py
```diff
@@ -237,16 +237,20 @@ def _compute_root_mask(self, start_date, end_date, extra_rows):
...
 
-    def _mask_and_dates_for_term(self, term, workspace, graph, dates):
+    def _mask_and_dates_for_term(self, term, workspace, graph, all_dates):
         """
         Load mask and mask row labels for term.
         """
         mask = term.mask
         mask_offset = graph.extra_rows[mask] - graph.extra_rows[term]
+
+        # This offset is computed against _root_mask_term because that is what
```
ssanderson: :+1:

tests/pipeline/test_engine.py
```diff
@@ -8,13 +8,15 @@
 
 from nose_parameterized import parameterized
 from numpy import (
+    append,
```
ssanderson: FWIW, the already-imported `np.concatenate` does the same thing that you're accomplishing with `append`, except it takes a list of arrays instead of two positional args.

ssanderson:
Travis build is failing.

tests/pipeline/test_engine.py
dmichalowicz: Ah yes meant to do that, good catch

tests/pipeline/test_engine.py
```diff
@@ -354,6 +371,78 @@ def test_numeric_factor(self):
...
+        # Day 3         False           True          False           True
+        #
+        alternating_mask = (AssetIDPlusDay() % 2).eq(0)
+
+        for mask in (cascading_mask, alternating_mask):
+            # Test running a pipeline with a single masked factor.
+            columns = {'factor1': OpenPrice(mask=mask), 'mask': mask}
+            pipeline = Pipeline(columns=columns)
+            results = engine.run_pipeline(pipeline, dates[0], dates[-1])
+            mask_results = results['mask'].unstack()
```
ssanderson: Can we assert that the masks are actually as expected here?

ssanderson:
LGTM;  merge on green from Travis.

PR #1096, opened 2016-03-31
DEV: warn and return None if ordering an asset between its delist and autoclose dates

jbredeche:


zipline/algorithm.py
```diff
@@ -114,6 +115,38 @@
 DEFAULT_CAPITAL_BASE = float("1.0e5")
 
 
+log = logbook.Logger("ZiplineLog")
+
+
+def check_asset_lifetime(f):
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        trading_algorithm = args[0]
```
ssanderson: You can just make the signature here `self, sid, **kwargs`.

zipline/algorithm.py
```diff
@@ -907,6 +940,7 @@ def _calculate_order_value_amount(self, asset, value):
         return value / (last_price * value_multiplier)
 
     @api_method
+    @check_asset_lifetime
     def order(self, asset, amount,
```
ssanderson: It looks like we updated just this method to have the first parameter be named `asset` on this branch.  The other `order_*` methods still take a parameter named `sid`.  We should probably pick one or the other.

zipline/algorithm.py
ssanderson: See note below regarding sid/asset naming inconsistency in our other order methods.

zipline/algorithm.py
jbredeche: agreed.  OK with me making it all `asset`?  hopefully that doesn't cause too many legacy breakages.

tests/test_algorithm.py
```diff
@@ -3194,3 +3195,79 @@ def transactions_for_date(date):
...
+            env=self.env,
+            sim_params=SimulationParameters(
+                period_start=pd.Timestamp("2016-01-07", tz='UTC'),
+                period_end=pd.Timestamp("2016-01-07", tz='UTC'),
+                env=self.env,
+                data_frequency="minute"
+            )
+        )
+
+        log_catcher = TestHandler()
```
richafrank: We should use [`make_test_handler`] (https://github.com/quantopian/zipline/blob/6b3560ade8ac9294fd4b5db71948881b6cc673ab/zipline/testing/core.py#L905-L905), so the handler doesn't leak stack frames.

zipline/algorithm.py
```diff
@@ -114,6 +115,38 @@
...
+        if asset.auto_close_date:
+            if asset.end_date < current_dt < asset.auto_close_date:
+                # we are between the asset's end date and auto close date,
+                # so warn the user that they can't place an order for this
+                # asset, and return None.
+                log.warn("Cannot place order for {0}, as it has de-listed. "
+                         "Any existing positions for this asset will be "
+                         "liquidated on "
+                         "{1}.".format(asset.symbol, asset.auto_close_date))
+                return None
```
ssanderson: Thoughts on returning an `InvalidOrder` or `CancelledOrder` value here rather than using None?  The main benefit there would be that it would be immediately obvious what happened if you try to use the return value of an order and failed.  As is, if you want to get the number of shares placed from an order_target_* call, you might reasonably do something like:

```
amount = order_target_percent(sid(24), 0.5).amount
```
that will **almost always** work, but will fail with something like `NoneType object has no attribute 'amount'`, which is n't the friendliest.

ssanderson: (We actually have this problem for order_target_percent already, if you already hold the amount you want.  I wonder if we should just return a filled order of 0 for that case.)

jbredeche: Yeah, we ran into this problem with some legacy algorithms.  Since we already have this problem, and this bandaid is likely going to be rarely applied, I'd rather contain scope for now and spend time coming up with a good design.

PR #1098, opened 2016-03-31
BUG: Stop blowing up when trying to do week_start with trading day zero

abhijeetkalyan:
Small bugfix for when we tried to call week_start.should_trigger(trading_days[0]). 

@jbredeche mind taking a look?


jbredeche:
LGTM

PR #1101, opened 2016-03-31
Fixes to context.portfolio and context.portfolio crashes

lianga888:
- Minute bar reader is now compatible with non trading minutes
- Calls to context.portfolio and context.account in BTS should adjust for overnight splits

IMPORTANT: Only the first commit passes right now. The second commit breaks tests for daily mode algos

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -1066,7 +1066,13 @@ class TradingAlgorithm(object):
 
         self.portfolio_needs_update = True
         self.account_needs_update = True
-        self.performance_needs_update = True
+
+        # Do not update performance if we are at midnight. Portfolio and
+        # Account have already been adjusted for splits
+        if not normalize_date(dt) == dt:
+            self.performance_needs_update = True
+        else:
+            self.performance_needs_update = False
 
     @api_method
     def get_datetime(self, tz=None):
--- a/zipline/data/_minute_bar_internal.pyx
+++ b/zipline/data/_minute_bar_internal.pyx
@@ -37,10 +37,50 @@ def minute_value(ndarray[long_t, ndim=1] market_opens,
     return market_opens[q] + r
 
 def find_position_of_minute(ndarray[long_t, ndim=1] market_opens,
+                            ndarray[long_t, ndim=1] market_closes,
                             long_t minute_val,
                             short minutes_per_day):
     """
     Finds the position of a given minute in the given array of market opens.
+    If not a market minute, adjusts to the last market minute.
+
+    Parameters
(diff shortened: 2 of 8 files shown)


zipline/finance/performance/tracker.py
```diff
@@ -194,7 +194,7 @@ def get_portfolio(self, performance_needs_update, dt):
         if performance_needs_update:
             self.position_tracker.sync_last_sale_prices(dt)
             self.update_performance()
-            self.account_needs_update = True
+        self.account_needs_update = True
```
jbredeche: was this change intentional?

zipline/algorithm.py
```diff
@@ -1066,7 +1066,13 @@ def on_dt_changed(self, dt):
 
         self.portfolio_needs_update = True
         self.account_needs_update = True
-        self.performance_needs_update = True
+
+        # Do not update performance if we are at midnight. Portfolio and
+        # Account have already been adjusted for splits
+        if not normalize_date(dt) == dt:
```
jbredeche: All these flags are ripe for refactoring, as @ehebert suggested yesterday, but that's out of scope for now.

Regardless, I'll have to profile this to see if we can afford an extra 390 `normalize_date` calls per day (I think probably not).  If not, I have some ideas of how to avoid this check.

zipline/data/_minute_bar_internal.pyx
```diff
@@ -37,10 +37,50 @@ def minute_value(ndarray[long_t, ndim=1] market_opens,
     return market_opens[q] + r
 
 def find_position_of_minute(ndarray[long_t, ndim=1] market_opens,
+                            ndarray[long_t, ndim=1] market_closes,
```
jbredeche: this whole diff confused me until I realized you had simply renamed the original `find_position_of_minute ` method to `find_position_of_minute_half_day_unadjusted` and added some logic in there to be resilient against day barriers.  I was confused why you had been changing the `IntervalTree` logic elsewhere.

I think it would be easier to parse if we kept `find_position_of_minute` with the same name (and add the small piece of new logic in there), and create a new method called `find_position_of_minute_with_halfdays` that is the totally new method.  I can do this later today if needed.

jbredeche: Do we still need the older logic at all?  Why not make `find_position_of_minute` always be aware of half days instead of having two different methods?

jbredeche:
@aliang can you close this PR and reopen a new one, now that the first part is already merged?

lianga888:
This PR is split in two now

PR #1106, opened 2016-04-04
13d in pipeline

mtydykov:
Adds 13d dataset to pipeline.


tests/pipeline/test_13d_filings.py
```diff
@@ -0,0 +1,105 @@
...
+        PERCENT_SHARES: [10, 20],
+        TS_FIELD_NAME: pd.to_datetime(['2014-01-05', '2014-01-10']),
+        DISCLOSURE_DATE: pd.to_datetime(['2014-01-04', '2014-01-09'])
+    }),
+    pd.DataFrame(
+        columns=[NUM_SHARES,
+                 PERCENT_SHARES,
+                 DISCLOSURE_DATE,
+                 TS_FIELD_NAME],
+        dtype='datetime64[ns]'
```
kglowinski: Doesn't this force all columns in the empty df to have the dtype of dt64[ns]?

tests/pipeline/test_13d_filings.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @classmethod
+    def get_dataset(cls):
+        return {sid: frame
+                for sid, frame
+                in enumerate(_13d_filngs_cases)}
+
+    loader_type = _13DFilingsLoader
+
+    def setup(self, dates):
```
kglowinski: This is potentially a nosetests thing, but what calls setup and gets the cols return?

zipline/pipeline/factors/events.py
```diff
@@ -203,3 +204,14 @@ class BusinessDaysSincePreviousExDate(
     zipline.pipeline.factors.BusinessDaysSinceDividendAnnouncement
     """
     inputs = [DividendsByExDate.previous_date]
+
+
+class BusinessDaysSince13DFilingsDate(
+    BusinessDaysSincePreviousEvents
+):
+    """
+    Factor returning the number of **business days** (not trading days!) since
```
kglowinski: What's the distinction between business days and trading days here?

kglowinski: Oh. You mean actual business days vs trading days as defined in zipline?

mtydykov: I believe so... this is the same for all factors that have been defined

tests/pipeline/test_13d_filings.py
mtydykov: Totally true! Good catch

tests/pipeline/test_13d_filings.py
mtydykov: This is part of the current architecture of event dataset loader tests - they all subclass `WithPipelineEventDataLoader`, which calls `setup` in `test_compute`.

mtydykov:
@kglowinski any more comments on this one?

zipline/pipeline/loaders/_13d_filings.py
```diff
@@ -0,0 +1,54 @@
...
+    :class:`zipline.pipeline.data._13DFilings`.
+
+    events_by_sid: dict[sid -> pd.DataFrame(knowledge date,
+    disclosure date, percent shares, number of shares)]
+
+    """
+    expected_cols = frozenset([DISCLOSURE_DATE,
+                               PERCENT_SHARES,
+                               NUM_SHARES])
+    event_date_col = DISCLOSURE_DATE
```
kglowinski: What does having this column as the `event_date_col` do? Was this not something that was needed before that we recently added?

mtydykov: Yeah, this was introduced in a very recent PR to zipline. Each event dataset now needs to explicitly define an event_date_col which implies certain behavior for other parts of the code.

kglowinski:
Good to go from me. 👍

PR #1108, opened 2016-04-04
MAINT: make the data loading apis more consistent.

llllllllll:
Changes BcolzDailyBarWriter to not be an abc, data is passed as an
iterator of (sid, dataframe) pairs to the write method.

Changes the AssetsDBWriter to be a single class which accepts an engine
at construction time and has a `write` method for writing dataframes for
the various tables. We no longer support writing the various other data
types, callers should coerce their data into a dataframe themselves. See
zipline.assets.synthetic for some helpers to do this.

Adds a lot of test fixtures and updates many tests to use these. This
was done in the process of updating the tests that used one of the
writers above.

PR #1100, opened 2016-03-31
update writers for consistency

llllllllll:
I am splitting this out of the quandl data loader work.

PR #1109, opened 2016-04-04
MAINT: make the data loading apis more consistent.

llllllllll:
Changes BcolzDailyBarWriter to not be an abc, data is passed as an
iterator of (sid, dataframe) pairs to the write method.

Changes the AssetsDBWriter to be a single class which accepts an engine
at construction time and has a `write` method for writing dataframes for
the various tables. We no longer support writing the various other data
types, callers should coerce their data into a dataframe themselves. See
zipline.assets.synthetic for some helpers to do this.

Adds a lot of test fixtures and updates many tests to use these. This
was done in the process of updating the tests that used one of the
writers above.

ref:  #1108. #1100


tests/test_bar_data.py
```diff
@@ -89,115 +88,100 @@ def check_internal_consistency(self, bar_data):
...
-            equity_minute_reader=cls.build_minute_data(),
-            adjustment_reader=cls.adjustments_reader
+        return merge(
+            {
+                sid: create_minute_df_for_asset(
+                    cls.env,
+                    cls.bcolz_minute_bar_days[0],
+                    cls.bcolz_minute_bar_days[-1],
+                )
+                for sid in (1, cls.SPLIT_ASSET_SID)
```
ssanderson: Might be nicer to use `cls.ASSET1` and `cls.ASSET2` here, since we're using symbolic names for the others.

tests/test_bar_data.py
```diff
@@ -89,115 +88,100 @@ def check_internal_consistency(self, bar_data):
...
-            50
+        cls.ILLIQUID_SPLIT_ASSET = cls.asset_finder.retrieve_asset(
+            cls.ILLIQUID_SPLIT_ASSET_SID,
+        )
+        cls.HILARIOUSLY_ILLIQUID_ASSET = cls.asset_finder.retrieve_asset(
+            cls.HILARIOUSLY_ILLIQUID_ASSET_SID,
         )
 
-        return BcolzMinuteBarReader(cls.tempdir.path)
+        cls.ASSETS = [cls.ASSET1, cls.ASSET2]
```
ssanderson: It seems a bit odd that this is `cls.ASSETS` when we have two other named assets here.

tests/test_bar_data.py
```diff
@@ -382,8 +371,12 @@ def test_spot_price_is_unadjusted(self):
     def test_spot_price_is_adjusted_if_needed(self):
         # on cls.days[1], the first 9 minutes of ILLIQUID_SPLIT_ASSET are
         # missing. let's get them.
-        day0_minutes = self.env.market_minutes_for_day(self.days[0])
```
ssanderson: Another thing we could have done here would be to set `self.days` in the testcase's `init_{instance,class}_fixtures`.  That way the tests still get a nice variable name, but we don't clutter the global fixture namespace.  Probably not worth changing now though.

tests/test_bar_data.py
```diff
@@ -522,158 +525,145 @@ def test_overnight_adjustments(self):
...
     @classmethod
-    def create_adjustments_reader(cls):
-        path = cls.tempdir.getpath("test_adjustments.db")
-
-        adj_writer = SQLiteAdjustmentWriter(
-            path,
-            cls.env.trading_days,
-            MockDailyBarReader()
+    def make_equity_info(cls):
+        return pd.DataFrame.from_dict(
```
ssanderson: Is there a reason this isn't just `make_simple_equity_info` with a symbols list?

tests/test_bar_data.py
```diff
@@ -522,158 +525,145 @@ def test_overnight_adjustments(self):
...
-                'sid': cls.ILLIQUID_SPLIT_ASSET.sid
-            }
+                'sid': cls.ILLIQUID_SPLIT_ASSET_SID,
+            },
         ])
 
-        mergers = pd.DataFrame([
+    @classmethod
+    def make_mergers_data(cls):
+        return pd.DataFrame([
```
ssanderson: Is there a reason we changed `make_splits_data` to `DataFrame.from_records`, but left this unchanged?

tests/test_bar_data.py
```diff
@@ -522,158 +525,145 @@ def test_overnight_adjustments(self):
...
+                'sid': cls.ILLIQUID_MERGER_ASSET_SID,
             }
         ])
 
-        # we're using a fake daily reader in the adjustments writer which
-        # returns every daily price as 100, so dividend amounts of 2.0 and 4.0
-        # correspond to 2% and 4% dividends, respectively.
-        dividends = pd.DataFrame([
+    @classmethod
+    def make_dividends_data(cls):
```
ssanderson: ditto above note.

zipline/testing/fixtures.py
```diff
@@ -315,72 +333,218 @@ def init_class_fixtures(cls):
...
+
+    For more advanced configuration, `make_adjustment_writer` may be
+    overwritten directly.
+
+    See Also
+    --------
+    zipline.testing.MockDailyBarReader
+    """
+    @classmethod
+    def _make_data(cls):
```
ssanderson: Thoughts on making this provide empty frames with the right dtypes instead of `None` as the value for "no data"?

zipline/testing/fixtures.py
```diff
@@ -315,72 +333,218 @@ def init_class_fixtures(cls):
...
+    def make_adjustment_writer_daily_bar_reader(cls):
+        return cls.bcolz_daily_bar_reader
 
-        See, zipline.data.minute_bars.BcolzMinuteBarWriter
-        """
-        return {}
+    @classmethod
+    def init_class_fixtures(cls):
+        super(WithAdjustmentReader, cls).init_class_fixtures()
+        cls.adjustment_db_conn = conn = sqlite3.connect(':memory:')
```
ssanderson: Does anyone use this attribute?  Could we just make it an arg to `make_adjustment_writer`?  The only case where that wouldn't work would be if we wanted to test writing into an already-populated db, but in that case we probably shouldn't use the fixture.

zipline/testing/fixtures.py
```diff
@@ -315,72 +333,218 @@ def init_class_fixtures(cls):
...
+      :func:`zipline.pipeline.loaders.synthetic.make_minute_bar_data`.
+    - `cls.bcolz_minute_bar_reader` is a minute bar reader pointing to the
+      directory that was just written to.
 
+    See Also
+    --------
+    WithBcolzDailyBarReader
+    WithDataPortal
+    """
+    BCOLZ_MINUTE_BAR_PATH = 'minute_equity_pricing.bcolz'
```
ssanderson: Are there tests that need to configure this?

ssanderson: might be worth throwing a `test` in the name here somewhere so it's clearer where this came from if you stumble upon it

zipline/testing/fixtures.py
```diff
@@ -315,72 +333,218 @@ def init_class_fixtures(cls):
...
+            cls.asset_finder.sids,
+        )
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(WithBcolzMinuteBarReader, cls).init_class_fixtures()
+        cls.bcolz_minute_bar_path = p = cls.tmpdir.makedir(
+            cls.BCOLZ_MINUTE_BAR_PATH,
+        )
+        cls.bcolz_minute_bar_days = days = (
```
ssanderson: I think you just wanted an if-statement here...

zipline/testing/fixtures.py
```diff
@@ -315,72 +333,218 @@ def init_class_fixtures(cls):
...
-            cls.env.open_and_closes.market_open,
-            cls.env.open_and_closes.market_close,
-            cls.BCOLZ_MINUTES_PER_DAY,
+            days[0],
+            p,
+            cls.env.open_and_closes.market_open.loc[days],
+            cls.env.open_and_closes.market_close.loc[days],
+            US_EQUITIES_MINUTES_PER_DAY
+        )
+        cls.bcolz_minute_bar_data = df_dict = cls.make_minute_bar_data()
```
ssanderson: The multiple assignment here feels like more complexity than it's worth.

zipline/testing/fixtures.py
```diff
@@ -315,72 +333,218 @@ def init_class_fixtures(cls):
...
+        )
 
     @classmethod
     def init_class_fixtures(cls):
-        super(WithBcolzMinutes, cls).init_class_fixtures()
+        super(WithBcolzDailyBarReader, cls).init_class_fixtures()
+        cls.bcolz_daily_bar_path = p = cls.tmpdir.makedir(
+            cls.BCOLZ_DAILY_BAR_PATH,
+        )
+        cls.bcolz_daily_bar_days = days = (
```
ssanderson: I think this just wanted to be an if-statement.

zipline/testing/fixtures.py
```diff
@@ -315,72 +333,218 @@ def init_class_fixtures(cls):
...
+      directory that was just written to.
+
+    See Also
+    --------
+    WithBcolzMinuteBarReader
+    WithDataPortal
     """
+    BCOLZ_DAILY_BAR_PATH = 'daily_equity_pricing.bcolz'
+    BCOLZ_DAILY_BAR_LOOKBACK_DAYS = 0
+    BCOLZ_DAILY_BAR_FROM_CSVS = False
```
ssanderson: This option isn't documented.  When is it used?

tests/test_finance.py
```diff
@@ -174,33 +170,42 @@ def test_alternating_long_short(self):
...
+        # optional parameters
+        # ---------------------
+        # if present, alternate between long and short sales
+        alternate = params.get('alternate')
+
+        # if present, expect transaction amounts to match orders exactly.
+        complete_fill = params.get('complete_fill')
+
+        sid = 1
+        metadata = pd.DataFrame.from_dict(
```
ssanderson: Could this just be a `make_simple_equity_info` call?

llllllllll: sure

tests/test_api_shim.py
```diff
@@ -111,133 +112,77 @@ def handle_data(context, data):
...
-
-        cls.adj_reader = cls.create_adjustments_reader()
-
-        cls.sim_params = SimulationParameters(
-            period_start=cls.trading_days[0],
-            period_end=cls.trading_days[-1],
-            data_frequency="minute",
-            env=cls.env
+    @classmethod
+    def make_equity_info(cls):
```
ssanderson: It seems like most of the invocations of `make_equity_info` could be converted into `make_simple_equity_info` calls.

llllllllll: these equities have custom symbols. I do agree that most calls will probably want to be a `make_*_equity_info` call though.

zipline/testing/core.py
```diff
@@ -465,31 +327,25 @@ def make_trade_data_for_asset_info(dates,
     volumes = (volume_sid_deltas + volume_date_deltas[:, None]) + volume_start
 
     for j, sid in enumerate(sids):
-        start_date = asset_info[sid]['start_date']
-        end_date = asset_info[sid]['end_date']
+        start_date = asset_info.loc[sid, 'start_date']
```
ssanderson: :+1:

zipline/testing/core.py
```diff
@@ -465,31 +327,25 @@ def make_trade_data_for_asset_info(dates,
...
-        df = pd.DataFrame({
-            "open": prices[:, j],
-            "high": prices[:, j],
-            "low": prices[:, j],
-            "close": prices[:, j],
-            "volume": volumes[:, j],
-            date_field: dates_arr
-        })
-
-        if frequency == 'minute':
```
ssanderson: Looks like we weren't setting the index before if the frequency was 'daily'.  Was that just a bug?

zipline/testing/core.py
```diff
@@ -499,148 +355,6 @@ def make_trade_data_for_asset_info(dates,
...
-        root_symbols=root_symbols,
-        years=years,
-        notice_date_func=lambda dt: dt - MonthBegin(2) + nineteen_days,
-        expiration_date_func=lambda dt: dt - MonthBegin(1) + nineteen_days,
-        start_date_func=lambda dt: dt - one_year,
-        month_codes=month_codes,
-    )
-
-
-def make_simple_asset_info(assets, start_date, end_date, symbols=None):
```
ssanderson: This function was never actually used.

zipline/testing/core.py
llllllllll: no, I changed this to use the index instead of the name `day` to make them the same

zipline/testing/core.py
```diff
@@ -853,22 +539,32 @@ def write_minute_data_for_asset(env, writer, start_dt, end_dt, sid,
         for minute in minute_blacklist:
             df.loc[minute] = 0
 
-    writer.write(sid, df)
+    return df
+
+
+def write_minute_data_for_asset(env, writer, start_dt, end_dt, sid,
```
ssanderson: This seems like it's just shuffling arguments around.  Is this just for backward compat with something else?

zipline/testing/core.py
```diff
@@ -853,22 +539,32 @@ def write_minute_data_for_asset(env, writer, start_dt, end_dt, sid,
...
+                                interval=1, start_val=1):
+    writer.writer(
```
ssanderson: Also, should this be `writer.write`?  (Is this actually called anywhere?)

zipline/testing/core.py
```diff
@@ -853,22 +539,32 @@ def write_minute_data_for_asset(env, writer, start_dt, end_dt, sid,
...
+        sid,
+        create_minute_df_for_asset(env, start_dt, end_dt, interval, start_val),
+    )
 
 
 def create_daily_df_for_asset(env, start_day, end_day, interval=1):
     days = env.days_in_range(start_day, end_day)
     days_count = len(days)
-    days_arr = np.array(range(2, days_count + 2))
+    days_arr = np.arange(2, days_count + 2)
```
ssanderson: Even simpler here might be `np.arange(days_count) + 2`

zipline/testing/core.py
```diff
@@ -883,37 +579,39 @@ def create_daily_df_for_asset(env, start_day, end_day, interval=1):
...
+            volumes.append(trade["volume"])
+
+        yield sidint, pd.DataFrame(
+            {
+                "open": opens,
+                "high": highs,
+                "low": lows,
+                "close": closes,
+                "volume": volumes,
+                "day": [day.value for day in index]
```
ssanderson: Not your doing, but this is just `index.values`, right?

zipline/testing/core.py
```diff
@@ -1356,16 +1100,34 @@ def decorator(f):
...
-            # Hackery to make the dtypes correct on an empty frame.
-            'effective_date': np.array([], dtype=int),
-            'ratio': np.array([], dtype=float),
-            'sid': np.array([], dtype=int),
-        },
+        np.array(
+            [],
+            dtype=[
+                ('effective_date', 'int64'),
+                ('ratio', 'float32'),
```
ssanderson: Should this be float64?

zipline/testing/core.py
```diff
@@ -1406,3 +1168,177 @@ def make_test_handler(testcase, *args, **kwargs):
...
+    join(realpath(dirname(__file__)), '..', '..'),
+)
+
+
+def test_resource_path(*path_parts):
+    return os.path.join(zipline_git_root, 'tests', 'resources', *path_parts)
+
+
+@contextmanager
+def patch_os_environment(remove=None, **values):
```
ssanderson: I don't think this is used anywhere.

zipline/testing/core.py
```diff
@@ -1406,3 +1168,177 @@ def make_test_handler(testcase, *args, **kwargs):
...
+                try:
+                    del os.environ[key]
+                except KeyError:
+                    pass
+            else:
+                # Restore the old value.
+                os.environ[old_key] = old_value
+
+
+class tmp_dir(object):
```
ssanderson: This might not actually be an improvement over what it's wrapping.

zipline/testing/core.py
```diff
@@ -1406,3 +1168,177 @@ def make_test_handler(testcase, *args, **kwargs):
...
+        The data to write.
+    path : str, optional
+        The path to the directory to write the data into. If not given, this
+        will be a unique name.
+
+    See Also
+    --------
+    tmp_bcolz_daily_bar_reader
+    """
+    _reader = BcolzMinuteBarReader
```
ssanderson: I might call this `_reader_cls` to be clear that it's the type, not the reader itself.

zipline/testing/fixtures.py
```diff
@@ -126,7 +152,7 @@ def enter_instance_context(self, context_manager):
         return self._instance_teardown_stack.enter_context(context_manager)
 
     @final
-    def add_instance_callback(self, callback):
+    def add_instance_callback(self, callback, *args, **kwargs):
```
ssanderson: Commented this on the original PR: we're not forwarding *args/**kwargs here.

zipline/testing/fixtures.py
```diff
@@ -315,72 +333,218 @@ def init_class_fixtures(cls):
...
     @classmethod
     def init_class_fixtures(cls):
-        super(WithTempdir, cls).init_class_fixtures()
+        super(WithTmpDir, cls).init_class_fixtures()
+        cls.tmpdir = cls.enter_class_context(tmp_dir(cls.TMP_DIR_PATH))
 
-        cls.tempdir = TempDirectory(path=cls.TEMPDIR_PATH)
 
-        cls.add_class_callback(cls.tempdir.cleanup)
+class WithInstanceTmpDir(object):
```
ssanderson: Should we give the fixture generated here a different name so we don't collide if you use both?  I could imagine a test that needs a daily bar reader and also needs a separate per-test tempdir (though maybe not a whole suite of them, in which case maybe we should just make it an error to use both and say you should use TempDirectory directly if you already have the class-level one.)

zipline/testing/fixtures.py
ssanderson: Actually, it looks like all these config options need docs.

zipline/assets/asset_writer.py
```diff
@@ -216,171 +232,174 @@ def write_version_info(version_table, version_value):
...
 
 
-class AssetDBWriter(with_metaclass(ABCMeta)):
+class _empty(object):
+    columns = ()
+
+
+class AssetDBWriter(object):
     """
     Class used to write arbitrary data to SQLite database.
```
ssanderson: This docstring is out of date now.

zipline/assets/asset_writer.py
```diff
@@ -216,171 +232,174 @@ def write_version_info(version_table, version_value):
...
-
-    @abstractmethod
-    def defaultval(self):
-        raise NotImplementedError
-
-    def write_all(self,
-                  engine,
-                  allow_sid_assignment=True):
-        """ Write pre-supplied data to SQLite.
+    def __init__(self, engine, allow_sid_assignment=True):
```
ssanderson: allow_sid_assignment is never used

zipline/assets/asset_writer.py
```diff
@@ -216,171 +232,174 @@ def write_version_info(version_table, version_value):
...
-        self._write_df_to_table(root_symbols, self.futures_root_symbols, bind)
-
-    def _write_futures(self, futures, bind):
-        self._write_assets(futures, self.futures_contracts, 'future', bind)
-
-    def _write_equities(self, equities, bind):
-        self._write_assets(equities, self.equities, 'equity', bind)
-
-    def check_for_tables(self, engine):
+    def _check_for_tables(self, txn):
```
ssanderson: Can we call this `has_all_tables` or something like it?  `check_for_tables` doesn't tell me what is being checked.

zipline/data/us_equity_pricing.py
```diff
@@ -119,54 +104,112 @@ class NoDataOnDate(Exception):
...
+        The names of the columns to check.
+
+    Returns
+    -------
+    truncated : pd.DataFrame
+        ``df`` with rows that contain values that do not fit into a uint32
+        zeroed out.
+    """
+    columns = (column,) + columns
+    df[reduce(op.or_, (df[c] > UINT32_MAX for c in columns))] = 0
```
ssanderson: This might be clearer as `(df[list(columns)] > UINT32_MAX).any(axis=1)`

ssanderson: ^ that should in theory be more performant as well, though it's pandas, so who knows.

zipline/data/us_equity_pricing.py
```diff
@@ -119,54 +104,112 @@ class NoDataOnDate(Exception):
...
+        # we already have a ctable so do nothing
+        return raw_data
+
+    columns = {}
+    for colname in OHLC:
+        coldata = raw_data[colname]
+        check_uint32_safe(coldata.max() * 1000, colname)
+        columns[colname] = coldata * 1000
+
+    winsorise_uint32(raw_data, 'volume', *OHLC)
```
ssanderson: Do we want this to be the unconditional behavior?  I could imagine wanting to treat this as an error.  At the very least, we should probably fire a warning about dropped data.

zipline/data/us_equity_pricing.py
```diff
@@ -119,54 +104,112 @@ class NoDataOnDate(Exception):
...
+    columns = valmap(op.methodcaller('astype', 'uint32'), columns)
+    raw_data.index = dates = raw_data.index.values.astype('datetime64[s]')
+    check_uint32_safe(dates.max().view(int), 'day')
+    columns['day'] = dates.astype('uint32')
+    columns['volume'] = raw_data.volume.astype('uint32')
+    return ctable_from_dict(columns)
+
+
+def _snd_to_ctable(tup):
```
ssanderson: I read this as `send`.  Vowels are your friends.

ssanderson: This is also only used once and it's in a map.  Feels like it should just be a lambda maybe?

tests/pipeline/test_engine.py
```diff
@@ -146,10 +149,23 @@ def compute(self, today, assets, out, *inputs):
         out[:] = sum(inputs).sum(axis=0)
 
 
-class ConstantInputTestCase(TestCase):
+class ConstantInputTestCase(WithTradingEnvironment, ZiplineTestCase):
+    asset_ids = 1, 2, 3
```
ssanderson: These feel like they want to be an argument (i.e. an overridable class-level attribute) to `WithAssetFinder`.  Right now, this test is relying on the fact that these are the default sids for `tmp_asset_finder`, which is pretty far away from here.

tests/pipeline/test_engine.py
```diff
@@ -796,27 +784,30 @@ def test_drawdown(self):
         assert_frame_equal(expected, result)
 
 
-class ParameterizedFactorTestCase(TestCase):
+class ParameterizedFactorTestCase(WithTradingEnvironment, ZiplineTestCase):
+    sids = Int64Index([1, 2, 3])
+
+    @classmethod
+    def make_equity_info(cls):
+        cls.equity_info = ret = make_simple_equity_info(
```
ssanderson: Almost all of the usages of `make_equity_info` are calling `make_simple_equity_info` with a fixed set of sids and a fixed start/end date.  Thoughts on making that the default with configurable ALL_CAPS args to `WithAssetFinder`?

tests/pipeline/test_pipeline_algo.py
```diff
@@ -361,87 +339,77 @@ def spot_price(self, sid, day, column):
...
 
     @classmethod
-    def setUpClass(cls):
-        cls.AAPL = 1
-        cls.MSFT = 2
-        cls.BRK_A = 3
-        cls.assets = [cls.AAPL, cls.MSFT, cls.BRK_A]
-        asset_info = make_simple_equity_info(
+    def make_equity_info(cls):
+        return make_simple_equity_info(
```
ssanderson: See note above re: making this pattern the default behavior via parameters.  Might also include an optional `ASSET_FINDER_EQUITY_SYMBOLS` for this case.

tests/pipeline/test_us_equity_pricing_loader.py
```diff
@@ -257,41 +253,46 @@
...
-        cls.assets = TEST_QUERY_ASSETS
-        cls.asset_info = EQUITY_INFO
-        cls.bcolz_writer = SyntheticDailyBarWriter(
-            cls.asset_info,
-            cls.calendar_days,
+    @classmethod
+    def make_dividends_data(cls):
+        return DIVIDENDS
+
+    class make_adjustment_writer_daily_bar_reader(object):
```
ssanderson: Everywhere else that we mock this out, we just return an instance of `MockDailyBarReader`.

tests/test_bar_data.py
llllllllll: it would need to be ASSET1_SID which is just 1, idk if that actually helps

tests/test_bar_data.py
llllllllll: this is how it was, I imagine some tests use only those two

zipline/testing/fixtures.py
llllllllll: idk, I think `None` to signal empty input is reasonable.

zipline/testing/fixtures.py
llllllllll: that sounds good.

zipline/testing/fixtures.py
llllllllll: this would be a subdir of one of the tmpdirs so idk if people will be confused.

llllllllll: also, I don't think any tests currently configure this but I think it is nice to have control over the shared tmpdir for a suite

zipline/testing/fixtures.py
llllllllll: there is a test that loads test data from a directory mapping sid->path-to-csv

zipline/testing/core.py
llllllllll: yeah, this existed and then I split it into a function that creates the data without writing.

zipline/testing/core.py
llllllllll: this is likely dead code because there is no `writer` method here. I will check and probably remove.

zipline/testing/core.py
llllllllll: I don't think we even need the day field here. buy yeah

zipline/testing/core.py
llllllllll: idk, I will check with the schema

zipline/testing/core.py
llllllllll: This was added to test the new feature that this change started on. We will use this soon.

zipline/testing/core.py
llllllllll: kk

zipline/testing/fixtures.py
llllllllll: we probably should, maybe just `instance_tmpdir`?

tests/test_bar_data.py
llllllllll: probably not, I think the from_records is a bit more clear here. What do you think it should be?

ssanderson:
Comments for the files that github won't show because of the diff size:

### zipline.utils.compat

- Nothing is using the newly-added `mappingproxy`.

### zipline.utils.control_flow

- Nothing is using `invert_unique`.

### zipline.utils.functional

- I don't think any of the additions here are being used.

### zipline.dispatch

I'm a little confused about what this module is for.  I understand generally
that it's defining a namespace for multiple-dispatched functions, but I don't
follow why we're using the module's globals as the namespace.

zipline/data/us_equity_pricing.py
ssanderson: (or just a genexpr)

zipline/testing/core.py
ssanderson: I also sometimes write this specific pattern as `start_date, end_date = df.loc[sid, ['start_date', 'end_date']]`.

zipline/data/us_equity_pricing.py
llllllllll: to be fair there is prior art for the name `snd`: http://hackage.haskell.org/package/base-4.8.1.0/docs/Prelude.html#v:snd

tests/test_bar_data.py
```diff
@@ -100,15 +101,12 @@ class TestMinuteBarData(WithBarDataChecks,
...
-                    'symbol': "ASSET{0}".format(sid)
-                } for sid in [1, 2, 3, 4, 5]
-            },
-            orient='index',
+        sids = [1, 2, 3, 4, 5]
+        return make_simple_equity_info(
+            sids,
+            cls.SIM_PARAMS_START,
+            cls.SIM_PARAMS_END,
+            map('ASSET{}'.format, sids),
```
ssanderson: This is going to to generate a DataFrame full of map objects in Python 3.  I have made this mistake several times.

tests/pipeline/test_engine.py
llllllllll: since we already have make_simple_equity_info, I am not sure how much more clear the tests will be if we lift all the arguments to the class level.

tests/pipeline/test_us_equity_pricing_loader.py
llllllllll: I think I wrote this before i moved this to core, I will update this now.

llllllllll:
- `mappingproxy` was added for the quandl stuff
- I can kill invert_unique.
- I am using unzip (I think), if not I have usages in my other branch, sorry I did a poor job isolating these pieces
- re dispatch: I can make it not use the globals as the namespace, but I think it is nice to be able to import from that single module if you intend on adding a dispatch to an already existing dispatcher with register.

zipline/testing/fixtures.py
```diff
@@ -283,11 +315,26 @@ class WithSimParams(WithTradingEnvironment):
     by putting ``SIM_PARAMS_{argname}`` in the class dict except for the
     trading environment which is overridden with the mechanisms provided by
     ``WithTradingEnvironment``.
+
+    Attributes
+    ----------
+    SIM_PARAMS_YEAR : int
```
ssanderson: :+1:  On using Attributes to document the pseudo-args.

zipline/testing/core.py
llllllllll: ooh, I like that a lot. thank you

zipline/testing/fixtures.py
```diff
@@ -384,10 +445,38 @@ class WithBcolzDailyBarReader(WithTmpDir, WithSimParams):
...
+    Attributes
+    ----------
+    BCOLZ_DAILY_BAR_PATH : str
+        The path inside the tmpdir where this will be written.
+    BCOLZ_DAILY_BAR_LOOKBACK_DAYS : int
+        The number of days of data to add before the first day in the
+        sim_params. This is used when a test needs to use history, in which
+        case this should be set to the largest history window that will be
+        requested.
+    BCOLZ_DAILY_BAR_FROM_CSVS : bool
```
ssanderson: This feels like it just belongs in a subclass that overrides `make_daily_bar_data`.

zipline/testing/fixtures.py
```diff
@@ -359,14 +415,19 @@ class WithInstanceTmpDir(object):
...
 
     def init_instance_fixtures(self):
         super(WithInstanceTmpDir, self).init_instance_fixtures()
-        self.tmpdir = self.enter_instance_context(tmp_dir(self.TMP_DIR_PATH))
+        self.instance_tmpdir = self.enter_instance_context(
+            tmp_dir(self.TMP_DIR_PATH),
+        )
 
 
 class WithBcolzDailyBarReader(WithTmpDir, WithSimParams):
```
ssanderson: It seems odd to me that this inherits from `WithSimParams`.  That to me implies that you need a sim params to create a daily bar reader, which isn't true.

tests/test_bar_data.py
ssanderson: I don't care deeply either way, but it seems like we should be consistent here.  Maybe slight preference for `from_records`?

tests/test_bar_data.py
ssanderson: It just seems strange to me to refer to some of the assets symbolically but refer to others by number.  I don't care strongly though.

tests/finance/test_slippage.py
```diff
@@ -35,37 +35,28 @@
...
     WithDataPortal,
+    WithSimParams,
     ZiplineTestCase,
 )
 
 
-class SlippageTestCase(WithDataPortal, ZiplineTestCase):
-    SIM_PARAMS_START = pd.Timestamp('2006-01-05 14:31', tz='utc')
-    SIM_PARAMS_END = pd.Timestamp('2006-01-05 14:36', tz='utc')
+class SlippageTestCase(WithSimParams, WithDataPortal, ZiplineTestCase):
```
ssanderson: This test feels much cleaner to me now :+1: .

tests/finance/test_slippage.py
```diff
@@ -35,37 +35,28 @@
...
+    START_DATE = pd.Timestamp('2006-01-05 14:31', tz='utc')
+    END_DATE = pd.Timestamp('2006-01-05 14:36', tz='utc')
     SIM_PARAMS_CAPITAL_BASE = 1.0e5
     SIM_PARAMS_DATA_FREQUENCY = 'minute'
     SIM_PARAMS_EMISSION_RATE = 'daily'
 
-    sids = 133,
+    ASSET_FINDER_EQUITY_SIDS = 133,
```
ssanderson: You know what my comment here is...

tests/pipeline/test_engine.py
```diff
@@ -517,17 +506,9 @@ def expected_for_col(col):
...
-
-    @classmethod
-    def make_equity_info(cls):
-        return make_simple_equity_info(
-            cls.asset_ids,
-            start_date=cls.start,
-            end_date=cls.end,
-        )
+    asset_ids = ASSET_FINDER_EQUITY_SIDS = 1, 2, 3
+    start = ASSET_FINDER_EQUITY_START_DATE = Timestamp('2015-01-01', tz='utc')
```
ssanderson: Does this test specifically want ASSET_FINDER_START_*, or could it just set the default START_DATE / END_DATE?

tests/test_algorithm.py
```diff
@@ -1010,7 +987,7 @@ def make_minute_bar_data(cls):
 
     @classmethod
     def make_daily_bar_data(cls):
-        for sid in (1, 2):
+        for sid in cls.ASSET_FINDER_EQUITY_SIDS:
```
ssanderson: :+1:

tests/test_algorithm.py
```diff
@@ -1202,9 +1179,10 @@ def handle_data(context, data):
 
 class TestAlgoScript(WithLogger,
                      WithDataPortal,
+                     WithSimParams,
                      ZiplineTestCase):
-    SIM_PARAMS_END = None
```
ssanderson: Why the change from None to a fixed end?

tests/test_algorithm.py
```diff
@@ -1617,25 +1590,12 @@ def handle_data(context, data):
...
 
 
-class TestTradingControls(WithDataPortal, ZiplineTestCase):
-    SIM_PARAMS_END = None
-    SIM_PARAMS_NUM_DAYS = 4
+class TestTradingControls(WithSimParams, WithDataPortal, ZiplineTestCase):
+    START_DATE = pd.Timestamp('2006-01-03', tz='utc')
+    END_DATE = pd.Timestamp('2006-01-06', tz='utc')
 
     sid = 133
```
ssanderson: Is `sid` used in this test suite?

tests/test_algorithm.py
llllllllll: I am just avoiding the num_days constructor because it doesn't play as nicely with `WithDefaultDateBounds`

tests/test_algorithm.py
```diff
@@ -1952,7 +1913,8 @@ def test_asset_date_bounds(self):
             'start_date': self.sim_params.period_start,
             'end_date': '2020-01-01',
         }])
-        with tmp_dir() as tempdir, tmp_trading_env(equities=metadata) as env:
+        with TempDirectory() as tempdir, \
```
ssanderson: I think this `with` and the one above it really want to be `with temp_data_portal(...)`.

tests/test_algorithm.py
```diff
@@ -2002,18 +1966,11 @@ def test_asset_date_bounds(self):
...
+    END_DATE = pd.Timestamp('2006-01-06', tz='utc')
 
-    @classmethod
-    def make_equity_info(cls):
-        return make_simple_equity_info(
-            [cls.sidint],
-            cls.SIM_PARAMS_START,
-            cls.SIM_PARAMS_START + timedelta(days=7),
-        )
+    sidint, = ASSET_FINDER_EQUITY_SIDS = 133,
```
ssanderson: Ibid

tests/test_algorithm.py
```diff
@@ -2157,10 +2114,10 @@ def handle_data(algo, data):
...
 
 
-class TestFutureFlip(WithDataPortal, ZiplineTestCase):
-    SIM_PARAMS_START = pd.Timestamp('2006-01-09', tz='utc')
-    SIM_PARAMS_END = pd.Timestamp('2006-01-10', tz='utc')
-    sid = 1
+class TestFutureFlip(WithSimParams, WithDataPortal, ZiplineTestCase):
+    START_DATE = pd.Timestamp('2006-01-09', tz='utc')
+    END_DATE = pd.Timestamp('2006-01-10', tz='utc')
+    sid, = ASSET_FINDER_EQUITY_SIDS = 1,
```
ssanderson: Ibid

tests/test_algorithm.py
```diff
@@ -2244,10 +2201,14 @@ def analyze(context, perf):
...
+                           WithSimParams,
                            ZiplineTestCase):
 
-    SIM_PARAMS_START = pd.Timestamp('2016-01-05', tz='utc')
-    SIM_PARAMS_END = pd.Timestamp('2016-01-07', tz='utc')
+    START_DATE = pd.Timestamp('2016-01-05', tz='utc')
+    END_DATE = pd.Timestamp('2016-01-07', tz='utc')
+
+    ASSET_FINDER_EQUITY_SIDS = 1,
+    ASSET_FINDER_EQUITY_SYMBOLS = 'ASSET1',
```
ssanderson: Ibid

tests/test_perf_tracking.py
```diff
@@ -624,7 +604,7 @@ def test_long_position_receives_dividend(self):
             env=self.env
         )
 
-        dbpath = self.tmpdir.getpath('adjustments.sqlite')
+        dbpath = self.instance_tmpdir.getpath('adjustments.sqlite')
```
ssanderson: :+1:

tests/test_security_list.py
```diff
@@ -258,7 +258,7 @@ def test_algo_without_rl_violation_after_delete(self):
             'start_date': sim_params.period_start,
             'end_date': sim_params.period_end,
         }])
-        with tmp_dir() as new_tempdir, \
+        with TempDirectory() as new_tempdir, \
```
ssanderson: This is another one that's really `with temp_data_portal` I think.

zipline/assets/asset_writer.py
```diff
@@ -237,25 +237,19 @@ class _empty(object):
...
     ----------
     engine : Engine or str
-        An SQLAlchemy engine to a SQL database.
-    allow_sid_assignment: bool, optional
-        If True then the instance can assign sids where necessary.
+        An SQLAlchemy engine or path to a SQL database.
     """
     DEFAULT_CHUNK_SIZE = SQLITE_MAX_VARIABLE_NUMBER
 
-    def __init__(self, engine, allow_sid_assignment=True):
```
ssanderson: :+1:

zipline/data/us_equity_pricing.py
```diff
@@ -111,13 +118,17 @@ def check_uint32_safe(value, colname):
         )
 
 
-def winsorise_uint32(df, column, *columns):
+@expect_element(invalid_data_behavior={'warn', 'raise', 'ignore'})
+@preprocess(columns=expect_non_empty)
```
ssanderson: Do you think this signature is preferable to just doing:

```
def func(arg1, arg2, column, *columns):
```
?

zipline/data/us_equity_pricing.py
```diff
@@ -111,13 +118,17 @@ def check_uint32_safe(value, colname):
...
+def winsorise_uint32(df, invalid_data_behavior, *columns):
     """Drops any record where a value would not fit into a uint32.
 
     Parameters
     ----------
     df : pd.DataFrame
         The dataframe to winsorise.
+    invalid_data_behavior : {'warn', 'raise', 'ignore'}
```
ssanderson: This feels a little verbose to me, but I didn't think of a better name in 5 minutes.  Maybe `onerror`?

zipline/data/us_equity_pricing.py
```diff
@@ -127,8 +138,14 @@ def winsorise_uint32(df, column, *columns):
...
         zeroed out.
     """
-    columns = (column,) + columns
-    df[reduce(op.or_, (df[c] > UINT32_MAX for c in columns))] = 0
+    mask = (df[list(columns)] > UINT32_MAX).any(axis=1)
+    if mask.any():
+        if invalid_data_behavior == 'raise':
+            raise ValueError('Values out of bounds for uint32: %r' % df[mask])
+        if invalid_data_behavior == 'warn':
+            warnings.warn('Values out of bounds for uint32: %r' % df[mask])
```
ssanderson: This might want to explicitly say that we're ignoring those values.  Also, it'd be nice to include the count of bad values here (`mask.sum()` should work for this).

zipline/testing/fixtures.py
```diff
@@ -164,6 +166,39 @@ def add_instance_callback(self, callback):
...
+
+    Parameters
+    ----------
+    attr_name : str
+        The name of the attribute to alias.
+
+    Returns
+    -------
+    p : classproperty
+        A class property that does the propert aliasing.
```
ssanderson: Missing a "y" in property here.

zipline/testing/fixtures.py
```diff
@@ -164,6 +166,39 @@ def add_instance_callback(self, callback):
...
+    """
+    return classproperty(flip(getattr, attr_name))
```
ssanderson: A usage example in the docstring here would be nice for the non-functionally-inclined.

zipline/data/us_equity_pricing.py
llllllllll: idk how I feel about that because it doesn't tell me what the error is. This is not an argument you will be passing often so I think it is okay to be a very clear here.

zipline/testing/fixtures.py
```diff
@@ -224,18 +272,36 @@ class WithAssetFinder(object):
     zipline.testing.make_future_info
     zipline.testing.make_commodity_future_info
     """
+    ASSET_FINDER_EQUITY_SIDS = tuple(map(ord, 'ABC'))
```
ssanderson: This is substantially less clear than just `ord('A'), ord('B'), ord('C')`.  In particular, it's not obvious at a glance that there are 3 sids here.

zipline/data/us_equity_pricing.py
llllllllll: good call

zipline/testing/fixtures.py
```diff
@@ -224,18 +272,36 @@ class WithAssetFinder(object):
...
+    ASSET_FINDER_EQUITY_SYMBOLS = None
+    ASSET_FINDER_EQUITY_START_DATE = alias('START_DATE')
+
+    @classproperty
+    def ASSET_FINDER_EQUITY_END_DATE(cls):
+        # add one day to end to make sure the asset exists through the end
+        # of the data
+        return cls.END_DATE + pd.Timedelta(days=1)
```
ssanderson: Is there a reason we add a 1 day buffer at the end but not the start?

llllllllll: I don't think assets are available on their end date, I think it is `[start, end)`. I could be wrong in which case I will remove the extra day.

zipline/testing/fixtures.py
```diff
@@ -480,8 +548,10 @@ class WithBcolzDailyBarReader(WithTmpDir, WithSimParams):
     """
     BCOLZ_DAILY_BAR_PATH = 'daily_equity_pricing.bcolz'
     BCOLZ_DAILY_BAR_LOOKBACK_DAYS = 0
-    BCOLZ_DAILY_BAR_FROM_CSVS = False
     BCOLZ_DAILY_BAR_USE_FULL_CALENDAR = False
+    BCOLZ_DAILY_BAR_START_DATE = alias('START_DATE')
+    BCOLZ_DAILY_BAR_END_DATE = alias('END_DATE')
+    _write_method_name = 'write'
```
ssanderson: Might be worth commenting that this exists to be overridden in a subclass, and that tests that want to change the write method should inherit from the other mixin rather than overridding this attribute (which is why the name is lowercase).

zipline/utils/data.py
```diff
@@ -399,82 +390,3 @@ def _update_buffer(self, frame):
             self.buffer.loc[non_nan_items, :, non_nan_cols])
 
         self.buffer = new_buffer
-
-
-class SortedDict(MutableMapping):
```
ssanderson: Are we no longer using this?

zipline/testing/fixtures.py
llllllllll: I think the bigger indication that this is not the supported way to modify this is that it is private. I can add a quick comment explaining why this exists though.

zipline/utils/input_validation.py
```diff
@@ -130,6 +133,19 @@ def ensure_timezone(func, argname, arg):
     )
 
 
+def expect_non_empty(func, argname, arg):
```
ssanderson: So far all of the other functions here named `expect_<thing>` take **kwargs that get mapped to the arguments.  That seems like a worthwhile pattern to maintain.  Maybe this should be `expect_length_between` and take **lower_and_upper_bounds as pairs of ints?

zipline/utils/data.py
llllllllll: dead

zipline/utils/preprocess.py
```diff
@@ -84,26 +84,19 @@ def _decorator(f):
...
-                "Can't validate functions that take *args: %s" % argspec
-            )
+            args_defaults.append((varargs, NO_DEFAULT))
         if varkw:
-            raise TypeError(
-                "Can't validate functions that take **kwargs: %s" % argspec
-            )
+            args_defaults.append((varkw, NO_DEFAULT))
+
+        argset = set(args) | {varargs, varkw} - {None}
```
ssanderson: Can we put parens on the RHS here?  I don't have immediate intuition for whether `|` or `-` binds tighter.

zipline/utils/input_validation.py
llllllllll: I thought the more important pattern was between `expect` and `ensure`. I could do that but I still would want a named version for `non_empty` which is a more common case imho

zipline/utils/preprocess.py
llllllllll: this is the same regardless of how it binds.

zipline/utils/preprocess.py
```diff
@@ -172,21 +171,29 @@ def make_processor_assignment(arg, processor_name):
...
     assignments = []
+    star_map = {
+        varargs: '*',
+        varkw: '**',
+    }
+
+    def argname(arg):
+        return star_map.get(arg, '') + arg
+
     for arg, default in args_defaults:
```
ssanderson: Should we just `keymap` argname over `args_defaults`?

ssanderson: Ah, nvm, there's an if-statement where we need the bound name below.

zipline/utils/preprocess.py
```diff
@@ -172,21 +171,29 @@ def make_processor_assignment(arg, processor_name):
     signature = []
     call_args = []
     assignments = []
+    star_map = {
+        varargs: '*',
+        varkw: '**',
+    }
+
+    def argname(arg):
```
ssanderson: I might call this `name_as_arg` to help clarify that this is the name as it will appear in the signature, as distinguished from the name as it will be used in the body of the function.

zipline/utils/preprocess.py
ssanderson: fair point

zipline/testing/fixtures.py
ssanderson: `end_date` is the last day on which we have trades for an asset.

zipline/utils/data.py
ssanderson: RIP in pieces

tests/test_algorithm.py
llllllllll: yes, the first usage is on 1603, then it is used in a lot of the tests.

tests/test_algorithm.py
llllllllll: I agree but I think we can write this later.

PR #1110, opened 2016-04-05
Env markers

richafrank:
Use environment markers instead of a separate requirements file for py2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,19 +107,37 @@ STR_TO_CMP = {
     '>=': ge,
 }
 
+SYS_VERSION = '.'.join(list(map(str, sys.version_info[:3])))
 
-def _filter_requirements(lines_iter, filter_names=None):
+
+def _filter_requirements(lines_iter, filter_names=None,
+                         filter_sys_version=False):
     for line in lines_iter:
         line = line.strip()
-        if line and not line.startswith('#'):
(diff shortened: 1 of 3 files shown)


richafrank:
@llllllllll What do you think of this replacement for requirements_py2?

setup.py
```diff
@@ -160,30 +157,40 @@ def _with_bounds(req):
...
+            return 'numpy x.x'
+
+        formatted = '%s %s%s' % ((name,) + m.group('comp', 'spec'))
+        pycomp, pyspec = m.group('pycomp', 'pyspec')
+        if pyspec:
+            comp = STR_TO_CMP[pycomp]
+            pyver_spec = StrictVersion(pyspec)
+            selector = []
+
+            for pyver in ('2.7', '3.4'):
```
llllllllll: I think we could just do something like: `selector = 'int(py) %s %s' % (pycomp, pyver_spec)`. Conda selectors have the name `py` available as a 2 character string like `'34'`.

llllllllll:
just had one comment about the translation to a conda selector, but otherwise lgtm. I wonder if we should make a package like `versioneer` that has our cython stuff and now our conda translation. I feel like this is a problem other people also have.

setup.py
richafrank: Oh good point - the selector is just a string of python!

PR #1112, opened 2016-04-05
FIX: time_rules should trigger only at dt specified

lianga888:
Previously, in Q1, a `time_rule` triggered when the dt specified has passed. This was to accommodate the possibility that `handle_data` would not run if there was no new event or data for a particular bar. In Q2, we necessarily check every bar to see if `time_rule` triggers. Thus, we should check that the current dt equals the rule dt.

As a result, in the case of a live algo starting mid-day, any function scheduled for earlier that day will not run. For example:
  - live algo starts at 2 pm and has a function with a time rule of 1 hr after open --> function will not run on first day
  - live algo starts at 2 pm and has a function with a date rule of every day (no time rule means it runs at market open) --> function will not run on first day

### DIFF
--- a/tests/utils/test_events.py
+++ b/tests/utils/test_events.py
@@ -16,7 +16,6 @@ from collections import namedtuple
 import datetime
 from functools import partial
 from inspect import isabstract
-from itertools import islice
 import random
 from unittest import TestCase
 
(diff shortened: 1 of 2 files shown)


jbredeche:
looks good to me.

PR #1114, opened 2016-04-07
ENH: make handle_data optional

jbredeche:
`handle_data` is no longer strictly needed, as one can use `schedule_function` to define trading logic that doesn't need to happen every minute.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -318,6 +318,8 @@ class TradingAlgorithm(object):
             create_context=kwargs.pop('create_event_context', None),
         )
 
+        self._handle_data = None
+
         if self.algoscript is not None:
             filename = kwargs.pop('algo_filename', None)
             if filename is None:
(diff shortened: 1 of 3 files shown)


zipline/algorithm.py
```diff
@@ -407,7 +407,8 @@ def before_trading_start(self, data):
         self._in_before_trading_start = False
 
     def handle_data(self, data):
-        self._handle_data(self, data)
+        if self._handle_data:
```
jbredeche: I would have preferred to check if `self._handle_data` was defined at the time of scheduling the `Always()` call to `handle_data`, but we have various tests that attach a `handle_data` after the `TradingAlgorithm` is created.  Easier to leave it this way for now.

zipline/test_algorithms.py
```diff
@@ -784,6 +784,11 @@ def handle_data(context, data):
     pass
 """
 
+no_handle_data = """
+def initialize(context):
+    pass
```
richafrank: Do we really want this to be a valid strategy?  Should we require that you have handle_data OR a scheduled function?

jbredeche: I see what you're saying, but I don't think it's worth the trouble to have those requirements.  If you have neither `handle_data` nor a scheduled function, your strategy just won't do anything - which while not very useful, is ok I think.

PR #1115, opened 2016-04-07
Fix windows/conda builds

richafrank:
Fixes 32-bit issues as well as tests/simulations leaking handles to adjustments dbs (via data portals).

### DIFF
--- a/tests/test_memoize.py
+++ b/tests/test_memoize.py
@@ -1,6 +1,8 @@
 """
 Tests for zipline.utils.memoize.
 """
+from collections import defaultdict
+import gc
 from unittest import TestCase
 
 from zipline.utils.memoize import remember_last
@@ -32,3 +34,63 @@ class TestRememberLast(TestCase):
         # Calling the old value should still increment the counter.
         self.assertEqual((func(1), call_count[0]), (1, 3))
         self.assertEqual((func(1), call_count[0]), (1, 3))
+
+    def test_remember_last_method(self):
+        call_count = defaultdict(int)
+
+        class clz(object):
+            @remember_last
+            def func(self, x):
+                call_count[(self, x)] += 1
+                return x
+
+        inst1 = clz()
+        self.assertEqual((inst1.func(1), call_count), (1, {(inst1, 1): 1}))
+
+        # Calling again with the same argument should just re-use the old
+        # value, which means func shouldn't get called again.
+        self.assertEqual((inst1.func(1), call_count), (1, {(inst1, 1): 1}))
+
+        # Calling with a new value should increment the counter.
+        self.assertEqual((inst1.func(2), call_count), (2, {(inst1, 1): 1,
+                                                           (inst1, 2): 1}))
+        self.assertEqual((inst1.func(2), call_count), (2, {(inst1, 1): 1,
+                                                           (inst1, 2): 1}))
+
+        # Calling the old value should still increment the counter.
+        self.assertEqual((inst1.func(1), call_count), (1, {(inst1, 1): 2,
+                                                           (inst1, 2): 1}))
+        self.assertEqual((inst1.func(1), call_count), (1, {(inst1, 1): 2,
+                                                           (inst1, 2): 1}))
+
+        inst2 = clz()
+        self.assertEqual((inst2.func(1), call_count),
+                         (1, {(inst1, 1): 2, (inst1, 2): 1,
+                              (inst2, 1): 1}))
+        self.assertEqual((inst2.func(1), call_count),
+                         (1, {(inst1, 1): 2, (inst1, 2): 1,
+                              (inst2, 1): 1}))
+
+        self.assertEqual((inst2.func(2), call_count),
+                         (2, {(inst1, 1): 2, (inst1, 2): 1,
+                              (inst2, 1): 1, (inst2, 2): 1}))
+        self.assertEqual((inst2.func(2), call_count),
+                         (2, {(inst1, 1): 2, (inst1, 2): 1,
+                              (inst2, 1): 1, (inst2, 2): 1}))
+
+        self.assertEqual((inst2.func(1), call_count),
+                         (1, {(inst1, 1): 2, (inst1, 2): 1,
+                              (inst2, 1): 2, (inst2, 2): 1}))
+        self.assertEqual((inst2.func(1), call_count),
+                         (1, {(inst1, 1): 2, (inst1, 2): 1,
+                              (inst2, 1): 2, (inst2, 2): 1}))
+
+        # Remove the above references to the instances and ensure that
+        # remember_last has not made its own.
+        del inst1, inst2
+        call_count.clear()
+        while gc.collect():
+            pass
+
+        self.assertFalse([inst for inst in gc.get_objects()
+                          if type(inst) == clz])
--- a/zipline/utils/memoize.py
+++ b/zipline/utils/memoize.py
@@ -1,8 +1,13 @@
 """
 Tools for memoization of function results.
 """
-from zipline.utils.compat import lru_cache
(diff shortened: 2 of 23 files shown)


zipline/utils/memoize.py
```diff
@@ -84,4 +87,194 @@ def __get__(self, instance, owner):
...
+        return self.data.pop(_WeakArgs(key), *args)
+
+
+class _WeakArgsOrderedDict(_WeakArgsDict, object):
+    def __init__(self):
+        super(_WeakArgsOrderedDict, self).__init__()
+        self.data = OrderedDict()
+
+    def popitem(self, last=True):
+        while True:
```
llllllllll: why is this in a while true

zipline/utils/memoize.py
```diff
@@ -84,4 +87,194 @@ def __get__(self, instance, owner):
...
+    Works with _WeakArgsDict to provide a weak cache for function args.
+    When any of those args are gc'd, the pair is removed from the cache.
+    """
+    def __init__(self, items, dict_remove=None):
+        def remove(k, selfref=ref(self), dict_remove=dict_remove):
+            self = selfref()
+            if self is not None and dict_remove is not None:
+                dict_remove(self)
+
+        super(_WeakArgs, self).__init__(tuple(self._try_ref(item, remove)
```
llllllllll: do we need to call `tuple` on the genexpr?

zipline/utils/memoize.py
```diff
@@ -84,4 +87,194 @@ def __get__(self, instance, owner):
...
+
+    def move_to_end(self, key):
+        """Move an existing element to the end.
+
+        Raises KeyError if the element does not exist.
+        """
+        self[key] = self.pop(key)
+
+
+class instance_weak_lru_cache(lazyval):
```
llllllllll: can we make `lru_cache` a class and then move this `__get__` there. I think that users should not need to know that there is a different decorator for instance methods vs free functions.

zipline/utils/memoize.py
richafrank: Was a vestige of copying from `WeakKeyDictionary.popitem`, where the keys are actual weakrefs, so here we would check if it's still alive.  I think we probably want that behavior ported here instead of missing, in that we should check if all of the weakrefs in the `_WeakArgs` are alive.

zipline/utils/memoize.py
richafrank: Probably not...

zipline/utils/memoize.py
richafrank: That does sound nicer.

zipline/utils/memoize.py
```diff
@@ -84,4 +87,194 @@ def __get__(self, instance, owner):
...
+                hits[0] = misses[0] = 0
+
+        wrapper.cache_info = cache_info
+        wrapper.cache_clear = cache_clear
+        return wrapper
+
+    return decorating_function
+
+
+class _WeakArgs(list):
```
llllllllll: why not make this a subclass of `tuple` if we want it to be hashable?

zipline/utils/memoize.py
```diff
@@ -84,4 +87,194 @@ def __get__(self, instance, owner):
...
+
+class _WeakArgsDict(WeakKeyDictionary, object):
+    def __delitem__(self, key):
+        del self.data[_WeakArgs(key)]
+
+    def __getitem__(self, key):
+        return self.data[_WeakArgs(key)]
+
+    def __repr__(self):
+        return "<%s at %s>" % (type(self), id(self))
```
llllllllll: isn't this the default repr?

zipline/utils/memoize.py
richafrank: Instances of subclasses of `tuple` aren't weak reference-able :disappointed: .

zipline/utils/memoize.py
richafrank: `WeakKeyDictionary` has one that hard codes its name.  Sounds good.

zipline/utils/memoize.py
llllllllll: hmm, could we just make a box around a tuple then, this currently supports mutation which could cause strange behavior.

zipline/utils/memoize.py
```diff
@@ -84,4 +87,194 @@ def __get__(self, instance, owner):
...
+        try:
+            return self.__hash
+        except AttributeError:
+            h = self.__hash = hash(tuple(self))
+            return h
+
+
+class _WeakArgsDict(WeakKeyDictionary, object):
+    def __delitem__(self, key):
+        del self.data[_WeakArgs(key)]
```
llllllllll: what if `key` is not iterable?

zipline/utils/memoize.py
richafrank: We could do that.  I was thinking this would be internal to the cache implementation, so the mutability wouldn't matter.  Thoughts?

zipline/utils/memoize.py
richafrank: `key` is constructed in the two `wrapper` functions above from `args` and `kwargs`, so it should always be iterable.

zipline/utils/memoize.py
llllllllll: just seeing a subclass of list implies that the intention is to mutate the object in place at some point. I don't think we are getting much out of having this be a list subclass over just `collections.Sequence`

tests/test_memoize.py
```diff
@@ -1,19 +1,21 @@
...
 
 class TestRememberLast(TestCase):
 
-    def test_remember_last(self):
+    def test_weak_lru_cache(self):
 
         # Store the count in a list so we can mutate it from inside `func`.
         call_count = [0]
 
-        @remember_last
```
llllllllll: changing this implies that we will also need to change all the usages of `remember_last`, or is it only ever used as on instance methods?

richafrank: It's currently only used on instance methods, but I think you had a suggestion for making it work in both cases.

zipline/utils/memoize.py
```diff
@@ -1,11 +1,13 @@
...
 """
-from collections import OrderedDict
-from functools import wraps, partial
+from collections import OrderedDict, Sequence
+from functools import wraps
+from itertools import compress
 from weakref import WeakKeyDictionary, ref
 
 from six.moves._thread import allocate_lock as Lock
+from toolz.sandbox import unzip
```
llllllllll: we also have a version in zipline.utils.functional.unzip, not sure what the difference is

zipline/utils/memoize.py
```diff
@@ -191,21 +178,37 @@ def remove(k, selfref=ref(self), dict_remove=dict_remove):
...
-            return ref(item, callback)
+            return ref(item, callback), True
         except TypeError:
-            return item
+            return item, False
+
+    @property
+    def alive(self):
+        return all(item is not None
+                   for item in compress(self._items, self._selectors))
```
llllllllll: +1 for compress

zipline/utils/memoize.py
richafrank: Can you point me to that?  I don't see it on master.

zipline/utils/memoize.py
```diff
@@ -84,4 +89,210 @@ def __get__(self, instance, owner):
         return super(classlazyval, self).__get__(owner, owner)
 
 
-remember_last = lru_cache(1)
+def _weak_lru_cache(maxsize=100):
+    # Users should only access the lru_cache through its public API:
```
llllllllll: this can just be the docstring

zipline/utils/memoize.py
llllllllll: I don't think it is on master, nevermind.

PR #1116, opened 2016-04-07
BUG: Ensure matched input length to minute writer.

ehebert:
When the dts and length of cols are mismatched the writer behaves in
unintended ways. e.g. in a case where a consumer passed dts which had
minutes with no trades removed, but regular (market minute for day)
sized arrays for the data with `0`'s on minutes without trades, the non
trade minutes from cols are written to slots in the output where a trade
is intended.

Protect against this misuse by checking that all lengths are equal when
using the `write_cols` method.

Make a separate `_write_cols` method for use by both `write_cols` and
`write`, since the `write` method which takes a DataFrame has the
matched input length enforced by the DataFrame.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -43,6 +43,10 @@ class BcolzMinuteOverlappingData(Exception):
     pass
 
 
+class BcolzMinuteWriterColumnMismatch(Exception):
+    pass
+
+
 def _calc_minute_index(market_opens, minutes_per_day):
     minutes = np.zeros(len(market_opens) * minutes_per_day,
                        dtype='datetime64[ns]')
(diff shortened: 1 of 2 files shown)


zipline/data/minute_bars.py
```diff
@@ -465,7 +471,7 @@ def write(self, sid, df):
             'volume': df.volume.values,
         }
         dts = df.index.values
-        self.write_cols(sid, dts, cols)
+        self._write_cols(sid, dts, cols)
```
yankees714: Could you add a comment to explain why we don't need to call `write_cols` here, basically what you said in the PR description?
> Make a separate _write_cols method for use by both write_cols and write, since the write method which takes a DataFrame has the matched input length enforced by the DataFrame.

ehebert: Done.

PR #1117, opened 2016-04-07
FIX: Check types of args passed to api methods on data

lianga888:
Ensures that for methods `current`, `history`, `can_trade` and `is_stale`:
  - assets must be an Asset object or str (or iterable of either) for `current` and `history` but only Asset (or iterable of Asset) for the other two
  - fields must be str or iterable or str
  - bar_count must be int
  - frequency must be str

This solution is not particularly elegant, but we have a new decorator `assert_types` that takes the expected types, and checks that the args passed to the function are of the correct type

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -107,7 +107,22 @@ from zipline.test_algorithms import (
     call_without_kwargs,
     call_with_bad_kwargs_current,
     call_with_bad_kwargs_history,
+    bad_type_history_assets,
+    bad_type_history_fields,
+    bad_type_history_bar_count,
+    bad_type_history_frequency,
+    bad_type_current_assets,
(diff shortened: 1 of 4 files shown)


jbredeche:
@lianga888 can you estimate the perf impact of these changes?

lianga888:
how would i generally go about doing that?

abhijeetkalyan:
@lianga888 one option is to use a more comprehensive tool like qprof, but if you want something simple, an option is to just use timeit like this:
```python
tdays = env.trading_days[2000:3000]

def test_trigger(tdays, rule):
     for tday in tdays:
         mins = env.market_minutes_for_day(tday)
         for min in mins:
             my_func()

%timeit test_trigger(tdays, my_func)
```

PR #1119, opened 2016-04-08
Support multiple outputs for custom factors

dmichalowicz:
Add an optional `outputs` parameter to CustomFactor allowing multiple outputs to be computed and returned from a single CustomFactor, each output of which is itself a Factor.

### DIFF
--- a/tests/pipeline/test_engine.py
+++ b/tests/pipeline/test_engine.py
@@ -4,6 +4,7 @@ Tests for SimplePipelineEngine
 from __future__ import division
 from collections import OrderedDict
 from itertools import product
+from operator import add, sub
 
 from nose_parameterized import parameterized
 from numpy import (
@@ -11,6 +12,7 @@ from numpy import (
     array,
     concatenate,
     float32,
+    float64,
     full,
     log,
     nan,
@@ -38,19 +40,9 @@ from toolz import merge
 
 from zipline.assets.synthetic import make_rotating_equity_info
 from zipline.lib.adjustment import MULTIPLY
-from zipline.pipeline.loaders.synthetic import PrecomputedLoader
-from zipline.pipeline import Pipeline
-from zipline.pipeline.data import USEquityPricing, DataSet, Column
-from zipline.pipeline.loaders.equity_pricing_loader import (
-    USEquityPricingLoader,
-)
-from zipline.pipeline.loaders.synthetic import (
-    make_daily_bar_data,
-    expected_daily_bar_values_2d,
-)
+from zipline.pipeline import CustomFactor, Pipeline
+from zipline.pipeline.data import Column, DataSet, USEquityPricing
 from zipline.pipeline.engine import SimplePipelineEngine
-from zipline.pipeline.loaders.frame import DataFrameLoader
-from zipline.pipeline import CustomFactor
 from zipline.pipeline.factors import (
     AverageDollarVolume,
     EWMA,
@@ -60,6 +52,16 @@ from zipline.pipeline.factors import (
     MaxDrawdown,
     SimpleMovingAverage,
 )
+from zipline.pipeline.loaders.equity_pricing_loader import (
+    USEquityPricingLoader,
+)
+from zipline.pipeline.loaders.frame import DataFrameLoader
+from zipline.pipeline.loaders.synthetic import (
+    expected_daily_bar_values_2d,
+    make_daily_bar_data,
+    PrecomputedLoader,
+)
+from zipline.pipeline.term import NotSpecified
 from zipline.testing import (
     product_upper_triangle,
     check_arrays,
@@ -110,6 +112,28 @@ class OpenPrice(CustomFactor):
 
     def compute(self, today, assets, out, open):
         out[:] = open
+
+
+class MultipleOutputs(CustomFactor):
+    window_length = 1
+    inputs = [USEquityPricing.open, USEquityPricing.close]
+    outputs = ['open', 'close']
+
+    def compute(self, today, assets, out, open, close):
+        out.open[:] = open
+        out.close[:] = close
+
+
+class OpenCloseSumAndDiff(CustomFactor):
+    """
+    Used for testing a CustomFactor with multiple outputs operating over a non-
+    trivial window length.
+    """
+    inputs = [USEquityPricing.open, USEquityPricing.close]
+
+    def compute(self, today, assets, out, open, close):
+        out.sum_[:] = open.sum(axis=0) + close.sum(axis=0)
+        out.diff[:] = open.sum(axis=0) - close.sum(axis=0)
 
 
 def assert_multi_index_is_product(testcase, index, *levels):
@@ -407,9 +431,9 @@ class ConstantInputTestCase(WithTradingEnvironment, ZiplineTestCase):
 
         alternating_mask = (AssetIDPlusDay() % 2).eq(0)
         expected_alternating_mask_result = array(
-            [[False,  True, False,   True],
-             [True,  False,  True,  False],
-             [False,  True, False,   True]],
+            [[False,  True, False,  True],
+             [True,  False,  True, False],
+             [False,  True, False,  True]],
             dtype=bool,
         )
 
@@ -509,6 +533,176 @@ class ConstantInputTestCase(WithTradingEnvironment, ZiplineTestCase):
                     data=full(result_shape, const, dtype=float),
                 ),
             )
+
+    def test_factor_with_single_output(self):
+        """
+        Test passing an `outputs` parameter of length 1 to a CustomFactor.
+        """
+        dates = self.dates[5:10]
+        assets = self.assets
+        num_dates = len(dates)
+        open = USEquityPricing.open
+        open_values = [self.constants[open]] * num_dates
+        open_values_as_tuple = [(self.constants[open],)] * num_dates
+        engine = SimplePipelineEngine(
+            lambda column: self.loader, self.dates, self.asset_finder,
+        )
+
+        single_output = OpenPrice(outputs=['open'])
+        pipeline = Pipeline(
+            columns={
+                'open_instance': single_output,
+                'open_attribute': single_output.open,
+            },
+        )
+        results = engine.run_pipeline(pipeline, dates[0], dates[-1])
+
+        # The instance `single_output` itself will compute a numpy.recarray
+        # when added as a column to our pipeline, so we expect its output
+        # values to be 1-tuples.
+        open_instance_expected = {
+            asset: open_values_as_tuple for asset in assets
+        }
+        open_attribute_expected = {asset: open_values for asset in assets}
+
+        for colname, expected_values in (
+                ('open_instance', open_instance_expected),
+                ('open_attribute', open_attribute_expected)):
+            column_results = results[colname].unstack()
+            expected_results = DataFrame(
+                expected_values, index=dates, columns=assets, dtype=float64,
+            )
+            assert_frame_equal(column_results, expected_results)
+
+    def test_factor_with_multiple_outputs(self):
+        dates = self.dates[5:10]
+        assets = self.assets
+        asset_ids = self.asset_ids
+        constants = self.constants
+        open = USEquityPricing.open
+        close = USEquityPricing.close
+        engine = SimplePipelineEngine(
+            lambda column: self.loader, self.dates, self.asset_finder,
+        )
+
+        def create_expected_results(expected_value, mask):
+            expected_values = where(mask, expected_value, nan)
+            return DataFrame(expected_values, index=dates, columns=assets)
+
+        cascading_mask = AssetIDPlusDay() < (asset_ids[-1] + dates[0].day)
+        expected_cascading_mask_result = array(
+            [[True,   True,  True, False],
+             [True,   True, False, False],
+             [True,  False, False, False],
+             [False, False, False, False],
+             [False, False, False, False]],
+            dtype=bool,
+        )
+
+        alternating_mask = (AssetIDPlusDay() % 2).eq(0)
+        expected_alternating_mask_result = array(
+            [[False,  True, False,  True],
+             [True,  False,  True, False],
+             [False,  True, False,  True],
+             [True,  False,  True, False],
+             [False,  True, False,  True]],
+            dtype=bool,
+        )
+
+        expected_no_mask_result = array(
+            [[True, True, True, True],
+             [True, True, True, True],
+             [True, True, True, True],
+             [True, True, True, True],
+             [True, True, True, True]],
+            dtype=bool,
+        )
+
+        masks = cascading_mask, alternating_mask, NotSpecified
+        expected_mask_results = (
+            expected_cascading_mask_result,
+            expected_alternating_mask_result,
+            expected_no_mask_result,
+        )
+        for mask, expected_mask in zip(masks, expected_mask_results):
+            open_price, close_price = MultipleOutputs(mask=mask)
+            pipeline = Pipeline(
+                columns={'open_price': open_price, 'close_price': close_price},
+            )
+            if mask is not NotSpecified:
+                pipeline.add(mask, 'mask')
+
+            results = engine.run_pipeline(pipeline, dates[0], dates[-1])
+            for colname, case_column in (('open_price', open),
+                                         ('close_price', close)):
+                if mask is not NotSpecified:
+                    mask_results = results['mask'].unstack()
+                    check_arrays(mask_results.values, expected_mask)
+                output_results = results[colname].unstack()
+                output_expected = create_expected_results(
+                    constants[case_column], expected_mask,
+                )
+                assert_frame_equal(output_results, output_expected)
+
+    def test_instance_of_factor_with_multiple_outputs(self):
+        """
+        Test adding a CustomFactor instance, which has multiple outputs, as a
+        pipeline column directly. Its computed values should be tuples
+        containing the computed values of each of its outputs.
+        """
+        dates = self.dates[5:10]
+        assets = self.assets
+        num_dates = len(dates)
+        num_assets = len(assets)
+        constants = self.constants
+        engine = SimplePipelineEngine(
+            lambda column: self.loader, self.dates, self.asset_finder,
+        )
+
+        open_values = [constants[USEquityPricing.open]] * num_assets
+        close_values = [constants[USEquityPricing.close]] * num_assets
+        expected_values = [list(zip(open_values, close_values))] * num_dates
+        expected_results = DataFrame(
+            expected_values, index=dates, columns=assets, dtype=float64,
+        )
+
+        multiple_outputs = MultipleOutputs()
+        pipeline = Pipeline(columns={'instance': multiple_outputs})
+        results = engine.run_pipeline(pipeline, dates[0], dates[-1])
+        instance_results = results['instance'].unstack()
+        assert_frame_equal(instance_results, expected_results)
+
+    def test_custom_factor_outputs_parameter(self):
+        dates = self.dates[5:10]
+        assets = self.assets
+        num_dates = len(dates)
+        num_assets = len(assets)
+        constants = self.constants
+        engine = SimplePipelineEngine(
+            lambda column: self.loader, self.dates, self.asset_finder,
+        )
+
+        def create_expected_results(expected_value):
+            expected_values = full(
+                (num_dates, num_assets), expected_value, float64,
+            )
+            return DataFrame(expected_values, index=dates, columns=assets)
+
+        for window_length in range(1, 3):
+            sum_, diff = OpenCloseSumAndDiff(
+                outputs=['sum_', 'diff'], window_length=window_length,
+            )
+            pipeline = Pipeline(columns={'sum_': sum_, 'diff': diff})
+            results = engine.run_pipeline(pipeline, dates[0], dates[-1])
+            for colname, op in ('sum_', add), ('diff', sub):
+                output_results = results[colname].unstack()
+                output_expected = create_expected_results(
+                    op(
+                        constants[USEquityPricing.open] * window_length,
+                        constants[USEquityPricing.close] * window_length,
+                    )
+                )
+                assert_frame_equal(output_results, output_expected)
 
     def test_loader_given_multiple_columns(self):
 
--- a/tests/pipeline/test_term.py
+++ b/tests/pipeline/test_term.py
@@ -10,10 +10,17 @@ from zipline.errors import (
     WindowedInputToWindowedTerm,
     NotDType,
     TermInputsNotSpecified,
+    TermOutputsEmpty,
     UnsupportedDType,
     WindowLengthNotSpecified,
 )
-from zipline.pipeline import Classifier, Factor, Filter, TermGraph
+from zipline.pipeline import (
+    Classifier,
+    CustomFactor,
+    Factor,
+    Filter,
+    TermGraph,
+)
 from zipline.pipeline.data import Column, DataSet
 from zipline.pipeline.data.testing import TestingDataSet
 from zipline.pipeline.term import AssetExists, NotSpecified
@@ -65,6 +72,19 @@ class DateFactor(Factor):
 class NoLookbackFactor(Factor):
     dtype = float64_dtype
     window_length = 0
+
+
+class GenericCustomFactor(CustomFactor):
+    dtype = float64_dtype
+    window_length = 5
+    inputs = [SomeDataSet.foo]
+
+
+class MultipleOutputs(CustomFactor):
+    dtype = float64_dtype
+    window_length = 5
+    inputs = [SomeDataSet.foo, SomeDataSet.bar]
+    outputs = ['alpha', 'beta']
 
 
 def gen_equivalent_factors():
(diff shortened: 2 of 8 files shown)


tests/pipeline/test_engine.py
```diff
@@ -510,6 +520,42 @@ def test_rolling_and_nonrolling(self):
                 ),
             )
 
+    def test_factor_with_multiple_outputs(self):
```
dmichalowicz: TODO: Add better test cases

zipline/pipeline/factors/factor.py
```diff
@@ -1227,6 +1227,54 @@ def compute(self, today, assets, out, data):
...
+            missing_value=factor.missing_value,
+        )
+        cls.siblings.setdefault(factor, set()).add(new_instance)
+        return new_instance
+
+    def _init(self, factor, attribute, *args, **kwargs):
+        self.factor = factor
+        self.attribute = attribute
+        self.compute = factor.compute
+        self._compute = factor._compute
```
dmichalowicz: I'm thinking maybe this should implement its own `_compute` method instead of adopting the parent CustomFactor

zipline/pipeline/factors/factor.py
```diff
@@ -1227,6 +1227,54 @@ def compute(self, today, assets, out, data):
...
+    def __iter__(self):
+        if not self.outputs:
+            raise ValueError('This factor does not have multiple outputs.')
+        RecarrayFactor_ = partial(RecarrayFactor, self)
+        return iter(map(RecarrayFactor_, self.outputs))
+
+
+class RecarrayFactor(Factor):
+
+    siblings = {}
```
ssanderson: This dict is going to be shared by all instances of `RecarrayFactor`.  You probably want to just have this be `NotSpecified` and make a new dict.

zipline/pipeline/engine.py
```diff
@@ -348,12 +348,19 @@ def compute_chunk(self, graph, dates, assets, initial_workspace):
...
                     assets,
                     mask,
                 )
+                if hasattr(term, 'siblings'):
+                    siblings = term.siblings[term.factor]
+                    while siblings:
+                        sibling = siblings.pop()
+                        workspace[sibling] = term_output[sibling.attribute]
+                else:
+                    workspace[term] = term_output
```
dmichalowicz: Probably should move this into a helper function

zipline/pipeline/engine.py
```diff
@@ -348,12 +348,19 @@ def compute_chunk(self, graph, dates, assets, initial_workspace):
...
-                workspace[term] = term._compute(
+                term_output = term._compute(
                     self._inputs_for_term(term, workspace, graph),
                     mask_dates,
                     assets,
                     mask,
                 )
+                if hasattr(term, 'siblings'):
+                    siblings = term.siblings[term.factor]
+                    while siblings:
```
ssanderson: This will break if you compute the same term twice.

ssanderson: As a general rule, any term we use in the Pipeline API needs to be immutable.

tests/pipeline/test_engine.py
```diff
@@ -112,6 +112,16 @@ def compute(self, today, assets, out, open):
...
 
 
+class MultipleOutputs(CustomFactor):
+    window_length = 1
+    inputs = [USEquityPricing.open, USEquityPricing.close]
+    outputs = ['double_open', 'double_close']
+
+    def compute(self, today, assets, out, open, close):
+        out.double_open[:] = open * 2
+        out.double_close[:] = close * 2
```
twiecki: Nice API!

zipline/errors.py
```diff
@@ -433,10 +433,21 @@ class TermInputsNotSpecified(ZiplineError):
     msg = "{termname} requires inputs, but no inputs list was passed."
 
 
+class TermOutputsNotSpecified(ZiplineError):
+    """
+    Raised if a user attempts to construct a CustomFactor without specifying
+    enough outputs and that factor does not have class-level default outputs.
+    """
+    msg = (
+        "{termname} requires at least two outputs, but was given "
```
ssanderson: "at least two" seems more specific than this should be?

zipline/pipeline/factors/factor.py
```diff
@@ -1130,13 +1130,27 @@ class CustomFactor(PositiveWindowLengthMixin, CustomTermMixin, Factor):
...
+        An iterable of strings which represent the names of each output this
+        factor should compute and return. If this argument is not passed to the
+        CustomFactor constructor, we look for a class-level attribute named
+        `outputs`.
     window_length : int, optional
         Number of rows to pass for each input.  If this argument is not passed
         to the CustomFactor constructor, we look for a class-level attribute
         named `window_length`.
 
+    Returns
```
ssanderson: Generally you don't put a `Returns` block in a class, since the assumption (which is true here?) is that the return value of the class is an instance of the class.

zipline/pipeline/mixins.py
```diff
@@ -109,7 +111,14 @@ def _compute(self, windows, dates, assets, mask):
         compute = self.compute
         missing_value = self.missing_value
         params = self.params
-        out = full_like(mask, missing_value, dtype=self.dtype)
+        outputs = self.outputs
+        if outputs is not NotSpecified and len(outputs) > 1:
```
ssanderson: If you specify outputs at all, I'd say you should always get a `recarray` with all the fields of your outputs.

zipline/pipeline/term.py
```diff
@@ -368,6 +370,13 @@ def __new__(cls,
             # normalize to a tuple so that inputs is hashable.
             inputs = tuple(inputs)
 
+        if outputs is NotSpecified:
+            outputs = cls.outputs
+        if outputs is NotSpecified:
+            outputs = tuple()
```
ssanderson: You can get an empty tuple with 5 fewer characters :).

zipline/pipeline/factors/factor.py
```diff
@@ -1224,9 +1240,79 @@ def compute(self, today, assets, out, data):
...
     '''
     dtype = float64_dtype
 
+    def __getattr__(self, name):
+        if name in self.outputs:
+            return RecarrayFactor(factor=self, attribute=name)
+        else:
+            raise AttributeError(
+                "This factor has no output called '{}'.".format(name)
+            )
```
dmichalowicz: This is for allowing something like:

``` python
class MyFactor(CustomFactor):
    outputs = ['alpha', 'beta']
    ...
my_factor = MyFactor
alpha, beta = my_factor.alpha, my_factor.beta
```

Should we keep both this and `__iter__` and allow users to choose how to access the outputs, or just pick one single option?

ssanderson: I think having both of these makes sense.

zipline/pipeline/factors/factor.py
dmichalowicz: Yeah; wasn't sure the best way to convey the possibility of returning a RecarrayFactor. Will remove for now.

zipline/errors.py
dmichalowicz: Yes; I'll make RecarrayFactor support single outputs as well

ssanderson: I actually meant that this might want to be `at least N`, where N is an arg, but I don't know if there are cases where we know we need more than two outputs.

dmichalowicz: Ah ok. So for the purpose of RecarrayFactors as we have them it makes sense, but wouldn't hurt to be more generalizable in case we come across such a case.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,73 @@ def compute(self, today, assets, out, data):
...
+    def static_identity(cls, attribute, *args, **kwargs):
+        return (
+            super(RecarrayFactor, cls).static_identity(*args, **kwargs),
+            attribute,
+        )
+
+    def _validate(self):
+        super(RecarrayFactor, self)._validate()
+        if not self.outputs:
+            raise TermOutputsNotSpecified(termname=type(self).__name__)
```
dmichalowicz: In thinking about it, I don't think a RecarrayFactor even needs to specify any outputs. At the end of the day it acts how all other regular factors do with a single out. So this can all be removed.

dmichalowicz: Actually, this check should be performed on the inputted factor, i.e. check that the parent's `factor.outputs` is not empty.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,73 @@ def compute(self, today, assets, out, data):
...
+    def __iter__(self):
+        if len(self.outputs) < 2:
+            raise ValueError('This factor does not have multiple outputs.')
+        RecarrayFactor_ = partial(RecarrayFactor, self)
+        return iter(map(RecarrayFactor_, self.outputs))
+
+
+class RecarrayFactor(SingleInputMixin, Factor):
+
+    def __new__(cls, factor, attribute):
```
ssanderson: This should probably forward the mask from its parent.

dmichalowicz:
@ssanderson Ready for another look

tests/pipeline/test_engine.py
```diff
@@ -510,6 +519,107 @@ def test_rolling_and_nonrolling(self):
...
+            constants[open], full((num_dates, num_assets), True),
+        )
+        second_output_expected = create_expected_results(
+            constants[close], full((num_dates, num_assets), True),
+        )
+
+        assert_frame_equal(first_output_results, first_output_expected)
+        assert_frame_equal(second_output_results, second_output_expected)
+
+        # Now test the same outputs again, but with different masks.
```
dmichalowicz: Should this be a separate test entirely? I lumped it in here to save on doing the setup again.

dmichalowicz: Also, is it worth asserting the output of each mask, as was done in `test_masked_factor`, or would that be redundant?

tests/pipeline/test_engine.py
```diff
@@ -510,6 +519,107 @@ def test_rolling_and_nonrolling(self):
                 ),
             )
 
+    def test_factor_with_single_output(self):
+        dates = self.dates[5:10]
+        assets = self.assets
+        num_dates = len(dates)
+        open = USEquityPricing.open
+        open_values = array([self.constants[open]] * num_dates, dtype=float)
```
ssanderson: `float` here should be `float64`.  `float` will be a 32-bit float in windows, which has caused us problems in the past.

tests/pipeline/test_engine.py
```diff
@@ -510,6 +519,107 @@ def test_rolling_and_nonrolling(self):
...
+        multiple_outputs = MultipleOutputs()
+        open_price, close_price = MultipleOutputs()
+
+        # Ensure that both methods of accessing our outputs return the same
+        # things.
+        self.assertIs(open_price, multiple_outputs.open)
+        self.assertIs(close_price, multiple_outputs.close)
+
+        pipeline = Pipeline(
+            columns={'open_price': open_price, 'close_price': close_price},
```
ssanderson: What happens if I add `multiple_outputs` itself as a column.  If that works, we should probably have a test for the behavior.  If it doesn't, then we should fail immediately if you try to set it as an output.

tests/pipeline/test_engine.py
```diff
@@ -510,6 +519,107 @@ def test_rolling_and_nonrolling(self):
...
+        )
+
+        results = engine.run_pipeline(pipeline, dates[0], dates[-1])
+        first_output_results = results['open_price'].unstack()
+        second_output_results = results['close_price'].unstack()
+
+        first_output_expected = create_expected_results(
+            constants[open], full((num_dates, num_assets), True),
```
ssanderson: You probably want dtype=bool here.

tests/pipeline/test_engine.py
```diff
@@ -510,6 +519,107 @@ def test_rolling_and_nonrolling(self):
...
+        # Ensure that both methods of accessing our outputs return the same
+        # things.
+        self.assertIs(open_price, multiple_outputs.open)
+        self.assertIs(close_price, multiple_outputs.close)
+
+        pipeline = Pipeline(
+            columns={'open_price': open_price, 'close_price': close_price},
+        )
+
+        results = engine.run_pipeline(pipeline, dates[0], dates[-1])
```
ssanderson: There's a lot of duplication here that might be simplified as a loop like:

```
for colname, case_column in ('open_price', open), ('close_price', close):
    # Do shared checks.
```
```

File: tests/pipeline/test_engine.py [hunk 19]
```diff
@@ -112,6 +111,16 @@ def compute(self, today, assets, out, open):
         out[:] = open
 
 
+class MultipleOutputs(CustomFactor):
+    window_length = 1
```
ssanderson: Do we have a test that runs a multi-output factor with a nontrivial window_length?

tests/pipeline/test_engine.py
```diff
@@ -112,6 +111,16 @@ def compute(self, today, assets, out, open):
...
+    inputs = [USEquityPricing.open, USEquityPricing.close]
+    outputs = ['open', 'close']
```
ssanderson: We probably want a test that exercises the behavior of passing `outputs` dynamically to the constructor.

tests/pipeline/test_engine.py
```diff
@@ -510,6 +519,107 @@ def test_rolling_and_nonrolling(self):
...
+        def create_expected_results(expected_value, mask):
+            expected_values = where(mask, expected_value, nan)
+            return DataFrame(expected_values, index=dates, columns=assets)
+
+        multiple_outputs = MultipleOutputs()
+        open_price, close_price = MultipleOutputs()
+
+        # Ensure that both methods of accessing our outputs return the same
+        # things.
+        self.assertIs(open_price, multiple_outputs.open)
```
ssanderson: These checks probably belong in `test_term.py`.

zipline/pipeline/factors/factor.py
```diff
@@ -1130,8 +1130,13 @@ class CustomFactor(PositiveWindowLengthMixin, CustomTermMixin, Factor):
     inputs : iterable, optional
         An iterable of `BoundColumn` instances (e.g. USEquityPricing.close),
         describing the data to load and pass to `self.compute`.  If this
-        argument is passed to the CustomFactor constructor, we look for a
+        argument is not passed to the CustomFactor constructor, we look for a
         class-level attribute named `inputs`.
+    outputs : iterable, optional
```
ssanderson: I sometimes write this is as `outputs : iterable[str], optional`.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,68 @@ def compute(self, today, assets, out, data):
...
+            inputs = [USEquityPricing.close]
+            outputs = ['alpha', 'beta']
+            window_length = N
+
+            def compute(self, today, assets, out, close):
+                computed_alpha, computed_beta = some_function(close)
+                out.alpha[:] = computed_alpha
+                out.beta[:] = computed_beta
+
+        # Each output is returned as its own Factor upon instantiation.
```
ssanderson: You might also want to show the attribute-access style here.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,68 @@ def compute(self, today, assets, out, data):
...
+        alpha, beta = MultipleOutputs()
     '''
     dtype = float64_dtype
 
+    def __getattr__(self, name):
+        if name in self.outputs:
+            return RecarrayFactor(factor=self, attribute=name)
+        else:
+            raise AttributeError(
+                "This factor has no output called '{}'.".format(name)
```
ssanderson: The more idiomatic way to get the quotes here is `{!r}`.

This error message should include the name of the factor that's failing.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,68 @@ def compute(self, today, assets, out, data):
...
+            )
+
+    def __iter__(self):
+        if len(self.outputs) < 2:
```
ssanderson: I think this should still work if outputs has length 1.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,68 @@ def compute(self, today, assets, out, data):
...
+            raise ValueError('This factor does not have multiple outputs.')
```
ssanderson: Same not as above "this factor" is much less useful than "FactorThatIThoughtHadMultipleOutputs"

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,68 @@ def compute(self, today, assets, out, data):
...
+            attribute=attribute,
+            inputs=[factor],
+            window_length=0,
+            mask=factor.mask,
+            dtype=factor.dtype,
+            missing_value=factor.missing_value,
+        )
+
+    def _init(self, attribute, *args, **kwargs):
+        self.attribute = attribute
```
ssanderson: Generally I've been keeping internal factor state in `_variables`.  See, for example, `Rank` and `GroupedRowTransform`.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,68 @@ def compute(self, today, assets, out, data):
...
+            )
+
+    def __iter__(self):
+        if len(self.outputs) < 2:
+            raise ValueError('This factor does not have multiple outputs.')
+        RecarrayFactor_ = partial(RecarrayFactor, self)
+        return iter(map(RecarrayFactor_, self.outputs))
+
+
+class RecarrayFactor(SingleInputMixin, Factor):
```
ssanderson: I would probably call this `RecarrayField` to make it clear that it's not producing a recarray, it's taking a slice of one.

zipline/pipeline/mixins.py
```diff
@@ -109,7 +111,16 @@ def _compute(self, windows, dates, assets, mask):
         compute = self.compute
         missing_value = self.missing_value
         params = self.params
-        out = full_like(mask, missing_value, dtype=self.dtype)
+        outputs = self.outputs
+        if outputs:
+            out = recarray(
+                mask.shape,
+                formats=[self.dtype.str] * len(outputs),
```
ssanderson: We might at some point want to support different dtypes for the different fields.  It's probably a fine restriction to start with that multi-output factors have to have a fixed dtype for all fields, but we should call that out somewhere in the docs.

zipline/pipeline/mixins.py
```diff
@@ -109,7 +111,16 @@ def _compute(self, windows, dates, assets, mask):
...
+                names=outputs,
+            )
+            out[:] = array([missing_value])
```
ssanderson: Do we need to cast to an array here?  It seems to work to just assign the missing value directly:

```
In [2]: recarray((2, 2), formats=['i', 'i'], names=['a', 'b'])
Out[2]:
rec.array([[(-396702552, 32663), (-396702552, 32663)],
       [(-392389200, 32663), (-392592440, 32663)]],
      dtype=[('a', '<i4'), ('b', '<i4')])

In [3]: r = recarray((2, 2), formats=['i', 'i'], names=['a', 'b'])

In [4]: r[:] = 0

In [5]: r
Out[5]:
rec.array([[(0, 0), (0, 0)],
       [(0, 0), (0, 0)]],
      dtype=[('a', '<i4'), ('b', '<i4')])
```

zipline/pipeline/term.py
```diff
@@ -368,6 +370,13 @@ def __new__(cls,
             # normalize to a tuple so that inputs is hashable.
             inputs = tuple(inputs)
 
+        if outputs is NotSpecified:
+            outputs = cls.outputs
+        if outputs is NotSpecified:
+            outputs = ()
```
ssanderson: I think we should just leave this as `NotSpecified` here, otherwise we can't detect and raise if a user does:

```
class MyDumbFactor(CustomFactor):
    outputs = ()
```

zipline/pipeline/factors/factor.py
ssanderson: See note below re: using `NotSpecified` if a user doesn't supply outputs.  I would make this check:

```
if self.outputs is NotSpecified:
    raise ValueError(...)
```

tests/pipeline/test_engine.py
ssanderson: Other things to verify in `test_term`:

- Constructing two instances of the same factor with different output strings (including re-orderings of the same set of strings) should produce different factor instances.  The RecarraySlices produced by those factors should be the same **even if the names overlap**.
- We should fail with a useful error at construction time if the user passes an empty sequence as output.

tests/pipeline/test_engine.py
ssanderson: One option here might be to parameterize the entire suite on mask with values of `alternating_mask, cascading_mask, NotSpecified`, and then just always pass mask to the factor constructors.  This makes the logic in `create_expected_results` slightly more complex, but it would cut the size of the test in half.

docs/source/whatsnew/0.9.1.txt
```diff
@@ -27,6 +27,13 @@ Enhancements
   which manages expiration of entries based on the `dt` supplied to the `get`
   method.
 
+* Implemented :class:`zipline.pipeline.factors.RecarrayFactor`, a new pipeline
+  term designed to be the output type of a CustomFactor with multiple outputs.
```
ssanderson: This should reference the issue number.

Also, I think the next release is going to be 1.0, not 0.9.1, so we should probably move the notes here to 1.0.0.txt

richafrank: Looks like we need to merge the two and get rid of 0.9.1...

zipline/pipeline/mixins.py
dmichalowicz: Huh, could've sworn I tried that. You're right though, it works

tests/pipeline/test_engine.py
dmichalowicz: > The RecarraySlices produced by those factors should be the same even if the names overlap.

Do you mean they should be different?

ssanderson: Yes

tests/pipeline/test_engine.py
dmichalowicz: Added in `test_custom_factor_outputs_parameter`

tests/pipeline/test_engine.py
dmichalowicz: Added in `test_custom_factor_outputs_parameter`

tests/pipeline/test_engine.py
dmichalowicz: Added `test_instance_of_factor_with_multiple_outputs`

zipline/pipeline/factors/factor.py
dmichalowicz: Added

zipline/pipeline/mixins.py
dmichalowicz: Added in `CustomFactor` doc string

docs/source/whatsnew/0.9.1.txt
dmichalowicz: I can do this move; should it go in a separate PR though?

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,86 @@ def compute(self, today, assets, out, data):
...
+                    factor=self.__class__.__name__, attr=attribute_name,
+                )
+            )
+
+    def __iter__(self):
+        if self.outputs is NotSpecified:
+            raise ValueError(
+                '{factor} does not have multiple outputs.'.format(
+                    factor=self.__class__.__name__,
+                )
```
dmichalowicz: Modified the language here and above

dmichalowicz:
@ssanderson Should be ready for another look

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,86 @@ def compute(self, today, assets, out, data):
...
+
+    Note: If a CustomFactor has multiple outputs, all outputs must be of the
+    same dtype. For instance, in the example above, if alpha is a float then
+    beta must also be a float.
     '''
     dtype = float64_dtype
 
+    def __getattr__(self, attribute_name):
+        if self.outputs is NotSpecified:
+            raise AttributeError()
```
ssanderson: This should be `AttributeError(attribute_name)`.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,86 @@ def compute(self, today, assets, out, data):
...
+        if attribute_name in self.outputs:
+            return RecarrayField(factor=self, attribute=attribute_name)
+        else:
+            raise AttributeError(
+                'Instance of {factor} has no output called {attr!r}.'.format(
+                    factor=self.__class__.__name__, attr=attribute_name,
```
ssanderson: I tend to prefer `type(self).__name__` for this.  Not a big deal though.

tests/pipeline/test_term.py
```diff
@@ -243,6 +288,30 @@ class SomeFactor(Factor):
...
+
+        # Different outputs.
+        self.assertIsNot(
+            MultipleOutputs(), MultipleOutputs(outputs=['beta', 'gamma']),
+        )
+
+        # Reordering outputs.
+        self.assertIsNot(
+            multiple_outputs,
+            MultipleOutputs(
```
ssanderson: Another option for this would be `list(reversed(MultipleOutputs.outputs))`

tests/pipeline/test_term.py
```diff
@@ -243,6 +288,30 @@ class SomeFactor(Factor):
 
         self.assertIsNot(orig_foobar_instance, SomeFactor())
 
+    def test_instance_non_caching_multiple_outputs(self):
+        multiple_outputs = MultipleOutputs()
```
ssanderson: Might want to verify that this produces the same result if you do `MultipleOutputs(outputs=MultipleOutputs.outputs)`.

zipline/errors.py
```diff
@@ -433,10 +433,20 @@ class TermInputsNotSpecified(ZiplineError):
     msg = "{termname} requires inputs, but no inputs list was passed."
 
 
+class TermOutputsNotSpecified(ZiplineError):
```
ssanderson: This should be `TermOutputsEmpty`, not `NotSpecified`.  `NotSpecified` means they weren't passed at all.

zipline/pipeline/factors/factor.py
```diff
@@ -1229,9 +1236,86 @@ def compute(self, today, assets, out, data):
...
+        # Each output is returned as its own Factor upon instantiation.
+        alpha, beta = MultipleOutputs()
+
+        # Equivalently, we can create a single factor instance and access each
+        # output as an attribute of that instance.
+        multiple_outputs = MultipleOutputs()
+        alpha = multiple_outputs.alpha
+        beta = multiple_outputs.beta
+
+    Note: If a CustomFactor has multiple outputs, all outputs must be of the
```
ssanderson: Grammar nit: "must be of the same" is a bit awkward.  "must have the same" is cleaner IMO.

ssanderson:
LGTM.  @dmichalowicz feel free to merge when appveyor finishes.

PR #1121, opened 2016-04-11
BUG: bar_data.history too limiting on iterable types

jbredeche:
In before_trading_start, history needs to call `DataPortal.get_adjustments`, and that method wasn’t correctly checking for iterables.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -859,6 +859,23 @@ class MinuteEquityHistoryTestCase(HistoryTestCaseBase):
         # last 5 minutes should not be adjusted
         np.testing.assert_array_equal(np.array(range(782, 787)), window3[395:])
 
+    def test_passing_iterable_to_history_regular_hours(self):
+        # regular hours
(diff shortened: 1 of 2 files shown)


tests/test_history.py
```diff
@@ -859,6 +859,31 @@ def test_minute_dividends(self):
...
+        bar_data.history(df.index, "high", 5, "1m")
+
+    def test_passing_iterable_to_history_bts(self):
+        # before market hours
+        current_dt = pd.Timestamp("2015-01-07 8:45", tz='US/Eastern')
+        bar_data = BarData(self.data_portal, lambda: current_dt, "minute")
+
+        df = pd.DataFrame(
+            [1, 2],
+            index=[self.ASSET1, self.ASSET2]
```
ehebert: Could also construct this more directly as `pd.Index64([self.ASSET1, self.ASSET2])`

jbredeche: good call, I'll do that.

PR #1123, opened 2016-04-12
DOC: Docstring and whatsnew for masked custom factors.

dmichalowicz:


docs/source/whatsnew/0.9.1.txt
```diff
@@ -17,7 +17,11 @@ None
 Enhancements
 ~~~~~~~~~~~~
 
-None
+* Added masking to `CustomFactor`.
```
ssanderson: This should be `:class:\`zipline.pipeline.CustomFactor`.

zipline/pipeline/factors/factor.py
```diff
@@ -1136,6 +1136,11 @@ class CustomFactor(PositiveWindowLengthMixin, CustomTermMixin, Factor):
         Number of rows to pass for each input.  If this argument is not passed
         to the CustomFactor constructor, we look for a class-level attribute
         named `window_length`.
+    mask : zipline.pipeline.Filter, optional
+        When given a mask, a CustomFactor will only compute over those stocks
```
ssanderson: Generally the way these comments are structured is that the first sentence is a fragment describing what the input is, and then second sentence describes it in more detail if necessary.

I'd write this as something like:
```
mask : zipline.pipeline.Filter, optional
    A Filter describing the assets on which we should compute each day.
    Each call to ``CustomFactor.compute`` will only receive assets for which ``mask`` produced True on the day for
    which compute is being called.
```

ssanderson:
Minor nitpicks, otherwise looks good.

PR #1124, opened 2016-04-12
BUG: Properly log partially filled sell orders.

jbredeche:
We were logging partially filled sells as completely unfilled, which is incorrect.


lianga888:
looks good to me

PR #1125, opened 2016-04-12
BUG: Enforce sorted order on minutes to delete.

ehebert:
The intervals are returned as a set, so order is not guaranteed,
which becomes exposed when reading windows which span multiple years.

The deletion of values from the regular sized minute array assumes that
intervals can be reversed to delete the array from the back.


yankees714:
LGTM

PR #1126, opened 2016-04-12
BUG/PERF: Remove extra minute bar loader cache.

ehebert:
The minute history loader caching was incorrectly mimicking the daily
history loader caching.

Where caching the adjusted array on the last dt helps an access pattern
of repeated calling history windows on the same day (which has an end_dt
of the previous day), with minute windows the end dt is always moving
forward, so the cached values are seldom used. (Would only be used if
`history` was called with same parameters twice on the same simulation time.)


jbredeche:
lgtm

PR #1127, opened 2016-04-13
PERF: Remove extra daily history caching.

ehebert:
The cache in data portal was added before the change to using a
CachedObject to wrap the window_blocks in the USEquityHistoryLoader.

Removing this extra layer saves some cycles.

Does not fix current memory investigaton (since only one sids/dts pair
per column was cached in `_equity_daily_reader_array_data` at a time),
but removing should make it more clear where needed references are being
held.


jbredeche:
lgtm

PR #1128, opened 2016-04-14
Fix regression on retrieval of minute positions.

ehebert:
The main change in this PR is:

PERF: Remove memoization from minute bar position.

Instead of using the `remember_last` memoization on all calls to
`_find_position_of_minute`, add an instance local cache which is only
used by the `get_value` call. The `get_value` call is very hot, so any
extra overhead (e.g. creating the WeakArgs on every invocation) becomes
costly. The current usage `get_value` also has the property that it is
called with monotonically increasing, but with a high repeat count on
each value. (A further improvement could making a `get_value` which
supports being used by many sids, for use by the update portfolio
positions.)

The caching is not done at the `_find_position_of_minute_level` because
`unadjusted_window` always uses two positions on the tape (start and end
of range) which would cause the entries and removal into the cache which
would be invalidated both between the calls of start and end, and next
call of the function.

Also removes a parameter which was noticed while investigating the regression:

MAINT: Remove adjusted arg to position of minute.

The argument was only needed for mapping the positions which need to be
removed on adjusted windows. The start and end position of each range
can be derived from the early closes' positions and the market open,
respectively.

Remove to reduce moving parts.

### DIFF
--- a/zipline/data/_minute_bar_internal.pyx
+++ b/zipline/data/_minute_bar_internal.pyx
@@ -38,8 +38,7 @@ def minute_value(ndarray[long_t, ndim=1] market_opens,
 def find_position_of_minute(ndarray[long_t, ndim=1] market_opens,
                             ndarray[long_t, ndim=1] market_closes,
                             long_t minute_val,
-                            short minutes_per_day,
-                            bool adjust_half_day_minutes):
+                            short minutes_per_day):
     """
     Finds the position of a given minute in the given array of market opens.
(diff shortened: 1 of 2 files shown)


jbredeche:
nice.  we're back to a hit ratio of ~98% on that cached value for an algo that orders a lot, which triggers a lot of lookups to `get_value`.

jbredeche:
LGTM, thanks

PR #1129, opened 2016-04-14
BUG: support passing an empty list to `data` methods.

jbredeche:
Our type checking code was a bit too aggressive.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -1905,6 +1905,21 @@ def handle_data(context, data):
 
         self.assertEqual(expected, cm.exception.args[0])
 
+    def test_empty_asset_list_to_history(self):
+        algo = TradingAlgorithm(
+            script=dedent("""
+                def initialize(context):
+                    pass
+
+                def handle_data(context, data):
+                    data.history([], "price", 5, '1d')
+                """),
+            sim_params=self.sim_params,
+            env=self.env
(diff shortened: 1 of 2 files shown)


tests/test_algorithm.py
```diff
@@ -1905,6 +1905,21 @@ def test_arg_types(self, name, inputs):
...
         self.assertEqual(expected, cm.exception.args[0])
 
+    def test_empty_asset_list_to_history(self):
+        algo = TradingAlgorithm(
+            script=dedent("""
+                def initialize(context):
+                    pass
+
+                def handle_data(context, data):
+                    data.history([], "price", 5, '1d')
```
lianga888: are we allowing `fields` to be an empty iterable as well? If so, maybe the test should have `[]` in place of `price`

jbredeche: technically yes, but we haven't defined well that behavior yet: https://github.com/quantopian/qexec/issues/9821

I want to get this fix in first because some existing algorithms are breaking due a pattern of getting a list of assets from `context.portfolio.positions.keys()` and passing that into `data.history`, and sometimes that list is empty.

a more definitive solution will come from https://github.com/quantopian/qexec/issues/9821.

lianga888: Okay, makes sense. Looks good to me then

PR #1130, opened 2016-04-14
ENH: Add expiring cache.

ehebert:
Add a cache interface which supports expirable entries with a changeable
backend for the cache into which they are entered.

The default cache is a `dict` but could swapped for
`cachetools.LRUCache` or any other cache which supports `__get__`,
`__set__`, and `__del__`.

So that consumers can change the use of `CachedObjects` stored in a
cache from:

```
self._cache = {}

...

try:
    obj = self._cache[key]
    try:
        return obj.unwrap(dt)
    except Expired:
        pass
except KeyError:
    pass

...

self._cache[key] = CachedObject(value, new_expiration)
```

to:

```
self._cache = ExpiringCache(LRUCache(maxsize=6))

...

try:
    return self._cache.get(key, dt)
except KeyError:
    # Get fresh value
    ...

    self._cache.set(key, value, new_expiration)
```

### DIFF
--- a/tests/utils/test_cache.py
+++ b/tests/utils/test_cache.py
@@ -2,7 +2,7 @@ from unittest import TestCase
 
 from pandas import Timestamp, Timedelta
 
-from zipline.utils.cache import CachedObject, Expired
+from zipline.utils.cache import CachedObject, Expired, ExpiringCache
 
 
 class CachedObjectTestCase(TestCase):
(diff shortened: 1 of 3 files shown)


tests/utils/test_cache.py
```diff
@@ -19,3 +19,37 @@ def test_cached_object(self):
...
+        cache.set('foo', 1, expiry)
+        cache.set('bar', 2, after)
+
+        self.assertEqual(cache.get('foo', before), 1)
+        # Unwrap on expiry is allowed.
+        self.assertEqual(cache.get('foo', expiry), 1)
+
+        with self.assertRaises(KeyError) as e:
+            self.assertEqual(cache.get('foo', after))
+            self.assertEqual(e.exception.args, ('foo',))
```
ssanderson: This line should be outside the with block.

tests/utils/test_cache.py
```diff
@@ -19,3 +19,37 @@ def test_cached_object(self):
...
+
+        # Should raise same KeyError after deletion.
+        with self.assertRaises(KeyError) as e:
+            self.assertEqual(cache.get('foo', after))
```
ssanderson: Should this be `expires` and/or `before`?

tests/utils/test_cache.py
```diff
@@ -19,3 +19,37 @@ def test_cached_object(self):
...
+            self.assertEqual(e.exception.args, ('foo',))
+
+        # Second value should still exist.
+        self.assertEqual(cache.get('bar', after), 2)
```
ssanderson: This reads a little misleadingly.  I'd just give `bar` a different expiration entirely.

ehebert: Good point, will do.

tests/utils/test_cache.py
ehebert: Will fix.

ssanderson:
Minor test notes, otherwise LGTM.

ehebert:
Squashed in the test notes.

ssanderson:
Might be worth throwing in a whatsnew entry.

PR #1131, opened 2016-04-14
BUG: week_end rule with offset=0 skips every other week

lianga888:
- previously, we calculated the new trigger period (9:30 to 16:00 EST for a given date) after the current trigger period has passed. We fast-forward 7 calendar days from the current dt and calculate the new period for that week
- however, if the trigger is the last day of the week, we recalculate on the first bar of the following week, and after fast-forwarding 7 calendar days, we end up in the week following that (effectively skipping a week)
- to fix this, we recalculate every time we trigger, because of the assumption in Q2 that we run handle_data every minute (and thus we will never miss a trigger)
- accompanying fixes to address shortened weeks:
  - when finding the first or last day of the week, if it is not a trading day, adjust accordingly
  - when applying the offset, if there's not enough days in the week, skip and go onto the next week
- existing tests are modified to reflect that every time a trigger hits, it is recalculated
- new test for week_end and week_start on shortened weeks

### DIFF
--- a/zipline/utils/events.py
+++ b/zipline/utils/events.py
@@ -443,6 +443,18 @@ class TradingDayOfWeekRule(six.with_metaclass(ABCMeta, StatelessRule)):
             )
         )
 
+        # If after applying the offset to the start/end day of the week, we get
+        # day in a different week, skip this week and go on to the next
+        while next_trading_day.date().isocalendar()[1] != \
+                dt.date().isocalendar()[1]:
+            dt += datetime.timedelta(days=7)
+            next_trading_day = _coerce_datetime(
+                env.add_trading_days(
+                    self.td_delta,
+                    self.date_func(dt, env),
+                )
+            )
+
         next_open, next_close = env.get_open_and_close(next_trading_day)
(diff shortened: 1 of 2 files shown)


jbredeche:
lgtm

zipline/utils/events.py
```diff
@@ -443,6 +443,18 @@ def calculate_start_and_end(self, dt, env):
             )
         )
 
+        # If after applying the offset to the start/end day of the week, we get
+        # day in a different week, skip this week and go on to the next
+        while next_trading_day.date().isocalendar()[1] != \
```
richafrank: How about `next_trading_day.weekday()`?

lianga888: how would that let me know if the two dates are in the different weeks?

richafrank: Ah, right! week number is what you're looking for.  

Since a datetime has all the information of a date, you actually don't need to call `date()` throughout here.

richafrank: e.g. `next_trading_day.isocalendar()[1]`

PR #1134, opened 2016-04-15
BUG: Recalculate trigger for week rule if we miss the first one

lianga888:
If we start the simulation on a day so that we miss the trigger
(the first for the sim) for that week, recalculate the trigger
for next week

### DIFF
--- a/tests/utils/test_events.py
+++ b/tests/utils/test_events.py
@@ -369,15 +369,29 @@ class TestStatelessRules(RuleTestCase):
                 self.assertEqual(m, first_of_week + timedelta(days=4-n))
         self.assertEqual(n_triggered, 1)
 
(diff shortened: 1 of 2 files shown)


tests/utils/test_events.py
```diff
@@ -369,15 +369,29 @@ def test_NDaysBeforeLastTradingDayOfWeek(self, n):
                 self.assertEqual(m, first_of_week + timedelta(days=4-n))
         self.assertEqual(n_triggered, 1)
 
-    @parameter_space(n=(0, 1, 2, 3, 4), type=('week_start', 'week_end'))
-    def test_edge_cases_for_TradingDayOfWeek(self, n, type):
+    @parameter_space(
+        rule_offset=(0, 1, 2, 3, 4),
+        start_offset=(0, 1, 2, 3, 4),
```
richafrank: Nice way to test these cases.

PR #1135, opened 2016-04-15
Custom estimates in pipeline

mtydykov:
Adds a custom dataset and test for earnings estimates data.


kglowinski:
Is this branch forked off of one of your other branches? It seems like there are a lot of files that were touched here that aren't related to estimize. Would it be possible separate those out and do a separate PR?

mtydykov:
No, this is off master. Since this is almost exactly like the earnings calendar dataset, I realized I was going to be duplicating a bunch of code, so I ended up refactoring.

tests/pipeline/test_buyback_auth.py
```diff
@@ -34,16 +34,16 @@
...
     zip_with_floats,
     zip_with_dates
 )
 from zipline.testing.fixtures import (
     WithPipelineEventDataLoader, ZiplineTestCase
 )
 
-date_intervals = [[None, '2014-01-04'], ['2014-01-05', '2014-01-09'],
-                  ['2014-01-10', None]]
+date_intervals = [
```
kglowinski: Why is this nested one level deeper now? The docs for the `get_sids_to_frames` seemed to indicate that it was just an iterable of date ranges. This appears to be an iter[iter[date ranges]].

tests/pipeline/test_buyback_auth.py
```diff
@@ -150,13 +140,16 @@ def setup(self, dates):
         cols = {}
         cols[
```
kglowinski: How do you feel about these being of the form `cols[NAME] = \` with the rest on the line below? I know the form is to try to keep from doing a line continuation, but in this case, I actually got confused about what I was looking at because of where the brackets were.

llllllllll: this should just be a dict literal

tests/pipeline/test_buyback_auth.py
mtydykov: Because of the refactoring. There happens to only be a single sid with non-NaN data in this test, but in other tests, the list has 3 levels, and I'm using the same method on all of them.

kglowinski:
More generally in the tests, is there a way to represent the "this will not return bad things" case? I'm trying to model the error Seong was running into, where with the default loader, we would have gotten back multiple results from any number of quarters, and now we should get back only one.

tests/pipeline/test_buyback_auth.py
kglowinski: What should be a dict literal?

tests/pipeline/test_estimates.py
```diff
@@ -0,0 +1,332 @@
...
+            )
+
+        cols = {}
+        cols[PREVIOUS_RELEASE_DATE] = _expected_previous_release_date
+        cols[NEXT_RELEASE_DATE] = _expected_next_release_date
+        for field_name in field_name_to_expected_col:
+            cols[field_name] = self.get_sids_to_frames(
+                zip_with_floats, field_name_to_expected_col[field_name],
+                self.prev_date_intervals
+                if field_name.startswith(PREVIOUS_PREFIX)
```
kglowinski: Let's either move the previous prefix into the common file, or if it's only going to be used once, and unlikely to be needed again elsewhere, hardcode it in here to avoid the variable assignment for a single use.

zipline/testing/fixtures.py
```diff
@@ -455,6 +458,41 @@ def pipeline_event_setup_engine(self, dates):
...
+        Parameters
+        ----------
+        zip_date_index_with_vals: callable
+            A function that returns a series of `vals` repeated based on the
+            number of days in the date interval for each val, indexed by the
+            dates in `dates`.
+        vals: iterable
+            An iterable with values that correspond to each interval in
+            `date_intervals`.
+        date_intervals: list
```
kglowinski: Per your response to my previous comment, can we update the docs to just note that it's nested one layer deeper?

zipline/testing/fixtures.py
```diff
@@ -520,3 +558,122 @@ def test_compute(self, dates):
...
+    ZiplineTestCase mixin extending common functionality for event data
+    loader tests that have both next and previous events.
+
+    `base_cases` should be used as the template to test cases that combine
+    knowledge date (timestamp) and some 'other_date' in various ways.
+    `next_date_intervals` gives the date intervals for the next event based
+    on the dates given in `base_cases`.
+    `next_dates` gives the next date from `other_date` which is known about at
+    each interval.
+    `prev_date_intervals` gives the date intervals for the next event based
```
kglowinski: s/next/previous

zipline/testing/fixtures.py
```diff
@@ -455,6 +458,41 @@ def pipeline_event_setup_engine(self, dates):
...
+        dates: DatetimeIndex
+            The dates which will serve as the index for each Series for each
+            sid in the DataFrame.
+        """
+        frame = pd.DataFrame({sid: get_values_for_date_ranges(
+            zip_date_index_with_vals,
+            vals[sid],
+            date_intervals[sid],
+            dates
+        ) for sid in self.get_sids()[:-1]})
```
kglowinski: Why do we only get the last one here?

kglowinski: Er, rather, everything BUT the last one.

mtydykov:
For that case, we want to make sure that the data we were getting back was based on today and the last revision that we knew about <= today and not on the asof_date (updated). So what I'm missing in this test is an asof_date. I'll add that.

zipline/testing/fixtures.py
mtydykov: Because for all tests, we want the last case to be an empty DataFrame to make sure that that edge case works.

kglowinski:
@joshpayne- @mtydykov and I were discussing the naming of this dataset in pipeline. In their API, we take elements from their `releases` endpoint and also `estimates` endpoint. In our database/DBZ API, we refer to this as `revisions`, but in the store it's `Consensus Estimates`.

Personally, I find earnings estimates confusing, but I'm willing to be overruled if you guys feel that best reflects the data. Though I would note that there are revenue estimates in there as well.

tests/pipeline/test_estimates.py
```diff
@@ -0,0 +1,332 @@
...
+        ).pipeline_event_loader_args(dates)
+        frames = []
+        for sid, df in iteritems(mapping):
+            frame = df.copy()
+            frame[SID_FIELD_NAME] = sid
+            frames.append(frame)
+        return bz.data(pd.concat(frames).reset_index(drop=True)),
+
+
+class BlazeEarningsEstimatesLoaderNotInteractiveTestCase(
```
kglowinski: What's the purpose of the non-interactive symbol test case vs the normal test case? I feel like I may have missed something with the pipeline explanation.

mtydykov: It uses a non-interactive symbol & swaps resources in. For more details probably @llllllllll knows...

mtydykov:
About my previous comment to add asof_date - never mind. At this point in the pipeline (distinct from Pipeline) we can only provide the columns that the loader expects; asof_date is not one of them, so it isn't an issue.

kglowinski:
Can we do a fixup on the commit entitled "MAINT: rename" since the Travis build failed for that one, and it wouldn't be a valid state?

kglowinski:
👍  from me.

PR #1138, opened 2016-04-18
ExchangeCalendars and TradingSchedules

jfkirk:
This PR adds two new base classes: ExchangeCalendar and TradingSchedule.

An ExchangeCalendar is an object that manages timing information for a given asset exchange. It can be queried to find out if an exchange is currently trading, or to get more general information about the trading on the exchange.

A TradingSchedule is an object that manages timing information for a TradingAlgorithm. It tells the algorithm, via the clock, when to execute.

This PR contains ExchangeCalendars for some exchanges, including NYSE, and an implementation of TradingSchedule that wraps an ExchangeCalendar for backwards compatibility.


jfkirk:
Next step is to delete the old tradingcalendar module.

zipline/utils/calendars/exchange_calendar.py
```diff
@@ -0,0 +1,497 @@
...
+        The name of the ExchangeCalendar to be retrieved.
+    """
+    # First, check if the calendar is already registered
+    if name not in _static_calendars:
+        # The calendar is not registered, so check if it is a lazy calendar
+        if name not in _lazy_calendar_names:
+            # It's not a lazy calendar, so raise an exception
+            raise InvalidCalendarName(calendar_name=name)
+
+        if name is 'NYSE':
```
ssanderson: Please don't use `is` for string comparisons.  This is effectively a random number generator.

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+    next_scheduled_minute,
+    previous_scheduled_minute,
+)
+
+
+class TradingSchedule(object):
+    """
+    A TradingSchedule defines the execution timing of a TradingAlgorithm.
+    """
+    __metaclass__ = ABCMeta
```
ssanderson: This doesn't do anything in Python3.  Use `six.with_metaclass`

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+
+    def __init__(self):
+        # Assign the partial calendar helpers
+        self.next_execution_day = partial(
```
ssanderson: Why are these all implemented as partials?  Seems like they could just be methods that access members of `self`?

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+        Returns
+        -------
+        DataFrame
+            A DataFrame, with a DatetimeIndex of trading dates, containing
+            columns of trading starts and ends in this TradingSchedule.
+        """
+        raise NotImplementedError()
+
+    @property
+    @remember_last
```
ssanderson: This is just `lazyval`.

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+        Parameters
+        ----------
+        date : Timestamp
+            The UTC-canonicalized calendar date whose data availability time
+            is needed.
+
+        Returns
+        -------
+        Timestamp or None
+            The data availability time on the given date, or None if there is
```
ssanderson: When would this happen?

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+
+        Returns
+        -------
+        (Timestamp, Timestamp)
+            The start and end for the given date.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def is_executing_on_minute(self, dt):
```
ssanderson: Seems like this should be called `should_execute_on_minute`, given that that's the phrasing used everywhere in the docs and description.

zipline/utils/calendars/trading_schedule.py
jfkirk: The methods are shared between the ExchangeCalendar and the TradingSchedule, and the names of the member methods are different between those objects. 

This is one of the reasons that we may remove the TradingSchedule before merging. We will be discussing that in the next few days.

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+        Returns
+        -------
+        bool
+            True if the TradingAlgorithm should be executed at dt,
+            otherwise False.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def is_executing_on_day(self, dt):
```
ssanderson: ditto above.

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+            is_scheduled_day_hook=self.is_executing_on_day,
+            open_and_close_hook=self.start_and_end,
+            previous_open_and_close_hook=self.previous_start_and_end,
+        )
+
+    @abstractproperty
+    def day(self):
+        """
+        A CustomBusinessDay defining those days on which the algorithm is
+        usually trading.
```
ssanderson: what does "usually" mean in this context?

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+        Returns
+        -------
+        bool
+            True if the TradingAlgorithm should be executed at dt,
+            otherwise False.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def minute_window(self, start, count, step=1):
```
ssanderson: Do we not have enough information to implement this in terms of the abstract intefaces of the class?  (It seems like we should be able to do so)

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+    @abstractproperty
+    def early_ends(self):
+        """
+        Returns a DatetimeIndex containing the session dates on-which there is
+        an early end to trading.
+        """
+        raise NotImplementedError()
+
+
+class ExchangeTradingSchedule(TradingSchedule):
```
ssanderson: Why does this class exist if it's just a proxy around an ExchangeCalendar?

zipline/utils/calendars/trading_schedule.py
```diff
@@ -0,0 +1,379 @@
...
+        return self._exchange_calendar.minute_window(start=start,
+                                                     count=count,
+                                                     step=step)
+
+    @property
+    def early_ends(self):
+        return self._exchange_calendar.early_closes
+
+
+class NYSETradingSchedule(ExchangeTradingSchedule):
```
ssanderson: Ditto above.

zipline/utils/cme_exchange_calendar.py
```diff
@@ -0,0 +1,454 @@
...
+MonTuesThursBeforeIndependenceDay = Holiday(
+    # When July 4th is a Tuesday, Wednesday, or Friday, the previous day is a
+    # half day.
+    'Mondays, Tuesdays, and Thursdays Before Independence Day',
+    month=7,
+    day=3,
+    days_of_week=(MONDAY, TUESDAY, THURSDAY),
+    start_date=Timestamp("1995-01-01"),
+)
+FridayAfterIndependenceDayExcept2013 = Holiday(
```
ssanderson: Most of these oddities were developed by hand by checking against the history of the NYSE.  Does the CME follow exactly the NYSE schedule?

zipline/utils/cme_exchange_calendar.py
```diff
@@ -0,0 +1,454 @@
...
+    # When Christmas is a Saturday, the 24th is a full holiday.
+    days_of_week=(MONDAY, TUESDAY, WEDNESDAY, THURSDAY),
+)
+
+
+# http://en.wikipedia.org/wiki/Aftermath_of_the_September_11_attacks
+September11Closings = date_range('2001-09-11', '2001-09-16', tz='UTC')
+
+# http://en.wikipedia.org/wiki/Hurricane_sandy
+HurricaneSandyClosings = date_range(
```
ssanderson: Was the CME closeed for sandy?

zipline/utils/calendars/trading_schedule.py
ssanderson: What does that have to do with implementing these as partials?  I'm asking why the implementation of this isn't something like:

```
def next_execution_day(self, date):
    return next_scheduled_day(date, last_trading_day=self.last_execution_day, is_scheduled_day_hook=self.is_executing_on_day)
```

zipline/utils/calendars/trading_schedule.py
llllllllll: the cache will be invalidated if a subclass accesses it, so it is slightly different, though I think that is not intended.

zipline/utils/calendars/trading_schedule.py
jfkirk: Oh, sorry - misunderstood the statement. Do you believe that implementing this as in your comment would be clearer? If so, I will modify them accordingly.

zipline/utils/events.py
```diff
@@ -54,6 +58,9 @@
 MAX_WEEK_RANGE = 5
 
 
+_static_nyse_cal = get_calendar('NYSE')
```
llllllllll: does this mean that none of the schedule function rules work if you use a different calendar?

zipline/utils/simfactory.py
```diff
@@ -48,6 +49,11 @@ def create_test_zipline(**config):
     else:
         order_amount = 100
 
+    if 'trading_schedule' in config:
+        trading_schedule = config['trading_schedule']
+    else:
+        trading_schedule = default_nyse_schedule
```
llllllllll: `config.get('trading_schedule', default_nyse_schedule)`

zipline/utils/calendars/trading_schedule.py
ssanderson: They way these are written, we can't have docstrings for the methods, and you can't introspect them from the class, because they're not defined until you construct an instance.

Some (most?) of the implementations here also feel not worth delegating at all.  `{next,previous}_open_and_close` are literally just calling the two functions you're passing in.  Breaking that out into another method is just adding overhead.

`next,previous_scheduled_day` are much less efficient than they should be because of this interface as well.  By only exposing an "are we executing today?" predicate, you're forcing the lookup algorithm to be a linear search.

zipline/utils/calendars/calendar_helpers.py
```diff
@@ -0,0 +1,213 @@
...
+        return None
+    j = bisect.bisect_left(all_days, second_date)
+    if j == len(all_days):
+        return None
+    distance = j - 1
+    assert distance >= 0
+    return distance
+
+
+def minutes_for_day(day, open_and_close_hook):
```
ssanderson: I don't think this is adding any meaningful value as a standalone function.

zipline/utils/calendars/trading_schedule.py
jfkirk: Makes sense - thank you. I will re-write accordingly.

zipline/utils/calendars/calendar_helpers.py
```diff
@@ -0,0 +1,213 @@
...
+    start, end = open_and_close_hook(day)
+    return pd.date_range(start, end, freq='T')
+
+
+def days_in_range(start, end, all_days):
```
ssanderson: Was this the previously-existing implementation?  This is doing four linear scans of a large array, when it should be doing two binary searches.  The right implementation of this is:

```
return all_days[all_days.slice_indexer(start_date, end_date)]
```

zipline/utils/calendars/nyse_exchange_calendar.py
```diff
@@ -0,0 +1,522 @@
...
+        Parameters
+        ----------
+        dt : Timestamp
+
+        Returns
+        -------
+        bool
+            True if  exchange is open at any time during the day containing @dt
+        """
+        dt_normalized = normalize_date(dt)
```
ssanderson: When do we use `session_date` and when do we use `normalize_date`?

jfkirk:
I am proposing that we scrap the TradingSchedule class, and roll its usages over to using the ExchangeCalendar API. We would then implement the 24/5 trading schedule as an ExchangeCalendar object.

I believe that this would create the clearest relationship between the class and its consumers. It would also solve some outstanding issues with the previous design re: schedule_function and history calls.

The reason that we did not conflate these two originally is because we expected that there were many different usages for the two different classes, and it would improve clarity to have them implemented separately. Having now implemented the two all-the-way through Zipline, I realize that they are nearly identical in usage.

This requires:
1. Remove TradingSchedule and usages, switching them over to an ExchangeCalendar.
2. Remove ExchangeTradingSchedule and NYSETradingSchedule, as they are now entirely redundant.
3. Add 'data_availability_time' to ExchangeCalendar.
4. TradingAlgorithm stores an ExchangeCalendar rather than a TradingSchedule.

Benefits:
1. Removes massive redundancy between the two classes (all of the calendar_helpers methods).
2. The TradingAlgorithm's ExchangeCalendar can be used as the default calendar backing all schedule_function rules (though eventually the plan is for this to be a required arg).
3. Removes the parallel-but-not-identical naming between the two classes that may cause confusion.
4. Removes the ExchangeTradingSchedule class which is a thin, potentially-performance-expensive wrapper around an ExchangeCalendar.
5. Generally much less complex and clearer for future development.

Drawbacks:
1. It is slightly un-intuitive for an ExchangeCalendar to have a `data_availability_time` method, though it is straightforward to implement as returning the `market_open`.
2. It is slightly un-intuitive to implement a 24/5 ExchangeCalendar, as it is not really representative of an Exchange. One option would be to rename ExchangeCalendar, but I don't view this as necessary.

zipline/utils/cme_exchange_calendar.py
StewartDouglas: I previously thought it was, but more digging revealed that trading was just 'disrupted', so the code as it stands is out of date. See this for more info: https://github.com/quantopian/dataloader/blob/futures-exp-dates/dataloader/resources/tradingcalendar_cme.py#L254-L274

zipline/algorithm.py
```diff
@@ -603,7 +620,7 @@ def run(self, data=None, overwrite_sim_params=True):
                 self.data_portal = DataPortal(
                     self.trading_environment,
```
jfkirk: This should be a trading_schedule.

zipline/utils/calendars/cme_exchange_calendar.py
```diff
@@ -0,0 +1,374 @@
...
+
+from zipline.utils.calendars import ExchangeCalendar
+
+# Useful resources for making changes to this file:
+# http://www.nyse.com/pdfs/closings.pdf
+# http://www.stevemorse.org/jcal/whendid.html
+
+MONDAY, TUESDAY, WEDNESDAY, THURSDAY, FRIDAY, SATURDAY, SUNDAY = range(7)
+
+US_CENTRAL = timezone('Americas/Chicago')
```
yankees714: I think this should be `'America/Chicago'`

StewartDouglas: My bad - fixing

zipline/utils/calendars/cme_exchange_calendar.py
```diff
@@ -0,0 +1,374 @@
...
+
+    holidays_calendar = CMEHolidayCalendar()
+    special_opens_calendars = ()
+    special_closes_calendars = []
+
+    holidays_adhoc = chain(
+        September11Closings,
+        HurricaneSandyClosings,
+        USNationalDaysofMourning,
+    )
```
yankees714: This is expected to be `list(chain(...))`

zipline/utils/calendars/cme_exchange_calendar.py
```diff
@@ -0,0 +1,374 @@
...
+        dt : Timestamp
+            A timezone-aware Timestamp.
+
+        Returns
+        -------
+        Timestamp
+            The date of the exchange session in which dt belongs.
+        """
+        dt_utc = dt.tz_convert('UTC').tz_convert(None)
+        return dt_utc.replace(hour=0, minute=0, second=0)
```
yankees714: This returns a Timestamp without a timezone.

Could this just be `dt.tz_convert('UTC').normalize()`?

zipline/utils/calendars/cme_exchange_calendar.py
```diff
@@ -0,0 +1,374 @@
...
+    @property
+    def tz(self):
+        """
+        The native timezone of the exchange.
+
+        SD: Not clear that this needs to be exposed.
+        """
+        return self.native_timezone
+
+    def is_open(self, dt):
```
yankees714: The abstract class requires both an `is_open_on_minute

zipline/utils/calendars/cme_exchange_calendar.py
```diff
@@ -0,0 +1,374 @@
...
+
+        Returns
+        -------
+        (Timestamp, Timestamp)
+            The open and close for the given date.
+        """
+        # Generalised logic for the case of trading pauses.
+        # Note: this logic is ~3-4 times slower than that used for the NYSE
+        # (pass a list, to ensure we get a DataFrame returned)
+        return self.schedule.loc[[session]]
```
yankees714: This is supposed to be returning a tuple of Timestamps, instead of a dataframe. There's a `_get_open_and_close` helper method of `NYSEExchangeCalendar` that's general enough that it seems like it would be useful to share through `ExchangeCalendar`.

zipline/utils/calendars/exchange_calendar.py
```diff
@@ -0,0 +1,503 @@
...
+    next_scheduled_minute,
+    previous_scheduled_minute,
+    minute_window,
+)
+
+start_default = pd.Timestamp('1990-01-01', tz='UTC')
+end_base = pd.Timestamp('today', tz='UTC')
+# Give an aggressive buffer for logic that needs to use the next trading
+# day or minute.
+end_default = end_base + pd.Timedelta(days=365)
```
yankees714: This really doesn't give enough of a buffer for futures. We should probably either adjust this date, or pass a separate one as the `end` arg when constructing the `CMEExchangeCalendar` in `get_calendar`.

ssanderson: @yankees714 when do you need a calender that extends longer than a year past the current date?

yankees714: The example I happened upon was computing the notice date (which requires a calendar) for a future that expires more than a year out.

yankees714: That being said, if this is a sufficiently special case, it probably makes sense to keep the default at a 1 year buffer and specify the end date in this particular instance.

tests/data/test_minute_bars.py
```diff
@@ -55,27 +57,24 @@
...
-    def setUpClass(cls):
-        cls.env = TradingEnvironment()
-        all_market_opens = cls.env.open_and_closes.market_open
-        all_market_closes = cls.env.open_and_closes.market_close
-        indexer = all_market_opens.index.slice_indexer(
-            start=TEST_CALENDAR_START,
-            end=TEST_CALENDAR_STOP
+    def init_class_fixtures(cls):
+        super(BcolzMinuteBarTestCase, cls).init_class_fixtures()
+        trading_days = get_calendar('NYSE').trading_days(
```
ehebert: Could `trading_days` be provided by default by the `WithTradingSchedule` fixture?

zipline/data/data_portal.py
```diff
@@ -494,12 +494,14 @@ class DataPortal(object):
     """
     def __init__(self,
                  env,
```
ehebert: This could now be `asset_finder` instead of `env`.

zipline/finance/performance/tracker.py
```diff
@@ -78,28 +78,28 @@ class PerformanceTracker(object):
     """
     Tracks the performance of the algorithm.
     """
-    def __init__(self, sim_params, env):
+    def __init__(self, sim_params, trading_schedule, env):
```
ehebert: Another place where `env` can now be `asset_finder`.

zipline/utils/calendars/calendar_helpers.py
```diff
@@ -0,0 +1,297 @@
...
+    ----------
+    date : Timestamp
+        The date whose following date is needed.
+
+    Returns
+    -------
+    Timestamp
+        The next scheduled date after the provided date.
+    """
+    dt = normalize_date(date)
```
ehebert: A few CPU cycles and levels of abstraction could be saved if we allow the calendar to define this method itself, e.g.:

```
def next_scheduled_day(self, date):
    dt = normalized_date(date)
    loc = self.schedule.get_loc(dt)
    try:
        return self.schedule[loc + 1]
    except IndexError:
        raise CalendarExceeded(dt, self.schedule[-1])
```

ehebert: Or since we are storing the CustomBusinessDay.

```
def next_scheduled_day(self, date):
    dt = normalized_date(date)
    return dt + self.day
```
```

zipline/finance/performance/tracker.py
jfkirk: I was planning to leave this as env because the perf tracker uses both the asset finder and the treasury curves that are part of the environment.

jfkirk:
Nice.

PR #1141, opened 2016-04-18
BUG: Week rule plus time rule doesn't work

lianga888:
A previous commit (1ee3c5f04968ab2b1ab4ff334c1cdefe487e7927), changed the schedule function week rules so that every time it triggers, we recalculate the next trigger. As a result, the rule only triggers once a week (on the first bar of the day it's supposed to trigger). This works if we only have a week rule, but it won't work if we combine a week and time rule, and the time rule triggers after the first bar of the day, because the week rule keeps getting recalculated before the time rule triggers.

So, we revert back to having the week rule trigger for the entire day, but recalculate based on the last triggered date, as opposed to the date after the last trigger

### DIFF
--- a/zipline/utils/events.py
+++ b/zipline/utils/events.py
@@ -466,18 +466,15 @@ class TradingDayOfWeekRule(six.with_metaclass(ABCMeta, StatelessRule)):
             # of the simulation start
             self.calculate_start_and_end(dt, env)
 
-            # If we've missed the first trigger because it occurs before the
-            # simulation starts, recalculate for the next week
-            if dt > self.next_date_end:
-                self.calculate_start_and_end(dt + datetime.timedelta(days=7),
-                                             env)
+        # If we've passed the trigger, calculate the next one
+        if dt > self.next_date_end:
+            self.calculate_start_and_end(self.next_date_end +
+                                         datetime.timedelta(days=7),
+                                         env)
 
-        # if the given dt is within the next matching day, return true. Also
-        # calculate the start and end dates for the next trigger
+        # if the given dt is within the next matching day, return true.
         if self.next_date_start <= dt <= self.next_date_end or \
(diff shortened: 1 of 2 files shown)


zipline/utils/events.py
```diff
@@ -466,18 +466,15 @@ def should_trigger(self, dt, env):
...
-            # If we've missed the first trigger because it occurs before the
-            # simulation starts, recalculate for the next week
-            if dt > self.next_date_end:
-                self.calculate_start_and_end(dt + datetime.timedelta(days=7),
-                                             env)
-
-        # if the given dt is within the next matching day, return true. Also
-        # calculate the start and end dates for the next trigger
+        # If we've passed the trigger, calculate the next one
+        if dt > self.next_date_end:
```
richafrank: In the previous PR, we had considered changing `calculate_start_and_end` to make this guarantee for the consumer by adding this condition to the while loop.  Do you think that would be any cleaner now?

lianga888: This is about guaranteeing that `calculate_start_and_end` returns a useful value, correct? I don't think it would be necessary. In any case, we still need the `if dt> self.next_date_end: ...` because that's what allows us to recalculate throughout the simulation. Adding this to `calculate_start_and_end` would be unnecessary, since we do it here (in `should_trigger`) anyway.

The way I see it, the responsibility of `calculate_start_and_end` is, given a date and an offset, to find the day of the week to set the trigger. It should not care whether or not this trigger has expired. `should_trigger`, on the other hand, should be responsible for recalculating the trigger once it has passed

richafrank: Ok, that division of responsibilities makes sense to me.

richafrank:
LGTM @lianga888

jbredeche:
talked to @lianga888, merging on his behalf.

PR #1143, opened 2016-04-19
ENH: add actual value column to estimates dataset.

mtydykov:
Adds a column representing the previous "actual" value of the estimated field for the estimates dataset.

### DIFF
--- a/zipline/pipeline/common.py
+++ b/zipline/pipeline/common.py
@@ -1,6 +1,7 @@
 """
 Common constants for Pipeline.
 """
+ACTUAL_VALUE_FIELD_NAME = 'actual_value'
 AD_FIELD_NAME = 'asof_date'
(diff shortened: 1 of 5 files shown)


zipline/pipeline/common.py
```diff
@@ -1,6 +1,7 @@
 """
 Common constants for Pipeline.
 """
+ACTUAL_VALUE_FIELD_NAME = 'actual_value'
```
kglowinski: Is actual value the agreed upon name? This is the latest known metric (eps/revenue) value, right?

mtydykov: I don't think we really agreed on a specific name - seemed like a logical choice to me, since the other columns (mean, low, high, stdev) are estimates of this value, which is the "actual" value.

PR #1149, opened 2016-04-20
ENH: make BcolzMinuteBarWriter.write take iterable

llllllllll:
Updates the BcolzMinuteBarWriter.write api to allow users to pass their
data as a stream instead of requiring that they loop over their data
externally. This matches the API presented by BcolzDailyBarWriter.

### DIFF
--- a/zipline/testing/core.py
+++ b/zipline/testing/core.py
@@ -349,7 +349,7 @@ def make_trade_data_for_asset_info(dates,
         )
 
         if writer:
-            writer.write(sid, df)
+            writer.write_sid(sid, df)
 
         trade_data[sid] = df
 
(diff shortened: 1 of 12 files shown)


zipline/testing/core.py
```diff
@@ -492,20 +490,17 @@ def create_data_portal(env, tempdir, sim_params, sids, adjustment_reader=None):
         )
 
 
-def write_bcolz_minute_data(env, days, path, df_dict):
+def write_bcolz_minute_data(env, days, path, data):
```
ehebert: Confirming my understanding, main intent of this patch is to enable `data` here as any given iterable of `sid, frame`.

Looks good to me.

llllllllll: correct. I am working on a feature where I am passing all of our writers to a user defined function and it makes it easier to understand the API if we minimize the differences between the minute and daily writers

ssanderson:
@ehebert @llllllllll is this still something we're planning on merging?

llllllllll:
oh, I thought this was merged, I would still like to merge this. I can rebase and merge this later assuming this still looks good.

llllllllll:
This code is on master, it was merged in the wiki loader pr.

PR #1150, opened 2016-04-20
Use __slots__ in Order object to save memory

jbredeche:
for a 2year backtest that places an order every minute (= ~205k orders), this takes max memory usage from 250MB to 163MB.

### DIFF
--- a/zipline/finance/order.py
+++ b/zipline/finance/order.py
@@ -12,7 +12,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from copy import copy
 import math
 import uuid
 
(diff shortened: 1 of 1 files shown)


zipline/finance/order.py
```diff
@@ -70,11 +76,23 @@ def make_id(self):
         return uuid.uuid4().hex
 
     def to_dict(self):
-        py = copy(self.__dict__)
-        for field in ['type', 'direction', '_status']:
-            del py[field]
-        py['status'] = self.status
-        return py
+        # all the properties, minute "type", "direction", and "_status".
```
richafrank: typo (minus)

zipline/finance/order.py
```diff
@@ -70,11 +76,23 @@ def make_id(self):
...
+        # we use the `self.status` property to fill in the order status.
+        return {
```
richafrank: Do you think we should filter out the 3 explicitly, like we did before?
```python
dct = {name: getattr(self, name)
       for name in self.__slots__ 
       if name not in {'type', 'direction', '_status'}}
dct['status'] = self.status
return dct
```

zipline/finance/order.py
jbredeche: thanks. I've typed "minute" a bit too much recently.

zipline/finance/order.py
jbredeche: sounds good to me.

PR #1153, opened 2016-04-21
Filter nulls in expected cols

mtydykov:
Changes in this PR include:

- Generalized idea of a "key date" for each event. Each event loader must define an event date column which corresponds to this key date. The event loader filters out any rows where this event date is null in the incoming raw data.

- Changes to test mechanics - changing to a more efficient pandas function and vectorizing date range calculations to speed up tests.


tests/pipeline/test_dividends.py
```diff
@@ -184,19 +184,19 @@ def get_dataset(cls):
...
             ],
             [
-                [None, '2014-01-04'], ['2014-01-05', '2014-01-14'],
-                ['2014-01-15', None]
+                ['2014-01-01', '2014-01-04'], ['2014-01-05', '2014-01-14'],
+                ['2014-01-15', '2014-01-31']
             ],
             [
-                [None, '2014-01-04'], ['2014-01-05', None]
+                ['2014-01-01', '2014-01-04'], ['2014-01-05', '2014-01-31']
```
ssanderson: Why did we remove the None's here?

mtydykov: I just wanted to have some specific begin/end date to search for in the date index. I could also put the `None`'s back and change the search logic to assume the first/last element of the date index.

zipline/pipeline/loaders/events.py
```diff
@@ -148,7 +152,8 @@ def __init__(self,
                 raise ValueError(
                     WRONG_MANY_COL_DATA_FORMAT_ERROR.format(sid=k)
                 )
-
+        self.events_by_sid = {sid: df.dropna(subset=[self.event_date_col]) for
```
ssanderson: I would have expected this filtering to hapen in `BlazeEventsLoader` at query time, rather than here.  Is there a reason we want the null rows to make it this far?

mtydykov: We don't know what the columns are that we want to filter out nulls for until this point.

zipline/pipeline/loaders/utils.py
```diff
@@ -307,10 +300,10 @@ def get_values_for_date_ranges(zip_date_index_with_vals,
         returns a pd.Series with the values indexed by the dates.
     vals_for_date_intervals : list
         A list of values for each date interval in `date_intervals`.
-    date_intervals : list
-        A list of pairs of dates, where each pair represents a date interval
-        that corresponds to the value at the same index in
-        `vals_for_date_intervals`.
+    starts : list
```
ssanderson: Can we just pass a DatetimeIndex instead of parsing all the strings individually here?

zipline/testing/fixtures.py
```diff
@@ -1120,24 +1121,24 @@ class WithNextAndPreviousEventDataLoader(WithPipelineEventDataLoader):
     ]
 
     next_date_intervals = [
-        [[None, '2014-01-04'],
```
ssanderson: Why did we remove all these None values?

mtydykov: Same reason as above. Doesn't work in a straightforward way with searchsorted (when the end date is a NaT, it maps to index 0, for example). I could change it back and modify the way `get_values_for_date_ranges` works if we decide that's better than just providing concrete start/end dates.

zipline/pipeline/loaders/events.py
mtydykov: Note: put this back for the time being, since moving the filtering to the blaze loader didn't work (blaze didn't know how to compute the query when I added filtering for nulls to it).

ssanderson: @mtydykov can you confer with @llllllllll about how you were doing the filtering here?  This case seems like it should be bread and butter for blaze...

mtydykov: We conferred last week and I tried the solution he suggested, but something was still broken and required more debugging. Talked to @joshpayne at standup on Friday and he suggested just doing this instead of debugging blaze to fix this (for now).

PR #1154, opened 2016-04-21
Add new correlation and regression factors

dmichalowicz:


zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+    x = [-.03, -.02, -.01, 0, .01] and y = [.03, -.03, .02, -.02, .04], and it
+    produced a slope of .3 and an intercept of .011.
+
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.CorrelationFactor`
+    """
+    params = ('asset', 'dependent')
+
+    def __new__(cls,
```
dmichalowicz: Not sure if overriding `__new__` is correct here.

zipline/pipeline/graph.py
```diff
@@ -70,7 +70,9 @@ def offset(self):
 
             Max number of extra rows needed by any term depending on `input`
             minus
-            Number of extra rows needed by `term`.
+            Number of rows needed by `input` for `term`
```
dmichalowicz: This is necessary to accommodate passing `Returns`, a windowed term, as an input to another factor.

dmichalowicz:
@ssanderson Ready for a first review

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+                asset,
+                returns_length,
+                regression_length,
+                dependent=False,
+                mask=NotSpecified,
+                dtype=NotSpecified,
```
ssanderson: I don't think there are any cases where it's useful to be able to customize `dtype` or `missing_value` here.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+                missing_value=NotSpecified,
+                **kwargs):
+        new_instance = super(SingleRegressionFactor, cls).__new__(
+            cls,
+            asset=asset,
+            dependent=dependent,
+            inputs=[Returns(window_length=returns_length)],
+            outputs=['alpha', 'beta'],
```
ssanderson: Is there a reason this isn't just a class-level attribute?

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+
+    Parameters
+    ----------
+    asset : zipline.assets.Asset
+        The asset to regress against all other assets.
+    returns_length : int > 1
+        Length of the lookback window over which to compute returns.
+    regression_length : int > 1
+        Length of the lookback window over which to compute each regression.
+    dependent : bool, optional
```
ssanderson: When would you want to change this?

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+    params = ['asset']
+
+    def compute(self, today, assets, out, data, asset):
+        asset_col = where(assets == asset.sid)[0][0]
+        # Use the transpose here because we want the correlations between
+        # columns, but `corrcoef` calculates between rows.
+        out[:] = corrcoef(data.T)[asset_col]
+
+
+class SingleRegressionFactor(SingleInputMixin, CustomFactor):
```
ssanderson: I'm not sure I understand what the name here is supposed to convey.  It seems like this should call out the fact that it calculates regressions against `Returns`.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+        2017-03-13    -.03     .03    .04
+        2017-03-14    -.02    -.03    .02
+        2017-03-15    -.01     .02    .01
+        2017-03-16       0    -.02    .01
+        2017-03-17     .01     .04   -.01
+        2017-03-20     .02    -.03   -.02
+        2017-03-21     .03     .01   -.02
+        2017-03-22     .04    -.02   -.02
+
+    Suppose we are interested in predicting each stock's returns from AAPL's
```
ssanderson: Regressing against SPY might be a more plausible example.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+            window_length=regression_length,
+            mask=mask,
+            dtype=dtype,
+            missing_value=missing_value,
+            **kwargs
+        )
+        return new_instance
+
+    def compute(self, today, assets, out, returns, asset, dependent):
+        asset_col = where(assets == asset.sid)[0][0]
```
ssanderson: You just want `searchsorted` here.

ssanderson: I think this is going to barf if the regressed asset is masked out.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+        my_asset = returns[:, asset_col]
+        for i in range(len(out)):
+            other_asset = returns[:, i]
+            if dependent:
+                regr_results = linregress(y=my_asset, x=other_asset)
+            else:
+                regr_results = linregress(y=other_asset, x=my_asset)
+            # `linregress` returns its results in the following order:
+            # slope, intercept, r-value, p-value, stderr
```
ssanderson: Is there a reason we're not supplying these values as well?

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
         out[:] = nanmean(close * volume, axis=0)
 
 
+class CorrelationFactor(SingleInputMixin, CustomFactor):
```
ssanderson: I wouldn't include the name `Factor` in either of these.  The user is already importing these from `zipline.pipeline.factors`.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.SingleRegressionFactor`
+    """
+    params = ['asset']
+
+    def compute(self, today, assets, out, data, asset):
+        asset_col = where(assets == asset.sid)[0][0]
+        # Use the transpose here because we want the correlations between
```
ssanderson: I think you can just pass `rowvar=0`:

```
rowvar : int, optional
    If `rowvar` is non-zero (default), then each row represents a
    variable, with observations in the columns. Otherwise, the relationship
    is transposed: each column represents a variable, while the rows
    contain observations.
```

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+    2017-03-17 (5, 5, 4, 6, 3) and MSFT's five close prices (6, 4, 9, 7, 4).
+
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.SingleRegressionFactor`
+    """
+    params = ['asset']
+
+    def compute(self, today, assets, out, data, asset):
+        asset_col = where(assets == asset.sid)[0][0]
```
ssanderson: See notes below re: searchsorted and barfing if our asset is masked.

zipline/pipeline/factors/technical.py
ssanderson: I'd also be hesitant to just call this `Correlation`, since there are several mathematical formulae that have some claim to that name.  I think what we're actually calculating here is the [Pearson Product-Moment Correlation](https://statistics.laerd.com/statistical-guides/pearson-correlation-coefficient-statistical-guide.php).  I think `scipy.stats` calls this `pearsonr`.  It also cautions that it should only be used on normally-distributed data, which I don't think is true of most of our inputs.  @CaptainKanuk thoughts on names for this?  I would lean toward something like `PearsonCorrelationCoefficient`.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +148,205 @@ def compute(self, today, assets, out, close, volume):
...
+class SingleRegressionFactor(SingleInputMixin, CustomFactor):
+    """
+    Perform an ordinary least-squares regression predicting the returns of all
+    other assets on the given asset.
+
+    **Default Inputs:** [Returns]
+
+    Parameters
+    ----------
+    asset : zipline.assets.Asset
```
ssanderson: Not having true 1-dimensional columns makes this feel really janky...I'm wondering if we should just bite the bullet and make single columns of data a first-class citizen.

zipline/pipeline/factors/technical.py
ssanderson: Another question for @CaptainKanuk and/or @twiecki:

- Should we expose other correlation measures? Various internet resources also mention [Spearman Rank Correlation](http://docs.scipy.org/doc/scipy-0.17.0/reference/generated/scipy.stats.spearmanr.html) and [Kendall's Tau Coefficient](http://docs.scipy.org/doc/scipy-0.17.0/reference/generated/scipy.stats.kendalltau.html#scipy.stats.kendalltau).

zipline/pipeline/graph.py
ssanderson: Can you explain a bit more about why this is necessary?  I don't have much intuition for why this is correct just from reading it.

zipline/pipeline/term.py
```diff
@@ -24,6 +24,8 @@
 from zipline.utils.sentinel import sentinel
 
 
+ALLOWED_WINDOWED_INPUTS = ['Returns']
```
ssanderson: This feels too brittle to me.  What do you think about just having `window_safe` attribute on `Term`, which defaults to False, and is set to True on `BoundColumn`s and on any Factors that we've explicitly marked as safe (other good candidates for window_safe expressions would be the outputs of `Factor.rank` and `Factor.zscore`).

zipline/pipeline/factors/technical.py
twiecki: Rank correlation certainly plays a big role in these factor-models, Kendall's Tau less so.

CaptainKanuk: I'm always in favor of limiting scope initially, but if it's a small amount of work Spearman Rank would certainly be something people would use.

CaptainKanuk: @ssanderson I recommended that David implement one 'Corrlation' factor with an argument to specify which type of correlation is being computed. Pearson is the default in nearly every analysis so it makes sense to have that be the default one. In terms of issues with normality, yes but that's something that a user is just going to have to learn through the lectures imo.

CaptainKanuk: I would also be okay with implementing multiple factors each named for their correlation, my only worry is that might be confusing for beginner users, adding to the complexity of pipeline. People are used to thinking about 'correlation', not the underlying type.

ssanderson: > In terms of issues with normality, yes but that's something that a user is just going to have to learn through the lectures imo.

If it's wrong to use Pearson in our domain, I don't think we should be making it easy to do an operation that's subtly wrong without thinking about it.  If we're giving people footguns, they should at least come with warning labels.

> my only worry is that might be confusing for beginner users, adding to the complexity of pipeline. People are used to thinking about 'correlation', not the underlying type.

Again, "the user doesn't know what they're doing" is not an argument for making it easy to make mistakes.  Giving things explicit names forces the user to make explicit choices, which sometimes forces them to learn about what the choices are.  We should optimize for the user's ability to make something they understand, not for their ability to make something that doesn't crash.

zipline/pipeline/factors/technical.py
dmichalowicz: @CaptainKanuk Would these be worth returning?

CaptainKanuk: I don't know how much memory/computational complexity this would add. It would be nice to have the other 3 columns, but I feel like a small percentage of use cases would need them. Generally you'd run your statistical tests at a higher level in the research environment, or a few broad statistical tests on your overall portfolio motion. Getting a p-value for each asset is gonna result in a ton of multiple comparisons bias, but could be useful for dropping assets on which the regression clearly doesn't work.

 If it doesn't slow down the method much, of it there's a way to flag whether or not the factor returns these, I would say do it.

zipline/pipeline/graph.py
```diff
@@ -70,7 +70,9 @@ def offset(self):
 
             Max number of extra rows needed by any term depending on `input`
             minus
-            Number of extra rows needed by `term`.
+            Number of extra rows needed by `input` for `term`
+            minus
+            Number of extra rows needed by `term`
```
dmichalowicz: > Can you explain a bit more about why this is necessary? I don't have much intuition for why this is correct just from reading it.

Since we now have windowed inputs, such as Returns, we get this sort of two-layered lookback effect where the Correlation factors need `X` days of extra Returns, but Returns itself needs `Y` days of extra close prices. The result is that when calculating Return's offset, we need to adjust for the fact that we need `X` extra days of a term for which we previously thought we didn't need any.

For example, if we have a Correlation factor with `returns_length = 3` and `correlation_length = 5`, then we end up with (3 - 1) + (5 - 1) = 6 extra days (that is, 6 days before our start_date) of close price in order to calculate the (5 - 1) = 4 extra days of Returns. Previously, Return's offset in this case would be 6 (the max number of extra days needed for close price) minus 2 (the number of extra rows of close price needed by Returns) = 4. If Returns has a `date_offset` (given in `SimplePipelineEngine._mask_and_dates_for_term`) of 2, the total offset for Returns is 6, bringing us back to our start_date. However, we wanted 2 extra days of Returns, so we subtract from the offset the "Number of extra rows needed by term", which for Returns is 2.

Hopefully this example helps, otherwise I can try to elaborate in a more general sense.

zipline/pipeline/factors/technical.py
dmichalowicz: Removed this

zipline/pipeline/factors/technical.py
dmichalowicz: Nope, changed this

zipline/pipeline/factors/technical.py
dmichalowicz: Removed this

tests/pipeline/test_engine.py
```diff
@@ -1237,3 +1244,174 @@ def test_dollar_volume(self):
...
+        my_asset_filter = (AssetID() != (my_asset_column + 1))
+        num_assets = len(self.sids)
+        num_days = end_date_index - start_date_index + 1
+
+        # Our correlation factors require that their baseline asset is not
+        # filtered out, so make sure that masking out our baseline asset does
+        # not take effect. That is, a filter which filters out only our
+        # baseline asset should produce the same result as if no mask was
+        # passed at all.
+        for mask in (NotSpecified, my_asset_filter):
```
dmichalowicz: Added tests with masks to make sure we don't filter out the baseline asset.

zipline/pipeline/factors/factor.py
```diff
@@ -647,12 +647,14 @@ def zscore(self, mask=NotSpecified, groupby=NotSpecified):
...
 
-        return GroupedRowTransform(
+        zscore_factor = GroupedRowTransform(
             transform=zscore,
             factor=self,
             mask=mask,
             groupby=groupby,
         )
+        zscore_factor.window_safe = True
+        return zscore_factor
```
dmichalowicz: I added `window_safe` as a class attribute to `Term`, which I think is typically what we want, but in this specific case we only want to overwrite it for an instance of `GroupedRowTransform`, not the class itself.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
         out[:] = nanmean(close * volume, axis=0)
 
 
+class RollingPearsonOfReturns(SingleInputMixin, CustomFactor):
```
dmichalowicz: These names are still subject to change.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+        for i in range(len(out)):
+            other_asset = returns[:, i]
+            regr_results = linregress(y=other_asset, x=my_asset)
+            # `linregress` returns its results in the following order:
+            # slope, intercept, r-value, p-value, stderr
+            out.alpha[i] = regr_results[1]
+            out.beta[i] = regr_results[0]
+            out.r_value[i] = regr_results[2]
+            out.p_value[i] = regr_results[3]
+            out.stderr[i] = regr_results[4]
```
dmichalowicz: I added all 5 outputs here, I think it's worth having because you can easily just use the specific outputs you need if you don't want them all.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+            inputs=[Returns(window_length=returns_length)],
+            window_length=regression_length,
+            mask=mask,
+            **kwargs
+        )
+
+    def compute(self, today, assets, out, returns, baseline_asset):
+        asset_col = searchsorted(list(assets), baseline_asset.sid)
+        my_asset = returns[:, asset_col]
+        for i in range(len(out)):
```
dmichalowicz: I couldn't find a good vectorized way of doing this, but I don't think the looping should really be any slower since it would have to compute all the regressions regardless. This is also probably easier to read as it's pretty explicit what its doing.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+
+    def __new__(cls,
+                baseline_asset,
+                returns_length,
+                correlation_length,
+                mask=NotSpecified,
+                **kwargs):
+        if mask is not NotSpecified:
+            # Make sure we do not filter out the asset of interest.
+            mask = mask | SingleAsset(asset=baseline_asset)
```
dmichalowicz: > I think this is going to barf if the regressed asset is masked out.

Added

dmichalowicz:
@ssanderson All set for a second look

tests/pipeline/test_engine.py
```diff
@@ -1242,6 +1249,177 @@ def test_dollar_volume(self):
...
+                    (RollingSpearmanOfReturns, spearmanr)):
+                correlation = CorrelationFactor(
+                    baseline_asset=my_asset,
+                    returns_length=returns_length,
+                    correlation_length=correlation_length,
+                    mask=mask,
+                )
+
+                results = self.engine.run_pipeline(
+                    Pipeline(columns={'correlation': correlation}),
```
ssanderson: Should we test the outputs here as well?

tests/pipeline/test_engine.py
```diff
@@ -1242,6 +1249,177 @@ def test_dollar_volume(self):
...
+                # On each day, calculate the expected correlation coefficient
+                # between the asset we are interested in and each other asset.
+                # Each correlation is calculated over `correlation_length`
+                # days.
+                expected_correlation_results = full_like(
+                    correlation_results, nan,
+                )
+                for day in range(num_days):
+                    my_asset_returns = returns_results.ix[
+                        day:day+correlation_length, my_asset_column,
```
ssanderson: flake8 is going to yell at you about not having whitespace around the `+` here I think.

tests/pipeline/test_engine.py
ssanderson: Is there a reason we're running these in separate pipelines?  The indentation from the for-loops makes the body here a fair amount harder to read.

tests/pipeline/test_engine.py
```diff
@@ -1242,6 +1249,177 @@ def test_dollar_volume(self):
...
+
+            # On each day, calculate the expected regression results for Y ~ X
+            # where Y is the asset we are interested in and X is each other
+            # asset. Each regression is calculated over `regression_length`
+            # days of data.
+            for day in range(num_days):
+                my_asset_returns = returns_results.ix[
+                    day:day+regression_length, my_asset_column,
+                ]
+                for asset in range(num_assets):
```
ssanderson: `iteritems` on a DataFrame will iterate over the columns.  Clearer here might be to slice once and iterate over the columns, ala:

```
for asset, returns in returns_results.iloc[day: day + regression_length].iteritems():
    ...
```

zipline/lib/adjusted_array.py
```diff
@@ -243,6 +243,20 @@ def inspect(self):
         )
 
 
+def ensure_adjusted_array(ndarray_or_adjusted_array, missing_value):
+    if isinstance(ndarray_or_adjusted_array, AdjustedArray):
+        return ndarray_or_adjusted_array
+    elif isinstance(ndarray_or_adjusted_array, ndarray):
+        return AdjustedArray(
+            ndarray_or_adjusted_array, None, {}, missing_value,
```
ssanderson: You might want to pass `AdjustedArray.NOMASK` (which is an alias of `None`) for clarity here.

tests/pipeline/test_engine.py
dmichalowicz: Surprisingly it doesn't, but will change for readability

zipline/pipeline/factors/factor.py
ssanderson: I would add this to the signature of `Term` and give it the same defaulting behavior of all the other class-level attributes.

ssanderson: (In which case this should be passed as a keyword-argument to `GroupedRowTransform`.

tests/pipeline/test_engine.py
dmichalowicz: No particular reason. I can try to consolidate all the cases into one pipeline run, or maybe keep just the mask loop.

tests/pipeline/test_engine.py
dmichalowicz: I'm not sure what you mean. The correlation factors don't define outputs.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
         out[:] = nanmean(close * volume, axis=0)
 
 
+class RollingPearsonOfReturns(SingleInputMixin, CustomFactor):
+    """
+    Calculates the Pearson product-moment correlation coefficient of the
+    returns of the given asset with the returns of all other assets.
+
+    **Default Inputs:** [Returns]
```
ssanderson: This isn't really a default anymore; it's just the only option.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
         out[:] = nanmean(close * volume, axis=0)
 
 
+class RollingPearsonOfReturns(SingleInputMixin, CustomFactor):
+    """
+    Calculates the Pearson product-moment correlation coefficient of the
+    returns of the given asset with the returns of all other assets.
```
ssanderson: We probably should say somewhere in this doc that this is what most people mean when they just say "correlation coefficient".

zipline/pipeline/factors/technical.py
ssanderson: cc @twiecki @CaptainKanuk for name bikeshedding here.

- Is the `Rolling` here necessary?
- Is `Against` a better preposition than `Of`?
- Should this be `PearsonR` ?
- Should this have the word `Correlation` and/or `Coefficient` somewhere?

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+
+    **Default Inputs:** [Returns]
+
+    Parameters
+    ----------
+    baseline_asset : zipline.assets.Asset
+        The asset to correlate with all other assets.
+    returns_length : int > 1
```
ssanderson: It occurs to me reading this that `Returns(window_length=1)` does not do what I think most people would expect it to do (namely, compute daily returns)... :/

This should probably either add 1 to returns_length, or enforce that it's greater than 2.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+        Length of the lookback window over which to compute returns.
+    correlation_length : int > 1
```
ssanderson: Ditto above; this probably needs to be at least 2 for it to be even remotely meaningful.

tests/pipeline/test_engine.py
ssanderson: nvm, I got mixed up with the regression case (which it looks like we're testing below).

zipline/pipeline/filters/filter.py
```diff
@@ -427,3 +427,12 @@ class Latest(LatestMixin, CustomFilter):
...
     pass
+
+
+class SingleAsset(CustomFilter):
+    inputs = []
+    window_length = 1
+    params = ['asset']
+
+    def compute(self, today, assets, out, asset):
+        out[:] = (assets == asset.sid)
```
ssanderson: This probably shouldn't be a `CustomFilter`, since it's always computing a constant over the input.  I would just make this a direct subclass of Filter.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+            mask=mask,
+            **kwargs
+        )
+
+    def compute(self, today, assets, out, data, baseline_asset):
+        asset_col = searchsorted(list(assets), baseline_asset.sid)
+        out[:] = corrcoef(data, rowvar=0)[asset_col]
+
+
+class RollingSpearmanOfReturns(RollingPearsonOfReturns):
```
ssanderson: I don't think this should be a subclass of `RollingPearson`.  If these have shared behavior, we should have a private `_RollingRegression` base class (or just a shared function).

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+        Length of the lookback window over which to compute returns.
+    regression_length : int > 1
+        Length of the lookback window over which to compute each regression.
+
+    Note
+    ----
+    This factor is designed to return five outputs:
+        - alpha, a factor that computes the intercepts of the regressions.
+        - beta, a factor that computes the slopes of the regressions.
+        - r_value, the correlation coefficient
```
ssanderson: The latter three values should probably still make clear that they're for the regression between each asset and the target.

ssanderson: Sidenote: `target` might be an alternative name choice for `baseline_asset`.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+    outputs = ['alpha', 'beta', 'r_value', 'p_value', 'stderr']
+    params = ['baseline_asset']
+
+    def __new__(cls,
+                baseline_asset,
+                returns_length,
+                regression_length,
+                mask=NotSpecified,
+                **kwargs):
+        if mask is not NotSpecified:
```
ssanderson: I think we want to do this `|` unconditionally.  If I pass my own custom mask, and it removes the target asset, it seems like this is going to fail pretty nastily.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +150,238 @@ def compute(self, today, assets, out, close, volume):
...
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.RollingSpearmanOfReturns`
+    """
+    def compute(self, today, assets, out, data, baseline_asset):
+        asset_col = searchsorted(list(assets), baseline_asset.sid)
+        out[:] = spearmanr(data)[0][asset_col]
+
+
+class RollingRegressionOfReturns(SingleInputMixin, CustomFactor):
```
ssanderson: The name here probably wants the word `Linear` and/or `OLS` somewhere.

zipline/pipeline/term.py
```diff
@@ -46,6 +46,9 @@ class Term(with_metaclass(ABCMeta, object)):
     # no params.
     params = ()
 
+    # Determines if a term is safe to be used as a windowed input.
+    window_safe = False
```
dmichalowicz: > I would add this to the signature of Term and give it the same defaulting behavior of all the other class-level attributes.

So should this have a default value of False or NotSpecified?

ssanderson: This should be False.

zipline/pipeline/term.py
```diff
@@ -436,7 +439,12 @@ def _validate(self):
         if self.window_length:
             for child in self.inputs:
                 if child.windowed:
```
ssanderson: I don't think you want the `.windowed` check anymore.  You just want `if not child.window_safe: error()`.

zipline/pipeline/factors/technical.py
dmichalowicz: Do you mean that this `|` needs to be applied in the case of a class-level defined mask? I'm not sure how this could be applied unconditionally, i.e. if no mask is passed

zipline/pipeline/factors/technical.py
dmichalowicz: Or maybe `target_asset`?

zipline/pipeline/factors/technical.py
ssanderson: I think I read the conditional here backwards the first time.  Ignore me...

zipline/pipeline/factors/technical.py
```diff
@@ -41,6 +46,16 @@ class Returns(CustomFactor):
...
+    window_safe = True
+
+    def _validate(self):
+        super(Returns, self)._validate()
+        if self.window_length < 2:
+            raise ValueError(
+                "'Returns' expected a window length of at least 2, but was "
+                "given {window_length}. For daily returns, use a window "
+                "length of 2.".format(window_length=self.window_length)
+            )
```
dmichalowicz: Added validation here.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.RollingSpearmanOfReturns`
+    """
+    def compute(self, today, assets, out, data, baseline_asset):
+        asset_col = searchsorted(list(assets), baseline_asset.sid)
+        out[:] = spearmanr(data)[0][asset_col]
+
+
+class RollingLinearRegressionOfReturns(CustomFactor, SingleInputMixin):
```
dmichalowicz: Changed this name, but might be too long now? cc @twiecki @CaptainKanuk

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+            **kwargs
+        )
+
+
+class RollingPearsonOfReturns(_RollingCorrelationOfReturns):
+    """
+    Calculates the Pearson product-moment correlation coefficient of the
+    returns of the given asset with the returns of all other assets. Note
+    that the Pearson correlation coeffiecient is often what is meant when
+    simply referring to a 'correlation coefficient'.
```
dmichalowicz: Updated the language here

zipline/pipeline/filters/filter.py
```diff
@@ -427,3 +428,29 @@ class Latest(LatestMixin, CustomFilter):
...
+    @classmethod
+    def static_identity(cls, asset, *args, **kwargs):
+        return (
+            super(SingleAsset, cls).static_identity(*args, **kwargs), asset,
+        )
+
+    def _compute(self, arrays, dates, assets, mask):
+        asset_column = searchsorted(list(assets), self._asset.sid)
+        mask[:, asset_column] = True
+        return mask
```
dmichalowicz: Is this the correct way to create this Filter, since it's no longer a CustomFilter?

zipline/pipeline/factors/factor.py
```diff
@@ -576,6 +576,7 @@ def zscore(row):
             factor=self,
             mask=mask,
             groupby=groupby,
+            window_safe=True,
```
dmichalowicz: Now passing `window_safe` as a keyword argument.

zipline/pipeline/factors/technical.py
CaptainKanuk: Maybe RollingLinearReturnsRegression?

twiecki: `RollingRegressionOfReturns`?

zipline/pipeline/factors/technical.py
ssanderson: @dmichalowicz the docs here still need updating to reflect the new constraint.

zipline/pipeline/filters/filter.py
ssanderson: Yes, though I think this will still fail if the requested asset isn't in the assets array (say, because the asset doesn't exist during the requested timeframe.)

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+    x = [-.03, -.02, -.01, 0, .01] and y = [.03, -.03, .02, -.02, .04], and it
+    produced a slope of .3 and an intercept of .011.
+
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.RollingPearsonOfReturns`
+    :class:`zipline.pipeline.factors.technical.RollingSpearmanOfReturns`
+    """
+    outputs = ['alpha', 'beta', 'r_value', 'p_value', 'stderr']
+    params = ['baseline_asset']
```
ssanderson: Thoughts on renaming this to `target`.  `baseline_asset` is a bit of mouthful, and you have to pass this by keyword.

zipline/pipeline/filters/filter.py
```diff
@@ -427,3 +428,29 @@ class Latest(LatestMixin, CustomFilter):
...
+        self._asset = asset
+        return super(SingleAsset, self)._init(*args, **kwargs)
+
+    @classmethod
+    def static_identity(cls, asset, *args, **kwargs):
+        return (
+            super(SingleAsset, cls).static_identity(*args, **kwargs), asset,
+        )
+
+    def _compute(self, arrays, dates, assets, mask):
```
ssanderson: Notes:

0. This shouldn't mutate the mask in place.  The input arrays here are shared with other computations, so mutating them will break all kinds of things and make pipeline results non-deterministic on the order in which they're computed.

1. You don't actually need the index to build the mask.  You can just do `zipline.utils.numpy_utils.repeat_first_axis(assets == self._asset.sid)`.

e.g.:
```
In [14]: sids = arange(5)

In [15]: sids
Out[15]: array([0, 1, 2, 3, 4])

In [16]: sid = 3

In [17]: sids == sid
Out[17]: array([False, False, False,  True, False], dtype=bool)

In [18]: repeat_first_axis(sids == sid, 10)
Out[18]:
array([[False, False, False,  True, False],
       [False, False, False,  True, False],
       [False, False, False,  True, False],
       [False, False, False,  True, False],
       [False, False, False,  True, False],
       [False, False, False,  True, False],
       [False, False, False,  True, False],
       [False, False, False,  True, False],
       [False, False, False,  True, False],
       [False, False, False,  True, False]], dtype=bool)
```

2. We still need to assert that the asset is in `assets`.

The right implementation of this is probably something like:

```
def _compute(self, arrays, dates, assets, mask):
    is_my_asset = (assets == self._asset.sid)
    if is_my_asset.sum() != 1:
        raise YourAssetDidntExistYet(...)  # Probably need a better name for this.
    out = repeat_first_axis(is_my_asset, len(mask))
    # Maybe assert that mask is True everywhere that out is?
    # i.e. assert (out & mask).sum() == len(mask)
    return out
```

zipline/pipeline/filters/filter.py
ssanderson: See note above.

zipline/pipeline/term.py
```diff
@@ -435,8 +449,10 @@ def _validate(self):
 
         if self.window_length:
             for child in self.inputs:
-                if child.windowed:
-                    raise WindowedInputToWindowedTerm(parent=self, child=child)
+                if not child.window_safe:
+                    raise WindowedInputToWindowedTerm(
```
ssanderson: The name of this exception is probably wrong now.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+
+    Note that the column for SPY is all 1's, as the correlation of any data
+    series with itself is always 1. To understand how each of the other values
+    were calculated, take for example the .15 in MSFT's column. This is the
+    correlation coefficient between SPY's returns looking back from 2017-03-17
+    (-.03, -.02, -.01, 0, .01) and MSFT's returns (.03, -.03, .02, -.02, .04).
+
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.RollingReturnsRegression`
```
ssanderson: RollingSpearman probably also belongs here.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+            window_length=correlation_length,
+            mask=mask,
+            **kwargs
+        )
+
+
+class RollingPearsonOfReturns(_RollingCorrelationOfReturns):
+    """
+    Calculates the Pearson product-moment correlation coefficient of the
+    returns of the given asset with the returns of all other assets. Note
```
ssanderson: I would separate this second sentence from the first with a blank line so it's harder to miss.  I'd also probably drop the `Note that` in favor of just: "Pearson correlation is what most people mean when they say "correlation coefficient" or "R-value".

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+    that the Pearson correlation coeffiecient is often what is meant when
```
ssanderson: typo on coefficient here

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+    correlation_length : int > 1
+        Length of the lookback window over which to compute each correlation
+        coefficient.
+
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.RollingSpearmanOfReturns`
+    """
+    def compute(self, today, assets, out, data, baseline_asset):
+        asset_col = searchsorted(list(assets), baseline_asset.sid)
```
ssanderson: why are we casting to `list` here?

ssanderson: Also `searchsorted` will do the wrong thing here if `baseline_asset` isn't in the input.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+class RollingLinearRegressionOfReturns(CustomFactor, SingleInputMixin):
+    """
+    Perform an ordinary least-squares regression predicting the returns of all
+    other assets on the given asset.
+
+    Parameters
+    ----------
+    baseline_asset : zipline.assets.Asset
+        The asset to regress against all other assets.
+    returns_length : int > 1
```
ssanderson: This needs updating.

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,253 @@ def compute(self, today, assets, out, close, volume):
...
+        Length of the lookback window over which to compute returns.
```
ssanderson: This (and other notes like it) could probably do with another sentence noting that "one day" returns actually require a window_length of 2.

zipline/pipeline/factors/technical.py
dmichalowicz: > why are we casting to list here?

`searchsorted` was failing on a pandas `Int64Index` as the input. Going to change this to `assets.values`

zipline/pipeline/factors/technical.py
dmichalowicz: > or enforce that it's greater than 2.

The constraint on returns is greater than or equal to 2, so should I write this as `>=2`? I think "greater than 1" is correct.

zipline/pipeline/factors/technical.py
dmichalowicz: `target` is good by me

zipline/errors.py
```diff
@@ -617,3 +617,10 @@ class HistoryWindowStartsBeforeData(ZiplineError):
         "History window extends before {first_trading_day}. To use this "
         "history window, start the backtest on or after {suggested_start_day}."
         )
+
+
+class NonExistentAssetInTimeFrame(ZiplineError):
+    msg = (
+        "The target asset {asset!r} does not exist for the entire timeframe "
+        "between {start_date} and {end_date}."
+    )
```
dmichalowicz: Added this for the `SingleAsset` filter

zipline/pipeline/factors/technical.py
```diff
@@ -144,6 +159,260 @@ def compute(self, today, assets, out, close, volume):
...
+    Note that the column for SPY is all 1's, as the correlation of any data
+    series with itself is always 1. To understand how each of the other values
+    were calculated, take for example the .15 in MSFT's column. This is the
+    correlation coefficient between SPY's returns looking back from 2017-03-17
+    (-.03, -.02, -.01, 0, .01) and MSFT's returns (.03, -.03, .02, -.02, .04).
+
+    See Also
+    --------
+    :class:`zipline.pipeline.factors.technical.RollingSpearmanOfReturns`
+    :class:`zipline.pipeline.factors.technical.RollingLinearRegressionOfReturns`
```
dmichalowicz: Are docstrings exempt from flake8? It's not complaining about this line that's too long

zipline/pipeline/filters/filter.py
```diff
@@ -427,3 +428,35 @@ class Latest(LatestMixin, CustomFilter):
...
+        is_my_asset = (assets == self._asset.sid)
+        out = repeat_first_axis(is_my_asset, len(mask))
+        # Raise an exception if `self._asset` does not exist for the entirety
+        # of the timeframe over which we are computing.
+        if (is_my_asset.sum() != 1) or ((out & mask).sum() != len(mask)):
+            raise NonExistentAssetInTimeFrame(
+                asset=self._asset, start_date=dates[0], end_date=dates[-1],
+            )
+        return out
```
dmichalowicz: Updated

zipline/pipeline/graph.py
```diff
@@ -62,30 +62,49 @@ def __init__(self, terms):
...
+        Factor A needs 5 extra rows of USEquityPricing.close, and Factor B
+        needs 3 extra rows of Factor A, and Factor B needs 2 extra rows of
+        USEquityPricing.close.
+
+        We load 8 extra rows of USEquityPricing.close (enough to load 5 extra
+        rows of Factor A), and the following offsets get computed::
+
+            offset[Factor A, USEquityPricing.close] == (8 - 3) - 5 == 0
+            offset[Factor B, USEquityPricing.close] == (8 - 0) - 2 == 6
+            offset[Factor B, Factor A]              == (5 - 0) - 5 == 0
```
dmichalowicz: Just noticed this is incorrect, changing to (3 - 0) - 3 == 0

zipline/pipeline/term.py
```diff
@@ -435,8 +449,8 @@ def _validate(self):
 
         if self.window_length:
             for child in self.inputs:
-                if child.windowed:
-                    raise WindowedInputToWindowedTerm(parent=self, child=child)
+                if not child.window_safe:
+                    raise NonWindowSafeInput(parent=self, child=child)
```
dmichalowicz: Changed this name

dmichalowicz:
@ssanderson Good for a final look

zipline/pipeline/factors/technical.py
ssanderson: Looks like it doesn't complain if the 80 character line contains only one identifier.

ssanderson:
Looks like we're missing test coverage for running a correlation factor without a mask with a target that doesn't exist in the time period.

tests/pipeline/test_engine.py
```diff
@@ -1242,6 +1251,241 @@ def test_dollar_volume(self):
...
+                    Pipeline(columns={'spearman_factor': spearman_factor}),
+                    self.dates[start_date_index],
+                    self.dates[end_date_index],
+                )
+            with self.assertRaises(NonExistentAssetInTimeFrame):
+                self.engine.run_pipeline(
+                    Pipeline(columns={'regression_factor': regression_factor}),
+                    self.dates[start_date_index],
+                    self.dates[end_date_index],
+                )
```
dmichalowicz: Are these tests sufficient? A few questions:
- is there a better/more specific function call than `run_pipeline` on which we can assert the exception happens?
- should these test the actual exception string output?
- should this test the case of a partially existent asset? (not sure how to do this without modifying the asset finder fixture though)

Issue #1147, opened 2016-04-20
Testing runtime is dominated by calls to gc.collect().


ssanderson:
As of bc0b117dc94b8e93081818964e3b1bdbf9b33abb, we call gc.collect() in a loop at the end of every test.  This takes on the order of .1 seconds every time (it grows larger and larger as the test suite continues, because the object graph gets larger).  For parameterized tests that don't do a lot of work, this overwhelmingly dominates the test runtime.  For example, running `test_execution_styles` on master has this profile:

```
Wed Apr 20 11:43:24 2016    stats.dat

         99530 function calls (99080 primitive calls) in 6.107 seconds

   Ordered by: cumulative time
   List reduced from 199 to 25 due to restriction <25>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      4/1    0.000    0.000    6.107    6.107 suite.py:176(__call__)
      4/1    0.001    0.000    6.107    6.107 suite.py:197(run)
      217    0.000    0.000    6.077    0.028 case.py:44(__call__)
      217    0.001    0.000    6.076    0.028 case.py:115(run)
      217    0.001    0.000    6.057    0.028 case.py:142(runTest)
      217    0.000    0.000    6.055    0.028 case.py:392(__call__)
      217    0.003    0.000    6.055    0.028 case.py:295(run)
      218    0.001    0.000    5.987    0.027 fixtures.py:53(_take_out_the_trash)
      219    5.986    0.027    5.986    0.027 {gc.collect}
      217    0.003    0.000    5.966    0.027 fixtures.py:190(tearDown)
      217    0.004    0.000    0.032    0.000 fixtures.py:167(setUp)
        6    0.000    0.000    0.028    0.005 util.py:441(try_run)
        4    0.000    0.000    0.028    0.007 suite.py:322(tearDown)
        3    0.000    0.000    0.028    0.009 suite.py:355(teardownContext)
        1    0.000    0.000    0.028    0.028 fixtures.py:124(tearDownClass)
```

PR #1155, opened 2016-04-21
TEST: Take out the trash less often.

ssanderson:
We're spending upwards of 50% of the time in many test suites calling
gc.collect() so that windows builds don't fail trying to clean up file
handles.

Only do it in the tmpdir fixtures instead.

Partial fix for https://github.com/quantopian/zipline/issues/1147.


richafrank:
This looked fine, but closing in favor of #1160.

PR #1156, opened 2016-04-21
BUG: Fetcher wasn't working properly in `before_trading_start`.

jbredeche:
We were trying to use the previous day in before_trading_start because
we were looking for the previous market minute, then normalizing it.  That's
no longer the case, as we want to use today's date for fetcher lookups
in before_trading_start.

Also refactored a bit how dataportal determines if a query should be
routed to the fetcher data structures.

### DIFF
--- a/tests/test_fetcher.py
+++ b/tests/test_fetcher.py
@@ -39,6 +39,7 @@ from .resources.fetcher_inputs.fetcher_test_data import (
     MULTI_SIGNAL_CSV_DATA,
     NON_ASSET_FETCHER_UNIVERSE_DATA,
     PALLADIUM_DATA,
+    NFLX_DATA
 )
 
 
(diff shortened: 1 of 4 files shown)


jbredeche:
@lianga888 signed off offline.

tests/test_fetcher.py
```diff
@@ -552,3 +560,38 @@ def handle_data(context, data):
...
+        self.responses.add(
+            self.responses.GET,
+            'https://fake.urls.com/fetcher_nflx_data.csv',
+            body=NFLX_DATA,
+            content_type='text/csv',
+        )
+
+        sim_params = factory.create_simulation_parameters(
+            start=pd.Timestamp("2013-06-13", tz='UTC'),
+            end=pd.Timestamp("2013-011-15", tz='UTC'),
```
lianga888: there's a small typo here with with the 011, though it probably doesn't matter

PR #1157, opened 2016-04-21
PERF: Improve read time for smaller num of assets.

ehebert:
The BcolzDailyBarReader was optimized for the pipeline case of reading
all assets at once.

Now that the reader is also used to support daily history the case of
reading a data for a small number of assets is more common, particularly
in algorithms that use the history API which have a high rotation of
assets (e.g. an algorithm which pipeline uses to set the active
universe)

Remove the bottleneck in reading a small number of assets by
conditionally reading the slice for each asset from the carray, instead
of reading the data for all equities and then indexing into that full
array. On a certain number of assets, it is still better to read all the
data at once. On the Quantopian dataset, which holds data for 20000
about for the last 10 years of equity data (where not all equities trade
over the full range), stored in 118 blosc blp files per column, the
tipping point where the 'read all' mode wins out between 3000-4000
assets.

That number was tested by trying to exercise a worst case scenario where
the equities were spread out evenly across the blp files, by stepping
along a sorted list of assets that were alive over a query range which
spanned 70 trading days.
```
size = 3000
sids = [assets[i] for i in range(0, len(assets), len(assets) /
size)][:size]
```

Also, add parameter to WithBcolzDailyBarReader fixture which allows the
test to specify what the threshold count for reading all data should be,
so that the test_us_equity_pricing can be forced into either mode to
make sure that both branches in logic are covered by all test cases.

On local dev machine this patch improves the read time of `load_raw_array`
for one asset from 100 ms to 96.5 µs. (10^5 improvement.) With reading
only asset per call a being an observed common case when populating the
non-cached values in USEquityHistoryLoader.

** Other Notes

Here is the timing table from running the following:
```
In []:
start_date
Timestamp('2013-10-02 00:00:00+0000', tz='UTC')

In []:
end_date
Timestamp('2014-01-08 00:00:00+0000', tz='UTC')

In []:
(end_date - start_date) * 5 / 7
Timedelta('70 days 00:00:00')

In []:
size = 1 # Change this value per test.
sids = [assets[i] for i in range(0, len(assets), len(assets) / size)][:size]

In []:
%%timeit
reader.load_raw_arrays([USEquityPricing.close], start_date, end_date, sids)
10000 loops, best of 3: 96.9 µs per loop
```

```
1:    96.5 µs per loop
2:   311   µs per loop
5:   654   µs per loop
10:    1.2 ms per loop
100:  10.2 ms per loop
200:  16.1 ms per loop
300:  18.9 ms per loop
500:  26.2 ms per loop
750:  32.9 ms per loop
800:  35.5 ms per loop
900:  38.2 ms per loop
1000: 38   ms per loop
2000: 63.2 ms per loop
3000: 87.3 ms per loop
4000:109   ms per loop
````

### DIFF
--- a/zipline/data/_equities.pyx
+++ b/zipline/data/_equities.pyx
@@ -14,6 +14,7 @@
 # limitations under the License.
 import bcolz
 cimport cython
+from cpython cimport bool
 
 from numpy import (
     array,
@@ -31,6 +32,7 @@ from numpy cimport (
 )
 from numpy.math cimport NAN
 
+ctypedef object carray_t
 ctypedef object ctable_t
 ctypedef object Timestamp_t
 ctypedef object DatetimeIndex_t
@@ -134,7 +136,8 @@ cpdef _read_bcolz_data(ctable_t table,
                        list columns,
                        intp_t[:] first_rows,
                        intp_t[:] last_rows,
-                       intp_t[:] offsets):
+                       intp_t[:] offsets,
+                       bool read_all):
     """
     Load raw bcolz data for the given columns and indices.
 
@@ -151,6 +154,9 @@ cpdef _read_bcolz_data(ctable_t table,
     last_rows : ndarray[intp]
     offsets : ndarray[intp
         Arrays in the format returned by _compute_row_slices.
+    read_all : bool
+        Whether to read_all sid data at once, or to read a silce from the
+        carray for each sid.
 
     Returns
     -------
@@ -160,6 +166,7 @@ cpdef _read_bcolz_data(ctable_t table,
     cdef:
         int nassets
         str column_name
+        carray_t carray
         ndarray[dtype=uint32_t, ndim=1] raw_data
         ndarray[dtype=uint32_t, ndim=2] outbuf
         ndarray[dtype=uint8_t, ndim=2, cast=True] where_nan
@@ -172,20 +179,40 @@ cpdef _read_bcolz_data(ctable_t table,
         intp_t offset
         list results = []
 
+    ndays = shape[0]
     nassets = shape[1]
     if not nassets== len(first_rows) == len(last_rows) == len(offsets):
         raise ValueError("Incompatible index arrays.")
 
     for column_name in columns:
-        raw_data = table[column_name][:]
         outbuf = zeros(shape=shape, dtype=uint32)
-        for asset in range(nassets):
-            first_row = first_rows[asset]
-            last_row = last_rows[asset]
-            offset = offsets[asset]
-            for out_idx, raw_idx in enumerate(range(first_row, last_row + 1)):
-                outbuf[out_idx + offset, asset] = raw_data[raw_idx]
+        if read_all:
+            raw_data = table[column_name][:]
 
+            for asset in range(nassets):
+                first_row = first_rows[asset]
+                last_row = last_rows[asset]
+                offset = offsets[asset]
+                if first_row <= last_row:
+                    outbuf[offset:offset + (last_row + 1 - first_row), asset] =\
+                        raw_data[first_row:last_row + 1]
+                else:
+                    continue
+        else:
+            carray = table[column_name]
+
+            for asset in range(nassets):
+                first_row = first_rows[asset]
+                last_row = last_rows[asset]
+                offset = offsets[asset]
+                out_start = offset
+                out_end = (last_row - first_row) + offset + 1
+                if first_row <= last_row:
+                    outbuf[offset:offset + (last_row + 1 - first_row), asset] =\
+                        carray[first_row:last_row + 1]
+                else:
+                    continue
+                    
         if column_name in {'open', 'high', 'low', 'close'}:
             where_nan = (outbuf == 0)
             outbuf_as_float = outbuf.astype(float64) * .001
--- a/zipline/testing/fixtures.py
+++ b/zipline/testing/fixtures.py
@@ -597,6 +597,9 @@ class WithBcolzDailyBarReader(WithTradingEnvironment, WithTmpDir):
         If this flag is set the ``bcolz_daily_bar_days`` will be the full
         set of trading days from the trading environment. This flag overrides
         ``BCOLZ_DAILY_BAR_LOOKBACK_DAYS``.
+    BCOLZ_DAILY_BAR_READ_ALL_THRESHOLD : int
+        If this flag is set, use the value as the `read_all_threshold`
+        parameter to BcolzDailyBarReader, otherwise use the default value.
 
     Methods
     -------
@@ -618,6 +621,7 @@ class WithBcolzDailyBarReader(WithTradingEnvironment, WithTmpDir):
     BCOLZ_DAILY_BAR_USE_FULL_CALENDAR = False
     BCOLZ_DAILY_BAR_START_DATE = alias('START_DATE')
     BCOLZ_DAILY_BAR_END_DATE = alias('END_DATE')
+    BCOLZ_DAILY_BAR_READ_ALL_THRESHOLD = None
     # allows WithBcolzDailyBarReaderFromCSVs to call the `write_csvs` method
     # without needing to reimplement `init_class_fixtures`
     _write_method_name = 'write'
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by ehebert** [comment]
cc: @ssanderson @jbredeche

File: zipline/data/_equities.pyx [hunk 1]
```diff
@@ -134,7 +135,8 @@ cpdef _read_bcolz_data(ctable_t table,
                        list columns,
                        intp_t[:] first_rows,
                        intp_t[:] last_rows,
-                       intp_t[:] offsets):
+                       intp_t[:] offsets,
+                       read_all):
```
jbredeche: worth typing this parameter?

zipline/data/_equities.pyx
```diff
@@ -170,22 +176,42 @@ cpdef _read_bcolz_data(ctable_t table,
         intp_t first_row
         intp_t last_row
         intp_t offset
+        intp_t out_start
+        intp_t out_end
         list results = []
 
+    ndays = shape[0]
```
jbredeche: should you declare `ndays` in the `cdef` section above?

jbredeche: oh, `ndays` looks unused.

zipline/testing/fixtures.py
```diff
@@ -651,7 +655,11 @@ def init_class_fixtures(cls):
             cls._write_method_name,
         )(cls.make_daily_bar_data())
 
-        cls.bcolz_daily_bar_reader = BcolzDailyBarReader(t)
+        if cls.BCOLZ_DAILY_BAR_READ_ALL_THRESHOLD:
+            cls.bcolz_daily_bar_reader = BcolzDailyBarReader(
+                t, cls.BCOLZ_DAILY_BAR_READER)
```
ssanderson: Typo here.  I think this branch is untested.

zipline/data/_equities.pyx
ehebert: Will remove.

zipline/data/_equities.pyx
ssanderson: Guessing this is untyped because Cython doesn't know about `bool`?  You can type this as `int` and it will behave the way you want, since `bool` is a subclass of `int`.

ehebert: Will do.

zipline/testing/fixtures.py
ehebert: Also, while looking at this, needs to be explicitly `is not None`.

zipline/data/_equities.pyx
```diff
@@ -170,22 +176,42 @@ cpdef _read_bcolz_data(ctable_t table,
...
-                outbuf[out_idx + offset, asset] = raw_data[raw_idx]
-
+        if read_all:
+            raw_data = table[column_name][:]
+
+            for asset in range(nassets):
+                first_row = first_rows[asset]
+                last_row = last_rows[asset]
+                offset = offsets[asset]
+                for out_idx, raw_idx in enumerate(
```
ssanderson: I'd be inclined to just let this run over 80 characters (flake8 doesn't look at Cython files).

zipline/data/_equities.pyx
```diff
@@ -170,22 +176,42 @@ cpdef _read_bcolz_data(ctable_t table,
...
+                    outbuf[out_idx + offset, asset] = raw_data[raw_idx]
+        else:
+            carray = table[column_name]
+
+            for asset in range(nassets):
+                first_row = first_rows[asset]
+                last_row = last_rows[asset]
+                offset = offsets[asset]
+                out_start = offset
+                out_end = (last_row - first_row) + offset + 1
```
ssanderson: Why is this range different than the one above?

zipline/testing/fixtures.py
ehebert: Aha, there was a type here, which is why the value was never set.
`_THRESHLOLD` instead of `_THRESHOLD`.

```
class BcolzDailyBarAlwaysReadAllTestCase(BcolzDailyBarTestCase):
    """
    Force tests defined in BcolzDailyBarTestCase to always read the entire
    column into memory before selecting desired asset data, when invoking
    `load_raw_array`.
    """
    BCOLZ_DAILY_BAR_READ_ALL_THRESHLOLD = 0
```

zipline/data/_equities.pyx
ehebert: Sure, will change.

zipline/data/_equities.pyx
ehebert: It is the same range as the enumeration of the `range(first_row, last_row + 1)` plus the offset (The `out_idx` from above.)
Would `outbuf[offset:offset + (last_row + 1 - first_row), asset]` be more clear that it aligns with the outbuf above?

zipline/testing/fixtures.py
ehebert: Fixed and squashed.

zipline/data/_equities.pyx
ssanderson: :+1: on your suggested rewrite.

ehebert: Squashed in.

ehebert: Going to test performance if we use a slice on `raw_data`.

zipline/data/_equities.pyx
```diff
@@ -134,7 +136,8 @@ cpdef _read_bcolz_data(ctable_t table,
                        list columns,
                        intp_t[:] first_rows,
                        intp_t[:] last_rows,
-                       intp_t[:] offsets):
+                       intp_t[:] offsets,
+                       bool read_all):
```
ssanderson: Oh, TIL you can cimport this.

zipline/data/_equities.pyx
ehebert: The following has similar performance of around 100ms when forced into `read_all` mode (tested with assets length of 1, 1000, 5000, 7000). The equivalency makes sense, since the write to a contiguous slice should be performant.

```
diff --git a/zipline/data/_equities.pyx b/zipline/data/_equities.pyx
index ef6bba6..caa8ae6 100644
--- a/zipline/data/_equities.pyx
+++ b/zipline/data/_equities.pyx
@@ -193,8 +193,11 @@ cpdef _read_bcolz_data(ctable_t table,
                 first_row = first_rows[asset]
                 last_row = last_rows[asset]
                 offset = offsets[asset]
-                for out_idx, raw_idx in enumerate(range(first_row, last_row + 1)):
-                    outbuf[out_idx + offset, asset] = raw_data[raw_idx]
+                if first_row <= last_row:
+                    outbuf[offset:offset + (last_row + 1 - first_row), asset] =\
+                        raw_data[first_row:last_row + 1]
+                else:
+                    continue
         else:
             carray = table[column_name]
```

@ssanderson shall we go with applying this patch to make the implementation more consistent?

ehebert: Also, a note the `carray` extension object does not support the buffer interface, so we still need to branch on `read_all` to get the correct types, instead of sharing an `uint32_t[:]` definition.

ssanderson: :+1: on using slicing if there's no perf difference.

ssanderson:
LGTM pending your decision on the slicing change.

zipline/data/_equities.pyx
```diff
@@ -172,20 +179,37 @@ cpdef _read_bcolz_data(ctable_t table,
...
+        else:
+            carray = table[column_name]
+
+            for asset in range(nassets):
+                first_row = first_rows[asset]
+                last_row = last_rows[asset]
+                offset = offsets[asset]
+                out_start = offset
+                out_end = (last_row - first_row) + offset + 1
+                if first_row <= last_row:
```
ssanderson: Oh, can we comment here that this is an expected case if the asset died after the end of the query?

zipline/data/_equities.pyx
ehebert: Made the slicing change, will merge once CI passes.

PR #1158, opened 2016-04-21
BUG: Restoring 'broker_order_id' to Order's dict

jbredeche:
More long-term fix is coming later, this restores existing downstream
behavior.


jbredeche:
@richafrank signed off offline

PR #1160, opened 2016-04-21
TST: What if we don't gc...

richafrank:
Looks like we removed ref cycles elsewhere, so windows builds are passing without this.


llllllllll:
:recycle:

richafrank:
@ssanderson ok with this over https://github.com/quantopian/zipline/pull/1155 ?

richafrank:
For #1147

ssanderson:
Slightly nervous that we're going to have non-deterministic windows failures depending on when gc gets invoked, but we can revisit if that happens.

PR #1162, opened 2016-04-25
DEV: Don't log an error if we can't find a matching asset/field/day triple in Fetcher

jbredeche:


jdricklefs:
Looks good to me, thanks @jbredeche!

PR #1163, opened 2016-04-25
BLD: run flake8 first, before tests

jbredeche:


ssanderson:
:+1:

jbredeche:
Merging on the thumbs up from @ssanderson

llllllllll:
I had explicitly changed this, why are we changing the order here?

jbredeche:
So that flake8 violation causes a faster failure (rather than waiting for all the tests to pass first).  Does that cause an issue?  Thought it was a no-side-effect change.

ssanderson:
@llllllllll @richafrank @jbredeche I think Joe's argument is that it's better to let the tests run to completion before the flake8 failure, because that way you can see the other test failures as well and fix them along with any style changes you need.

The counterpoint to this is that it's annoying/time-wasting when you push a branch to Travis, wait 15 minutes for it to build, and then have it fail because of flake8, forcing you to wait another 15 minutes (plus who knows how long for appveyor), for a build with no functional changes.  This is exacerbated by the fact that we now block merging of PRs with failing builds.

I could see arguments for either side here.  This is a no-op for any build that would ultimately pass, so it's really just a tradeoff on cycle time for PRs that have only flake8 failures vs PRs that have flake8 and test failures.

jikamens:
In qexec we run all the various test like things, including flake8 and nosetests, in parallel and report on all of them and then fail the build if any of them failed. No reason why we can't do that for Zipline too, is there? Even if Travis makes it hard, can't we just wrap the build steps in a shell script that Travis runs?

PR #1165, opened 2016-04-25
BUG: don't allow ordering in before_trading_start

jbredeche:


tests/test_algorithm.py
```diff
@@ -759,6 +759,48 @@ def test_invalid_order_parameters(self):
         )
         algo.run(self.data_portal)
 
+    def test_cannot_order_in_before_trading_start(self):
+        algotext = """
+from zipline.api import sid
+from zipline.errors import OrderInBeforeTradingStart
```
ehebert: I think semantically (and for easier debugging/tracebacks etc.) the `assertRaises` should be done in the test case method.

This function could be templated like this:

```
 def initialize(context):
     context.asset = sid(133)
 
     {order_func}(context.asset, {order_arg_1})
```

Then called with:
```
algot_text = algo_text_template.format(order_func=order_percent,  order_arg_1=1)
algo = TradingAlgorithm(script=algotext, sim_params=self.sim_params,
                                       data_frequency='daily', env=self.env)
with self.assertRaises(OrderInBeforeTradingStart):
    algo.run(self.data_portal)
```

ehebert: The second part of https://github.com/quantopian/zipline/pull/1165/commits/bc443d78b95be2f8796b201de4ed04268da0d697#r60965562 could also use `nose.parameterized`.

zipline/algorithm.py
```diff
@@ -87,7 +88,7 @@
     require_initialized,
     require_not_initialized,
     ZiplineAPI,
-)
+    require_not_in_before_trading_start)
```
ehebert: How about `disallowed_in_before_trading_start`?

jbredeche: sounds good, changed

tests/test_algorithm.py
jbredeche: good idea, changed

PR #1166, opened 2016-04-26
BUG: Don't save empty positions when user access non-existent position

lianga888:
Previously, whenever we try to access a missing value on the Positions
dict, we return a default Position and save it to the dict. Instead,
just return the Position

### DIFF
--- a/zipline/finance/performance/position_tracker.py
+++ b/zipline/finance/performance/position_tracker.py
@@ -350,13 +350,15 @@ class PositionTracker(object):
                     pass
                 continue
 
-            # Note that this will create a position if we don't currently have
-            # an entry
-            position = positions[sid]
+            position = zp.Position(sid)
             position.amount = pos.amount
             position.cost_basis = pos.cost_basis
             position.last_sale_price = pos.last_sale_price
             position.last_sale_date = pos.last_sale_date
+
+            # Adds the new position if we didn't have one before, or overwrite
+            # one we have currently
+            positions[sid] = position
 
         return positions
 
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -154,6 +154,7 @@ from zipline.test_algorithms import (
     call_with_bad_kwargs_get_open_orders,
     call_with_good_kwargs_get_open_orders,
     call_with_no_kwargs_get_open_orders,
+    empty_positions,
     no_handle_data,
 )
 from zipline.utils.api_support import ZiplineAPI, set_algo_instance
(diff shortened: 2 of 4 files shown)


zipline/finance/performance/position_tracker.py
```diff
@@ -350,14 +350,17 @@ def get_positions(self):
...
+            # Returns the position if we have one, or a default position if we
+            # don't
             position = positions[sid]
             position.amount = pos.amount
             position.cost_basis = pos.cost_basis
             position.last_sale_price = pos.last_sale_price
             position.last_sale_date = pos.last_sale_date
 
+            # Adds the new position if we didn't have one before
+            positions[sid] = position
```
jbredeche: it's a bit confusing to be getting or creating the position via `positions[sid]`.  I think it would be clearer to explicitly create a new position on line 355, assign it all the properties, then put it into `positions` via the code on line 362.

jbredeche:
LGTM!

zipline/finance/performance/position_tracker.py
```diff
@@ -350,14 +350,16 @@ def get_positions(self):
                     pass
                 continue
 
-            # Note that this will create a position if we don't currently have
-            # an entry
-            position = positions[sid]
+            position = zp.Position(sid)
```
richafrank: Are there any other attributes added dynamically that this could drop?

richafrank: I ask b/c I noticed that last_sale_date doesn't exist on an empty Position.

jbredeche: if there are, I'd argue they should be defined in the Position class.

This is also a nice definition of what the exact fields are that we want to make visible to the end user.

richafrank: No argument with the solution, just trying to prevent bugs with this change.

jbredeche: yup, much appreciated!

I don't know of any other dynamically-added attributes.

richafrank: Ok great.  Sounds like we should add last_sale_date to the init then.  Or maybe consolidate with https://github.com/quantopian/zipline/blob/16fd6681a66101d3f06a0bb639ee699a8905ceda/zipline/finance/performance/position.py#L43-L43?  I'm a bit confused on when each is used.

lianga888: Is last_sale_date something that the user should be able to access? It doesn't happen to be in the docs

lianga888: Not in the docs, but not inconceivable that an algo might use it. So leaving it for now

PR #1168, opened 2016-04-27
FIX: Crashing on calculating benchmarking when no trading days

lianga888:
When we run a simulation that starts and ends on the same weekend,
return an empty series for the benchmark so as to not crash

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -155,6 +155,7 @@ from zipline.test_algorithms import (
     call_with_good_kwargs_get_open_orders,
     call_with_no_kwargs_get_open_orders,
     empty_positions,
+    set_benchmark_algo,
     no_handle_data,
 )
 from zipline.utils.api_support import ZiplineAPI, set_algo_instance
@@ -1818,6 +1819,28 @@ def handle_data(context, data):
         amounts = results.amounts
         self.assertTrue(all(num_positions == 0))
         self.assertTrue(all(amounts == 0))
+
+    @parameterized.expand([
+        ('noop_algo', noop_algo),
+        ('with_benchmark_set', set_benchmark_algo)]
+    )
+    def test_zero_trading_days(self, name, algocode):
+        """
+        Test that when we run a simulation with no trading days (e.g. beginning
+        and ending the same weekend), we don't crash on calculating the
+        benchmark
+        """
+        sim_params = factory.create_simulation_parameters(
+            start=pd.Timestamp('2006-01-14', tz='UTC'),
+            end=pd.Timestamp('2006-01-15', tz='UTC')
+        )
+
+        algo = TradingAlgorithm(
+            script=algocode,
+            sim_params=sim_params,
+            env=self.env
+        )
+        algo.run(self.data_portal)
 
 
 class TestGetDatetime(WithLogger,
--- a/zipline/sources/benchmark_source.py
+++ b/zipline/sources/benchmark_source.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import pandas as pd
+
 from zipline.errors import (
     InvalidBenchmarkAsset,
     BenchmarkAssetNotAvailableTooEarly,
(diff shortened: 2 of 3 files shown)


zipline/sources/benchmark_source.py
```diff
@@ -29,7 +31,9 @@ def __init__(self, benchmark_sid, env, trading_days, data_portal,
         self.emission_rate = emission_rate
         self.data_portal = data_portal
 
-        if self.benchmark_sid:
+        if len(trading_days) == 0:
+            self._precalculated_series = pd.Series()
```
richafrank: I noticed that we're not setting the `self.benchmark_asset` attribute in this case.  Should we set it to a default value?  Started to look at who uses it, and I'm confused by `_initialize_precalculated_series` getting passed `self.benchmark_asset` but also reading it directly.  Maybe we don't need a default value, but should pass it around in general?

lianga888:
@richafrank do you want to take another look at this? I incorporated your suggestion

richafrank:
Sure thing

zipline/sources/benchmark_source.py
```diff
@@ -146,7 +150,7 @@ def _initialize_precalculated_series(self, asset, env, trading_days,
 
             return benchmark_series.pct_change()[1:]
         else:
-            start_date = self.benchmark_asset.start_date
+            start_date = asset.start_date
```
richafrank: Nice cleanup!

tests/test_algorithm.py
```diff
@@ -1819,6 +1820,51 @@ def test_empty_positions(self):
...
+
+def handle_data(context, data):
+    pass
+        """
+        algo = TradingAlgorithm(
+            script=script,
+            sim_params=sim_params,
+            env=self.env
+        )
+        algo.run(self.data_portal)
```
richafrank: Are we ok to re-use a data portal across algo runs?  I thought they were consumed by each run.

tests/test_algorithm.py
```diff
@@ -1819,6 +1820,51 @@ def test_empty_positions(self):
...
```
richafrank: Should we assert anything about the output of `run` or is not crashing enough for this case?

richafrank: Was thinking about whether the empty Series is the correct default, and whether to test that here.

lianga888: I think the empty Series makes sense to me. In any case, the output of `run` is an empty DataFrame, which is understandable, since a trivial simulation is run

tests/test_algorithm.py
richafrank: Sounds good then.

lianga888: Unsure about how exactly the data portal is created using the new test fixtures, but it appears to work for all the other tests in this file

richafrank: Every test method is called with a new data portal b/c the fixture makes it in [`init_instance_fixtures`] (https://github.com/quantopian/zipline/blob/0191d9d90354d8a83366a2374f7615299c4d1025/zipline/testing/fixtures.py#L1041-L1041) (as opposed to `init_class_fixtures`, which is class level).  But here, we're using the portal twice in a single test.

lianga888: Fixed this to separate into two tests

tests/test_algorithm.py
```diff
@@ -1819,6 +1821,29 @@ def test_empty_positions(self):
...
+        ('with_benchmark_set', set_benchmark_algo)]
+    )
+    def test_zero_trading_days(self, name, algocode):
+        """
+        Test that when we run a simulation with no trading days (e.g. beginning
+        and ending the same weekend), we don't crash on calculating the
+        benchmark
+        """
+        sim_params = copy(self.sim_params)
+        sim_params.period_start = pd.Timestamp('2006-01-14', tz='UTC')
```
richafrank: If we're setting all these attributes on the copied sim_params to new values, should we just make a new SimParams instance entirely?

tests/test_algorithm.py
richafrank: Nice

richafrank:
Had one question about the test sim_params, but LGTM!

PR #1172, opened 2016-04-28
BUG: Don't crash on dataframes with assets in index.

ssanderson:


zipline/algorithm.py
```diff
@@ -632,42 +632,96 @@ def run(self, data=None, overwrite_sim_params=True):
     def _write_and_map_id_index_to_sids(self, identifiers, as_of_date):
         # Build new Assets for identifiers that can't be resolved as
         # sids/Assets
-        identifiers_to_build = set()
-        next_sid = max(self.asset_finder.sids or (0,)) + 1
+        def is_unknown(asset_or_sid):
+            sid = op.index(asset_or_sid)
+            return self.asset_finder.retrieve_asset(
```
richafrank: Should we try to batch up these queries?

ssanderson: I don't think it matters enough to be worth the extra complexity. This branch is only hit when you pass in a dataframe or panel as your dataset, which generally means you're running one of our examples, or you've loaded up your data in memory.  In either case, you probably don't have a ton of entries, and if you do have a ton of entries, you'd be much better off writing your data statically to disk.

This code is also all obsoleted by @llllllllll's data bundle work.

richafrank: Got it.  I think even with bundles, there will be example cases where people have a dataframe - should we add a unit test for this?

ssanderson: My understanding is that after the bundle PR, the frame/panel case will get intercepted much earlier so that we no longer need to mutate an existing AssetFinder in place, which renders this obsolete.  The fix being made here is for a specific pattern that arises only on Q research where you have asset objects loaded from an external resource, but you don't have the full asset finder available.  I think the right fix for that is to have a proper `run_algorithm` function that hides the details of resource construction from the user.

zipline/algorithm.py
```diff
@@ -632,42 +635,96 @@ def run(self, data=None, overwrite_sim_params=True):
...
+        elif new_sids:
+            frame_to_write = make_simple_equity_info(
+                new_sids,
+                start_date=self.sim_params.period_start,
+                end_date=self.sim_params.period_end,
+                symbols=symbols,
             )
+        elif new_symbols:
+            existing_sids = self.asset_finder.sids
+            first_sid = max(existing_sids) + 1 if existing_sids else 0
```
dmichalowicz: Should this be called `last_sid`?

zipline/algorithm.py
```diff
@@ -632,42 +635,96 @@ def run(self, data=None, overwrite_sim_params=True):
...
+                    exchange=map_getattr(new_assets, 'exchange'),
                 ),
+                index=map_getattr(new_assets, 'sid'),
+            )
+        elif new_sids:
+            frame_to_write = make_simple_equity_info(
+                new_sids,
+                start_date=self.sim_params.period_start,
+                end_date=self.sim_params.period_end,
+                symbols=symbols,
```
dmichalowicz: `symbols` is undefined. could this just be `map(str, sids)`?

zipline/algorithm.py
```diff
@@ -586,7 +585,11 @@ def run(self, data=None, overwrite_sim_params=True):
                         env=self.trading_environment
                     )
 
-                copy_panel = data.copy()
+                # Backwards compat for old column names.  These are still used
+                # on Quantopian Research.
+                copy_panel = data.rename(
+                    {'close_price': 'close', 'open_price': 'open'}
```
dmichalowicz: This needs `columns={'close_price': 'close', 'open_price': 'open'}`

dmichalowicz:
LGTM

PR #1173, opened 2016-04-28
Quandl wiki loader

llllllllll:


docs/source/beginner-tutorial.rst
```diff
@@ -106,60 +103,60 @@ After you installed zipline you should be able to execute the following
...
-don't have to supply the command line args all the time (see the .conf
-files in the examples directory).
+   Usage: __main__.py run [OPTIONS]
+
+     Run a backtest for the given algorithm.
+
+   Options:
+     -f, --algofile FILENAME         The file that contains the algorithm to run.
+     -t, --algotext TEXT             The algorithm script to run.
+     -D, --define TEXT               Define define a name to be bound in the
```
richafrank: typo "define" twice

scripts/run_algo.py
```diff
@@ -1,26 +0,0 @@
-#!/usr/bin/env python
```
richafrank: Should we keep run_algo for a while, but deprecated?

richafrank: It could be a passthrough to the new entry point.

zipline/data/__init__.py
```diff
@@ -1,8 +0,0 @@
-from . import loader
-from .loader import (
-    load_from_yahoo, load_bars_from_yahoo, load_prices_from_csv,
-    load_prices_from_csv_folder
-)
-
-__all__ = ['loader', 'load_from_yahoo', 'load_bars_from_yahoo',
-           'load_prices_from_csv', 'load_prices_from_csv_folder']
```
richafrank: Should we be deprecating these before removing them?

scripts/run_algo.py
llllllllll: I think that for the 1.0 release it would be okay to just rename this, I imagine most people type this out each time they use it so it won't break much

zipline/data/__init__.py
llllllllll: I could do that

zipline/utils/math_utils.py
```diff
@@ -12,11 +12,37 @@
...
+    --------
+    numpy.isclose
+
+    Notes
+    -----
+    This function is just a scalar version of numpy.isclose for performance.
+    See the docstring of ``isclose`` for more information about ``atol`` and
+    ``rtol``.
+    """
+    if equal_nan and a is nan and b is nan:
```
richafrank: Do we only care about numpy.nan?

```python
In [1]: float('nan') is float('nan')
Out[1]: False

In [2]: from numpy import nan

In [3]: float('nan') is nan
Out[3]: False
```

zipline/utils/paths.py
```diff
@@ -0,0 +1,223 @@
...
+"""
+from errno import EEXIST
+import os
+from os.path import exists, expanduser, join
+
+import pandas as pd
+
+
+def hidden(path):
+    """Check if a path is hidden.
```
richafrank: Does this need to be a cross-platform notion of hidden, e.g. using windows file attributes, or just looking at the name?

zipline/utils/cli.py
```diff
@@ -1,285 +1,118 @@
...
+        The timezone to parse the string as.
+        By default the timezone will be infered from the string or naiive.
+    """
+    def parser(self, value):
+        return super(Time, self).parser(value).time()
 
 
-def maybe_show_progress(it, show_progress, **kwargs):
-    """Optionally show a progress bar for the given iterator.
+class Timedelta(_DatetimeParam):
```
richafrank: Where is this used?

llllllllll: I don't think I ended up using it, We might want to add an `--older-than` flag to `clean` which would want this

zipline/utils/math_utils.py
llllllllll: I will use np.isnan

zipline/utils/paths.py
llllllllll: this is mainly to filter out the `.cache` dir we write. Until we really care about it being hidden I think this is fine

zipline/utils/run_algo.py
```diff
@@ -0,0 +1,317 @@
...
+    if bundle is not None:
+        bundle_data = load(
+            bundle,
+            environ,
+            bundle_timestamp,
+        )
+
+        env = TradingEnvironment(asset_db_path=str(
+            bundle_data.asset_finder.engine.url,
+        )[len('sqlite:///'):])
```
ssanderson: This feels a little brittle.  Could we make this a regex split?

zipline/utils/run_algo.py
```diff
@@ -0,0 +1,317 @@
...
+            adjustment_reader=bundle_data.adjustment_reader,
+        )
+
+    perf = TradingAlgorithm(
+        namespace=namespace,
+        capital_base=capital_base,
+        start=start,
+        end=end,
+        env=env,
+        **{
```
ssanderson: TIL that this is valid syntax.

docs/source/beginner-tutorial.rst
```diff
@@ -106,60 +103,59 @@ After you installed zipline you should be able to execute the following
...
+                                     they may refer to previously defined names.
+     --data-frequency [minute|daily]
+                                     The data frequency of the simulation.
+                                     [default: daily]
+     --capital-base FLOAT            The starting capital for the simulation.
+                                     [default: 10000000.0]
+     -b, --bundle BUNDLE-NAME        The data bundle to use for the simulation.
+                                     [default: quandl]
+     --bundle-timestamp TIMESTAMP    The date to lookup data on or before.
+                                     [default: 2016-04-28 20:14:38.252921+00:00]
```
richafrank: Is this date actually hard-coded?

llllllllll: no, that is `pd.Timestamp.utcnow()(`, I wonder if we can make this look nicer

llllllllll: I could set show_default to false and just say that it is the current time

docs/source/bundles.rst
```diff
@@ -0,0 +1,275 @@
...
+.. code-block:: bash
+
+   # clean everything older than <date>
+   $ python -m zipline clean <bundle> --before <date>
+
+   # clean everything newer than <date>
+   $ python -m zipline clean <bundle> --after <date>
+
+   # keep everything in the range of [before, after] and delete the rest
+   $ python -m zipline clean <bundle> --before <date> --after <after>
```
richafrank: Will it keep everything in the range `(,after] ∪ [before,)` if I specify a `before` that follows the `after`?

docs/source/bundles.rst
```diff
@@ -0,0 +1,275 @@
...
+
+   $ python -m zipline run --bundle <bundle> --algofile algo.py ...
+
+
+We may also specify the date to use to look up the bundle data with the
+``--bundle-date`` option. This will cause us to the the most recent bundle
+ingestion that is less than or equal to the ``bundle-date``. This is how we can
+run backtests with older data. The reason that this uses a less than or equal to
+relationship is that we can specify the date that we ran an old backtest and get
+the same data that would have been available to us on that date. the
```
richafrank: Should capitalize "The" at the end of this line.

docs/source/bundles.rst
```diff
@@ -0,0 +1,275 @@
...
+
+Quandl WIKI Bundle
+``````````````````
+
+By default zipline comes with the ``quandl`` data bundle which uses quandl's
+`WIKI dataset <https://www.quandl.com/data/WIKI>`_. This bundle includes daily
+pricing data, splits, cash dividends, and asset metadata. This is the bundle
+that ``run`` will use by default if no other bundle is specified. To ingest this
+data bundle we recommend creating an account on quandl.com to get an API key to
+be able to make more API requests. Once we have an API key we may run:
```
richafrank: "more API requests" - more than what?

docs/source/bundles.rst
llllllllll: right now that would delete everything:

```python
    def should_clean(name):
        dt = pd.Timestamp(name)

        return (
            (
                (before is not None and dt < before) or
                (after is not None and dt > after)
            ) and
            not in_last_n(dt)
        )
```

docs/source/bundles.rst
llllllllll: does "more per day" work here?

docs/source/bundles.rst
```diff
@@ -0,0 +1,275 @@
...
+Zipline also ships with a factory function for creating a data bundle out of a
+set of tickers from yahoo: :func:`~zipline.data.bundles.yahoo_equities`.
+This makes it easy to pre-download and cache the data for a set of equities from
+yahoo. This includes daily pricing data along with splits, cash dividends, and
+inferred asset metadata. To create a bundle from a set of equities, add the
+following to your ``~/.zipline/extensions.py`` file:
+
+.. code-block:: python
+
+   from zipline.bundles import register, yahoo_equities
```
ssanderson: unused import?

llllllllll: this should be `yahoo_equities(equities)` in the call to register

docs/source/bundles.rst
```diff
@@ -0,0 +1,275 @@
...
+
+though we may still run ``ingest`` as an anonymous quandl user (with no API
+key). We may also set the ``QUANDL_DOWNLOAD_ATTEMPTS`` environment variable to
+an integer which is the number of attempts that should be made to download data
+from quandls servers. By default this will be 5, meaning that we will retry each
+attempt 5 times.
+
+.. note::
+
+   This is not the total number of allowed failures, just the number of allowed
```
ssanderson: It's unclear what the word "this" refers to here.  There are several other usages of the word "This" in this paragraph that should probably be replaced with explicit nouns.  (e.g. "this bundle includes" vs "the quandl bundle includes.).

docs/source/bundles.rst
```diff
@@ -0,0 +1,275 @@
...
+          adjustment_writer,
+          calendar,
+          cache,
+          show_progress)
+
+``environ``
+```````````
+
+``environ`` is a mapping representing the environment variables to use. This is
+where any custom arguments needed for the ingestion should be passed, for
```
richafrank: Should we say that "this is where any custom arguments need for the ingestion WILL be passed"?  Or are we asking the bundle author to do something here?

docs/source/bundles.rst
```diff
@@ -0,0 +1,275 @@
...
+``asset_db_writer``
+```````````````````
+
+``asset_db_writer`` is an instance of :class:`~zipline.assets.AssetDBWriter`.
+This is the writer for the asset metadata which provides the asset lifetimes and
+the symbol to asset id (sid) mapping. This may also contain the asset name,
+exchange and a few other columns. To write data, invoke
+:meth:`~zipline.assets.AssetDBWriter.write`. This is passed dataframes for the
+various pieces of metadata, more information about the format of the data exists
+in the docs for write.
```
richafrank: "To write data, invoke write with dataframes for the various pieces of metadata.  More information about the format..."

docs/source/bundles.rst
```diff
@@ -0,0 +1,275 @@
...
+``calendar`` is a ``pandas.DatetimeIndex`` object holding all of the trading
+days that the bundle should load data for. This is to help some bundles generate
+queries for the days needed.
+
+``cache``
+`````````
+
+``cache`` is an instance of :class:`~zipline.utils.cache.dataframe_cache`. This
+object is a mapping from strings to dataframes. This object is provided in case
+an ingestion crashes part way through. The idea is that as the ingest function
```
richafrank: typo: extra "as"

docs/source/bundles.rst
richafrank: That makes more sense to me.

zipline/__init__.py
```diff
@@ -28,19 +29,18 @@
...
 
-try:
-    ip = get_ipython()  # flake8: noqa
-except NameError:
-    pass
-else:
-    ip.register_magic_function(utils.parse_cell_magic, "line_cell", "zipline")
-    del ip
+
+def load_ipython_extension(ipython):
```
richafrank: What calls this?

llllllllll: %load_ext

tests/resources/yahoo_samples/rebuild_samples.py
```diff
@@ -0,0 +1,78 @@
...
+def pricing_for_sid(sid):
+    modifier = {
+        'Low': 0,
+        'Open': 1,
+        'Close': 2,
+        'High': 3,
+        'Volume': 0,
+    }
+
+    def c(name):
```
richafrank: What does "c" stand for here?  In the previous module, it stood for "cache".  Could we use longer names?

tests/resources/rebuild_example_data
```diff
@@ -0,0 +1,88 @@
...
+
+from zipline import examples, run_algorithm
+from zipline.testing import test_resource_path, tmp_dir
+from zipline.utils.cache import dataframe_cache
+
+banner = """
+Please verify that the new perfomance is more correct than the old performance.
+
+To do this, please inspect `new` and `old` which are mappings from the name of
+the examle to the results.
```
richafrank: typo "example"

tests/resources/rebuild_example_data
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
+from code import InteractiveConsole
```
richafrank: TIL

tests/resources/rebuild_example_data
```diff
@@ -0,0 +1,88 @@
...
+import readline  # noqa
```
richafrank: Why the import?

zipline/__main__.py
```diff
@@ -0,0 +1,354 @@
...
+            else:
+                __import__(ext)
+        except Exception as e:
+            if strict_extensions:
+                # if `strict_extensions` we should raise the actual exception
+                # and fail
+                raise
+            # without `strict_extensions` we should just log the failure
+            click.echo(
+                'Failed to load extension: %r\n%s' % (extension, e),
```
richafrank: Should be logging `ext`

zipline/__main__.py
```diff
@@ -0,0 +1,354 @@
...
+        # to be told they need to pass the second argument also
+        ctx.fail(
+            "must specify dates with '-s' / '--start' and '-e' / '--end'",
+        )
+    if start is None:
+        ctx.fail("must specify a start date with '-s' / '--start'")
+    if end is None:
+        ctx.fail("must specify an end date with '-s' / '--end'")
+
+    if not (algotext is not None) ^ (algofile is not None):
```
richafrank: `==` also works as XNOR on booleans - possibly a little clearer.

ssanderson: I've also written things like this as:

`if sum([algotext is None, algofile is None]) != 1:`

zipline/data/bundles/core.py
```diff
@@ -0,0 +1,467 @@
...
+from zipline.utils.compat import mappingproxy
+from zipline.utils.input_validation import ensure_timestamp, optionally
+import zipline.utils.paths as pth
+from zipline.utils.preprocess import preprocess
+from zipline.utils.tradingcalendar import trading_days, open_and_closes
+
+
+def asset_db_path(bundle_name, timestr, environ=None):
+    return pth.data_path(
+        [bundle_name, timestr, 'assets-%d.sqlite' % ASSET_DB_VERSION],
```
richafrank: How do these locations compare to the current locations?

zipline/data/bundles/core.py
```diff
@@ -0,0 +1,467 @@
...
+    )
+
+
+_BundlePayload = namedtuple(
+    '_BundlePayload',
+    'calendar opens closes minutes_per_day ingest',
+)
+
+
+class UnknownBundle(click.ClickException, LookupError):
```
richafrank: Nice, I didn't know about LookupError.

zipline/utils/cache.py
```diff
@@ -104,11 +100,210 @@ def __init__(self, cache=None):
...
+    The file is committed on __exit__ if there are no exceptions.
+    """
+    def __init__(self, final_path, *args, **kwargs):
+        self.name = mkdtemp()
+        self._final_path = final_path
+
+    def commit(self):
+        """Sync the temporary directory to the final path.
+        """
+        copytree(self.name, self._final_path)
```
ssanderson: Might be worth noting in the docstring that this isn't truly transactional.  If the commit fails, we don't roll back to the original state.

zipline/data/bundles/core.py
```diff
@@ -0,0 +1,467 @@
...
+                transactional_file(
+                    adjustment_db_path(name, timestr, environ=environ),
+                ) as adjustment_db_file:
+            # we use `cleanup_on_failure=False` so that we don't purge the
+            # cache directory if the load fails in the middle
+            daily_bar_writer = BcolzDailyBarWriter(
+                daily_bars_dir.name,
+                bundle.calendar,
+            )
+            daily_bar_writer.write(())
```
richafrank: Why do we `write` here?

zipline/data/data_portal.py
```diff
@@ -232,15 +240,23 @@ def highs(self, assets, dt):
...
                         after_last = pd.Timestamp(
                             last_visited_dt + self._one_min, tz='UTC')
-                        window = self._minute_reader.unadjusted_window(
-                            ['high'], after_last, dt, [asset])
+                        window = self._minute_reader.load_raw_arrays(
+                            ['high'],
+                            after_last,
+                            dt,
+                            [asset],
+                        )[0].T
```
richafrank: Just pattern matching here, but why do some of these changes include a transposition, but not others?

zipline/data/minute_bars.py
```diff
@@ -877,7 +911,9 @@ def unadjusted_window(self, fields, start_dt, end_dt, sids):
                             excl_start - start_idx:excl_stop - start_idx + 1]
                         values = np.delete(values, excl_slice)
                 where = values != 0
-                out[i, where] = values[where]
+                # first slice down to len(where) because we might not have
+                # written data for all the minutes requested
+                out[:len(where), i][where] = values[where]
```
richafrank: Were these changes for consistency with other writers?

zipline/utils/paths.py
```diff
@@ -0,0 +1,223 @@
+"""
```
richafrank: Are there changes here besides the git move from zipline/data/paths.py?

zipline/data/us_equity_pricing.py
```diff
@@ -365,6 +364,7 @@ def iterator(iterator=iterator, assets=set(assets)):
         full_table.attrs['last_row'] = last_row
         full_table.attrs['calendar_offset'] = calendar_offset
         full_table.attrs['calendar'] = calendar.asi8.tolist()
+        full_table.flush()
```
richafrank: Why do we need to `flush` here?

zipline/examples/__init__.py
```diff
@@ -0,0 +1,17 @@
+from glob import glob
+from importlib import import_module
+import os
+
+for f in os.listdir(os.path.dirname(__file__)):
+    if not f.endswith('.py') or f == '__init__.py':
+        continue
+    modname = f[:-len('.py')]
+    globals()[modname] = import_module('.' + modname, package=__name__)
```
richafrank: What are we doing here that differs from the built-in import machinery?

zipline/utils/cache.py
```diff
@@ -104,11 +100,210 @@ def __init__(self, cache=None):
...
+        The directory path to the cache. Files will be written as
+        ``path/<keyname>``.
+    lock : Lock, optional
+        Thread lock for multithreaded/multiprocessed access to the cache.
+        If not provided no locking will be used.
+    clean_on_failure : bool, optional
+        Should the directory be cleaned up if an exception is raised in the
+        context manager.
+    serialize : {'msgpack', 'pickle'}, optional
+        How should the data be serialized.
```
richafrank: Could we include the defaults in the docstring for these optional parameters?

zipline/utils/cache.py
```diff
@@ -104,11 +100,210 @@ def __init__(self, cache=None):
...
+    The cache uses a temporary file format that is subject to change between
+    versions of zipline.
+    """
+    @expect_element(serialization={'msgpack', 'pickle'})
+    def __init__(self,
+                 path=None,
+                 lock=None,
+                 clean_on_failure=True,
+                 serialization='msgpack'):
+        self.path = path if path is not None else mkdtemp()
```
richafrank: I was surprised to see that we create a temp dir even if you don't enter the cache context.

zipline/utils/cache.py
```diff
@@ -104,11 +100,210 @@ def __init__(self, cache=None):
...
+
+        with self.lock:
+            rmtree(self.path)
+
+    def __getitem__(self, key):
+        with self.lock:
+            try:
+                return self.deserialize(self._keypath(key))
+            except UnboundLocalError:
+                # This is how pandas fails if the file doesn't exist! #pandas
```
richafrank: Wow

zipline/utils/cache.py
```diff
@@ -104,11 +100,210 @@ def __init__(self, cache=None):
...
+                raise KeyError(key)
+
+    def __setitem__(self, key, value):
+        with self.lock:
+            try:
+                del self[key]
```
richafrank: Why do we remove it first?

zipline/utils/run_algo.py
```diff
@@ -0,0 +1,326 @@
...
+                run_path(ext, run_name='<extension>')
+            else:
+                __import__(ext)
+        except Exception as e:
+            if strict:
+                # if `strict` we should raise the actual exception and fail
+                raise
+            # without `strict` we should just log the failure
+            warnings.warn(
+                'Failed to load extension: %r\n%s' % (extensions, e),
```
richafrank: `extensions` -> `ext` ?

zipline/utils/run_algo.py
```diff
@@ -0,0 +1,326 @@
...
+
+    if output == '-':
+        click.echo(str(perf))
+    elif output != os.devnull:  # make the zipline magic not write any data
+        perf.to_pickle(output)
+
+    return perf
+
+
+def _load_extensions(default, extensions, strict, environ):
```
richafrank: Should `cli` call this function instead of duplicating it?

zipline/utils/cache.py
```diff
@@ -104,11 +100,210 @@ def __init__(self, cache=None):
...
+    Notes
+    -----
+    The file is committed on __exit__ if there are no exceptions.
+    """
+    def __init__(self, final_path, *args, **kwargs):
+        self._tmpfile = NamedTemporaryFile(*args, **kwargs)
+        self._final_path = final_path
+
+    def commit(self):
+        """Sync the temporary file to the final path.
```
ssanderson: Same note as below.

docs/source/bundles.rst
llllllllll: This is where the bundle author should find any arguments that it needs. For example, quandl needs to know what the auth key is so it uses the env. Do you have a suggestion on how to word this?

tests/resources/yahoo_samples/rebuild_samples.py
llllllllll: I was just thinking `column`, it made the dataframe definition look nicer.

tests/resources/rebuild_example_data
llllllllll: this enables readline support for prompts in python like `input` and the interactive console.

zipline/data/bundles/core.py
llllllllll: There is no current location for these files on disk that I know of.

zipline/data/minute_bars.py
llllllllll: There are two changes here. The first is that we are transposing the output to be more consistent with the other readers. The second is that we are adding some extra slicing to make sure the indicies are aligned when write. This was causing warnings.

zipline/data/bundles/core.py
llllllllll: to ensure that the ctables exist when we try to create the adjustment reader which will eagerly open the tables.

zipline/data/us_equity_pricing.py
llllllllll: I wanted to ensure that this was flushed to make it easier to reason about the state of the files on disk when debugging.

zipline/examples/__init__.py
llllllllll: Do you mean `__import__`? If so, `__import__('a.b.c')` returns `a`, not `a.b.c`. I wanted to just get the last module.

zipline/examples/buy_and_hold.py
```diff
@@ -33,18 +30,12 @@ def handle_data(context, data):
         context.has_ordered = True
 
 
-if __name__ == '__main__':
+def _test_args():
+    """Extra arguments to use when testing this example.
```
llllllllll: does this docstring seem like enough?

zipline/utils/cache.py
llllllllll: numpy does not suggest including the defaults because you can see that if you inspect the function signature. The only time that the default should be included is if it is lazy, like `None` which creates an empty dict or something.

zipline/utils/cache.py
llllllllll: this is standard for context managers like this. `open` will open the file even if you don't enter the context.

zipline/utils/cache.py
llllllllll: I originally was using castra for this, I will just remove this.

zipline/utils/run_algo.py
llllllllll: yeah, I pulled this out so they could share. Looks like I missed a critical step ;)

zipline/__main__.py
llllllllll: I think the `==` is the most clear here, I sort of backed into the xnor there

zipline/utils/paths.py
llllllllll: I made the defaults for os.environ more standard. They used to be None and then bind to os.environ.copy().

zipline/data/data_portal.py
llllllllll: some places we are doing a ravel and taking the first element anyways.

tests/resources/yahoo_samples/rebuild_samples.py
richafrank: Would you be opposed to naming it `column` then?

zipline/data/bundles/core.py
richafrank: Is that because the default was in-memory?

zipline/data/bundles/core.py
richafrank: Where does that happen?  Could we have a comment here?

zipline/data/us_equity_pricing.py
richafrank: Any meaningful performance hit?

zipline/examples/__init__.py
richafrank: Sorry, I meant, why are we walking the file system looking for submodules to import and add as attributes of this module, when `import module.submodule` should do that already?

zipline/utils/cache.py
richafrank: Oh, good to know.

zipline/utils/cache.py
richafrank: That's fair.

zipline/examples/__init__.py
llllllllll: ah, I wanted to be able to use `dir(zipline.examples)` to find the examples in the tests and as a user

zipline/data/bundles/core.py
llllllllll: yeah

zipline/examples/buy_and_hold.py
richafrank: What do you think of "Extra arguments to use when zipline automated tests run this example."?

zipline/data/us_equity_pricing.py
llllllllll: this should normally get flushed soon anyways so it should be fine.

PR #1174, opened 2016-04-29
ENH: Add support for strings in Pipeline.

ssanderson:
- Adds a new class, ``LabelArray``, which is a subclass of np.ndarray.
  LabelArray is conceptually similar to pandas.Categorical, in that it
  stores data with many duplicate values as indices into an array of
  unique values.  For string data with many duplicates (e.g. time-series
  of tickers or or industry classifications), this provides multiple
  orders of magnitude of improvement when doing string operations,
  especially string comparison/matching operations.

- Adds a new generic object "specialization" for `AdjustedArrayWindow`,
  and a corresponding ObjectOverwrite adjustment.

- Adds a new ``postprocess`` method to ``zipline.pipeline.term.Term``.
  This method is called on the final result of any pipeline expression
  after screen filtering has occurred. The default implementation of
  ``postprocess`` is identity, but Classifier overrides it to coerce
  string columns into pandas.Categoricals before presenting them to the
  user.

### DIFF
--- a/tests/pipeline/test_adjusted_array.py
+++ b/tests/pipeline/test_adjusted_array.py
@@ -27,6 +27,7 @@ from zipline.lib.adjustment import (
 from zipline.lib.adjusted_array import AdjustedArray, NOMASK
 from zipline.lib.labelarray import LabelArray
 from zipline.testing import check_arrays, parameter_space
+from zipline.utils.compat import unicode
 from zipline.utils.numpy_utils import (
     coerce_to_dtype,
     datetime64ns_dtype,
@@ -84,7 +85,7 @@ def as_labelarray(initial_dtype, missing_value, array):
     """
     return LabelArray(
         array.astype(initial_dtype),
-        missing_value=initial_dtype.type(''),
+        missing_value=initial_dtype.type(missing_value),
     )
 
 
@@ -367,9 +368,9 @@ class AdjustedArrayTestCase(TestCase):
             ),
             _gen_unadjusted_cases(
                 'object_ndarray',
-                make_input=lambda a: a.astype(str).astype(object),
-                make_expected_output=as_labelarray(bytes_dtype, b''),
-                missing_value=b'',
+                make_input=lambda a: a.astype(unicode).astype(object),
+                make_expected_output=as_labelarray(unicode_dtype, u''),
+                missing_value='',
             ),
             # Test passing a LabelArray directly to AdjustedArray.
             _gen_unadjusted_cases(
@@ -380,17 +381,17 @@ class AdjustedArrayTestCase(TestCase):
             ),
             _gen_unadjusted_cases(
                 'unicode_labelarray',
-                make_input=as_labelarray(unicode_dtype, u''),
-                make_expected_output=as_labelarray(bytes_dtype, u''),
+                make_input=as_labelarray(unicode_dtype, None),
+                make_expected_output=as_labelarray(unicode_dtype, None),
                 missing_value=u'',
             ),
             _gen_unadjusted_cases(
                 'object_labelarray',
                 make_input=(
-                    lambda a: LabelArray(a.astype(str).astype(object), b'')
+                    lambda a: LabelArray(a.astype(unicode).astype(object), u'')
                 ),
-                make_expected_output=as_labelarray(bytes_dtype, b''),
-                missing_value=b'',
+                make_expected_output=as_labelarray(unicode_dtype, ''),
+                missing_value='',
             ),
         )
     )
@@ -442,8 +443,8 @@ class AdjustedArrayTestCase(TestCase):
                 ),
             ),
             # There are six cases here:
-            # Using np.bytes/np.unicode/python string arrays as inputs.
-            # Passing np.bytes/np.unicode/python string arrays to LabelArray,
+            # Using np.bytes/np.unicode/object arrays as inputs.
+            # Passing np.bytes/np.unicode/object arrays to LabelArray,
             # and using those as input.
             #
             # The outputs should always be LabelArrays.
@@ -461,9 +462,9 @@ class AdjustedArrayTestCase(TestCase):
             ),
             _gen_unadjusted_cases(
                 'object_ndarray',
-                make_input=lambda a: a.astype(str).astype(object),
-                make_expected_output=as_labelarray(bytes_dtype, b''),
-                missing_value=b'',
+                make_input=lambda a: a.astype(unicode).astype(object),
+                make_expected_output=as_labelarray(unicode_dtype, u''),
+                missing_value=u'',
             ),
             _gen_unadjusted_cases(
                 'bytes_labelarray',
@@ -474,16 +475,19 @@ class AdjustedArrayTestCase(TestCase):
             _gen_unadjusted_cases(
                 'unicode_labelarray',
                 make_input=as_labelarray(unicode_dtype, u''),
-                make_expected_output=as_labelarray(bytes_dtype, u''),
+                make_expected_output=as_labelarray(unicode_dtype, u''),
                 missing_value=u'',
             ),
             _gen_unadjusted_cases(
                 'object_labelarray',
                 make_input=(
-                    lambda a: LabelArray(a.astype(str).astype(object), b'')
+                    lambda a: LabelArray(
+                        a.astype(unicode).astype(object),
+                        None,
+                    )
                 ),
-                make_expected_output=as_labelarray(bytes_dtype, b''),
-                missing_value=b'',
+                make_expected_output=as_labelarray(unicode_dtype, u''),
+                missing_value=None,
             ),
         )
     )
--- a/tests/pipeline/test_adjusted_array.py
+++ b/tests/pipeline/test_adjusted_array.py
@@ -9,19 +9,23 @@ from nose_parameterized import parameterized
 from numpy import (
     arange,
     array,
+    asarray,
+    dtype,
     full,
     where,
 )
-from numpy.testing import assert_array_equal
 from six.moves import zip_longest
+from toolz import curry
 
 from zipline.errors import WindowLengthNotPositive, WindowLengthTooLong
 from zipline.lib.adjustment import (
     Datetime64Overwrite,
     Float64Multiply,
     Float64Overwrite,
+    ObjectOverwrite,
 )
 from zipline.lib.adjusted_array import AdjustedArray, NOMASK
+from zipline.lib.labelarray import LabelArray
 from zipline.testing import check_arrays, parameter_space
 from zipline.utils.numpy_utils import (
     coerce_to_dtype,
@@ -29,6 +33,7 @@ from zipline.utils.numpy_utils import (
     default_missing_value_for_dtype,
     float64_dtype,
     int64_dtype,
+    object_dtype,
 )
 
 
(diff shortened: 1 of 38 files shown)


zipline/lib/adjusted_array.py
```diff
@@ -82,19 +90,29 @@ def _normalize_array(data):
...
+        return data.astype(uint8), {'dtype': dtype(bool_)}
     elif data_dtype in FLOAT_DTYPES:
-        return data.astype(float64), dtype(float64)
+        return data.astype(float64), {'dtype': dtype(float64)}
     elif data_dtype in INT_DTYPES:
-        return data.astype(int64), dtype(int64)
-    elif data_dtype.name.startswith('datetime'):
+        return data.astype(int64), {'dtype': dtype(int64)}
+    elif data_dtype in CATEGORICAL_DTYPES:
+        if not isinstance(missing_value, (bytes, unicode)):
```
dmichalowicz: This will need to use `six.string_types`

ssanderson: I think we actually want this to continue to support `bytes` in py3. (`string_types` is just unicode in py3).

ssanderson: (Note that unicode here isn't the builtin name (since that's only defined in py2), it's a value imported from `zipline.utils.compat`.

zipline/lib/adjustment.pyx
```diff
@@ -530,9 +533,62 @@ cdef class Datetime64Overwrite(Datetime64Adjustment):
...
+    cdef:
+        readonly object value
+
+    def __init__(self,
+                 Py_ssize_t first_row,
+                 Py_ssize_t last_row,
+                 Py_ssize_t first_col,
+                 Py_ssize_t last_col,
+                 object value):
+        super(Adjustment, self).__init__(
```
dmichalowicz: Did you mean for this to be `super(_ObjectAdjustment, ...)`?

ssanderson: Yes.  I think we're being saved here by the fact that Cython implicitly calls all your parent's `__new__`s before yours.  Will update.

zipline/lib/labelarray.py
```diff
@@ -0,0 +1,482 @@
...
+    """
+    @preprocess(
+        values=coerce(list, partial(np.asarray, dtype=object)),
+    )
+    @expect_types(
+        values=np.ndarray,
+        categories=optional(list),
+    )
+    @expect_kinds(values=("O", "S", "U"))
+    def __new__(cls,
```
mtydykov: Should there be a docstring here describing the params?

zipline/lib/labelarray.py
```diff
@@ -0,0 +1,482 @@
...
+           In this case, obj will be None.  We treat this as an error case and
+           fail.
+
+        2. Someone (most likely our own __new__) calls
+           other_array.view(type=LabelArray).
+
+           In this case, `self` will be the new LabelArray instance, and
+           ``obj` will be the array on which ``view`` is being called.
+
+           The caller of ``obj.view`` is responsible for copying setting
```
mtydykov: wording

zipline/lib/labelarray.py
ssanderson: It probably belongs in the class docstring, but yeah, I can add that.

zipline/lib/labelarray.py
```diff
@@ -0,0 +1,482 @@
...
+
+        Parameters
+        ----------
+        suffix : str
+
+        Returns
+        -------
+        matches : np.ndarray[bool]
+            An array with the same shape as self indicating whether each
+            element of self ended with ``suffix``.w
```
mtydykov: cruft

zipline/lib/labelarray.py
```diff
@@ -0,0 +1,482 @@
...
+        if self.dtype != int64_dtype:
+            return "Invalid LabelArray: dtype={}, shape={}".format(
+                self.dtype, self.shape
+            )
+        repr_lines = repr(self.as_string_array()).splitlines()
+        repr_lines[0] = repr_lines[0].replace('array(', 'LabelArray(', 1)
+        repr_lines[-1] = repr_lines[-1].rsplit(',', 1)[0] + ')'
+        # The extra spaces here account for the difference in length between
+        # 'array(' and 'LabelArray('.
+        return '\n     '.join(repr_lines)
```
mtydykov: Should these extra spaces be parameterized/moved into a variable somewhere so that it's clearer how we expect this to look?

zipline/data/us_equity_loader.py
```diff
@@ -250,7 +251,7 @@ def _ensure_sliding_windows(self, assets, dts, field):
                     adjs = {}
                 window = Float64Window(
                     array[:, i].reshape(prefetch_len, 1),
-                    dtype_,
+                    view_kwargs,
```
mtydykov: What does `view_kwargs` do differently from `dtype_`?

ssanderson: The relevant corresponding change is https://github.com/quantopian/zipline/pull/1174/files#diff-9c48be544e3a72c4ea00f841fe7f9b7bL96.

Previously, the low-level array iterator would do:

```
self._data.view(self.viewtype)
```
it now instead does:
```
self._date.view(**self.view_kwargs)
```

The difference is that we used to always pass a dtype to `view`, but we now pass an empty dict (which is a no-op).  The alternative would have been to pass `None` or another sentinel and have `AdjustedArrayWindow` check its viewtype on every iteration, but it makes the downstream logic cleaner to just always receive a dictionary.

zipline/lib/labelarray.py
ssanderson: This would look something like:
```
separator = '\n' + ' ' * (len(type(self).__name__) - len('array'))
return separator.join(repr_lines)
```
I'm not sure that's actually much clearer (I would want the comment explaining what's happening either way), so I'd be inclined to just leave this as-is with the comment.

zipline/pipeline/engine.py
```diff
@@ -412,8 +417,18 @@ def _to_narrow(self, data, mask, dates, assets):
         resolved_assets = array(self._finder.retrieve_all(assets))
         dates_kept = repeat_last_axis(dates.values, len(assets))[mask]
         assets_kept = repeat_first_axis(resolved_assets, len(dates))[mask]
+
+        final_columns = {}
+        for name in data:
+            # Each term that computed an output has its postprocess method
+            # called on the filtered result.
+            #
+            # We use this convert LabelArrays into categoricals, among other
```
mtydykov: to convert?

zipline/lib/labelarray.py
ssanderson: Fixed.

zipline/lib/labelarray.py
ssanderson: Fixed.

zipline/lib/labelarray.py
ssanderson: Updated.

zipline/lib/adjustment.pyx
ssanderson: Updated.

zipline/pipeline/engine.py
ssanderson: Updated.

tests/pipeline/test_adjusted_array.py
```diff
@@ -62,12 +67,41 @@ def valid_window_lengths(underlying_buffer_length):
...
+
+@curry
+def as_labelarray(initial_dtype, missing_value, array):
+    """
+    Curried wrapper around LabelArray, that round-trips the input data through
+    `initial_dtype` first.
+    """
+    return LabelArray(
+        array.astype(initial_dtype),
+        missing_value=initial_dtype.type(''),
```
mtydykov: Why is this an empty string?

ssanderson: this should be `missing_value`

tests/pipeline/test_adjusted_array.py
```diff
@@ -156,84 +190,125 @@ def _gen_multiplicative_adjustment_cases(dtype):
...
+                                    dtype,
+                                    missing_value):
     """
     Generate test cases for overwrite adjustments.
 
     The algorithm used here is the same as the one used above for
     multiplicative adjustments.  The only difference is the semantics of how
     the adjustments are expected to modify the arrays.
+
+    This is parameterized on `make_input` and make_expected_output functions,
```
mtydykov: make_expected_output -> `make_expected_output`

zipline/lib/adjusted_array.py
```diff
@@ -82,19 +90,29 @@ def _normalize_array(data):
...
-        return data.astype(uint8), dtype(bool_)
+        return data.astype(uint8), {'dtype': dtype(bool_)}
     elif data_dtype in FLOAT_DTYPES:
-        return data.astype(float64), dtype(float64)
+        return data.astype(float64), {'dtype': dtype(float64)}
     elif data_dtype in INT_DTYPES:
-        return data.astype(int64), dtype(int64)
-    elif data_dtype.name.startswith('datetime'):
+        return data.astype(int64), {'dtype': dtype(int64)}
+    elif data_dtype in CATEGORICAL_DTYPES:
```
dmichalowicz: Why doesn't `CATEGORICAL_DTYPES` include 'S' and 'U'? I was under the impression that `LabelArray` can accept data of strings/unicode, not just dtype `object`.

ssanderson: `S` and `U` aren't actually dtypes; they're families of dtypes, one for each possible fixed length.  If you make an array with `dtype('S')`, it's actually equivalent to `dtype('S1')`, which has some unexpected behavior like silently truncating any string longer than one character.  I actually had 'S' and 'U' here in an earlier draft and removed them because I ended up having to guard defensively against the dtype being 'S' or 'U' all over the place.

mtydykov:
Looks good to me (whenever the build passes).

zipline/pipeline/mixins.py
```diff
@@ -103,27 +103,45 @@ def compute(self, today, assets, out, *arrays):
...
+        Allocate an output array whose rows should be passed to `self.compute`.
+
+        The resulting array must have a shape of ``shape``.
+
+        If we have standard outputs (i.e. self.outputs is NotSpecified), the
+        default is an empty ndarray whose dtype is ``self.dtype``.
+
+        If we have an outputs tuple, the default is an empty recarray with
+        ``self.outputs`` as field names. Each field will have dtype
+        ``self.dtype``, the default shape is ``self.shape``.
```
dmichalowicz: ``self.shape`` --> ``shape``

zipline/pipeline/data/dataset.py
```diff
@@ -164,15 +159,25 @@ def qualname(self):
...
-        if self.dtype == bool_dtype:
-            from zipline.pipeline.filters import Latest
-        elif self.dtype == int64_dtype:
-            from zipline.pipeline.classifiers import Latest
+        from zipline.pipeline.factors import Factor, Latest as LatestFactor
+        from zipline.pipeline.filters import Filter, Latest as LatestFilter
+        from zipline.pipeline.classifiers import (
+            Classifier,
+            Latest as LatestClassifier,
+        )
```
dmichalowicz: Could you maybe add a brief comment as to why these imports need to happen here?

ssanderson: I think these actually don't need to be lazy imports anymore (I don't remember why they were originally...).  Will move them out.

PR #1176, opened 2016-05-03
BUG: DAY_END action not emitted during minute emission

lianga888:
Refactor AlgorithmSimulator so that DAY_END is emitted for both
minute and daily emission, and that handling of end-of-minute
and end-of-day is separated

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -303,9 +303,7 @@ class PerformanceTracker(object):
 
     def handle_minute_close(self, dt, data_portal):
         """
-        Handles the close of the given minute. This includes handling
-        market-close functions if the given minute is the end of the market
-        day.
+        Handles the close of the given minute in minute emission.
 
         Parameters
         __________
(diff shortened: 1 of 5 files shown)


zipline/finance/performance/tracker.py
```diff
@@ -333,40 +329,37 @@ def handle_minute_close(self, dt, data_portal):
...
-
-        benchmark_value = self.all_benchmark_returns[completed_date]
 
-        self.cumulative_risk_metrics.update(
-            completed_date,
-            self.todays_performance.returns,
-            benchmark_value,
-            account.leverage)
+        if self.emission_rate == 'daily':
+            self.position_tracker.sync_last_sale_prices(dt, False, data_portal)
```
jbredeche: why is this block only done in daily emission?

lianga888: the method that was previously `handle_market_close_daily` and contains this block was only called in daily mode. I suspect running this block would not effectively update anything, as that would have been done in the latest `handle_minute_close`. Additionally, I believe lines like `benchmark_value = self.all_benchmark_returns[completed_date]` would crash if called in minute emission, since the `completed_date` is normalized and the benchmark would not have midnight

jbredeche: ok.  can you add a comment explaining that?

PR #1177, opened 2016-05-03
PERF: Cap memory usage by minute bar carrays.

ehebert:
Instead of letting the cache of carrays grow unbounded, use an LRUCache
to cap the number of equities for any given column.

Tested with the size 1000, on an algo that was using pipeline which was
using over 3000, runtimes were similar, but the memory usage was
successfully capped to around 1.2GB.

Also, tested with an algorithm which bought and hold just one equity and
no major slow down was seen when using the LRUCache vs. a dictionary.

We may want to follow this up with an extension to `carray` which is not
as memory hungry per column; e.g. by not loading repeated/similar
metadata or releasing the last read chunk after a certain amount of
time.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -16,6 +16,7 @@ import os
 from os.path import join
 from textwrap import dedent
 
+from cachetools import LRUCache
 import bcolz
 from bcolz import ctable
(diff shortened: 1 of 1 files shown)


jbredeche:
LGTM!

PR #1180, opened 2016-05-05
MAINT: Add validation for schedule_function params

lianga888:
Return a TypeError if we passes a date_rule for a time_rule and
vice-versa

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -1836,6 +1836,50 @@ def handle_data(context, data):
         )
         algo.run(self.data_portal)
 
+    def test_schedule_function_validation(self):
+        """
+        Test that we raise a TypeError when we input a date_rule for a
+        time_rule and vice-versa
+        """
+
+        sim_params = factory.create_simulation_parameters(
+            start=pd.Timestamp('2006-01-14', tz='UTC'),
+            end=pd.Timestamp('2006-01-15', tz='UTC'),
+            data_frequency='minute'
+        )
+
+        algo = TradingAlgorithm(
+            script=noop_algo,
+            sim_params=sim_params,
+            env=self.env
+        )
+
+        def nothing(context, data):
+            pass
+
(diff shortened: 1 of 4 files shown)


llllllllll:
I think a much more simple change that could be made here would be to have the default `time_rule` for `schedule_function` be `Always()`. In the old implementation, `market_open(1)` was "is the current time at least 1 minute after the market has opened" which was just always true. Making this change will preserve the old behavior.

lianga888:
I thought `schedule_function` was designed so that the function runs at most once a day. By have a `time_rule` of `Always()`, this is no longer the case

llllllllll:
It will still be the case because we construct the actual rule that gets registered as `OncePerDay(date_rule & time_rule)`

lianga888:
so are you saying we should be supporting a schedule function where the user is trying to use two time rules as in "when the time is exactly 3 hours after the market opens and exactly 1 minute after the market opens"?

llllllllll:
No, we cannot support that because we have changed rules to be exactly equal and not greater than or equal, which is more correct with the new scheduling stuff in zipline. The only reason that is the rule that is generated is that the default for time rule is `market_open(minutes=1)`. If we make that `Always` then the rule would be: " hours after the market open and always" which reduces to "3 hours after the market open"

lianga888:
are you talking about a case where is the user is trying to do `schedule_function(func, time_rules.market_open(hours=3))`? I would argue that this is not even valid, since you're putting a time_rule where a date_rule should be

llllllllll:
My point is that there is no actual difference between a time and date rule, the names are picked to reflect a common use case. Notice that many rules have meaning in both contexts.

jbredeche:
unassigning myself here, @lianga888 and @llllllllll can figure out what needs to be done.

lianga888:
closing in favor on https://github.com/quantopian/zipline/pull/1221

PR #1185, opened 2016-05-05
Redesign buyback auth

mtydykov:
Replaces Cash/Share buyback auth custom datasets with a single BuybackAuthorizations dataset and modifies tests accordingly.


kglowinski:
Looks like quite a few of the tests in travis are breaking, in addition to flake8.

tests/pipeline/test_13d_filings.py
```diff
@@ -24,9 +24,11 @@
 from zipline.testing.fixtures import WithPipelineEventDataLoader
 from zipline.testing.fixtures import ZiplineTestCase
 
-date_intervals = [[None, '2014-01-04'],
-                  ['2014-01-05', '2014-01-09'],
-                  ['2014-01-10', None]]
+date_intervals = [
+    [['2014-01-01', '2014-01-04'],
```
kglowinski: Do we no longer need the `None` interval boundaries?

tests/pipeline/test_buyback_auth.py
```diff
@@ -48,34 +39,39 @@
...
+        BUYBACK_TYPE_FIELD_NAME: ["New", "Additional"],
         TS_FIELD_NAME: pd.to_datetime(['2014-01-05', '2014-01-10']),
         BUYBACK_ANNOUNCEMENT_FIELD_NAME: pd.to_datetime(['2014-01-04',
                                                          '2014-01-09'])
     }),
     pd.DataFrame(
-        columns=[SHARE_COUNT_FIELD_NAME,
-                 CASH_FIELD_NAME,
+        columns=[VALUE_FIELD_NAME,
+                 VALUE_TYPE_FIELD_NAME,
```
kglowinski: What's the difference between the `VALUE_TYPE` and the `BUYBACK_TYPE`?

zipline/pipeline/common.py
```diff
@@ -8,6 +8,8 @@
 CASH_AMOUNT_FIELD_NAME = 'cash_amount'
 COUNT_FIELD_NAME = 'count'
 BUYBACK_ANNOUNCEMENT_FIELD_NAME = 'buyback_date'
+BUYBACK_TYPE_FIELD_NAME = 'buyback_type'
+CASH_AMOUNT_FIELD_NAME = 'cash_amount'
```
kglowinski: `CASH_AMOUNT_FIELD_NAME` doesn't appear to be getting used anywhere. Leftover?

zipline/pipeline/loaders/utils.py
```diff
@@ -116,9 +116,11 @@ def previous_event_frame(events_by_sid,
     next_date_frame
     """
     sids = list(events_by_sid)
+    populate_value = None if field_dtype == categorical_dtype else \
```
kglowinski: Can you explain what `categorical_dtype` is doing here?

zipline/testing/fixtures.py
```diff
@@ -887,7 +887,9 @@ def get_sids_to_frames(self,
                            zip_date_index_with_vals,
                            vals,
                            date_intervals,
-                           dates):
+                           dates,
+                           dtype_name,
+                           missing_dtype):
```
kglowinski: What does the `missing_dtype` do?

zipline/pipeline/loaders/utils.py
mtydykov: Basically, we were using the `missing_value` of columns before in order to populate the array with default values; we can't do this with categoricals because the current design is such that they don't have a default missing value, so we have to explicitly check if the field is a categorical, and if so, here we can just use `None` as the missing value.

zipline/pipeline/loaders/utils.py
```diff
@@ -140,6 +142,8 @@ def previous_event_frame(events_by_sid,
 
     frame = pd.DataFrame(out, index=date_index, columns=sids)
     frame.ffill(inplace=True)
+    if field_dtype == categorical_dtype:
+        frame[frame.isnull()] = missing_value
```
ssanderson: You probably want `frame.fillna` here.

tests/pipeline/test_buyback_auth.py
mtydykov: `VALUE_TYPE` is the unit - e.g., shares, USD, CAD, etc. I think I was trying to make it generic enough to use for other loaders, but it's probably not clear what it is; I can rename it to `VALUE_UNIT` or something like that.

`BUYBACK_TYPE` is just the type of buyback (that's the actual name of the field from EventVestor) - e.g., New, Additional, Reduction, etc.

tests/pipeline/test_13d_filings.py
mtydykov: I changed the way `get_values_for_date_ranges` works, so now we need concrete start/end dates. I can infer that logic within `get_values_for_date_ranges` itself (e.g., assume that `None` at the start refers to the start of the month, and `None` at the end refers to the end); I felt it was clearer to just explicitly do this where the date intervals are defined.

zipline/pipeline/common.py
mtydykov: Whoops - I put it there twice by accident (removed). It's used for dividends, though.

zipline/testing/fixtures.py
mtydykov: Added a comment that explains this.

tests/pipeline/test_buyback_auth.py
mtydykov: Done - renamed these for clarity.

tests/pipeline/test_dividends.py
```diff
@@ -206,10 +206,11 @@ def setup(self, dates):
         amounts = [['NaN', 1, 15], ['NaN', 7, 13], ['NaN', 3, 1], ['NaN', 23]]
         cols = {
             PREVIOUS_ANNOUNCEMENT: self.get_sids_to_frames(
-                zip_with_dates, announcement_dates, date_intervals, dates
+                zip_with_dates, announcement_dates, date_intervals, dates,
+                'datetime64[ns]', 'NaN'
```
kglowinski: Remind me: the dividend changes in here are because this PR is also the one that incorporates string support?

mtydykov: No - it's because this PR included some modifications to general test functions that also affected dividend tests.

zipline/pipeline/factors/events.py
```diff
@@ -132,7 +131,7 @@ class BusinessDaysSincePreviousEarnings(BusinessDaysSincePreviousEvents):
     inputs = [EarningsCalendar.previous_announcement]
 
 
-class BusinessDaysSinceCashBuybackAuth(
+class BusinessDaysSinceBuybackAuth(
     BusinessDaysSincePreviousEvents
 ):
     """
```
kglowinski: Remove `cash` from `most recent cash buyback authorization`.

kglowinski:
Assuming tests above pass: I just had a tiny comment and a question, then scott's remark about `fillna`. Otherwise, looks good to me. 👍

zipline/pipeline/loaders/utils.py
mtydykov: That won't work because `missing_value` is sometimes `None`.

tests/pipeline/test_dividends.py
kglowinski: Ah, gotcha.

zipline/pipeline/factors/events.py
mtydykov: Done.

PR #1186, opened 2016-05-05
Remove type restrictions for generic loader

mtydykov:
Remove filtering for float columns in the blaze core loader - allow any supported dtypes.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -270,19 +270,190 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
             NonPipelineField,
         )
 
-    def test_cols_with_missing_vals(self):
-        dates = (self.dates[0], self.dates[-1])
+    def test_cols_with_all_missing_vals(self):
+        """
+        Tests that when there is no known data, we get output where the
+        columns have the right dtypes and the right missing values filled in.
+
+        input (self.df):
+        Empty DataFrame
+        Columns: [sid, float_value, str_value, int_value, bool_value, dt_value,
+            asof_date, timestamp]
+        Index: []
+
+        output (expected)
+                                          str_value  float_value  int_value
+        2014-01-01 Equity(65 [A])      None          NaN          0
+                   Equity(66 [B])      None          NaN          0
+                   Equity(67 [C])      None          NaN          0
+        2014-01-02 Equity(65 [A])      None          NaN          0
+                   Equity(66 [B])      None          NaN          0
+                   Equity(67 [C])      None          NaN          0
+        2014-01-03 Equity(65 [A])      None          NaN          0
+                   Equity(66 [B])      None          NaN          0
+                   Equity(67 [C])      None          NaN          0
+
+                                  dt_value  bool_value
+        2014-01-01 Equity(65 [A])      NaT  False
+                   Equity(66 [B])      NaT  False
+                   Equity(67 [C])      NaT  False
+        2014-01-02 Equity(65 [A])      NaT  False
+                   Equity(66 [B])      NaT  False
+                   Equity(67 [C])      NaT  False
+        2014-01-03 Equity(65 [A])      NaT  False
+                   Equity(66 [B])      NaT  False
+                   Equity(67 [C])      NaT  False
+        """
+        df = pd.DataFrame(columns=['sid', 'float_value', 'str_value',
+                                   'int_value', 'bool_value', 'dt_value',
+                                   'asof_date', 'timestamp'])
+
+        expr = bz.data(
+            df,
+            dshape="""
+            var * {
+                 sid: int64,
+                 float_value: float64,
+                 str_value: string,
+                 int_value: int64,
+                 bool_value: bool,
+                 dt_value: datetime,
+                 asof_date: datetime,
+                 timestamp: datetime,
+            }""",
+        )
+        fields = OrderedDict(expr.dshape.measure.fields)
+
+        expected = pd.DataFrame({
+            "str_value": np.array([None,
+                                   None,
+                                   None,
+                                   None,
+                                   None,
+                                   None,
+                                   None,
+                                   None,
+                                   None],
+                                  dtype='object'),
+            "float_value": np.array([np.NaN,
+                                     np.NaN,
+                                     np.NaN,
+                                     np.NaN,
+                                     np.NaN,
+                                     np.NaN,
+                                     np.NaN,
+                                     np.NaN,
+                                     np.NaN],
+                                    dtype='float64'),
+            "int_value": np.array([0,
+                                   0,
+                                   0,
+                                   0,
+                                   0,
+                                   0,
+                                   0,
+                                   0,
+                                   0],
+                                  dtype='int64'),
+            "bool_value": np.array([False,
+                                    False,
+                                    False,
+                                    False,
+                                    False,
+                                    False,
+                                    False,
+                                    False,
+                                    False],
+                                   dtype='bool'),
+            "dt_value": [pd.NaT,
+                         pd.NaT,
+                         pd.NaT,
+                         pd.NaT,
+                         pd.NaT,
+                         pd.NaT,
+                         pd.NaT,
+                         pd.NaT,
+                         pd.NaT],
+        },
+            columns=['str_value', 'float_value', 'int_value', 'bool_value',
+                     'dt_value'],
+            index=pd.MultiIndex.from_product(
+                (self.dates, self.asset_finder.retrieve_all(
+                    self.ASSET_FINDER_EQUITY_SIDS
+                ))
+            )
+        )
+
+        self._test_id(
+            df,
+            var * Record(fields),
+            expected,
+            self.asset_finder,
+            ('float_value', 'str_value', 'int_value', 'bool_value',
+             'dt_value'),
+        )
+
+    def test_cols_with_some_missing_vals(self):
+        """
+        Tests the following:
+            1) Forward filling replaces missing values correctly for the data
+            types supported in pipeline.
+            2) We don't forward fill when the missing value is the actual value
+             we got for a date in the case of int/bool columns.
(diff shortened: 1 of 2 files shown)


ssanderson:
Change looks fine.  Looks like some of the blaze tests need to be updated though.

tests/pipeline/test_blaze.py
```diff
@@ -271,32 +270,56 @@ def test_non_numpy_field(self):
...
-            ds.a
-        self.assertIsInstance(
-            object.__getattribute__(ds, 'a'),
-            NonPipelineField,
-        )
+        fields = OrderedDict(expr.dshape.measure.fields)
+
+        with tmp_asset_finder() as finder:
+            expected = pd.DataFrame(
+                np.array([["a", 0],
```
ssanderson: This array is playing with pandas type-inference fire a bit (we have `NoneType`, `str`, `int`, and `np.float64` all in the same array.  You probably want to pass this as something like:

```
pd.DataFrame({'str_value': np.array(..., dtype=object), 'value': np.array(..., dtype=float64)})
```

tests/pipeline/test_blaze.py
```diff
@@ -271,32 +270,56 @@ def test_non_numpy_field(self):
...
+                          [None, np.NaN],
+                          ["a", 0],
+                          [None, np.NaN],
+                          [None, np.NaN],
+                          ["a", 0],
+                          ["b", 1],
+                          [None, np.NaN]]),
+                columns=['str_value', 'value'],
+                index=pd.MultiIndex.from_product(
+                    (self.dates, finder.retrieve_all(self.sids))
```
ssanderson: What is enforcing that the tmp finder is aligned with the sids for this test case?  Should we just use `WithAssetFinder`?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -312,21 +311,17 @@ def new_dataset(expr, deltas, missing_values):
...
-            if promote(type_, float64, promote_option=False) != float64:
-                raise NotPipelineCompatible()
             if isinstance(type_, Option):
                 type_ = type_.ty
-        except NotPipelineCompatible:
-            col = NonPipelineField(name, type_)
-        except TypeError:
-            col = NonNumpyField(name, type_)
-        else:
+            type_ = type_.to_numpy_dtype()
```
ssanderson: If we're not catching `TypeError` anymore, then the weird goto here isn't necessary anymore.  This is just:

```
type_ = type_.to_numpy_dtype()
if can_represent_dtype(type_):
    col = Column(...)
else:
    col = NonPipelineField(...)
```

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1011,6 +1006,14 @@ def last_in_date_group(df, reindex, have_sids=have_sids):
                     )
                 else:
                     last_in_group = last_in_group.reindex(dates)
+            # Unstack will fill all missing values with NaN; we need to fix
```
ssanderson: This will do the wrong thing for int columns as well.

ssanderson: Also, is this filling bool/datetime columns with NaN?

tests/pipeline/test_blaze.py
mtydykov: Nothing, really. All the tests in this suite assume the sids are A, B, and C, which are the sids used in `tmp_asset_finder`. We would probably want to update all of them, then.

zipline/pipeline/loaders/blaze/core.py
mtydykov: How do we want to handle the case for ints?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1011,12 +1011,47 @@ def last_in_date_group(df, reindex, have_sids=have_sids):
...
+                        raise ValueError("Must specify a missing value for "
+                                         "column %d." % column.name)
+                    # Need special logic in case we're replacing with Nones
+                    elif df[column.name].dtype == categorical_dtype:
+                        last_in_group[column.name] = last_in_group[
+                            column.name
+                        ].where(pd.notnull(last_in_group[column.name]),
+                                column.missing_value)
+                    # Need to convert from float col to datetime col
+                    elif df[column.name].dtype == datetime64ns_dtype:
```
ssanderson: Why are we writing this twice?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1011,12 +1011,47 @@ def last_in_date_group(df, reindex, have_sids=have_sids):
...
             return last_in_group
 
         sparse_deltas = last_in_date_group(non_novel_deltas, reindex=False)
         dense_output = last_in_date_group(sparse_output, reindex=True)
-        dense_output.ffill(inplace=True)
+        for column in columns:
+            if have_sids:
+                dense_output[column.name] = dense_output[
+                    column.name
+                ].apply(lambda x: x.replace(
```
ssanderson: Why can't we call ``ffill`` or ``replace`` on the sub-frame itself here?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1011,12 +1011,47 @@ def last_in_date_group(df, reindex, have_sids=have_sids):
                     )
                 else:
                     last_in_group = last_in_group.reindex(dates)
+            # Unstack will fill all missing values with NaN; we need to fix
+            # this for all types that are not float.
+            if not df.empty:
+                for column in columns:
+                    if isinstance(column.missing_value, NotSpecifiedType):
```
ssanderson: I don't think this can ever happen.  By the time we have a loadable column, we should always have a concrete missing value.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -311,22 +317,16 @@ def new_dataset(expr, deltas, missing_values):
...
-            if promote(type_, float64, promote_option=False) != float64:
-                raise NotPipelineCompatible()
-            if isinstance(type_, Option):
-                type_ = type_.ty
-        except NotPipelineCompatible:
-            col = NonPipelineField(name, type_)
-        except TypeError:
-            col = NonNumpyField(name, type_)
-        else:
+        if isinstance(type_, Option):
```
ssanderson: We could probably move the unwrapping of `Option` into `datashape_type_to_numpy`.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -276,6 +275,13 @@ def __str__(self):
 _new_names = ('BlazeDataSet_%d' % n for n in count())
 
 
+def datashape_type_to_numpy(type_):
```
ssanderson: This could use a docstring.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -169,16 +167,17 @@
...
+from zipline.lib.adjusted_array import AdjustedArray, can_represent_dtype
 from zipline.lib.adjustment import Float64Overwrite
 from zipline.utils.enum import enum
 from zipline.utils.input_validation import (
     expect_element,
     ensure_timezone,
     optionally,
 )
-from zipline.utils.numpy_utils import repeat_last_axis
+from zipline.utils.numpy_utils import categorical_dtype, repeat_last_axis, \
```
ssanderson: Style nit: we generally prefer wrapping imports with parens over line breaks elsewhere in zipline.

zipline/pipeline/loaders/blaze/core.py
mtydykov: What do you mean when you say that we're writing it twice? It's a separate condition.

ssanderson: The next two lines here are:

```
last_in_group[column.name] = last_in_group[
 +                            column.name
 +                        ].astype('datetime64[ns]')
 +                        last_in_group[column.name] = last_in_group[
 +                            column.name
 +                        ].fillna(column.missing_value)
```
We're pulling data out of the frame, changing its type, and writing it back into the frame.
We're then pulling the data we just wrote back **out** of the frame, ffilling it, and writing it in again.

So, two questions:

1. Why is the `astype` only necessary in the datetime64 case?
2. Can we avoid writing twice here (e.g. by doing `last_in_group[...] = `last_in_group[...].astype(...).fillna(...)`)?

ssanderson: (To be clear, my original comment was referring to the content of the `datetime64` block, not to the fact that there are separate blocks for each dtype.)

mtydykov: Yeah, good catch, I think that was there twice by accident. I also don't think I need that `fillna` there. 

I was doing it in the datetime64 case because it's the only way to fill in `NaT` for null values. We could do it for the others, but it didn't seem necessary since the actual output of this method is already doing that conversion for each column/column dtype.

zipline/pipeline/loaders/blaze/core.py
mtydykov: We can't call `ffill` because it wouldn't work for something like `bool`. How would `replace` work for this?

zipline/pipeline/loaders/blaze/core.py
mtydykov: Done.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1016,7 +1035,25 @@ def last_in_date_group(df, reindex, have_sids=have_sids):
 
         sparse_deltas = last_in_date_group(non_novel_deltas, reindex=False)
         dense_output = last_in_date_group(sparse_output, reindex=True)
-        dense_output.ffill(inplace=True)
+        dense_output = dense_output.ffill()
```
ssanderson: Is there a reason we changed this to no longer be inplace?

ssanderson:
LGTM

PR #1187, opened 2016-05-06
BUG: Fix multiple bugs in PanelDailyBarReader.

ssanderson:
- Return a value from `verify_all_indices_unique` so that `panel` isn't
  unconditionally `None` in `PanelDailyBarReader`.

- Fix a bug where we always set the volume of every asset to `1e9`.

- Add minimal suite of tests for get_spot_value, which catch both of the
  above.

NOTE: There are still several issues with `PanelDailyBarReader`.  The
docstring for `get_spot_value` claims that it will return -1 on days
where an asset didn't trade, which isn't the case.  It also claims that
it will raise `NoDataOnDate` when a request is made outside the panel
range, but it just raises a KeyError.  We also still have no coverage
for `load_raw_arrays`, so it's likely that there are more bugs lurking.

### DIFF
--- a/tests/test_panel_daily_bar_reader.py
+++ b/tests/test_panel_daily_bar_reader.py
@@ -13,16 +13,60 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from itertools import permutations
+from itertools import permutations, product
 
+import numpy as np
 import pandas as pd
 
 from zipline.data.us_equity_pricing import PanelDailyBarReader
 from zipline.testing import ExplodingObject
-from zipline.testing.fixtures import ZiplineTestCase
+from zipline.testing.fixtures import (
+    WithAssetFinder,
+    WithNYSETradingDays,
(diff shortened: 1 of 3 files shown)


tests/test_panel_daily_bar_reader.py
```diff
@@ -13,16 +13,60 @@
...
+            minor_axis=minor_axis,
+        )
+
+        cls.reader = PanelDailyBarReader(days, cls.panel)
+
+    def test_spot_price(self):
+        panel = self.panel
+        reader = self.reader
+
+        for asset, date, field in product(*panel.axes):
```
ehebert: :+1: on the coverage of all values.

ehebert:
LGTM, thanks for fixing these. And I agree we should get full coverage, the initial implementation had tunnel vision on 'just get the examples directory to pass'; so a more thorough examination of the class is needed.

PR #1188, opened 2016-05-06
api documentation

llllllllll:
Update the documentation for the `zipline.api` namespace as well as some supporting objects.

### DIFF
--- a/docs/source/release-process.rst
+++ b/docs/source/release-process.rst
@@ -161,6 +161,13 @@ latest master and run:
 This will build the documentation, checkout a fresh copy of the ``gh-pages``
 git branch, and copy the built docs into the zipline root.
 
+.. note::
+
+   The docs should always be built with **Python 3**. Many of our api functions
+   are wrapped by preprocessing functions which accept \*args and \**kwargs. In
+   Python 3, sphinx will respect the ``__wrapped__`` attribute and display the
+   correct arguments.
+
 Now, using our browser of choice, view the ``index.html`` page and verify that
 the docs look correct.
 
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -788,6 +788,8 @@ class TradingAlgorithm(object):
                   The platform that the code is running on. By default this
                   will be the string 'zipline'. This can allow algorithms to
                   know if they are running on the Quantopian platform instead.
+              * : dict[str -> any]
+                  Returns all of the fields in a dictionary.
 
         Returns
         -------
@@ -796,7 +798,7 @@ class TradingAlgorithm(object):
 
         Raises
         ------
-        KeyError
+        ValueError
             Raised when ``field`` is not a valid option.
         """
         env = {
@@ -810,7 +812,12 @@ class TradingAlgorithm(object):
         if field == '*':
             return env
         else:
-            return env[field]
+            try:
+                return env[field]
+            except KeyError:
+                raise ValueError(
+                    '%r is not a valid field for get_environment' % field,
+                )
 
     @api_method
     def fetch_csv(self,
@@ -825,7 +832,8 @@ class TradingAlgorithm(object):
                   symbol_column=None,
                   special_params_checker=None,
                   **kwargs):
-        """Fetch a csv from a remote url.
+        """Fetch a csv from a remote url and register the data so that it is
+        queryable from the ``data`` object.
 
         Parameters
         ----------
@@ -912,7 +920,7 @@ class TradingAlgorithm(object):
                           date_rule=None,
                           time_rule=None,
                           half_days=True):
-        """Schedules a function to be called with some timed rules.
+        """Schedules a function to be called according to some timed rules.
 
         Parameters
         ----------
@@ -1065,7 +1073,7 @@ class TradingAlgorithm(object):
         Raises
         ------
         SidsNotFound
-            When a requested sid is not found and default_none=False.
+            When a requested ``sid`` does not map to any asset.
         """
         return self.asset_finder.retrieve_asset(sid)
 
@@ -1213,8 +1221,9 @@ class TradingAlgorithm(object):
         asset : Asset
             The asset that this order is for.
         amount : int
-            The amount of shares to order. If this is negative, this is the
-            number of shares to sell or short.
+            The amount of shares to order. If ``amount`` is positive, this is
+            the number of shares to buy or cover. If ``amount`` is negative,
+            this is the number of shares to sell or short.
         style : ExecutionStyle, optional
             The execution style for the order.
 
@@ -1759,13 +1768,16 @@ class TradingAlgorithm(object):
         Parameters
         ----------
         asset : Asset
-            If passed, return only the open orders for the given asset instead
-            of all open orders.
+            If passed and not None, return only the open orders for the given
+            asset instead of all open orders.
 
         Returns
         -------
-        open_orders : list[Order]
-            The open orders.
+        open_orders : dict[list[Order]] or list[Order]
+            If no asset is passed this will return a dict mapping Assets
+            to a list containing all the open orders for the asset.
+            If an asset is passed then this will return a list of the open
+            orders for this asset.
         """
         if asset is None:
             return {
@@ -1887,12 +1899,12 @@ class TradingAlgorithm(object):
                              self.trading_client.current_data)
 
     @api_method
-    def set_max_leverage(self, max_leverage=None):
+    def set_max_leverage(self, max_leverage):
         """Set a limit on the maximum leverage of the algorithm.
 
         Parameters
         ----------
-        max_leverage : float, optional
+        max_leverage : float
             The maximum leverage for the algorithm. If not provided there will
             be no maximum.
         """
@@ -1988,7 +2000,7 @@ class TradingAlgorithm(object):
 
         Parameters
         ----------
-        restricted_list : set[Asset]
+        restricted_list : container[Asset]
             The assets that cannot be ordered.
         """
         control = RestrictedListOrder(restricted_list)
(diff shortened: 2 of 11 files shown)


docs/source/release-process.rst
```diff
@@ -161,6 +161,13 @@ latest master and run:
 This will build the documentation, checkout a fresh copy of the ``gh-pages``
 git branch, and copy the built docs into the zipline root.
 
+.. note::
+
+   The docs should always be built with **Python 3**. Many of our api functions
+   are wrapped by preprocessing functions which accept \*args and \**kwargs. In
+   Python 3, sphinx will respect the ``__wrapped__`` attribute and display the
+   correct arguments.
```
richafrank: 👍

zipline/algorithm.py
```diff
@@ -764,6 +764,41 @@ def _create_daily_stats(self, perfs):
 
     @api_method
     def get_environment(self, field='platform'):
+        """Query the execution environment.
+
+        Parameters
+        ----------
+        field : {'platform', 'arena', 'data_frequency',
+                 'start', 'end', 'capital_base', 'platform', '*'}
```
richafrank: What does the star mean here?

zipline/algorithm.py
```diff
@@ -789,6 +825,47 @@ def fetch_csv(self, url,
                   symbol_column=None,
                   special_params_checker=None,
                   **kwargs):
+        """Fetch a csv from a remote url.
```
richafrank: Could we mention that it automatically registers the returned source with the data portal?

zipline/algorithm.py
```diff
@@ -829,11 +912,26 @@ def schedule_function(self,
                           date_rule=None,
                           time_rule=None,
                           half_days=True):
+        """Schedules a function to be called with some timed rules.
```
richafrank: What do you think of "Schedules a function to be called according to some timed rules."?

zipline/algorithm.py
```diff
@@ -845,8 +943,18 @@ def schedule_function(self,
 
     @api_method
     def record(self, *args, **kwargs):
-        """
-        Track and record local variable (i.e. attributes) each day.
+        """Track and record values each day.
```
richafrank: 👍 on the wording change

zipline/algorithm.py
```diff
@@ -860,18 +968,48 @@ def record(self, *args, **kwargs):
...
+
+        Parameters
+        ----------
+        benchmark : Asset
+            The asset to set as the new benchmark.
+
+        Notes
+        -----
+        Any dividends payed out for that new benchmark asset will be
+        automatically reinvested.
```
richafrank: 👍 good note

zipline/algorithm.py
```diff
@@ -860,18 +968,48 @@ def record(self, *args, **kwargs):
...
+        """
         if self.initialized:
             raise SetBenchmarkOutsideInitialize()
 
-        self.benchmark_sid = benchmark_sid
+        self.benchmark_sid = benchmark
```
richafrank: Would it be difficult to change the attribute name too?

zipline/algorithm.py
```diff
@@ -860,18 +968,48 @@ def record(self, *args, **kwargs):
...
+        """Lookup an Equity by its ticker symbol.
+
+        Parameters
+        ----------
+        symbol_str : str
+            The ticker symbol for the equity to lookup.
+
+        Returns
+        -------
+        equity : Equity
```
richafrank: Only Equities or any Asset?

zipline/algorithm.py
```diff
@@ -885,19 +1023,51 @@ def symbol(self, symbol_str):
...
+
+        Returns
+        -------
+        asset : Asset
+            The asset with the given ``sid``.
+
+        Raises
+        ------
+        SidsNotFound
+            When a requested sid is not found and default_none=False.
```
richafrank: What's `default_none`?

zipline/algorithm.py
```diff
@@ -1031,12 +1200,34 @@ def _can_order_asset(self, asset):
...
-        Place an order using the specified parameters.
+        """Place an order.
+
+        Parameters
+        ----------
+        asset : Asset
+            The asset that this order is for.
+        amount : int
+            The amount of shares to order. If this is negative, this is the
+            number of shares to sell or short.
```
richafrank: I'm not an expert, but do you think we should say that a negative amount could mean to short, or rather that the position may be short?  Thinking about the asymmetry of positive and negative here, and the fact that I could buy 3 shares, but if I'm already short 6, then I'm still short.

zipline/algorithm.py
llllllllll: oops, I will add that. star means return everything in a dict.

zipline/algorithm.py
llllllllll: this only looks up equities, future_symbol looks up futures

zipline/algorithm.py
llllllllll: not sure. this was already here but I will remove it. I think that comment is taken from the asset finder.

zipline/algorithm.py
```diff
@@ -1289,14 +1570,53 @@ def order_percent(self, asset, percent,
...
+        example:
+
+        .. code-block:: python
+
+           order_target(sid(0), 10)
+           order_target(sid(0), 10)
+
+        This code will result in 20 shares of ``sid(0)`` because the first
+        call to ``order_target`` will not have been filled when the second
+        ``order_target`` call is made.
```
richafrank: 👍

zipline/algorithm.py
```diff
@@ -1362,6 +1754,19 @@ def order_target_percent(self, asset, target,
                         'get_open_orders. Use `asset` instead.')
     @api_method
     def get_open_orders(self, asset=None):
+        """Retrieve all of the current open orders.
+
+        Parameters
+        ----------
+        asset : Asset
+            If passed, return only the open orders for the given asset instead
```
richafrank: "If not None"

zipline/algorithm.py
```diff
@@ -1362,6 +1754,19 @@ def order_target_percent(self, asset, target,
...
+            of all open orders.
+
+        Returns
+        -------
+        open_orders : list[Order]
+            The open orders.
```
richafrank: Should we document here that in the default case, a dict is returned?

zipline/algorithm.py
```diff
@@ -1461,8 +1888,13 @@ def validate_account_controls(self):
...
     def set_max_leverage(self, max_leverage=None):
-        """
-        Set a limit on the maximum leverage of the algorithm.
+        """Set a limit on the maximum leverage of the algorithm.
+
+        Parameters
+        ----------
+        max_leverage : float, optional
+            The maximum leverage for the algorithm. If not provided there will
+            be no maximum.
```
richafrank: Seems like an error case to have no max when setting max leverage...

llllllllll: I agree, I don't think that we should have as many defaults as we do but I didn't want to change the semantics with this PR. I was only trying to document what exists.

richafrank: In fact, looks like a [ValueError will be raised] (https://github.com/quantopian/zipline/blob/16fd6681a66101d3f06a0bb639ee699a8905ceda/zipline/finance/controls.py#L381-L384) later.

zipline/algorithm.py
```diff
@@ -1519,25 +1971,33 @@ def set_max_order_size(self, asset=None, max_shares=None,
...
 
     @api_method
     def set_do_not_order_list(self, restricted_list):
-        """
-        Set a restriction on which assets can be ordered.
+        """Set a restriction on which assets can be ordered.
+
+        Parameters
+        ----------
+        restricted_list : set[Asset]
```
richafrank: Does it need to be a `set`, or just support `__contains__`?

zipline/finance/slippage.py
```diff
@@ -34,6 +34,8 @@ class LiquidityExceeded(Exception):
 
 
 class SlippageModel(with_metaclass(abc.ABCMeta)):
+    """Abstract interface for defining a new slippage model.
```
richafrank: I say we remove the word "new".

zipline/sources/requests_csv.py
```diff
@@ -138,8 +139,7 @@ def mask_requests_args(url, validating=False, params_checker=None, **kwargs):
     return request_pair(requests_kwargs, url)
 
 
-class PandasCSV(object):
-    __metaclass__ = ABCMeta
```
richafrank: Thinking about the failure mode here, I guess people could instantiate it on py3, nice.

zipline/algorithm.py
llllllllll: I would want to do that in a different change set since it would have downstream affects

zipline/algorithm.py
llllllllll: sorry, how do you want me to change the wording. I do mention that negative means short

zipline/algorithm.py
llllllllll: I think set connotes that it is an unordered container. I could change this to `container[Asset]`

zipline/sources/requests_csv.py
llllllllll: yeah, this is a common one to be py3 incompatible because in valid code it will not matter

zipline/algorithm.py
richafrank: Maybe also include "If this is positive, this is the number of shares to buy or cover."

zipline/algorithm.py
richafrank: Gotcha.  Ok either way.

PR #1190, opened 2016-05-09
PERF: takes ~14% off the time it takes to initialize a MinuteSimulationClock

jbredeche:
on my machine, ~350ms -> ~300ms

also, clarifies the public API for MinuteSimulationClock (now, only `__iter__` is exposed)

### DIFF
--- a/zipline/gens/sim_engine.pyx
+++ b/zipline/gens/sim_engine.pyx
@@ -30,7 +30,6 @@ cpdef enum:
 
 cdef class MinuteSimulationClock:
     cdef object trading_days
-    cdef object all_trading_days
     cdef bool minute_emission
     cdef np.int64_t[:] market_opens, market_closes
     cdef public dict minutes_by_day, minutes_to_day
@@ -39,19 +38,18 @@ cdef class MinuteSimulationClock:
                  trading_days,
                  market_opens,
                  market_closes,
-                 all_trading_days,
                  minute_emission=False):
         self.minute_emission = minute_emission
         self.market_opens = market_opens
         self.market_closes = market_closes
         self.trading_days = trading_days
-        self.all_trading_days = all_trading_days
         self.minutes_by_day = self.calc_minutes_by_day()
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
-    cpdef market_minutes(self, np.intp_t i):
+    cdef np.ndarray[np.int64_t, ndim=1] market_minutes(self, np.intp_t i):
         cdef np.int64_t[:] market_opens, market_closes
+
         market_opens = self.market_opens
         market_closes = self.market_closes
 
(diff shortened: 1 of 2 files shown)


zipline/gens/sim_engine.pyx
```diff
@@ -39,36 +38,39 @@ cdef class MinuteSimulationClock:
...
         self.market_opens = market_opens
         self.market_closes = market_closes
         self.trading_days = trading_days
-        self.all_trading_days = all_trading_days
         self.minutes_by_day = self.calc_minutes_by_day()
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
-    cpdef market_minutes(self, np.intp_t i):
+    cdef np.ndarray[np.int64_t, ndim=1] market_minutes(self, np.intp_t i):
```
ehebert: I believe we have been favoring returning MemoryViews, e.g. `cdef int64_t[:] market_minutes(...` instead.

Any opinions on a hard and fast rule @ssanderson ?

ehebert:
Just to make sure I am reading correctly, the core of the change is adding the return types?

One comment about the return type for the vector, but otherwise looks good to me.

zipline/gens/sim_engine.pyx
jbredeche: I actually tried that and got an error: 

```
/Users/jean/.virtualenvs/qexec/lib/python2.7/site-packages/pandas/tseries/tools.pyc in to_datetime(arg, errors, dayfirst, utc, box, format, exact, coerce, unit, infer_datetime_format)
    343         return _convert_listlike(arg, box, format)
    344 
--> 345     return _convert_listlike(np.array([ arg ]), box, format)[0]
    346 
    347 class DateParseError(ValueError):

/Users/jean/.virtualenvs/qexec/lib/python2.7/site-packages/pandas/tseries/tools.pyc in _convert_listlike(arg, box, format)
    331                 return DatetimeIndex._simple_new(values, None, tz=tz)
    332             except (ValueError, TypeError):
--> 333                 raise e
    334 
    335     if arg is None:

ValueError: Buffer has wrong number of dimensions (expected 1, got 2)
```

ehebert: Looks like that is because `pandas.tseries.tools.to_datetime` does not currently handle `MemoryViews`; the memory view fails the `is_list_like` condition and the fallback when no cases are met in `to_datetime` is try creating an array via `np.array([arg])` and passing that back to `convert_listlike`.

In that case, the `ndarray` should be fine.

ssanderson: @ehebert I don't have a strong feeling either way, though looking at that stack trace, it looks like Pandas is treating the memoryview you're returning as a scalar.  `to_datetime` is "smart" and will work on all manner of different data structures (arrays, lists, strings, ints, etc).  Since we know what we have (an array of ints), and what we want returned (a DatetimeIndex), I wonder if there's a more "direct" constructor we could use.

ssanderson: (One such option is `DatetimeIndex._simple_new`, though I don't know how public/stable that method is.)

ssanderson: Issue opened here: https://github.com/quantopian/zipline/pull/1190/files/fe5a2a888a498d838c3cc43de2c33ac08b20d2a7#r62506342

ssanderson: Woops.  Meant to link to https://github.com/pydata/pandas/issues/13120

PR #1191, opened 2016-05-09
BUG: fix inverted splits in quandl data

llllllllll:


ssanderson:
cc @paulperry, this is the cause of the issues we were seeing the other day.  (All the splits ratios in the Quandl dataset are inverted from what they should be.)

llllllllll:
assuming this looks good I will upload the corrections to the s3 bucket for use with the `quantopian-quandl` bundle so people don't need to re-download the quandl data

ssanderson:
LGTM, merge on pass

PR #1192, opened 2016-05-09
BUG: Set _assets_from_source as a list.

ssanderson:


zipline/algorithm.py
```diff
@@ -594,24 +595,18 @@ def run(self, data=None, overwrite_sim_params=True):
                 copy_panel.items = self._write_and_map_id_index_to_sids(
                     copy_panel.items, copy_panel.major_axis[0],
                 )
-                self._assets_from_source = \
-                    set(self.trading_environment.asset_finder.retrieve_all(
+                self._assets_from_source = (
+                    self.trading_environment.asset_finder.retrieve_all(
```
ehebert: Using a list instead of a set looks right, since the `retrieve_all` should be de-duped on `sid` already.

@jbredeche do you recall why it was a `set`? Looks like this was part of the Q1 - Q2 shimming, so want to make sure we aren't breaking anything for that compatibility.

jbredeche: I don't remember :(

PR #1193, opened 2016-05-10
DEV: pull remove-open-orders logic into its own method

jbredeche:
And test it.

### DIFF
--- a/zipline/finance/blotter.py
+++ b/zipline/finance/blotter.py
@@ -296,6 +296,9 @@ class Blotter(object):
             parameters.  If there are no commission events (because, for
             example, Zipline models the commission cost into the fill price
             of the transaction), then this is None.
+
+        closed_orders: List
+            closed_orders: list of all the orders that have filled.
         """
 
         closed_orders = []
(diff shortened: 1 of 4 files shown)


zipline/finance/blotter.py
```diff
@@ -329,20 +329,23 @@ def get_transactions(self, bar_data):
...
+
+    def prune_orders(self, closed_orders):
         # remove all closed orders from our open_orders dict
         for order in closed_orders:
             sid = order.sid
             try:
                 sid_orders = self.open_orders[sid]
                 sid_orders.remove(order)
-            except KeyError:
+            except ValueError:
```
richafrank: When do we expect this case?  Also, could we move the slice statement out of the try, so it's clearer that that isn't generating the error?

zipline/finance/blotter.py
```diff
@@ -329,20 +329,23 @@ def get_transactions(self, bar_data):
                     if not order.open:
                         closed_orders.append(order)
 
+        self.prune_orders(closed_orders)
```
richafrank: Since this was a refactoring to `get_transactions`, should we have test coverage for that method?

jbredeche: ideally, yes.  I don't have time right now.

zipline/finance/blotter.py
jbredeche: I'll add a comment, but we expect this in a downstream subclass that is connected to a broker, and the broker is canceling orders.  There's some scenario there that causes us to try to remove an order that we've already removed locally (probably due to EOD cancel policy on our own end), and thus is happening.

Either way, we're trying to remove something that is already removed, so we shouldn't blow up on it.

richafrank: > Either way, we're trying to remove something that is already removed, so we shouldn't blow up on it.

That's fair.  Just wondering if there's some bad state tracking elsewhere, in that we're trying to remove it at all, and whether that should be fixed instead, as opposed to fixing all those downstream of it.

zipline/finance/blotter.py
```diff
@@ -329,20 +332,29 @@ def get_transactions(self, bar_data):
...
+
+    def prune_orders(self, closed_orders):
+        """
+        Removes all given orders from the blotter's open_orders list.
+
+        Parameters
+        ----------
+
+        :param closed_orders:
+        :return:
```
richafrank: If we're not filling these out, let's just remove 'em.

richafrank:
LGTM!  Just mentioned the docstring.

Issue #604, opened 2015-06-15
Spereate commissions for equtites / futures?


npezolano:
Is it possible to set separate commissions for different asset classes?

warren-oneill:
As far as I'm aware no. The commission is set for an algorithm before a run with the [set_commission](https://github.com/quantopian/zipline/blob/master/zipline/algorithm.py#L910) function and is fixed for the entire run. It would make a great addition though!

npezolano:
@warren-oneill Is the idea that you can only backtest one asset class at a time or more than one? If both you would need this feature.

jfkirk:
This is a forthcoming feature. We are building toward allowing a single algorithm to trade multiple Asset classes, but you are absolutely right that this is a constraint in the current version. For now, you would have to either:

1. Use a commission rate somwhere between Equities and Futures (which is unrealistic)
2. Only trade on asset class per algo

jfkirk:
Self-assigning.

jgoode21:
Yeah I give a plus one for this, as I've written Zipline algorithms that trade equities and their options. The options have different commissions that their underlying equities. I just deducted the difference from my ending PnL, but it wasn't the cleanest work around.

jfkirk:
@jgoode21 Makes sense - you shouldn't have to do such a workaround, so we'll work toward getting you this feature.

PR #1194, opened 2016-05-10
New tiered commission model for options

jimgoo:
I've always wanted to have different commissions for stocks and options traded through Zipline, so I added another commission model that supports this. The equity commissions in the model work the same as the [existing PerShare model](https://github.com/quantopian/zipline/blob/master/zipline/finance/commission.py#L20), where you specify a per-share commission and a minimum commission for each trade.

The difference here is that you can also specify a per-contract commission and minimum per-trade commission for option trades. The commission scheme to apply to a given trade depends on the `sid` of the asset. The user supplies a list of sids to apply equity commissions to and another list of sids to apply option commissions to. 

The per-contract option commissions can be tiered like at Interactive Brokers. The `PerShareWithOptions.ib_default_tier`  dictionary corresponds to the [commissions listed here](https://gdcdyn.interactivebrokers.com/en/index.php?f=commission&p=options1) for trading less than 10K contracts per month:

|                                 | Commissions           | Minimum Per Order |
| ------------------------------- | --------------------- | -------- |            
| Premium >= USD 0.10             | USD 0.70 per contract | USD 1.00 |
| USD 0.05 <= Premium < USD 0.10  | USD 0.50 per contract | USD 1.00 |
| Premium < USD 0.05              | USD 0.25 per contract | USD 1.00 |

This is specified in the dictionary as:

```python
tier_dict = {}
tier_dict[(0.10, np.inf)] = 0.70
tier_dict[(0.05, 0.10)] = 0.50
tier_dict[(-np.inf, 0.05)] = 0.25
```

I think this solves issue https://github.com/quantopian/zipline/issues/604 provided that you don't have a problem specifying the sids of your stocks and options/futures up front.

### DIFF
--- a/zipline/finance/commission.py
+++ b/zipline/finance/commission.py
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import bisect
+import numpy as np
+
 DEFAULT_PER_SHARE_COST = 0.0075         # 0.75 cents per share
 DEFAULT_MINIMUM_COST_PER_TRADE = 1.0    # $1 per trade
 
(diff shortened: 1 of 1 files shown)


zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+    share cost with an optional minimum cost per trade.
+    Applies different set of fees to equities and underliers.
+    """
+
+    def __init__(self, equity_cost=DEFAULT_PER_SHARE_COST,
+                 equity_min_trade_cost=DEFAULT_MINIMUM_COST_PER_TRADE,
+                 option_costs=None,
+                 option_min_trade_cost=DEFAULT_MINIMUM_COST_PER_TRADE,
+                 equity_sids=None, option_sids=None):
+        """
```
llllllllll: for zipline we have started using numpy style docstrings. This would change this slightly to look like:

```
Parameters
----------
equity_cost : float, optional
    Cost of an equity trade per-share.
equity_min_trade_cost : float, optional
    The minimum trade cost for equities.
...
```

This should also be merged with the class docstring.
See below for some comments about the names of these parameters

A full guide can be found here: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+                                tier_dict[(-np.inf, 0.05)] = 0.25
+                          indicates that an option with a premium >= $0.10
+                          has a per-contract commission of $0.70. An option
+                          with a premium between $0.05 and $0.10 has a
+                          commission of $0.50 per-contract, and an option
+                          with a premium less than $0.05 has a per-contract
+                          commission of $0.25.
+            option_min_trade_cost: minimum trade cost for options
+            equity_sids: list of integer sids to apply equity commissions
+            option_sids: list of integer sids to apply option commissions
```
llllllllll: Ideally we could just use the type of the `Asset` object to know which type of commission to apply. Right now we only have `Future` and `Equity`. I think it might not be too hard to just add this class for use here.

Another option would to accept only `tiered_assets` as a set of a `Asset` objects to apply the tiered pricing to, maybe with some sentinel that says, "all of them". We might want to say `tiered_assets` instead of ``option_sids` because we should represent if something is an option or not at the type level.

I will explain the choice of set over list below.

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+
+            https://gdcdyn.interactivebrokers.com/en/index.php?f=commission&p=options1
+        """
+        self.equity_cost = float(equity_cost)
+
+        self.equity_min_trade_cost = None if equity_min_trade_cost is None\
+            else float(equity_min_trade_cost)
+
+        if option_costs is not None:
+            assert isinstance(option_costs, dict)
```
llllllllll: No need to verify this, if the object respects the mapping protocol it should be accepted.

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+        self.option_costs = option_costs
```
llllllllll: If none is supplied then we should just make an empty dict here. This will save us a None check each time we need to use this object.

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+        self.option_min_trade_cost = None if option_min_trade_cost is None\
+            else float(option_min_trade_cost)
+
+        # check that all sids are integers
+        assert isinstance(equity_sids, list)
```
llllllllll: Same thing here, no need to validate the arguments to this extent.

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+        assert isinstance(option_sids, list)
+        for sid in equity_sids:
+            assert(isinstance(sid, int))
+        for sid in option_sids:
+            assert(isinstance(sid, int))
+
+        # sort for later searching with bisect
+        self.equity_sids = sorted(equity_sids)
```
llllllllll: It looks like the order here is not important, using a `set` would ensure that no element appears twice and will give us `O(c)` lookups later instead of `O(log(n))`.

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+            .format(class_name=self.__class__.__name__,
+                    cost=self.cost,
+                    min_trade_cost=self.min_trade_cost)
+
+    def calculate(self, transaction):
+        """
+        returns a tuple of:
+        (per share commission, total transaction commission)
+        """
+        assert isinstance(transaction.sid, int)
```
llllllllll: This assertion is a little misleading because it is an `Asset` which is a subclass of int. It is okay to omit the assertion.

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+        sid = transaction.sid
+
+        if _list_index(self.equity_sids, sid) is not None:
```
llllllllll: Assuming we make the change to a `set` in the constructor, this may become:

```python
if sid in self.tiered_assets:
    ...
else:
   ...
```

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+        else:
+            raise Exception('Unable to find sid = %i in either the'
+                            ' equities or options.' % sid)
+
+        commission = abs(transaction.amount * cost)
+        if min_trade_cost is None:
+            return cost, commission
+        else:
+            commission = max(commission, min_trade_cost)
+            return abs(commission / transaction.amount), commission
```
llllllllll: I think the first element is the same as `cost`

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+
+    @staticmethod
+    def ib_default_tier():
+        """
+        Interactive Brokers option commissions tier for < 10,000
+        monthly contracts traded.
+            https://gdcdyn.interactivebrokers.com/en/index.php?f=commission&p=options1
+        """
+        tier_dict = {}
+        tier_dict[(0.10, np.inf)] = 0.70
```
llllllllll: The parens are not needed inside the brackets and are normally omitted.

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+        tier_dict[(0.05, 0.10)] = 0.50
+        tier_dict[(-np.inf, 0.05)] = 0.25
+        return tier_dict
+
+
+def _lookup_option_cost(tier_dict, premium):
+    """
+    Returns the per share option commission for a given premium.
+    """
+    if tier_dict is None:
```
llllllllll: Why is `None` treated different from a lookup error below?

zipline/finance/commission.py
```diff
@@ -108,3 +111,135 @@ def calculate(self, transaction):
...
+        return 0.0
+    for (min_val, max_val) in tier_dict:
+        if min_val <= premium < max_val:
+            return tier_dict[(min_val, max_val)]
+    raise Exception('Failed to lookup tiered option cost.')
```
llllllllll: This should either be some more special error or just a `KeyError`. `Exception` is too general and hard to catch. We should also include the failing cost in the error message to help users debug the failure.

llllllllll:
Thank you for providing this feature! This feels like it really wants a first class representation for `Option` like `Equity` and `Future`. Zipline does not yet support options so I am not sure how you would be hitting this case, can you tell us a little more about how you are using this? ping @jfkirk

zipline/finance/commission.py
jimgoo: Ah ok cool, I can swing this.

zipline/finance/commission.py
jimgoo: Yeah indeed, sets would be much better. I can swing this too.

zipline/finance/commission.py
jimgoo: Yeah I had the thought going in that the more proper way would be to have an `Option(Asset)`object. This was going to take me a lot more time to understand how to implement and add tests for, so I did this rougher version first.

zipline/finance/commission.py
jimgoo: Ah ok, agreed.

zipline/finance/commission.py
jimgoo: Agreed.

jfkirk:
In general, it seems that what we want here isn't necessarily an equities/options commission model, but rather a commission model that is able to switch underlying commission models based on their SID.

This would allow you to implement two separate commission models and have them fork properly depending on the asset being passed in. This would be valuable even in a strictly-equities algorithm, as some blocks of equities might have different commission structures than others (ETFs, non-US equities, etc).

It might look something like this:

```
equities_commissions = PerShare()
options_commissions = PerShare(cost=.10, min_trade_cost=5.0)
options_sids = [2000000, 2000001, 2000002] # and so on and so forth

mixed_commission_model = MixedCommissionModel(
    default=equities_commissions,
    other_models=[(options_sids, options_commissions)], # you could add more models here as sid/model tuples
)
```

What do you folks think?

jimgoo:
Ooo I like the idea of @jfkirk, that layout makes a lot of sense. I suppose you could also do a similar thing with slippage, a mixed `MixedSlippageModel` that would work the same way. There probably isn't as much desire for that by others, but it seems like a case might arise where you would want to give some assets (ie. low volume stocks, options, and/or futures) more slippage than others.

zipline/finance/commission.py
pmart123: @llllllllll Curious  to proposing why the type of asset would imply a commission type? For instance, foreign shares are generally traded in BPS, US fixed, but then some platforms have fixed -> tiered commission schedules on the retail side. On the futures side, commissions could vary based on seat license or volume schedules. Seems like a constraint that doesn't agree with the necessary complex business logic unfortunately.

PR #1196, opened 2016-05-11
Cli improvements

ssanderson:
Adds a progress bar to quantopian-quandl downloads and fixes a bug where `zipline bundles` would print tuples in Python 2.

### DIFF
--- a/zipline/data/bundles/quandl.py
+++ b/zipline/data/bundles/quandl.py
@@ -304,12 +304,6 @@ def quandl_bundle(environ,
     )
 
 
-QUANTOPIAN_QUANDL_URL = (
-    'https://s3.amazonaws.com/quantopian-public-zipline-data/quandl'
-)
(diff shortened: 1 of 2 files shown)


zipline/data/bundles/quandl.py
```diff
@@ -320,8 +374,16 @@ def quantopian_quandl_bundle(environ,
...
-        print('Downloading quandl data. This can take about 1 minute.')
+        data = download_with_progress(
+            QUANTOPIAN_QUANDL_URL,
+            chunk_size=ONE_MEGABYTE,
+            label="Downloading Bundle: quantopian-quandl",
+        )
+    else:
+        data = download_without_progress(QUANTOPIAN_QUANDL_URL)
+
     # use closing for py2 compat
```
llllllllll: this comment is not true anymore

PR #1198, opened 2016-05-11
Allow BcolzMinuteBarWriter to append to most recent day

StewartDouglas:
Minutely data can now be appended to bcolz files even when
minutes in the same day have already been written. For example,
previously attempting to write data for the minute 2016-05-11 16:30
would raise an exception if any OHLCV data for 2016-05-11 had been
written to the same file.
    
Trying to overwrite existing minutes still raises a
BcolzMinuteOverlappingData exception.
    
Note that previously all sids' bcolz files ended at the same time.
This is no longer necessarily the case. The last record in each
sid's bcolz file now corresponds to the latest minute for which
OHLCV data is provided to the writer.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -441,6 +441,13 @@ class BcolzMinuteBarWriter(object):
         assert new_last_date == date, "new_last_date={0} != date={1}".format(
             new_last_date, date)
 
+    def set_sid_attrs(self, sid, **kwargs):
+        """Write all the supplied kwargs as attributes of the sid's file.
+        """
+        table = self._ensure_ctable(sid)
+        for k, v in kwargs.items():
+            table.attrs[k] = v
+
     def write(self, data, show_progress=False):
         """Write a stream of minute data.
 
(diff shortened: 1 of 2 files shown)


zipline/data/minute_bars.py
```diff
@@ -441,6 +441,13 @@ def pad(self, sid, date):
         assert new_last_date == date, "new_last_date={0} != date={1}".format(
             new_last_date, date)
 
+    def set_bcolz_attrs(self, sid, **kwargs):
```
ehebert: This is in support of adding the tick size for each future?

I would recommend taking `bcolz` out of the name, instead something like `set_sid_attrs` or `set_sid_metadata` so that the public writer interface is not tied to implementation.

We should add a test covering this method.

ehebert:
We should add some tests in `test_minute_bar_writer` exercising the new behavior, e.g. a test that writes up to some value before the market close, and then successfully writes another value in that same day.

zipline/data/minute_bars.py
StewartDouglas: It's in support of an adjustment factor, so that we can store a greater range of floats without truncating any decimal places.

Agreed on the renaming and test coverage.

zipline/data/minute_bars.py
```diff
@@ -441,6 +441,13 @@ def pad(self, sid, date):
         assert new_last_date == date, "new_last_date={0} != date={1}".format(
             new_last_date, date)
 
+    def set_sid_attrs(self, sid, **kwargs):
```
ehebert: If this is only being used for the sid adjustment factor, should we call this `set_sid_adjustment_factor`?

StewartDouglas: No, the plan is to use this for some other attributes associated with the sid.

zipline/data/minute_bars.py
```diff
@@ -756,6 +771,18 @@ def _open_minute_file(self, field, sid):
 
         return carray
 
+    def get_sid_attr(self, sid, name):
+        return self._get_sid_attr(sid, name)
```
llllllllll: why are we adding this level of indirection?

StewartDouglas: good catch, it's not necessary.

PR #1201, opened 2016-05-13
Dividends add currency col

mtydykov:
Adds "currency" and "type" columns to dividends dataset.


tests/pipeline/test_dividends.py
```diff
@@ -204,6 +245,10 @@ def setup(self, dates):
                               ['NaT', '2014-01-04', '2014-01-14'],
                               ['NaT', '2014-01-04']]
         amounts = [['NaN', 1, 15], ['NaN', 7, 13], ['NaN', 3, 1], ['NaN', 23]]
+        currency_types = [[None, "$", "EUR"], [None, "EUR", "$"],
+                          [None, "$", "EUR"], [None, "EUR"]]
```
kglowinski: Should the last list here have a 3rd element?

tests/pipeline/test_earnings.py
```diff
@@ -55,8 +55,7 @@ def setup(self, dates):
                 'datetime64[ns]', 'NaN'
             ),
             NEXT_ANNOUNCEMENT: self.get_expected_next_event_dates(
-                dates, 'datetime64[ns]', 'NaN'
-            ),
+                dates, 'datetime64[ns]', 'NaN'),
```
kglowinski: kruft, I suspect? Put it back, please.

kglowinski:
Should we start to somehow subdivide common? I feel like the strings in there tend to only be used to one "type" of pipeline loader. But they're all placed into the same file.

tests/pipeline/test_dividends.py
mtydykov: Nope, just 2. It's because in this last case, we don't know anything until the 5th, and then we find out about an announcement in euros, and we don't get any other announcements, so that's what we know until the end of our time frame.

mtydykov:
Yeah, maybe the constants can go into the specific loaders/tests where they belong. Initially I thought they should have a common place because some are used in qexec, but not all of them are actually used there, so maybe splitting them up makes sense.

kglowinski:
Squash WIP commit, then LGTM.

PR #1203, opened 2016-05-13
ENH: Allow leveraged etf as param to security list

ehebert:
To enable reading the leveraged etf data before creation of an algorithm
instance, add a parameter to `SecurityListSet` which passes in the
`leveraged_etf_data`.

(`SecurityListSet` in its current design requires to be created post
algorithm init since the `current_date_func` is the `get_datetime`
method of an algorithm instance.)

### DIFF
--- a/tests/test_security_list.py
+++ b/tests/test_security_list.py
@@ -24,7 +24,10 @@ LEVERAGED_ETFS = load_from_directory('leveraged_etf_list')
 
 class RestrictedAlgoWithCheck(TradingAlgorithm):
     def initialize(self, symbol):
-        self.rl = SecurityListSet(self.get_datetime, self.asset_finder)
+        self.rl = SecurityListSet(
+            self.get_datetime,
+            self.asset_finder,
+            load_from_directory('leveraged_etf_list'))
         self.set_do_not_order_list(self.rl.leveraged_etf_list)
         self.order_count = 0
         self.sid = self.symbol(symbol)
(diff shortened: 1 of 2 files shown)


tests/test_security_list.py
```diff
@@ -24,7 +24,10 @@
 
 class RestrictedAlgoWithCheck(TradingAlgorithm):
     def initialize(self, symbol):
-        self.rl = SecurityListSet(self.get_datetime, self.asset_finder)
+        self.rl = SecurityListSet(
+            self.get_datetime,
+            self.asset_finder,
+            load_from_directory('leveraged_etf_list'))
```
ssanderson: Is there a reason we're loading this in every test rather than just relying on `LEVERAGED_ETFS`?

ehebert: (We discussed this in person, but answering here for posterity.)

The current test design includes a context which copies the leveraged etf directories to a tmp directory and then patches the `security_list` module with the path to the temp directory. For the tests which add test leveraged etf, the `add_security_data` then writes a file to that temp directory. The call to `load_from_directory` on each test (which was previously being done further down in the `leveraged_etf` property) is necessary to get a fresh list for those tests that exercise exclusions.

Going to add a patch to this PR to add a fixture for this suite as we had discussed.

richafrank:
@ehebert Noticed this PR hasn't been touched in a while - what do we want to do with it?

ehebert:
Let's close it for now. The end goal of this work was to allow changing the ordering of how various resources are setup internally, but that ended up not being required for using fixtures throughout the internal codebase.

PR #1206, opened 2016-05-16
cleanup for sprint

llllllllll:


Dockerfile
```diff
@@ -2,58 +2,50 @@
...
 # set up environment
-# 
+#
 ENV PROJECT_DIR=/projects \
     NOTEBOOK_PORT=8888 \
     SSL_CERT_PEM=/root/.jupyter/jupyter.pem \
     SSL_CERT_KEY=/root/.jupyter/jupyter.key \
     PW_HASH="u'sha1:31cb67870a35:1a2321318481f00b0efdf3d1f71af523d3ffc505'" \
-    CONFIG_PATH=/root/.jupyter/jupyter_notebook_config.py
+    CONFIG_PATH=/root/.jupyter/jupyter_notebook_config.
```
richafrank: Intentional?

llllllllll: oops, nope

llllllllll:
hmm, so the dev experience here is that you should just mount your zipline root over /zipline so that you can edit the code outside of the container. Does this seem okay? I think it makes sense to just run by default but if you want to develop you need to pass the directory to the new source.

richafrank:
I think that sounds ok, as long as it's clear how to do it.

Issue #1205, opened 2016-05-15
PyCharm: zipline.api.x => Unresolved reference 'symbol' 


fredfortier:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

* Operating System: Mac OS El Capitain
* Python Version: 2.7.10
* Python Bitness: 64-bit
* How did you install Zipline: pip -- individually installed each library using version info in /zipline/etc/requirements.txt
* Zipline version: master

Now that you know a little about me, let me tell you about the issue I am
having:

I am using PyCharm to author my algorithms. I am able to run the simple the examples that ship with zipline. I can even analyze my results in pyfolio. As far as I can tell, zipline is installed correctly and runs happily. My problem is that I get unresolved reference errors in PyCharm for most of the zipline.api modules.

To illustrate the point, here is the contentious import statement:
from zipline.api import symbol, record, attach_pipeline, schedule_function, date_rules, \
    time_rules, order_value, order_target_percent

The PyCharm code analyzer has this to say regarding most of them:
Unresolved reference 'order_value'
Unresolved reference 'order_target_percent'

This does not prevent my algo from running. However, I am attached to the "code navigation" features of my IDE. After reviewing "api.py", I have some intuition about the cause. But I am still learning Python so I lack knowledge required to easily resolve this myself. Any suggestion would be appreciated.

Sincerely,
Frederic

llllllllll:
Many of the functions in this namespace are dynamically added. This is because they actually resolve to bound methods of some registered `TradingAlgorithm` instance. Here is what is statically in `zipline.api`: https://github.com/quantopian/zipline/blob/master/zipline/api.py#L16 and here is the decorator's definition: https://github.com/quantopian/zipline/blob/master/zipline/utils/api_support.py#L45. Inside `TradingAlgorithm` we enter a `ZiplineAPI` context like so: https://github.com/quantopian/zipline/blob/master/zipline/algorithm.py#L406 which makes it so that all of the api methods will resolve to the currently running algorithm. For example, here is how `order` gets added: https://github.com/quantopian/zipline/blob/master/zipline/algorithm.py#L1203.

This makes it very difficult for static analysis tools to understand the module; I get the same issue with Jedi, my code completion tool. I am not totally sure what the right fix is here.

fredfortier:
@llllllllll are you aware of any popular workaround which developers use to circumvent the limitation?

richafrank:
[...]

PR #1208, opened 2016-05-17
Api stubs

richafrank:
as well as tooling and docs to generate this for each release

For #1205 

screen shot 2016-05-17 at 10 34 42 am

### DIFF
--- a/docs/source/release-process.rst
+++ b/docs/source/release-process.rst
@@ -27,6 +27,32 @@ update the underline of the title to match the title's width.
 If you are renaming the release at this point, you'll need to git mv the file
 and also update releases.rst to reference the renamed file.
 
+Updating the Python stub files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+PyCharm and other linters and type checkers can use `Python stub files
+<https://www.python.org/dev/peps/pep-0484/#stub-files>`__ for type hinting. For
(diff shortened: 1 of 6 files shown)


etc/gen_type_stubs.py
```diff
@@ -0,0 +1,42 @@
+import inspect
+from operator import attrgetter
+from textwrap import dedent
+
+from zipline import api, TradingAlgorithm
+
+
+def main():
+    with open(api.__file__ + 'i', 'w') as stub:
```
llllllllll: we should rstrip 'c' from the filename because it could be `.pyc`

docs/source/release-process.rst
```diff
@@ -27,6 +27,21 @@ update the underline of the title to match the title's width.
 If you are renaming the release at this point, you'll need to git mv the file
 and also update releases.rst to reference the renamed file.
 
+Updating the Python stub files
```
llllllllll: should we comment that the imports may need to be updated?

etc/gen_type_stubs.py
richafrank: Agreed.

docs/source/release-process.rst
richafrank: Good idea.

llllllllll:
lgtm

PR #1209, opened 2016-05-17
Factor outputs name conflicts

dmichalowicz:


zipline/pipeline/term.py
```diff
@@ -452,8 +453,20 @@ def _validate(self):
         if self.inputs is NotSpecified:
             raise TermInputsNotSpecified(termname=type(self).__name__)
 
-        if not self.outputs:
+        if self.outputs is NotSpecified:
+            pass
```
dmichalowicz: Thought this `pass` might be clearer than a double negative check that says `if self.outputs is not NotSpecified`

dmichalowicz:
Closing for #1214

PR #1211, opened 2016-05-19
DOC: document limit and stop price args

llllllllll:


zipline/algorithm.py
```diff
@@ -1218,6 +1218,10 @@ def order(self,
             The amount of shares to order. If ``amount`` is positive, this is
             the number of shares to buy or cover. If ``amount`` is negative,
             this is the number of shares to sell or short.
+        limit_price : float, optional
+            The limit price for the order.
```
ssanderson: We might want to say somewhere that passing `limit_price=N` is equivalent to (and mutually-exclusive with) passing `style=LimitOrder(N)`, with the same going for `stop_price` with `StopOrder(N)`, and with `StopLimitOrder` when both are passed.

llllllllll: do you think this could go in a `Notes` section?

docs/Makefile
```diff
@@ -108,85 +113,85 @@ devhelp:
 	@echo "# ln -s $(BUILDDIR)/devhelp $$HOME/.local/share/devhelp/zipline"
 	@echo "# devhelp"
 
-epub:
+epub: build
```
ssanderson: Do we really need an epub target?

zipline/algorithm.py
```diff
@@ -1226,6 +1231,18 @@ def order(self,
         order_id : str
             The unique identifier for this order.
 
+        Notes
+        -----
+        The ``limit_price`` and ``stop_price`` arguments are mutually exclusive
+        with the ``style`` argument. ``limit_price`` and ``stop_price``
+        convenient shorthands for ``LimitOrder``, ``StopOrder``, and
+        ``StopLimitOrder`` execution styles. For example: ``limit_price=n`` is
+        equivalent to and mutually exclusive with
```
ssanderson: Using the phrase "mutually exclusive" twice  in consecutive sentences reads awkwardly.  How about this?

```
The ``limit_price`` and ``stop_price`` arguments provide shorthands for passing common execution styles.  Passing ``limit_price=N`` is equivalent to ``style=LimitOrder(N)``.  Similarly, passing ``stop_price=M`` is equivalent to ``style=StopOrder(M)``, and passing ``limit_price=N`` and ``stop_price=M`` is equivalent to ``style=StopLimitOrder(N, M)``.  It is an error to pass both a style and ``limit_price`` or ``stop_price``.
```

docs/Makefile
llllllllll: no, but I just did a find/replace for this.

PR #1212, opened 2016-05-19
DEV: Adjust performance calculations for capital changes

lianga888:
When we introduce a capital change (inflow or outflow of cash), make the appropriate adjustments to all the relevant performance fields on the `PerformancePeriod`.

Two types of capital changes:
- Capital Change between periods (daily periods)
  - When a capital change is specified outside of the algorithm's market hours
  - Because the capital change comes between daily periods, when a change is specified, just add that to the starting cash of that period (daily performance period starts with a greater portfolio value)
- Capital change within periods (daily periods or cumulative period)
  - if the capital change comes within the period (during market hours for daily periods, and anytime for the cumulative period), we need to segment this period by the datetime on which the changes come
  - example:
    - A cumulative period starts at t<sub>0</sub> with a PV of 1000, ends at t<sub>2</sub> with a PV of 3000
    - At t<sub>1</sub>, right before the capital change, it has a PV of 1500, and after the +1000 capital change at t<sub>1'</sub>, a PV of 2500
    - Returns = (PV<sub>t1</sub> - PV<sub>t0</sub>) * (PV<sub>t2</sub> - PV<sub>t1'</sub>) - 1 = 0.8   

- In `PerformancePeriod`, add a new method `subdivide_period`, which is called for the daily period when the change happens during market hours and for the cumulative period always:
  - Calculates the performance at that point and saves the pnl, returns and cash flow in memory
  - Adds the capital change, and then rolls over the ending value, exposure and cash to the new respective starting values and resets the cash flow, effectively starting a new 'subperiod'
  - When we calculate performance again, calculate the performance of the current and previous subperiod separately, and then incorporate them together
- In `PerformancePeriod`, `rollover` is called for the daily period when the change happens between days:
  - Adds the capital change, and rolls over a new period as before
- `TradingAlgorithm` get a `capital_changes` dictionary, with amount entries keyed by dt.
- `PerformanceTracker` has a new method `process_capital_changes`, which takes an amount, and calls appropriate methods on `PerformancePeriod` based whether its during the trading day or not
- `AlgorithmSimulator` calls `perf_tracker.process_capital_changes` at `DAY_START` and at `BAR`
- Tests:
  - performance in daily mode
  - performances in minute mode with (daily emission or minute emission) x (interperiod changes or intraperiod changes)

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -17,6 +17,7 @@ import datetime
 from datetime import timedelta
 from textwrap import dedent
 from unittest import TestCase, skip
+from copy import deepcopy
 
 import logbook
 import toolz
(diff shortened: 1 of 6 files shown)


zipline/algorithm.py
```diff
@@ -383,6 +384,40 @@ def __init__(self, *args, **kwargs):
 
         self.benchmark_sid = kwargs.pop('benchmark_sid', None)
 
+        self.capital_changes = kwargs.pop('capital_changes', {})
+        if self.capital_changes:
+            self.capital_changes = \
```
ssanderson: Why are we accepting these as strings?  If we want dates from the user, we should just have them pass dates.

lianga888: Doesn't really matter for me. @abhijeetkalyan suggested they be passed in this format

zipline/algorithm.py
```diff
@@ -383,6 +384,40 @@ def __init__(self, *args, **kwargs):
...
+                {pd.Timestamp(iso8601.parse_date(dt_str).
+                              strftime('%Y-%m-%d %H:%M'), tz='UTC'): v
+                 for dt_str, v in iteritems(self.capital_changes)}
+
+            for dt in copy(self.capital_changes):
```
ssanderson: IMO, all of these preprocessing and mutating steps don't belong here.  If we need data in a specific format, it should be up to the client to provide it in that format.

zipline/algorithm.py
```diff
@@ -418,6 +453,32 @@ def before_trading_start(self, data):
 
         self._in_before_trading_start = False
 
+    def process_capital_changes(self, dt):
```
ssanderson: Nothing about this depends on anything in the algorithm.  Can we just push this method down into `PerformanceTracker` and have `AlgorithmSimulator` call it directly?

zipline/finance/performance/period.py
```diff
@@ -143,14 +143,18 @@ def __init__(
         self.asset_finder = asset_finder
         self.data_frequency = data_frequency
 
+        # Start and end of the entire period
         self.period_open = period_open
         self.period_close = period_close
 
+        self.initialize(starting_cash=starting_cash,
```
ssanderson: What's the difference between `__init__` and `initialize`?  What logic belongs in `__init__` and what belongs in `initialize`?  Splitting them has the significant cost of making it harder to find all the instance attributes of this class, so if there isn't an important reason to split them, I'd argue that we should do all our initialization in one place.

zipline/algorithm.py
abhijeetkalyan: My main suggestion was that they be passed from quantopian's frontend to the executor as a string(ISO-8601 timestamp) because that's a format that can be easily sent over the wire. It could be that zipline works directly with the appropriate date object, and we take care of parsing strings to dates upstream

ssanderson: > It could be that zipline works directly with the appropriate date object, and we take care of parsing strings to dates upstream

Strong :+1: on moving this logic upstream (and removing all the non-midnight date normalization if possible).

As a general note, the question of "what serialization format should we use to do IPC at Quantopian?" is very different from "what interface should TradingAlgorithm present for providing this data".  We should make sure that we're clear about what layer of abstraction we're interested in when we're talking about how we represent data.

tests/test_perf_tracking.py
```diff
@@ -642,10 +643,12 @@ def test_long_position_receives_dividend(self):
         self.assertEqual(len(results), 6)
         cumulative_returns = \
             [event['cumulative_perf']['returns'] for event in results]
-        self.assertEqual(cumulative_returns, [0.0, 0.0, 0.1, 0.1, 0.1, 0.1])
+        assert_array_almost_equal(cumulative_returns,
```
ssanderson: Why did these all change?

zipline/finance/performance/period.py
lianga888: Mainly code duplication. We have to reset the values of some attributes when we roll over a period. It was previous done this way too, but it was `rollover` that initialized some values instead

tests/test_perf_tracking.py
lianga888: Due to floating point errors from some extra math we need to do. In this case, the returns used to be calculated as `1.0/10.0 = 0.1`. Now it's `(1 + 0.0) * (1 + 0.1) - 1 = 0.10000000000000009`

zipline/algorithm.py
```diff
@@ -383,6 +383,8 @@ def __init__(self, *args, **kwargs):
 
         self.benchmark_sid = kwargs.pop('benchmark_sid', None)
 
+        self.capital_changes = kwargs.pop('capital_changes', {})
```
ehebert: Could we put a comment either here or in the class docstring of the expected types of the key/value pairs? (I think it was mentioned in PR, but would be beneficial for quick reference here.)

zipline/gens/tradesimulation.py
```diff
@@ -145,6 +148,16 @@ def every_bar(dt_to_use, current_data=self.current_data,
 
         def once_a_day(midnight_dt, current_data=self.current_data,
                        data_portal=self.data_portal):
+
+            perf_tracker = algo.perf_tracker
```
ehebert: Since `every_bar` and `once_a_day` are all closures, this could be moved even further up to where `algo = self.algo` is defined. Though, not blocking for this PR.

zipline/gens/tradesimulation.py
```diff
@@ -190,10 +201,34 @@ def on_exit():
             if algo.data_frequency == 'minute':
                 def execute_order_cancellation_policy():
                     algo.blotter.execute_cancel_policy(DAY_END)
+
+                def process_minute_capital_changes(dt):
```
ehebert: Does defining this function require the processor or `ZiplineAPI` context to have been entered?

It is surprising that it is defined after `once_a_day` is defined.

lianga888: I'm not sure. Why does where it is defined with respect to `once_a_day` matter?

zipline/finance/performance/period.py
```diff
@@ -252,15 +306,24 @@ def calculate_performance(self):
 
         payout = self._get_payout_total(pt.positions)
 
-        total_at_start = self.starting_cash + self.starting_value
-        self.ending_cash = self.starting_cash + self.period_cash_flow + payout
+        # total at beginning of subperiod
```
ehebert: Checked this for idempotency, and looks like it is retained. :+1:
(It is currently possible to call `calculate_performance` multiple times per a given dt.)

zipline/finance/performance/period.py
```diff
@@ -212,6 +245,27 @@ def rollover(self):
             else:
                 del self._payout_last_sale_prices[asset]
 
+    def subdivide_period(self, capital_change):
+        self.calculate_performance()
+
+        # We are segmenting a previous subperiod that ran up until this point.
+        # Those performance stats are the total performance stats at this point
+        self._prev_subperiod_returns = self.total_returns
```
ehebert: These values operated on at the same time. The copy, especially, would benefit from having a single handle to the group of values which are affected by the period subdivisions. e.g. making a `namedtuple` like the `PeriodStats` used elsewhere. I do not have an immediate suggestion for what the name of that tuple should be.

(Not a merge blocker, but may help longterm.)

zipline/finance/performance/period.py
```diff
@@ -212,6 +245,27 @@ def rollover(self):
...
+        self._prev_subperiod_pnl = self.total_pnl
+        self._prev_subperiod_cash_flow = self.total_cash_flow
+
+        # Apply the capital change to the ending cash
+        self.ending_cash += capital_change
+
+        # We are segmenting a previous subperiod. Roll over the ending values
+        # of the previous into the starting values of the current. Reset cash
+        # flow
+        self._starting_value = self.ending_value
```
ehebert: https://github.com/quantopian/zipline/pull/1212/files#r64280750 applies here, though would this would be a different group of values.

zipline/finance/performance/period.py
```diff
@@ -193,15 +227,14 @@ def position_tracker(self, obj):
...
-    def rollover(self):
-        self.starting_value = self.ending_value
-        self.starting_exposure = self.ending_exposure
-        self.starting_cash = self.ending_cash
-        self.period_cash_flow = 0.0
-        self.pnl = 0.0
-        self.processed_transactions = {}
-        self.orders_by_modified = {}
-        self.orders_by_id = OrderedDict()
+    def rollover(self, capital_change=0):
```
ehebert: I do find how you are treating the daily period in a cumulative as a 'subperiod' easier to conceptualize.
The `rollover` naming is getting in the way of that new conceptualization. what do you think of `start_new_subperiod` or something along those lines?

ehebert: Can we make `capital_change`'s default `None` instead of 0, to make it clear that there are cases where the `ending_cash` is change or not?
And then make the change to cash as:
```
if capital_change is not None:
    self.ending_cash += capital_change
```

Though having said that, if we only change the value in the case of `if is_interday` in `process_capital_changes`, I think it would be more clear to make `process_capital_changes` do the processing of the capital_change in the scope of that function, i.e..:

```
if is_interday:
   # Change comes between days
   self.ending_cash += capital_change
   self.todays_performance.rollover() # or self.todays_performance.start_new_subperiod()
```

lianga888: I'm not exactly trying to treat a daily period in a cumulative as a 'subperiod`. To me, daily and cumulative periods are unrelated and treat capital changes differently. A 'subperiod` is just defined by a capital change that comes within a period, and this is applicable for both daily and cumulative. A subperiod in a cumulative period can be several days. `rollover` is not really a method meant to create a new subperiod, it's to make a new period. We `rollover` when we are at the end of a daily period and want to transfer those end values to the start values of the new period

zipline/gens/tradesimulation.py
ehebert: In this context, for legibility. That `once_a_day` refers to this function, and that `once_a_day` is  created earlier in `transform` and outside of the context, would imply (at least to me) that the deferred creation is necessary (but I am likewise unsure if it is.)

Wouldn't block this PR on it, but was mentioning since it is surprising.

tests/test_algorithm.py
```diff
@@ -1835,6 +1836,520 @@ def test_zero_trading_days(self, name, algocode):
...
+    def test_capital_changes_minute_mode_minute_emission(self, change):
+        sim_params = factory.create_simulation_parameters(
+            start=pd.Timestamp('2006-01-03', tz='UTC'),
+            end=pd.Timestamp('2006-01-05', tz='UTC'),
+            data_frequency='minute',
+            emission_rate='minute',
+            capital_base=1000.0
+        )
+
+        datestr = '2006-01-04 17:00' if change == 'intraday' else '2006-01-04'
```
ehebert: Should we test/support multiple `capital_changes` over the course of one day?

zipline/finance/performance/period.py
ehebert: Ok, becoming clearer now. (`initialize` and `subdivide` setting many of the same values was blurring that distinction.)

In that case, is it necessary to call `rollover` in `process_capital_changes` since `handle_daily_close` will also call `rollover`? (If the change is made to pull `self.ending_cash += capital_change` into `process_capital_changes`?) I don't think it is accurate to do 

Some other proposed changes:
- Make the "unset" values of `_prev_subperiod_pnl` etc. in `initialize` to be `None`. It is not that there was previous period with returns of 0, but that there was no previous period. (Am I understanding that correctly?)
- We may want to move `rollover` out of the shared `PerformancePeriod` class and make a `DailyPerformancePeriod` which clearly shows that rollover is only used in that context. (Would not block PR on breaking out 'rollover'.)

lianga888: I agree that `rollover` in it's entirety doesn't need to be called again in `process_capital_changes`. All we really need to do is increment the starting cash, which can be done as you suggested

zipline/finance/performance/period.py
```diff
@@ -212,6 +241,24 @@ def rollover(self):
...
+        self.calculate_performance()
+
+        # Apply the capital change to the ending cash
+        self.ending_cash += capital_change
+
+        # Increment the total capital change occurred within the period
+        self._total_intraperiod_capital_change += capital_change
+
+        # Divide the period into subperiods
+        self.subperiod_divider = SubPeriodDivider(
```
ehebert: Looks good to me. I think the divider object worked out well with making the before and after of the capital change clearer.

PR #1214, opened 2016-05-19
Factor outputs name conflicts

dmichalowicz:


zipline/pipeline/factors/factor.py
```diff
@@ -1205,15 +1205,20 @@ def compute(self, today, assets, out, close):
...
-        if attribute_name in self.outputs:
-            return RecarrayField(factor=self, attribute=attribute_name)
+            return getattr(super(CustomFactor, self), name)
         else:
             raise AttributeError(
                 'Instance of {factor} has no output called {attr!r}.'.format(
-                    factor=type(self).__name__, attr=attribute_name,
+                    factor=type(self).__name__, attr=name,
                 )
             )
```
dmichalowicz: I kept `__getattr__` here to make it clearer when we should show this error message

tests/pipeline/test_term.py
```diff
@@ -474,6 +478,13 @@ def test_bad_output_access(self):
             errmsg, "GenericCustomFactor does not have multiple outputs.",
         )
 
+        # Public method name, private method name, user-defined method name.
+        conflicting_output_names = ['compute', '_compute', 'some_method']
```
dmichalowicz: This is still broken for `inputs` as a name...

zipline/pipeline/term.py
```diff
@@ -466,8 +401,24 @@ def _validate(self):
...
+            disallowed_names = [
+                attr for attr in dir(ComputableTerm)
+                if not attr.startswith('_')
+            ]
+            for output in self.outputs:
+                if output.startswith('_') or output in disallowed_names:
+                    raise InvalidOutputName(
+                        output_name=output,
+                        termname=type(self).__name__,
+                        disallowed_names=disallowed_names,
```
dmichalowicz: Should this error be showing the list of prohibited names?

dmichalowicz: Note that `compute` is not in this list, so will still error if used as an output name. Without trying to make this too muddy, I'm thinking it should be added explicitly to the disallowed list

dmichalowicz:
@ssanderson Ready for a look

zipline/pipeline/term.py
```diff
@@ -543,6 +494,72 @@ def __repr__(self):
         )
 
 
+def validate_dtype(termname, dtype, missing_value):
+        """
```
ssanderson: indentation here is off

zipline/pipeline/factors/factor.py
```diff
@@ -1205,17 +1205,22 @@ def compute(self, today, assets, out, close):
...
 
-    def __getattr__(self, attribute_name):
+    def __getattribute__(self, name):
+        outputs = object.__getattribute__(self, 'outputs')
+        if outputs is not NotSpecified and name in outputs:
+            return RecarrayField(factor=self, attribute=name)
+        else:
+            return super(CustomFactor, self).__getattribute__(name)
+
+    def __getattr__(self, name):
```
ssanderson: I would probably just remove this method and make the logic in `__getattribute__` look like:

```
if outputs is NotSpecified:
    return super(...).__getattribute__(...)
if name in outputs:
    return RecarrayField(...)
raise AttributeError(...)
```

dmichalowicz: This won't work exactly, because if outputs is specified we would be raising an AttributeError for all other attributes. The best way I can think of to do everything in `__getattribute__` is something like:

```python
if outputs is NotSpecified:
    return super(...).__getattribute__(...)
elif name in outputs:
    return RecarrayField(...)
else:
    try:
        return super(...).__getattribute__(...)
    except AttributeError:
        raise AttributeError("No such output named...")
```

dmichalowicz: @ssanderson Do you think this is cleaner than keeping `__getattr__`?

ssanderson: I think it's probably slightly cleaner just because it keeps all the logic in one place.  I could go either way on this though, so do what you think makes sense.

PR #1221, opened 2016-05-20
Support the passing of a time rule positionally on the date_rule arg

lianga888:
When the user calls schedule_function(func, a_time_rule), assume that the user meant to specify a time rule but no date rule, instead of a date rule and no time rule as the signature suggests

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -16,6 +16,7 @@ from collections import namedtuple
 import datetime
 from datetime import timedelta
 from textwrap import dedent
+import warnings
 from unittest import TestCase, skip
 from copy import deepcopy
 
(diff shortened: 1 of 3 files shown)


tests/test_algorithm.py
```diff
@@ -1834,6 +1835,71 @@ def test_zero_trading_days(self, name, algocode):
...
+        for i, warning in enumerate(w):
+            self.assertIsInstance(warning.message, UserWarning)
+            self.assertEqual(
+                warning.message.message,
+                'Got a time rule for the second positional argument '
+                'date_rule. You should use keyword argument '
+                'time_rule= when calling schedule_function without '
+                'specifying a date_rule'
+            )
+            self.assertEqual(warning.lineno, 13 + i)
```
llllllllll: For zipline we can just define the functions outside of a string and then pass `handle_data` and `initalize` to `TradingAlgortithm`. A nice side effect of this is that we can check that the filename of the warning is `__file__`

tests/test_algorithm.py
```diff
@@ -1834,6 +1835,71 @@ def test_zero_trading_days(self, name, algocode):
...
+        def do_at_open(context, data):
+            context.done_at_open.append(context.get_datetime())
+
+        def do_at_close(context, data):
+            context.done_at_close.append(context.get_datetime())
+
+        def initialize(context):
+            context.done_at_open = []
+            context.done_at_close = []
+            schedule_function(do_at_open, time_rules.market_open())
```
llllllllll: it might help to put a comment on this line saying that it is line 13 this makes the assertion below more clear.

tests/test_algorithm.py
lianga888: I did that at first, but then the line number in the warning was the line in the test_algorithm.py script, and not the algocode

jbredeche:
@lianga888 @llllllllll what's the status of this?

lianga888:
@jean, waiting for travis then merging

PR #1226, opened 2016-05-23
MAINT: Make schedule function rules resilient for no trading day

lianga888:
Make the rules resilient to env.previous_trading_day or
env.next_trading_day being None

### DIFF
--- a/zipline/utils/events.py
+++ b/zipline/utils/events.py
@@ -20,6 +20,8 @@ import datetime
 import pandas as pd
 import pytz
 
+from zipline.errors import NoFurtherDataError
+
 from .context_tricks import nop_context
 
 
@@ -68,29 +70,6 @@ def ensure_utc(time, tz='UTC'):
     if not time.tzinfo:
         time = time.replace(tzinfo=pytz.timezone(tz))
     return time.replace(tzinfo=pytz.utc)
-
-
-def _coerce_datetime(maybe_dt):
-    if isinstance(maybe_dt, datetime.datetime):
-        return maybe_dt
-    elif isinstance(maybe_dt, datetime.date):
-        return datetime.datetime(
-            year=maybe_dt.year,
-            month=maybe_dt.month,
-            day=maybe_dt.day,
-            tzinfo=pytz.utc,
-        )
-    elif isinstance(maybe_dt, (tuple, list)) and len(maybe_dt) == 3:
-        year, month, day = maybe_dt
-        return datetime.datetime(
-            year=year,
-            month=month,
-            day=day,
-            tzinfo=pytz.utc,
-        )
-    else:
-        raise TypeError('Cannot coerce %s into a datetime.datetime'
-                        % type(maybe_dt).__name__)
 
 
 def _out_of_range_error(a, b=None, var='offset'):
@@ -434,23 +413,28 @@ class TradingDayOfWeekRule(six.with_metaclass(ABCMeta, StatelessRule)):
         raise NotImplementedError
 
     def calculate_start_and_end(self, dt, env):
-        next_trading_day = _coerce_datetime(
-            env.add_trading_days(
-                self.td_delta,
-                self.date_func(dt, env),
-            )
-        )
+        while True:
+            new_date = self.date_func(dt, env)
+            if not new_date:
+                return
 
-        # If after applying the offset to the start/end day of the week, we get
-        # day in a different week, skip this week and go on to the next
-        while next_trading_day.isocalendar()[1] != dt.isocalendar()[1]:
-            dt += datetime.timedelta(days=7)
-            next_trading_day = _coerce_datetime(
-                env.add_trading_days(
+            try:
+                next_trading_day = env.add_trading_days(
                     self.td_delta,
-                    self.date_func(dt, env),
+                    new_date,
                 )
-            )
+            except NoFurtherDataError:
+                return
+
+            if not next_trading_day:
+                return
+
+            # If after applying the offset to the start/end day of the week, we
+            # get day in a different week, skip this week and go on to the next
+            if next_trading_day.isocalendar()[1] == dt.isocalendar()[1]:
+                break
+            else:
+                dt += datetime.timedelta(days=7)
 
         next_open, next_close = env.get_open_and_close(next_trading_day)
         self.next_date_start = next_open
@@ -461,8 +445,11 @@ class TradingDayOfWeekRule(six.with_metaclass(ABCMeta, StatelessRule)):
         if self.next_date_start is None:
             # First time this method has been called. Calculate the midnight,
             # open, and close for the first trigger, which occurs on the week
-            # of the simulation start
+            # of the simulation start. Return False if there is not trigger
+            # that occurs on or before the max day
             self.calculate_start_and_end(dt, env)
+            if not self.next_date_start:
+                return False
 
         # If we've passed the trigger, calculate the next one
         if dt > self.next_date_end:
@@ -487,24 +474,16 @@ class NthTradingDayOfWeek(TradingDayOfWeekRule):
     def get_first_trading_day_of_week(dt, env):
         prev = dt
         dt = env.previous_trading_day(dt)
-        # If we're on the first trading day of the TradingEnvironment,
-        # calling previous_trading_day on it will return None, which
-        # will blow up when we try and call .date() on it. The first
-        # trading day of the env is also the first trading day of the
-        # week(in the TradingEnvironment, at least), so just return
-        # that date.
-        if dt is None:
-            return prev
-        while dt.date().weekday() < prev.date().weekday():
+        # Traverse backward until we hit a week border, then jump back to the
+        # previous trading day.
+        while dt and dt.weekday() < prev.weekday():
             prev = dt
             dt = env.previous_trading_day(dt)
-            if dt is None:
-                return prev
 
         if env.is_trading_day(prev):
-            return prev.date()
+            return prev
         else:
-            return env.next_trading_day(prev).date()
+            return env.next_trading_day(prev)
 
     date_func = get_first_trading_day_of_week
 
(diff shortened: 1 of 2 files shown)


zipline/utils/events.py
```diff
@@ -495,18 +507,16 @@ def get_first_trading_day_of_week(dt, env):
...
             prev = dt
             dt = env.previous_trading_day(dt)
-            if dt is None:
-                return prev
 
         if env.is_trading_day(prev):
             return prev.date()
-        else:
+        elif env.next_trading_day(prev):
             return env.next_trading_day(prev).date()
```
richafrank: How fast is `next_trading_day`?  Should we cache the result instead of calling it twice here?

zipline/utils/events.py
```diff
@@ -524,7 +534,7 @@ def get_last_trading_day_of_week(dt, env):
         dt = env.next_trading_day(dt)
         # Traverse forward until we hit a week border, then jump back to the
         # previous trading day.
-        while dt.date().weekday() > prev.date().weekday():
+        while dt and dt.date().weekday() > prev.date().weekday():
             prev = dt
             dt = env.next_trading_day(dt)
```
richafrank: Do we need the None check below here too for `previous_trading_day`?

zipline/utils/events.py
```diff
@@ -594,13 +605,14 @@ def get_nth_to_last_trading_day_of_month(self, dt, env):
...
 
-        if not self.td_delta:
-            self.day = self.get_last_trading_day_of_month(dt, env)
-        else:
-            self.day = env.add_trading_days(
-                self.td_delta,
-                self.get_last_trading_day_of_month(dt, env),
-            ).date()
+        self.day = self.get_last_trading_day_of_month(dt, env)
+        if self.td_delta and self.day:
```
richafrank: Now that this method has grown a bit, could we share its common code with `get_nth_trading_day_of_month`?

zipline/utils/events.py
```diff
@@ -436,23 +438,33 @@ def date_func(self, dt, env):
...
-                    self.date_func(dt, env),
+            try:
+                next_trading_day = _coerce_datetime(
+                    env.add_trading_days(
+                        self.td_delta,
+                        self.date_func(dt, env),
+                    )
                 )
-            )
+            except (NoFurtherDataError, TypeError):
```
richafrank: When do we expect a `TypeError`?  Since that's such a generic error type, is there any better way to handle it that won't potentially mask something else raising a `TypeError`?

zipline/utils/events.py
```diff
@@ -436,23 +438,33 @@ def date_func(self, dt, env):
...
+        try:
+            next_trading_day = _coerce_datetime(
+                env.add_trading_days(
+                    self.td_delta,
+                    new_date,
+                )
             )
-        )
+        except (NoFurtherDataError, TypeError):
+            return
```
richafrank: Instead of keeping this in sync with below, what do you think about making a starting sentinel value (like `None`) for `next_trading_day`?  Like elsewhere we have `while dt and ...`

zipline/utils/events.py
```diff
@@ -594,13 +605,14 @@ def get_nth_to_last_trading_day_of_month(self, dt, env):
...
-                self.td_delta,
-                self.get_last_trading_day_of_month(dt, env),
-            ).date()
+        self.day = self.get_last_trading_day_of_month(dt, env)
+        if self.td_delta and self.day:
+            try:
+                self.day = env.add_trading_days(self.td_delta, self.day)
+            except NoFurtherDataError:
+                self.day = None
+            if self.day:
```
richafrank: I think we could make this a bit less complex by replacing the conditional with an `else`:
```python
try:
    self.day = ...
except NoFurtherDataError:
    self.day = None
else:
    self.day = self.day.date()
```

zipline/utils/events.py
lianga888: in `events.py`, `_coerce_datetime` raises a `TypeError` when it tries to coerce a `None` dt. I could have a raise an new custom error, but is it worth it?

richafrank: In that case, what do you think about tying the error types to the expected producer, instead of expecting either type from any of the three function calls?  This currently says that a `TypeError` from `self.date_func` is expected.

zipline/utils/events.py
lianga888: I'm not sure I understand what you mean by this

zipline/utils/events.py
lianga888: `add_trading_days` could potentially return `None` as well, so we need to account for that

richafrank: Ah ok

zipline/utils/events.py
lianga888: Is it worth is to make a base class off which to subclass these two rules (like it is done with the week rule)? Or should it just share the method defined somewhere

richafrank: A base class sounds reasonable to me here.

zipline/utils/events.py
```diff
@@ -434,23 +413,28 @@ def date_func(self, dt, env):
...
                     self.td_delta,
-                    self.date_func(dt, env),
+                    new_date,
                 )
-            )
+            except NoFurtherDataError:
+                return
+
+            if not next_trading_day:
+                return
```
richafrank: Now that there are cases when `calculate_start_and_end` might return without changing `self.next_date_end`, could `should_trigger` raise an exception when `self.next_date_start` and `self.next_date_end` are `None` by comparing `dt > self.next_date_end`?

zipline/utils/events.py
```diff
@@ -522,19 +502,45 @@ def get_last_trading_day_of_week(dt, env):
...
+            return self.day
+
+        self.day = self.date_func(dt, env)
+        if self.td_delta and self.day:
+            try:
+                self.day = env.add_trading_days(self.td_delta, self.day)
+            except NoFurtherDataError:
+                self.day = None
+            if self.day:
+                self.day = self.day.date()
```
richafrank: Is it intentional that here `self.day` is assigned a `date`, but above it's assigned a `datetime`?

zipline/utils/events.py
```diff
@@ -522,19 +502,45 @@ def get_last_trading_day_of_week(dt, env):
...
+class TradingDayOfMonthRule(six.with_metaclass(ABCMeta, StatelessRule)):
+
+    def should_trigger(self, dt, env):
+        return self.get_trigger_day_of_month(dt, env) == dt.date()
+
+    @abstractmethod
+    def date_func(self, dt, env):
+        raise NotImplementedError
+
+    def get_trigger_day_of_month(self, dt, env):
```
richafrank: Glad this is being shared now!

zipline/utils/events.py
```diff
@@ -493,18 +477,14 @@ def get_first_trading_day_of_week(dt, env):
         # trading day of the env is also the first trading day of the
         # week(in the TradingEnvironment, at least), so just return
         # that date.
```
richafrank: I think this comment is outdated now.

zipline/utils/events.py
lianga888: Yeah, you're right. Will have `should_trigger` return False if `next_date_start` is still `None` after it tries to calculate

zipline/utils/events.py
lianga888: The naming here makes things very confusing. `self.date_func` actually returns a `datetime.date`. However `env.add_trading_days` will return a `datetime.datetime`, which needs to be coerced. I think everything about this is more complicated than needed. So, I'm going to leave everything as `datetime.datetime`, which are also all normalized. When we make the comparison in `should_trigger`, we'll compare to the normalized dt that's passed in

zipline/utils/events.py
```diff
@@ -522,19 +501,43 @@ def get_last_trading_day_of_week(dt, env):
...
+            return self.date
+
+        self.date = self.date_func(dt, env)
+        if self.td_delta and self.date:
+            try:
+                self.date = env.add_trading_days(self.td_delta, self.date)
+            except NoFurtherDataError:
+                self.date = None
+
+        return self.date
```
richafrank: If `self.month`, `self.td_delta`, and `self.date` are expected in all subclasses, you might move setting them up to the `__init__` of this base class.

zipline/utils/events.py
```diff
@@ -487,24 +474,16 @@ class NthTradingDayOfWeek(TradingDayOfWeekRule):
...
-        # will blow up when we try and call .date() on it. The first
-        # trading day of the env is also the first trading day of the
-        # week(in the TradingEnvironment, at least), so just return
-        # that date.
-        if dt is None:
-            return prev
-        while dt.date().weekday() < prev.date().weekday():
+        # Traverse backward until we hit a week border, then jump back to the
+        # previous trading day.
+        while dt and dt.date().weekday() < prev.date().weekday():
```
richafrank: Not sure how inner this loop is, but I think we can get `weekday()` from `dt` itself.

lianga888:
Yup, added a test for the comparison to `None` case, and the date versus datetime issue is covered by existing tests. Previously we were comparing two dates; now we're comparing two datetimes

zipline/utils/events.py
```diff
@@ -522,85 +501,74 @@ def get_last_trading_day_of_week(dt, env):
...
+
+class NDaysBeforeLastTradingDayOfMonth(TradingDayOfMonthRule):
     """
     A rule that triggers n days before the last trading day of the month.
     """
     def __init__(self, n=0):
-        if not 0 <= n < MAX_MONTH_RANGE:
-            raise _out_of_range_error(MAX_MONTH_RANGE)
+        super(NDaysBeforeLastTradingDayOfMonth, self).__init__(n)
         self.td_delta = -n
```
richafrank: Probably more consistent with the other attrs to have the base class do the assignment.  Could include a flag for whether to invert it.  What do you think?

PR #1227, opened 2016-05-23
ENH: improve performance of blaze core loader

llllllllll:
Breaking out what I have now so I don't build up a very large diff.

1. Only do the lower bound query for the base expression, not the deltas. This was extra work that is not needed.
1. `DatetimeIndex.time >= time` is _really_ slow. Implements a `mask_time_between` which will do the time comparison against the datetime index much more quickly.
1. Upgrade blaze ecosystem. One important change is to drop an unneeded copy in the blaze server/client.

This shaves about 1.5s off the tests which is nice.

### DIFF
--- a/zipline/pipeline/loaders/blaze/core.py
+++ b/zipline/pipeline/loaders/blaze/core.py
@@ -904,44 +904,12 @@ class BlazeLoader(dict):
             q : Expr
                 The query to run.
             """
-            def lower_for_col(column):
-                pred = e[TS_FIELD_NAME] <= lower_dt
-                colname = column.name
-                schema = e[colname].schema.measure
-                if isinstance(schema, Option):
-                    pred &= e[colname].notnull()
-                    schema = schema.ty
-                if schema in floating:
-                    pred &= ~e[colname].isnan()
-
-                filtered = e[pred]
-                lower = filtered[TS_FIELD_NAME].max()
-                if have_sids:
-                    # If we have sids, then we need to take the earliest of the
-                    # greatest date that has a non-null value by sid.
-                    lower = bz.by(
-                        filtered[SID_FIELD_NAME],
-                        timestamp=lower,
-                    ).timestamp.min()
-                return lower
-
-            lower = odo(
-                reduce(
-                    bz.least,
-                    map(lower_for_col, columns),
-                ),
-                pd.Timestamp,
-                **odo_kwargs
-            )
-            if lower is pd.NaT:
-                lower = lower_dt
             return e[
-                (e[TS_FIELD_NAME] >= lower) &
                 (e[TS_FIELD_NAME] <= upper_dt)
             ][added_query_fields + list(map(getname, columns))]
 
         def collect_expr(e):
-            """Execute and merge all of the per-column subqueries.
+            """Materialize the expression as a dataframe.
 
             Parameters
             ----------
--- a/zipline/pipeline/loaders/blaze/core.py
+++ b/zipline/pipeline/loaders/blaze/core.py
@@ -127,7 +127,7 @@ from __future__ import division, absolute_import
 from abc import ABCMeta, abstractproperty
 from collections import namedtuple, defaultdict
 from copy import copy
-from functools import partial, reduce
+from functools import partial
 from itertools import count
 import warnings
 from weakref import WeakKeyDictionary
@@ -137,7 +137,6 @@ from datashape import (
     Date,
     DateTime,
     Option,
-    floating,
     isrecord,
     isscalar,
 )
--- a/zipline/utils/pandas_utils.py
+++ b/zipline/utils/pandas_utils.py
@@ -1,6 +1,7 @@
 """
 Utilities for working with pandas objects.
 """
+from itertools import product
 import operator as op
 
 import pandas as pd
@@ -37,7 +38,7 @@ def sort_values(ob, *args, **kwargs):
 
 
 def _time_to_micros(time):
-    """Convert a time into milliseconds since midnight.
+    """Convert a time into microseconds since midnight.
 
     Parameters
     ----------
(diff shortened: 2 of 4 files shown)


zipline/pipeline/loaders/blaze/core.py
```diff
@@ -873,62 +873,73 @@ def _load_dataset(self, dates, assets, mask, columns):
...
-            lower = odo(
-                reduce(
-                    bz.least,
-                    map(lower_for_col, columns),
-                ),
-                pd.Timestamp,
-                **odo_kwargs
-            )
-            if lower is pd.NaT:
+            if not deltas:
```
llllllllll: The diff is messy here. The block was just indented by 4 inside the `if` block.

zipline/utils/pandas_utils.py
```diff
@@ -15,6 +17,86 @@ def explode(df):
...
+        return _series_sort_values(ob, *args, **kwargs)
+    raise ValueError(
+        'sort_values expected a dataframe or series, not %s: %r' % (
+            type(ob).__name__, ob,
+        ),
+    )
+
+
+def _time_to_micros(time):
+    """Convert a time into milliseconds since midnight.
```
ssanderson: The name here says micros, but the docstring says millis.  From reading the code, I think the docstring is correct?

zipline/utils/pandas_utils.py
```diff
@@ -15,6 +17,86 @@ def explode(df):
...
+
+    Parameters
+    ----------
+    time : datetime.time
+        The time to convert.
+
+    Returns
+    -------
+    ms : int
+        The number of milliseconds since midnight.
```
ssanderson: (Assuming no leap seconds or DST.  Does that affect the places where this is being used?)

zipline/utils/pandas_utils.py
```diff
@@ -15,6 +17,86 @@ def explode(df):
...
+    :meth:`pandas.DatetimeIndex.indexer_between_times`
+    """
+    time_micros = dts._get_time_micros()
+    start_micros = _time_to_micros(start)
+    end_micros = _time_to_micros(end)
+
+    lop, rop = _opmap[include_start, include_end]
+    if start_micros <= end_micros:
+        join_op = op.and_
+    else:
```
ssanderson: What's the intended behavior here if start is greater than end?

zipline/utils/pandas_utils.py
```diff
@@ -15,6 +17,86 @@ def explode(df):
...
+        Inclusive on ``end``.
+
+    Returns
+    -------
+    mask : np.ndarray[bool]
+        A bool array masking ``dts``.
+
+    See Also
+    --------
+    :meth:`pandas.DatetimeIndex.indexer_between_times`
```
ssanderson: The method (at least on pandas 16) is `indexer_between_time` (no s).

zipline/utils/pandas_utils.py
```diff
@@ -15,6 +17,86 @@ def explode(df):
...
+    """
+    time_micros = dts._get_time_micros()
```
ssanderson: We should probably comment here that this is extracted from `indexer_between_times` and give credit to the original author(s).  (Looks like Wes McKinney, Chang She, and Grant Roch).  We could also ask jreback if there's a preferred citation method for code that's ported out of pandas.

llllllllll: Good call, I will put something in the Notes and check with Jeff.

ssanderson: I'm not sure it needs to be in the docstring, since where the code came from isn't really public-facing info.

ssanderson: We're also still using `DatetimeIndex._get_time_micros`, which isn't public.

The implementation of that method (in 0.16 at least) is:

```
    def _get_time_micros(self):
        utc = _utc()
        values = self.asi8
        if self.tz is not None and self.tz is not utc:
            values = self._local_timestamps()
        return tslib.get_time_micros(values)
```

Where `get_time_micros` is a Cython function that looks like a vectorized version of _time_to_micros.
Should we at least use the tslib function instead of relying on a private DatetimeIndex method?

```
def get_time_micros(ndarray[int64_t] dtindex):
    """
    Datetime as int64 representation to a structured array of fields
    """
    cdef:
        Py_ssize_t i, n = len(dtindex)
        pandas_datetimestruct dts
        ndarray[int64_t] micros

    micros = np.empty(n, dtype=np.int64)

    for i in range(n):
        pandas_datetime_to_datetimestruct(dtindex[i], PANDAS_FR_ns, &dts)
        micros[i] = 1000000LL * (dts.hour * 60 * 60 +
                                 60 * dts.min + dts.sec) + dts.us

    return micros
```

zipline/utils/pandas_utils.py
llllllllll: the docstring is wrong, `1000000 * seconds` is micro

zipline/utils/pandas_utils.py
llllllllll: I put this in the note, it does not affect the usage.

zipline/utils/pandas_utils.py
llllllllll: This is actually how we are using it: if we unroll our particular case of `8:45 <= time < 0:00` we get:

```python
(start_micros <= time_micros) or (time_micros < end_micros)
```

since time_micros = 0 we reduce to:

`start_micros <= time` which is correct.

zipline/utils/pandas_utils.py
llllllllll: are you suggesting inlining this to rely only on the tslib function? Isn't tslib also "private" but just not prefixed with "_"? If that is the case I would leave it like this to make it clear we are using some priivate methods.

PR #1231, opened 2016-05-25
LOG: Log the dt and amount when each capital change executes

lianga888:


abhijeetkalyan:
Can we log this on the top-level entrypoint, as soon as the data comes in from qf?

lianga888:
Yeah, that'll be in the qexec though

PR #1232, opened 2016-05-25
BUG: Fixes bcolz padding to not always pad 390 minutes

yankees714:
If minutes already exist for the last existing day, adjust the number of minutes padded to account for them. Previously we would always pad 390, leading to a mismatch in the number of rows.


ehebert:
LGTM.

PR #1233, opened 2016-05-25
MAINT: Make env a required argument to algo.

ehebert:
Remove creation of TradingEnvironment during init, in favor of passing
it explicitly.

N.B, Do not merge before internal branch is compatible.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -54,7 +54,6 @@ from zipline.errors import (
     UnsupportedSlippageModel,
     CannotOrderDelistedAsset, UnsupportedCancelPolicy, SetCancelPolicyPostInit,
     OrderInBeforeTradingStart)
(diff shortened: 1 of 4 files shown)


zipline/algorithm.py
```diff
@@ -193,10 +192,12 @@ class TradingAlgorithm(object):
         default: 'zipline'
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, env, *args, **kwargs):
         """Initialize sids and other state variables.
 
         :Arguments:
+        :Required:
```
llllllllll: the __init__ docstring should be removed the args should be documents in the class docstring

ehebert:
Talked with @ssanderson and heeding his advice to hold off on this until at least after next release because it is a breaking API change.

PR #1242, opened 2016-06-02
MAINT: Adds first_trading_day arg to DataPortal

yankees714:
Instead of inferring it from the minute/daily writer, we now require the first trading day to be passed explicitly, so the creator of the `DataPortal` controls what is used as the first trading day.


ehebert:
LGTM. Removes a bunch of guesswork. :+1:

PR #1246, opened 2016-06-02
BUG: Don't delete namespace if import fails.

jeremyblow:
datashape should be optional, this ensures it is actually optional.


llllllllll:
merging on pass! Thank you!

Issue #1235, opened 2016-05-25
Remove Cycles Between TradingAlgorithm and AlgorithmSimulator


ssanderson:
AppVeyor builds mysteriously started failing on https://github.com/quantopian/zipline/pull/1230 trying to delete the adjustments db in a temporary directory.

The root of the issue, ultimately, is the fact that we create a cycle between TradingAlgorithm and AlgorithmSimulator, which prevents deterministic collection of any resources referenced by the Algorithm.

I tried debugging this using the [objgraph](http://mg.pov.lt/objgraph/objgraph.html#module-objgraph) package, but was initially stymied because I was getting empty graphs when trying to find referrers to `SQLiteAdjustmentReader`.  Reading the source, I learned that most objgraph functions call `gc.collect` before building their graphs, which cleaned up the reader I was trying to debug.  The solution to this was to monkey-patch `gc.collect` when generating reference graphs.

For future reference, anyone trying to debug a nondeterministic gc collection issue with objgraph should use something like the following:

```
        import gc; import objgraph
        gc.collect = lambda: 0
        objgraph.show_backrefs(
            objgraph.by_type('SQLiteAdjustmentReader'),
            max_depth=5,
            too_many=20,
            filename='/home/ssanderson/backrefs.svg',
        )
```

The result of which is:

![backrefs2](https://cloud.githubusercontent.com/assets/1588357/15556143/69576404-229a-11e6-894a-fdd277db7878.png)

ssanderson:
The short-term fix for this is judicious use of `gc.collect` in test_examples.  The real fix is to break the cycles between AlgorithmSimulator and TradingAlgorithm by making the algo no longer responsible for constructing the simulator.

richafrank:
Made a different short-term fix by removing the algo instance reference from the simulator at the end of the simulation, instead of gc'ing in the tests.

PR #1250, opened 2016-06-03
MAINT: Break reference cycle between algorithm and simulator

richafrank:
when simulation finishes, so that resources are cleaned up
deterministically.

See #1235.


ssanderson:
LGTM

PR #1253, opened 2016-06-03
DEV: VagrantFile for Ubuntu 14.04 that includes the development envir…

Ericlafevers:
…onment.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,11 +1,82 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-Vagrant.configure("2") do |config|
-  config.vm.box = "precise64"
-  config.vm.box_url = "http://files.vagrantup.com/precise64.box"
-  config.vm.provider :virtualbox do |vb|
-    vb.customize ["modifyvm", :id, "--memory", 2048, "--cpus", 2]
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure(2) do |config|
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -1,11 +1,82 @@
...
+  # Example for VirtualBox:
+  #
+  config.vm.provider "virtualbox" do |vb|
+  #   # Display the VirtualBox GUI when booting the machine
+  #   vb.gui = true
+  #
+    # Customize the amount of memory on the VM:
+    vb.memory = "2048"
   end
-  config.vm.provision "shell", path: "vagrant_init.sh"
```
richafrank: Thanks for the updates!

Looking over the changes here now, I notice that we used to run `vagrant_init.sh` - how does the inline shell command below compare to that?  Could we merge the two?  Would be nice not to have the legacy `vagrant_init.sh` hanging around, if we're not using it, and also we don't want to lose any functionality that was provided there.  I wonder if we should update `vagrant_init.sh` as necessary, continuing to call it as before, so we can clearly see the diff here.

richafrank: @Ericlafevers Do you think that's something you'll have time to clean up?

Ericlafevers: Hi Richard,

Unfortunately not in the immediate future. My apologies for the delay, I
have been swamped since returning from PyCon and I have not had time to get
back to this.

richafrank: No worries!  Would it be helpful if we cleaned it up for you?

Ericlafevers: Yes, please feel free to use/modify/scrap any part of it however you'd
like.

richafrank:
@Ericlafevers, I'm going to close this, since I merged #1310 as a replacement PR that includes and updates your changes.  Thanks so much for the improvement!

PR #1256, opened 2016-06-03
BUG: Apply latest adjustment for minute `1d`

ehebert:
Fix behavior in minute mode history with frequency `1d`, where on the
day immediately following an adjustment action, the overnight adjustment
would not apply. (However the adjustment would be applied after a 1 day
lag.)

The root cause of the bug was that the history data for minute mode when
using `1d` stitches together a sliding window of the daily data for
previous  and the current minute. That daily data sliding window and
corresponding adjustments was being read as if the data was being viewed
from on the last day of the window; however in this case the data is
being viewed from the day after the window has completed. The difference
in view points requires the adjustments to popped and applied by the
adjusted array one index earlier. The fix uses the `extra_slot` value as
signifier on whether the data is being viewed on the following day and
then accordingly adjusts the index of the mulitpy object.

Also, change the split and merger test data ratios to have different values,
to ensure that different adjustment values are applied; as opposed to
doubling up on just one of the values.

### DIFF
--- a/zipline/data/us_equity_loader.py
+++ b/zipline/data/us_equity_loader.py
@@ -189,11 +189,11 @@ class USEquityHistoryLoader(with_metaclass(ABCMeta)):
             'splits', sid)
         for s in splits:
             dt = s[0]
-            if field == 'volume':
-                ratio = 1.0 / s[1]
-            else:
-                ratio = s[1]
             if start < dt <= end:
+                if field == 'volume':
+                    ratio = 1.0 / s[1]
+                else:
+                    ratio = s[1]
                 end_loc = dts.searchsorted(dt)
                 adj_loc = end_loc
                 if is_perspective_after:
--- a/zipline/data/us_equity_loader.py
+++ b/zipline/data/us_equity_loader.py
@@ -103,7 +103,8 @@ class USEquityHistoryLoader(with_metaclass(ABCMeta)):
     def _array(self, start, end, assets, field):
         pass
 
-    def _get_adjustments_in_range(self, asset, dts, field):
+    def _get_adjustments_in_range(self, asset, dts, field,
+                                  is_perspective_after):
         """
         Get the Float64Multiply objects to pass to an AdjustedArrayWindow.
 
@@ -126,6 +127,12 @@ class USEquityHistoryLoader(with_metaclass(ABCMeta)):
             The days for which adjustment data is needed.
         field : str
             OHLCV field for which to get the adjustments.
+        is_perspective_after : bool
+            see: `USEquityHistoryLoader.history`
+            If True, the index at which the Multiply object is registered to
+            be popped is calculated so that it applies to the last slot in the
+            sliding window  when the adjustment occurs immediately after the dt
+            that slot represents.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


zipline/data/us_equity_loader.py
```diff
@@ -126,6 +127,13 @@ def _get_adjustments_in_range(self, asset, dts, field):
             The days for which adjustment data is needed.
         field : str
             OHLCV field for which to get the adjustments.
+        is_perspective_after : bool
+            see: _ensure_sliding_window
+            If True, the index at which the Mulitply object should be popped
```
jbredeche: some grammatical and spelling issues here, maybe from stitching together different versions of the docstring. 

maybe remove `the value before the time at which the adjustment is marked` ?  and clean up the typos.

zipline/data/us_equity_loader.py
```diff
@@ -176,18 +190,22 @@ def _get_adjustments_in_range(self, asset, dts, field):
                 ratio = s[1]
             if start < dt <= end:
                 end_loc = dts.searchsorted(dt)
+                adj_loc = end_loc
```
jbredeche: out of scope here, but the work done in lines 187-190 should probably be moved into this `if` block, to avoid doing unnecessary work.

zipline/data/us_equity_loader.py
```diff
@@ -207,6 +225,13 @@ def _ensure_sliding_windows(self, assets, dts, field):
             in the calendar.
         field : str
             The OHLCV field for which to retrieve data.
+        is_perspective_after : bool
+            True if the is window being viewed immediately after the last dt
```
jbredeche: typo - looks like `is` is in the wrong place?

zipline/data/us_equity_loader.py
```diff
@@ -207,6 +225,13 @@ def _ensure_sliding_windows(self, assets, dts, field):
...
+            in the sliding window.
+            False if the window is being last dt.
```
jbredeche: another typo here: `is being last dt`

zipline/data/us_equity_loader.py
```diff
@@ -278,13 +305,22 @@ def history(self, assets, dts, field):
             in the calendar.
         field : str
             The OHLCV field for which to retrieve data.
-
+        is_perspective_after : bool
+            True if the is window being viewed immediately after the last dt
```
jbredeche: same typos as above

zipline/data/us_equity_loader.py
```diff
@@ -142,30 +150,36 @@ def _get_adjustments_in_range(self, asset, dts, field):
                 dt = m[0]
                 if start < dt <= end:
                     end_loc = dts.searchsorted(dt)
+                    adj_loc = end_loc
+                    if is_perspective_after:
+                        adj_loc -= 1
```
jbredeche: can you add a comment explaining why `-=1` is the right thing to do here?

jbredeche:
looks good, bunch of docstring issues and a request for some more comments in the main area of change.

zipline/data/us_equity_loader.py
ehebert: Agreed, can follow up with that.

zipline/data/us_equity_loader.py
ehebert: Yes.

ehebert:
Fixed docstrings and moved the ratio adjustment.

zipline/data/us_equity_loader.py
```diff
@@ -126,6 +127,12 @@ def _get_adjustments_in_range(self, asset, dts, field):
             The days for which adjustment data is needed.
         field : str
             OHLCV field for which to get the adjustments.
+        is_perspective_after : bool
+            see: `USEquityHistoryLoader.history`
+            If True, the index at which the Mulitply object is registered to
```
jbredeche: This is pretty hard to parse, I've read it five times and still am not totally sure what you are saying.  Can you break it up into smaller sentences?

also, typo "Mulitply"

zipline/data/us_equity_loader.py
```diff
@@ -278,13 +308,22 @@ def history(self, assets, dts, field):
             in the calendar.
         field : str
             The OHLCV field for which to retrieve data.
-
+        is_perspective_after : bool
+            True if the window is being viewed immediately after the last dt
+            in the sliding window.
+            False if the window is being last dt.
```
jbredeche: typo "is being last dt"

zipline/data/us_equity_loader.py
```diff
@@ -278,13 +308,70 @@ def history(self, assets, dts, field):
...
-
+        is_perspective_after : bool
+            True, if the window is being viewed immediately after the last dt
+            in the sliding window.
+            False, if the window is viewed on the last dt.
+
+            This flag is used for handling the case where the last dt in the
+            requested window immediately precedes a corporate action, e.g.:
+
+            - is_perpsective_after is True
```
jbredeche: minor typo, but docs look much better, thanks!

PR #1257, opened 2016-06-03
MAINT: Remove unused data conversion script.

ehebert:
The file format converted by this script has no support for reading in
Zipline. Remove since it requires import of a library not defined in
requirements.


llllllllll:
Assuming nothing is using this I am fine removing it.

ehebert:
Grepped code base and found no references.

PR #1259, opened 2016-06-03
ENH: Adds LinearWeightedMovingAverage Factor

swacad:


zipline/pipeline/factors/technical.py
```diff
@@ -101,6 +102,34 @@ def compute(self, today, assets, out, data):
...
+    def compute(self, today, assets, out, data):
+        num_days = data.shape[0]
+
+        # Initialize weights array
+        weights = np.arange(1, num_days + 1, dtype=float)
+
+        # Compute normalizer
+        normalizer = (num_days * (num_days + 1)) / 2
+
+        for i in range(data.shape[1]):
```
llllllllll: We can take advantage of numpy broadcasting rules here and vectorize this loop. It looks like we want to multiply each column by the weights array. We can get numpy to do the looping in C by turning our weights into a column vector and then multiplying it by the array like:
```python
data * weights[:, np.newaxis]
```

We can also get `nansum` to loop over these columns in C by passing the `axis` keyword argument to sum over a column. For example:
```python
nansum(data * weights[:, np.newaxis], axis=0)
```

This will return a 1d array of length `data.shape[1]` which is the weighted sum of the columns. Finally, we can do the division by the normalizer in constant space by doing an inplace div:
```python
out[:] = nansum(data * weights[:, np.newaxis], axis=0)
out /= normalizer
```

This does the same as the loop but should run much faster.

tests/pipeline/test_technical.py
```diff
@@ -138,3 +141,20 @@ def test_bollinger_bands_output_ordering(self):
...
+    def test_wma(self):
+        wma = LinearWeightedMovingAverage(
+            inputs=(USEquityPricing.close,),
+            window_length=10
+        )
+
+        today = pd.Timestamp('2014')
+        assets = np.arange(5, dtype=np.int64)
+
+        data = np.ones((10, 5))
```
llllllllll: I think the constant test is good, but can we add a test with something like arange, or some handcrafted inputs?

zipline/pipeline/factors/technical.py
swacad: I'd be happy to make the suggested changes.  What's the best way to implement changes and get them pushed?

llllllllll: If you push another commit to the same branch then the github will add it to the PR and the CI will run the tests on the new changes.

swacad: Okay.  This week is busy, but I'll try to get to it this weekend.

llllllllll:
Do you mind rebasing this so we can merge, if not I can rebase, run the tests then merge.

swacad:
Joe:  Can you tell me what I have to do in rebase process to set this up properly to merge?  This is my first big open source project I've worked on, so forgive my ignorance.

llllllllll:
No problem, to rebase against master, you need to pull the newest version of master from the zipline upstream repo , then you run rebase and make any needed changes.

```bash
$ git checkout master
$ git pull upstream master
$ git checkout -
$ git rebase master
```

This will tell you that there are confilcts in some files. These conflicts are because people added other indicators to the same files. You should leave both your indicator and the new indicator on master by removing all of the lines with `>>>>>...`, `=====....`, and `<<<<<.....`. Hopefully this was clear, I can also rebase and push for you if you want.

swacad:
Joe:  Thanks for the explanation.  I will give that a try this week and I'll let you know if I have further questions.  Part of the reason why I'm doing this is to just learn the process of the contributing to big open source projects.

swacad:
@llllllllll Thanks for the help.  I've just done the rebase and pushed the new changes with fully vectorized code.  Runs much faster in my testing!

swacad:
@llllllllll My git rebase seems to keep failing on me. After I run the commands you gave me and run "git rebase --continue" I get a message saying that "no changes - did you forget to use 'git add'?"  Any ideas what this might be?

llllllllll:
that means there are no more changes in the commit because all the differences are on master, you can you `git rebase--skip` to skip that commit

swacad:
@llllllllll Well I've tried multiple times doing the steps that you've told me to do.  The merge keeps failing the tests.  Can you go ahead and push this for me?  I have no idea why the merge keeps failing.

llllllllll:
closing in favor of #1325

PR #1260, opened 2016-06-04
ENH: add groupby to rank, top, and bottom

andportnoy:
no tests yet, making request to let the CI services run


ssanderson:
@andportnoy do you think you'll have a chance to get back to this in the near future?  The code here looks mostly correct to me: I think all that's missing is a set of tests for grouped ranks.

If you don't think you'll be able to get to this soon (which is totally fine!), I'll probably pull this into a new branch, add some tests, and merge.

andportnoy:
@ssanderson sure, I'll work on this sometime this weekend.

andportnoy:
Still working on this, just in case.

andportnoy:
@ssanderson 
I apologize for the git madness. 

There are only two commits really: 
- the first fixes the missing handling of the ``ascending`` parameter
- the second adds 2 tests for ascending and descending grouped ranking (float64 dtype only)

zipline/pipeline/factors/factor.py
```diff
@@ -632,7 +632,7 @@ def rank(self,
 
         else:
             def rank(row):
-                return rankdata(row, method=method)
+                return rankdata(row if ascending else -row, method=method)
```
andportnoy: Is it ok that ``ascending`` is not a local for ``rankdata``?

zipline/pipeline/factors/factor.py
```diff
@@ -619,7 +626,21 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
...
         :class:`zipline.pipeline.factors.factor.Rank`
         """
-        return Rank(self, method=method, ascending=ascending, mask=mask)
+
+        if groupby is NotSpecified:
+            return Rank(self, method=method, ascending=ascending, mask=mask)
+
+        else:
+            def rank(row):
+                return rankdata(row if ascending else -row, method=method)
```
ssanderson: Noticed your question re: this not being a local got clobbered.

From a correctness perspective, it shouldn't matter, and from a performance perspective, the difference between loading a local variable and loading a closure variable is small (compared to loading a global, for example, which is a dict lookup).  The main cost of this strategy is that we're branching on every row even though we only need to do the check once.  If we want to be really tight here, I would probably write this as:

```
from functools import partial

if groupby is NotSpecified:
    return Rank(...)
elif ascending:
    rank = partial(rankdata, method=method)
else:
    def rank(row):
        return rankdata(-row, method=method)    
```

Another *possible* speedup would be to pass `-self` as `factor` instead of negating the row each time.  Naively, that should consume less CPU, since we can do the negation vectorized on a larger array, but that also means that we'll temporarily have a full copy of the (negated) array in memory at, which might actually degrade performance by requiring more allocations.  I'd have to profile to see if one were better than the other.

andportnoy:
And for zipline in general, how do you usually resolve this tradeoff problem (CPU vs. memory)?

ssanderson:
@andportnoy sorry for the long silence, I've been on vacation for a few weeks.  What's the current status of this?  I'd like to get this in this week if possible.  I'm happy to finish it up if there's still work that you think needs doing.

ssanderson:
@andportnoy I rebased what was here and finished it off in https://github.com/quantopian/zipline/pull/1349.

PR #1263, opened 2016-06-05
Ichimoku Cloud

llllllllll:
The more interesting feature here is providing default arguments for params.


ssanderson:
@llllllllll I don't really know how to judge the Ichimoku stuff here, but :+1: on the default params API/impl.

PR #1264, opened 2016-06-06
Speed up calculation of all trading minutes

jbredeche:
Take advantage of the fact that we've already calculated all the opens and closes.


jbredeche:
Hold off on review, looking into whether I need to make some more changes.

yankees714:
Looks good, we just need to remove the reference to `trading_minute_window` in `ExchangeCalendarTestBase.test_minute_window`

tests/test_trading_schedule.py
```diff
@@ -56,3 +59,51 @@ def test_nyse_execution_time(self):
                 "Mismatch between schedule: %s and calendar: %s at time %s"
                 % (cal_open, sched_exec, dt)
             )
+
+    def test_execution_minute_window_forward(self):
```
ssanderson: :heart:

PR #1265, opened 2016-06-06
MAINT: Removes the set_first_trading_day method of DataPortal

yankees714:
Since the first trading day is now passed directly to the `DataPortal` on init, there's no need for a method that does this. Moves all the additional logic/assignments into the init. Also corrects an issue where we would never create certain attributes if `self._first_trading_day` was `None`.

Adds the ability to specify the first trading day for a data portal in a test case when using the `WithDataPortal` fixture.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -27,7 +27,8 @@ from zipline.testing import (
 from zipline.testing.fixtures import (
     WithBcolzMinuteBarReader,
     WithDataPortal,
-    ZiplineTestCase
+    ZiplineTestCase,
+    alias,
 )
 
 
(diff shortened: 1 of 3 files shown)


tests/test_history.py
```diff
@@ -449,6 +449,11 @@ class MinuteEquityHistoryTestCase(WithHistory, ZiplineTestCase):
     BCOLZ_DAILY_BAR_SOURCE_FROM_MINUTE = True
 
     @classmethod
+    def init_class_fixtures(cls):
```
ssanderson: This could just be `DATA_PORTAL_FIRST_TRADING_DAY = alias('TRADING_START_DT')`.

ssanderson: (where alias is `zipline.testing.fixtures.alias`)

yankees714: Ah cool, made that change, thanks!

PR #1267, opened 2016-06-07
Pipeline single column capabilities

dmichalowicz:
- Adds a new `Slice` class, which is a term that returns a single column of a Factor/Filter/Classifer. Slices can be created by indexing into a Factor/Filter/Classifer with the desired asset column, e.g. `Returns(...)[Equity(24)]`

- Allow custom factors to compute single-column outputs by specifying `ndim = 1`; i.e. `out` is a single column (and at every iteration of `compute` is a singleton). However, "1D" output factors cannot currently be added as pipeline columns.


zipline/pipeline/slice.py
```diff
@@ -0,0 +1,55 @@
...
+            ndim=1,
+        )
+
+    def __repr__(self):
+        return "{type}({parent_term}, column={asset})".format(
+            type=type(self).__name__,
+            parent_term=type(self.inputs[0]).__name__,
+            asset=self._asset,
+        )
+
```
ssanderson: I think that rather than using a Mixin here, we might just want to have Slice be a direct subclass of `ComputableTerm`.

Rationale:

- If we're using it as a mixin, then we have to have separate `FactorSlice`, `FilterSlice` and `ClassifierSlice` types.  If it's just a computable term, then we only have one class.
- Using slice as a mixin means that `FactorSlice` and friends inherit all the methods from `Factor`, but most of them don't work on Slice.

zipline/pipeline/slice.py
```diff
@@ -0,0 +1,55 @@
...
+    should construct instances via indexing, e.g. `MyFactor()[Asset(24)]`.
+    """
+
+    def __new__(cls, term, asset):
+        return super(Slice, cls).__new__(
+            cls,
+            asset=asset,
+            inputs=[term],
+            window_length=0,
+            mask=term.mask,
```
ssanderson: I wonder if we should do `term.mask[asset]` here to keep the mask dims in sync with the term dims?

tests/pipeline/test_slice.py
```diff
@@ -0,0 +1,322 @@
...
+from zipline.pipeline.loaders.frame import DataFrameLoader
+from zipline.pipeline.slice import Slice
+from zipline.testing import check_arrays, parameter_space
+from zipline.testing.fixtures import WithTradingEnvironment, ZiplineTestCase
+
+from .test_engine import (
+    AssetID,
+    AssetIDPlusDay,
+    OpenPrice,
+)
```
dmichalowicz: @ssanderson Is it bad practice to import from another test file?

ssanderson: Yes.

On Mon, Jun 13, 2016 at 3:03 PM, David Michalowicz <notifications@github.com

zipline/pipeline/term.py
```diff
@@ -500,6 +525,64 @@ def __repr__(self):
...
+    def _static_identity(cls, asset, *args, **kwargs):
+        return (super(Slice, cls)._static_identity(*args, **kwargs), asset)
+
+    def _compute(self, windows, dates, assets, mask):
+        asset = self._asset
+        asset_column = searchsorted(assets.values, asset.sid)
+        if assets[asset_column] != asset.sid:
+            raise NonExistentAssetInTimeFrame(
+                asset=asset, start_date=dates[0], end_date=dates[-1],
+            )
```
dmichalowicz: I added a check here to make sure we don't quietly return the wrong column of data for a non existent asset

zipline/pipeline/term.py
```diff
@@ -500,6 +525,64 @@ def __repr__(self):
...
+    Users should rarely construct instances of `Slice` directly. Instead, they
+    should construct instances via indexing, e.g. `MyFactor()[Asset(24)]`.
+    """
+    def __new__(cls, term, asset):
+        return super(Slice, cls).__new__(
+            cls,
+            asset=asset,
+            inputs=[term],
+            window_length=0,
+            mask=term.mask[asset] if term.mask else AssetExists(),
```
dmichalowicz: Since we are taking the slice of our term's mask, this triggers a recursive call to `__new__`, so the `else` is the base case of when `term` is `AssetExists` and `term.mask` is `None`. Is this the correct way to handle this?

docs/source/whatsnew/1.0.1.txt
```diff
@@ -17,6 +17,10 @@ Enhancements
 - Added support for non-float columns to Blaze-backed Pipeline
   datasets (:issue:`1201`).
 
+- Added :class:`zipline.pipeline.slice.Slice`, a new pipeline term designed to
```
ssanderson: :+1:

tests/pipeline/test_engine.py
```diff
@@ -1258,19 +1214,39 @@ def test_correlation_factors(self, returns_length, correlation_length):
         `RollingSpearmanOfReturns`.
         """
         my_asset_column = 0
-        start_date_index = 6
-        end_date_index = 10
+        start_date_index = 14
```
ssanderson: Why did this change?

tests/pipeline/test_engine.py
```diff
@@ -1424,23 +1429,22 @@ def test_regression_of_returns_factor(self,
...
 
     def test_correlation_and_regression_with_bad_asset(self):
         """
         Test that `RollingPearsonOfReturns`, `RollingSpearmanOfReturns` and
         `RollingLinearRegressionOfReturns` raise the proper exception when
         given a nonexistent target asset.
         """
-        start_date_index = 6
-        end_date_index = 10
+        start_date_index = 14
```
ssanderson: Same question.  I don't see what else changed in these tests to necessitate a different date range.

tests/pipeline/test_slice.py
```diff
@@ -0,0 +1,304 @@
...
+            cls.dates,
+            cls.asset_finder,
+        )
+
+    def test_slice(self):
+        """
+        Test that slices can be created by indexing into a term, and that they
+        have the correct shape when used as inputs.
+        """
+        my_asset_column = 0
```
ssanderson: Might be worth parameterizing this over `my_asset_column` and `window_length`.

tests/pipeline/test_slice.py
```diff
@@ -0,0 +1,304 @@
...
+        )
+
+    def test_adding_slice_column(self):
+        """
+        Test that slices cannot be added as a pipeline column.
+        """
+        my_asset = self.asset_finder.retrieve_asset(self.sids[0])
+        open_slice = OpenPrice()[my_asset]
+
+        with self.assertRaises(UnsupportedPipelineColumn):
```
ssanderson: I might call this `UnsupportedPipelineTerm` rather than column.  I'm not sure how we want to refer to slices.

ssanderson: (This also matches `NonSliceableTerm`.)

tests/pipeline/test_slice.py
```diff
@@ -0,0 +1,304 @@
...
+            )
+
+        with self.assertRaises(NonWindowSafeInput):
+            returns.rolling_pearsonr(
+                target_slice=my_factor_slice,
+                correlation_length=10,
+            )
+
+    @parameter_space(returns_length=[2, 3], correlation_length=[3, 4])
+    def test_factor_correlation_methods(self,
```
ssanderson: These tests aren't super representative since they're being tested against `arange` data, which means the correlation coefficients should be 1 for everything I think.  This whole suite might be a good use-case for `WithSeededRandomPipelineEngine`

tests/pipeline/test_slice.py
```diff
@@ -0,0 +1,304 @@
...
+
+        with self.assertRaises(NonWindowSafeInput):
+            self.engine.run_pipeline(
+                Pipeline(columns={'uses_sliced_input': UsesSlicedInput()}),
+                self.start_date,
+                self.end_date,
+            )
+
+        # Slices of Returns *are* window safe.
+        returns = Returns(window_length=2)
```
ssanderson: For the purpose of this test, I think we care more about the fact that the term here is window_safe than that it's actually `Returns`.  I would probably make a new `MySafeFactor` here with window_safe=True and test that.

zipline/pipeline/engine.py
```diff
@@ -361,7 +361,6 @@ def compute_chunk(self, graph, dates, assets, initial_workspace):
                     assets,
                     mask,
                 )
-                assert(workspace[term].shape == mask.shape)
```
ssanderson: We probably should still make assertions here.  How about this?

```
if term.ndim == 2:
    <original assertion>
else:
    assert workspace[term].shape == (mask.shape[0], 1)
```
```

File: zipline/pipeline/factors/factor.py [hunk 14]
```diff
@@ -621,6 +627,120 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
...
+            A Filter describing which assets should have their correlation with
+            the target slice computed each day.
+
+        Returns
+        -------
+        correlations : zipline.pipeline.factors.RollingPearson
+            A new Factor that will compute correlations between `target_slice`
+            and the columns of `self`.
+        """
+        return RollingPearson(
```
ssanderson: Can we make assertions about the dtypes of `self` and `target_slice` here?  (Or are those assertions made in `RollingPearson`?)

zipline/pipeline/factors/factor.py
```diff
@@ -621,6 +627,120 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
...
+        target_slice=Slice,
+        correlation_length=int,
+        mask=(Filter, NotSpecifiedType),
+    )
+    def rolling_spearmanr(self,
+                          target_slice,
+                          correlation_length,
+                          mask=NotSpecified):
+        """
+        Construct a new Factor that computes rolling spearman correlation
```
ssanderson: I would probably add "spearman **rank** correlation".

zipline/pipeline/factors/factor.py
```diff
@@ -621,6 +627,120 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
...
+            A Filter describing which assets should be regressed with the
+            target slice each day.
+
+        Returns
+        -------
+        regressions : zipline.pipeline.factors.RollingLinearRegression
+            A new Factor that will compute linear regressions of `target_slice`
+            against the columns of `self`.
+        """
+        return RollingLinearRegression(
```
ssanderson: Same question as above re: type checking.

zipline/pipeline/factors/factor.py
```diff
@@ -621,6 +627,120 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
...
+            correlation_length=correlation_length,
+            mask=mask,
+        )
+
+    @expect_types(
+        target_slice=Slice,
+        correlation_length=int,
+        mask=(Filter, NotSpecifiedType),
+    )
+    def rolling_spearmanr(self,
```
ssanderson: I'm not sure we need the `rolling_` prefix on these.  cc @twiecki @jmccorriston @CaptainKanuk for naming thoughts.

zipline/pipeline/factors/factor.py
```diff
@@ -621,6 +627,120 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
         """
         return Rank(self, method=method, ascending=ascending, mask=mask)
 
+    @expect_types(
+        target_slice=Slice,
+        correlation_length=int,
+        mask=(Filter, NotSpecifiedType),
+    )
+    def rolling_pearsonr(self,
+                         target_slice,
```
ssanderson: `target_slice` feels a little verbose here.  Do you think `target` is too vague?

zipline/pipeline/factors/statistical.py
```diff
@@ -0,0 +1,270 @@
+
```
ssanderson: We should probably move the base correlation and regression factors here as well.  (I think that means you'll need to do a lazy import in the factor methods that emit those types, which is okay here.)

zipline/pipeline/term.py
ssanderson: I think this should just be `term.mask`.

zipline/pipeline/term.py
```diff
@@ -500,6 +525,64 @@ def __repr__(self):
         )
 
 
+class Slice(ComputableTerm):
```
ssanderson: This probably wants `SingleInputMixin`.

zipline/pipeline/term.py
```diff
@@ -500,6 +525,64 @@ def __repr__(self):
...
+        asset = self._asset
+        asset_column = searchsorted(assets.values, asset.sid)
+        if assets[asset_column] != asset.sid:
+            raise NonExistentAssetInTimeFrame(
+                asset=asset, start_date=dates[0], end_date=dates[-1],
+            )
+
+        # Return a 2D array with one column rather than a 1D array of the
+        # column.
+        col = windows[0][:, asset_column]
```
ssanderson: I think this is equivalent to `windows[0][:, [asset_column]]`.

zipline/pipeline/term.py
```diff
@@ -500,6 +525,64 @@ def __repr__(self):
...
+        return col[:, None]
```
ssanderson: We can do this in one step with `windows[0][:, [asset_column]]`.

zipline/testing/core.py
```diff
@@ -1156,6 +1158,72 @@ def create_empty_splits_mergers_frame():
     )
 
 
+def make_alternating_boolean_array(shape, first=True):
```
ssanderson: I might call this `first_value` for clarity.  Seeing `first=True` makes me think it's a flag.

zipline/testing/core.py
```diff
@@ -1381,3 +1449,39 @@ def patched_read_csv(filepath_or_buffer, *args, **kwargs):
...
+
+class AssetID(CustomFactor):
+    """
+    CustomFactor that returns the AssetID of each asset.
+
+    Useful for providing a Factor that produces a different value for each
+    asset.
+    """
+    window_length = 1
+    # HACK: We currently decide whether to load or compute a Term based on the
```
ssanderson: This isn't true anymore.  We should be able to do `inputs=()` and remove `close` from the compute signature now.

zipline/testing/core.py
```diff
@@ -1381,3 +1449,39 @@ def patched_read_csv(filepath_or_buffer, *args, **kwargs):
...
+    # length of its inputs. This means we have to provide a dummy input.
+    inputs = [USEquityPricing.close]
+
+    def compute(self, today, assets, out, close):
+        out[:] = assets
+
+
+class AssetIDPlusDay(CustomFactor):
```
ssanderson: Same deal here as for `AssetID`.

zipline/pipeline/factors/factor.py
CaptainKanuk: I'm not convinced the prefix is necessary, pipeline as a whole is rolling so I guess it's implicit that any statistical metric would also be rolling. I would prefer very helpful docstrings on the objects.

tests/pipeline/test_engine.py
dmichalowicz: I changed this so that I could run the pipeline over a contiguous set of 5 days (i.e. with no weekend) while also having plenty of days to look back on. It was previously starting at index 6 only because that was the first day that gave enough days to look back on (specifically, you need at least as many days as the window_length of `returns` plus the correlation/regression length). I made this change above as well for consistency and it also simplified things when computing the expected masks to have a contiguous 5 days.

zipline/errors.py
```diff
@@ -666,3 +666,21 @@ class ScheduleFunctionWithoutCalendar(ZiplineError):
         "To use schedule_function, the TradingAlgorithm must be running on an "
         "ExchangeTradingSchedule, rather than {schedule}."
     )
+
+
+class UnsupportedPipelineColumn(ZiplineError):
```
dmichalowicz: I think I like this slightly better than `UnsupportedPipelineTerm`

zipline/pipeline/factors/factor.py
dmichalowicz: No I suppose `target` would be ok, since the `target_factor` here is just `self`, so there's no ambiguity as to what `target` is.

zipline/pipeline/factors/factor.py
dmichalowicz: So something to make sure we can compute correlations, like this?
```python
assert self.dtype in {float64_dtype, int64_dtype}
assert target.dtype in {float64_dtype, int64_dtype}
```
As in, we can't do it on datetimes

ssanderson: I wouldn't actually use `assert` here, but we should fail at construction time on bad input.

tests/pipeline/test_slice.py
```diff
@@ -0,0 +1,413 @@
...
+    parameter_space,
+)
+from zipline.testing.fixtures import (
+    WithSeededRandomPipelineEngine,
+    ZiplineTestCase,
+)
+from zipline.utils.numpy_utils import datetime64ns_dtype
+
+
+class SliceTestCase(WithSeededRandomPipelineEngine, ZiplineTestCase):
```
dmichalowicz: I changed this to use `WithSeededRandomPipelineEngine`

zipline/pipeline/factors/statistical.py
dmichalowicz: Done

zipline/pipeline/factors/statistical.py
```diff
@@ -0,0 +1,435 @@
...
+from .technical import Returns
+
+
+ALLOWED_DTYPES = (float64_dtype, int64_dtype)
+
+
+class _RollingCorrelation(CustomFactor, SingleInputMixin):
+
+    @expect_dtypes(target_factor=ALLOWED_DTYPES, target_slice=ALLOWED_DTYPES)
```
dmichalowicz: Am now doing type checking here

zipline/pipeline/factors/statistical.py
```diff
@@ -0,0 +1,435 @@
...
+    :class:`zipline.pipeline.factors.RollingLinearRegressionOfReturns`
+
+    Notes
+    -----
+    Most users should call Factor.linear_regression rather than directly
+    construct an instance of this class.
+    """
+    outputs = ['alpha', 'beta', 'r_value', 'p_value', 'stderr']
+
+    @expect_dtypes(target_factor=ALLOWED_DTYPES, target_slice=ALLOWED_DTYPES)
```
dmichalowicz: And here

tests/pipeline/test_slice.py
```diff
@@ -0,0 +1,413 @@
...
+    sids = ASSET_FINDER_EQUITY_SIDS = Int64Index([1, 2, 3])
+    START_DATE = Timestamp('2015-01-31', tz='UTC')
+    END_DATE = Timestamp('2015-03-01', tz='UTC')
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(SliceTestCase, cls).init_class_fixtures()
+
+        day = cls.trading_schedule.day
+        cls.dates = dates = date_range(
```
ssanderson: I would just have these use the calendary provided by `WithNYSETradingDays`.

zipline/pipeline/factors/factor.py
```diff
@@ -621,6 +621,193 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
         """
         return Rank(self, method=method, ascending=ascending, mask=mask)
 
+    @expect_types(
+        target=Slice, correlation_length=int, mask=(Filter, NotSpecifiedType),
```
ssanderson: We'll need to update `target` here if we add support for Factor <-> Factor correlations.  This is fine for now though.

dmichalowicz: Yup, that's already being done here: https://github.com/quantopian/zipline/pull/1291

zipline/pipeline/pipeline.py
```diff
@@ -37,6 +39,9 @@ def __init__(self, columns=None, screen=None):
 
         if columns is None:
             columns = {}
+        for term in columns.values():
```
ssanderson: Can we push this into a `validate_column` method?  (We're duplicating this logic in `add`.)

zipline/pipeline/pipeline.py
```diff
@@ -72,6 +77,9 @@ def add(self, term, name, overwrite=False):
             Whether to overwrite the existing entry if we already have a column
             named `name`.
         """
+        if term.ndim == 1:
+            raise UnsupportedPipelineOutput()
```
ssanderson: It'd be nice for this error to tell you what the term was (the name and the value).

zipline/pipeline/term.py
ssanderson: :+1: do we have test coverage for this?

zipline/pipeline/mixins.py
```diff
@@ -148,29 +152,38 @@ def _allocate_output(self, windows, shape):
...
 
         with self.ctx:
             for idx, date in enumerate(dates):
-                col_mask = mask[idx]
-                masked_out = out[idx][col_mask]
+                # Never apply a mask to 1D outputs.
+                col_mask = array([True]) if self.ndim == 1 else mask[idx]
                 masked_assets = assets[col_mask]
+                out_row = out[idx][col_mask]
+                inputs = self._format_inputs(windows, col_mask)
```
ssanderson: I'm a little nervous about how much work we're doing in this loop now.  On my machine a 1-year USEP spends about a second of actual execution time is just in `_format_inputs`.

zipline/pipeline/mixins.py
```diff
@@ -148,29 +152,38 @@ def _allocate_output(self, windows, shape):
...
+
+        shape = (len(mask), 1) if self.ndim == 1 else mask.shape
+        out = self._allocate_output(windows, shape)
 
         with self.ctx:
             for idx, date in enumerate(dates):
-                col_mask = mask[idx]
-                masked_out = out[idx][col_mask]
+                # Never apply a mask to 1D outputs.
+                col_mask = array([True]) if self.ndim == 1 else mask[idx]
```
ssanderson: I think this is only correct for the output col_mask.

If I pass a mask to an ndim=1 CustomFactor, I should still get my inputs masked.

PR #1269, opened 2016-06-09
Rewrite EventsLoader as a single class rather than a base class

ssanderson:
Still a few rough edges, but the basic functionality is all here:

Still TODO:
- Remove existing datasets modules.
- Fix a few straggling tests around error handling.
- Decide what to do with `common.py`.
- Port `next_event_indexer` and/or `previous_event_indexer` to Cython.  (This probably isn't necessary right now unless we get a major regression downstream.


zipline/pipeline/loaders/events.py
```diff
@@ -36,196 +42,128 @@
...
+        ``events`` must contain at least three columns::
+            sid : int64
+                The asset id associated with each event.
+
+            event_date : datetime64[ns]
+                The date on which the event occurred.
+
+            timestamp : datetime64[ns]
+                The date on which we learned about the event. Must be greater
+                than or equal to the frame index value.
```
mtydykov: If we require timestamp as a separate column, what is the DatetimeIndex for, and why does timestamp need to be greater than or equal to it?

zipline/pipeline/loaders/events.py
```diff
@@ -36,196 +42,128 @@
...
-        occurred along with other associated values. If the DataFrame
-        contains a "timestamp" column, that column is interpreted as the date
-        on which we learned about the event. If the DataFrames do not contain a
-         "timestamp" column, we assume we knew about the event on all prior
-         dates.  This mode is only supported if ``infer_timestamp`` is
-         explicitly passed as a truthy value.
-
-    infer_timestamps : bool, optional
-        Whether to allow omitting the "timestamp" column.
+
```
mtydykov: Need to update docs to remove `all_dates` and `dataset` and include `next_value_columns` and `previous_value_columns`.

zipline/pipeline/loaders/utils.py
```diff
@@ -2,150 +2,85 @@
...
 from zipline.utils.pandas_utils import mask_between_time
 
 
-def next_event_frame(events_by_sid,
-                     dates,
-                     missing_value,
-                     field_dtype,
-                     event_date_field_name,
-                     return_field_name):
+def is_sorted_ascending(a):
```
mtydykov: Add docstring.

tests/pipeline/test_events.py
```diff
@@ -1,299 +1,425 @@
...
-                 dataset=EventDataSet):
-        super(EventDataSetLoaderNoExpectedCols, self).__init__(
-            all_dates,
-            events_by_sid,
-            infer_timestamps=infer_timestamps,
-            dataset=dataset,
-        )
+    Events for two different sids shouldn't interact in any way, so the
+    interesting cases are determined by the possible interleavings of
+    event_date and timestamp.
```
mtydykov: It's implied, but we should probably state explicitly that the interesting cases are for combinations of a single sid's event_date and timestamp.

tests/pipeline/test_events.py
```diff
@@ -1,299 +1,425 @@
...
+
+    Fix two events with dates e1, e2 and timestamps t1 and t2.
+
+    Without loss of generality, assume that e1 < e2. (If two events have the
+    same occurrence date, the behavior of next/previous event is undefined).
```
mtydykov: Is the behavior really undefined, though? What happens now if a sid has 2 events on the same date, and what's the expected behavior?

zipline/pipeline/loaders/events.py
ssanderson: These docs are old and out of date.  The frame index is ignored.

tests/pipeline/test_events.py
ssanderson: I don't think there's a way that we can deterministically choose an event if the events have the same date.  If the timestamps are different, we could make an arbitrary choice to take the event with the newer and/or older timestamp, but if both the event date and timestamp are the same, there's no way for us to distinguish the events here.

zipline/pipeline/loaders/utils.py
```diff
@@ -2,150 +2,85 @@
...
+                       all_sids,
+                       event_dates,
+                       event_timestamps,
+                       event_sids):
     """
-    Make a DataFrame representing the simulated next known dates or values
-    for an event.
+    Construct an indexer that will convert a column of event data into a 2D
+    array containing a value for the most recent next-known event for each sid
+    on each date.
```
mtydykov: I would probably clarify that the array contains an index into the event data column ('value' makes it sound like it contains the actual value from the column). We should also include parameters/return values in the docstring.

zipline/pipeline/loaders/utils.py
```diff
@@ -2,150 +2,85 @@
...
-        ] = df[previous_return_field]
-
-    frame = pd.DataFrame(out, index=date_index, columns=sids)
-    frame.ffill(inplace=True)
-    if field_dtype == categorical_dtype:
-        frame[frame.isnull()] = missing_value
-    return frame
+    Construct an indexer that will convert a column of event data into a 2D
+    array containing a value for the most recent previously-known event for
+    each sid on each date.
```
mtydykov: Same as comment for `next_event_indexer`.

zipline/pipeline/loaders/utils.py
```diff
@@ -2,150 +2,85 @@
...
-    events_by_sid : dict[int -> pd.Series]
-        Dict mapping sids to a series of dates. Each k:v pair of the series
-        represents the date we learned of the event mapping to the date the
-        event will occur.
-    event_date_field_name : str
-        The name of the date field that marks when the event occurred.
+    Locations where no next event was known will be filled with -1.
+    """
+    assert is_sorted_ascending(event_dates), "event dates must be sorted"
+    assert len(event_sids) == len(event_dates) == len(event_timestamps), \
```
mtydykov: We should probably refactor such that we make the same assertions in `prev_event_indexer` as we do here.

zipline/pipeline/loaders/utils.py
ssanderson: > I would probably clarify that the array contains an index into the event data column ('value' makes it sound like it contains the actual value from the column). 

The interpretation in the parenthetical is the correct one here.  The sentence here is saying that we're going to construct an index array that, when applied to a column of values, produces a 2D array of the values associated with the next event for each sid at each moment in time.

I'll probably replace what's here with what I just wrote above.

zipline/testing/fixtures.py
```diff
@@ -875,8 +870,7 @@ def init_class_fixtures(cls):
         cls.adjustment_reader = SQLiteAdjustmentReader(conn)
 
 
-class WithPipelineEventDataLoader(
-        with_metaclasses((type(ZiplineTestCase), ABCMeta), WithAssetFinder)):
+class WithPipelineEventDataLoader(ZiplineTestCase, WithAssetFinder):
```
ssanderson: I think this whole fixture can be removed.

tests/pipeline/test_events.py
```diff
@@ -1,299 +1,425 @@
...
+
+    def check_previous_value_results(self, column, results):
+        """
+        Check previous value results for a single column.
+        """
+        events = self.raw_events
+        # Remove timezone info from trading days, since the outputs
+        # from pandas won't be tz_localized.
+        dates = self.trading_days.tz_localize(None)
+        for asset, asset_result in results.iterkv():
```
mtydykov: Maybe we also want to assert here that results contains all sids that we expect? Otherwise, we're just checking the correctness of assets that we get back from the results, and the test probably shouldn't pass if those are not all there.

tests/pipeline/test_events.py
```diff
@@ -1,299 +1,425 @@
...
+
+    def check_next_value_results(self, column, results):
+        """
+        Check results for a single column.
+        """
+        events = self.raw_events
+        # Remove timezone info from trading days, since the outputs
+        # from pandas won't be tz_localized.
+        dates = self.trading_days.tz_localize(None)
+        for asset, asset_result in results.iterkv():
```
mtydykov: Same comment here as for `check_previous_value_results`.

tests/pipeline/test_events.py
```diff
@@ -1,299 +1,425 @@
...
-                   **kwargs)
+        for c in EventDataSet.columns:
+            if c in self.next_value_columns:
+                self.check_next_value_results(c, results[c.name].unstack())
+            elif c in self.previous_value_columns:
+                self.check_previous_value_results(c, results[c.name].unstack())
+            else:
+                raise AssertionError("Unexpected column %s." % c)
+
+    def check_previous_value_results(self, column, results):
```
mtydykov: Feels like a lot of this logic can be refactored to use the logic in `check_previous_event_indexer`.

tests/pipeline/test_events.py
```diff
@@ -1,299 +1,425 @@
...
+                    # If we haven't seen either event, then we should have
+                    # column.missing_value.
+                    assert_equal(
+                        computed_value,
+                        column.missing_value,
+                        # Coerce from Timestamp to datetime64.
+                        allow_datetime_coercions=True,
+                    )
+
+    def check_next_value_results(self, column, results):
```
mtydykov: As with `check_previous_value_results`, feels like a lot of this can be refactored to use logic in `check_next_event_indexer`.

tests/pipeline/test_events.py
mtydykov: In that case, is there a way we can either ignore all events in cases where this happens, or come up with some kind of logic to break ties? Either way, it's probably better to handle this explicitly.

zipline/testing/fixtures.py
mtydykov: Sounds good

tests/pipeline/test_events.py
ssanderson: I'm not sure we gain much of value by trying to make them share code.  The code is different enough that a shared version would be a fair amount more complicated, which would make it harder to follow what's actually being asserted here, and I don't expect this code to change frequently or at all.

tests/pipeline/test_events.py
ssanderson: Same reply as above.  I don't think we get much benefit from sharing here, and a shared implementation would be enough more complicated that I don't think it's worth trying to mash them together.

zipline/pipeline/loaders/events.py
```diff
@@ -1,15 +1,21 @@
...
+    EVENT_DATE_FIELD_NAME,
+    SID_FIELD_NAME,
+    TS_FIELD_NAME,
+)
+from zipline.pipeline.loaders.utils import (
+    next_event_indexer,
+    previous_event_indexer,
+)
+
+WRONG_COLS_ERROR = "Expected columns {expected_columns} for but " \
```
mtydykov: We can probably get rid of all the errors defined here other than the `WRONG_COLS_ERROR`.

zipline/pipeline/loaders/events.py
```diff
@@ -36,196 +42,135 @@
...
+            sid : int64
+                The asset id associated with each event.
+
+            event_date : datetime64[ns]
+                The date on which the event occurred.
+
+            timestamp : datetime64[ns]
+                The date on which we learned about the event.
+
+    next_value_columns : dict[BoundColumn -> str]
```
mtydykov: Also need an entry here for `previous_value_columns`.

zipline/pipeline/loaders/events.py
```diff
@@ -36,196 +42,135 @@
...
-            return getattr(self, "%s_loader" % column.name)
-        raise ValueError("Don't know how to load column '%s'." % column)
-
-    def load_adjusted_array(self, columns, dates, assets, mask):
-        return merge(
-            self.get_loader(column).load_adjusted_array(
-                [column], dates, assets, mask
+                 events,
+                 next_value_columns,
+                 previous_value_columns):
```
mtydykov: Maybe we should assert on construction of `EventsLoader` that the columns in  `next_value_columns` and `previous_value_columns` are contained in  `events` rather than waiting to break when we try to load the adjusted array.

tests/pipeline/test_events.py
ssanderson: The only thing we can use to deterministically break ties is `timestamp`, but if we have e1 == e2 and t1 == t2, then there's literally nothing else for us to use.

zipline/pipeline/loaders/events.py
ssanderson: We'll need a check at load time no matter what to give a useful error if the engine passes us columns we don't know how to load, but we could do a pre-emptive check on our events in the constructor to fail faster.

zipline/pipeline/loaders/utils.py
```diff
@@ -2,150 +2,91 @@
...
 
 
-    See Also
-    --------
-    previous_date_frame
+def next_event_indexer(all_dates,
+                       all_sids,
+                       event_dates,
+                       event_timestamps,
+                       event_sids):
```
mtydykov: Need to add parameters/return values to docstring for completeness.

zipline/pipeline/loaders/utils.py
```diff
@@ -2,150 +2,91 @@
...
+        out[start_ix:end_ix, sid_ixs[i]] = i
+
+    return out
+
+
+def previous_event_indexer(all_dates,
+                           all_sids,
+                           event_dates,
+                           event_timestamps,
+                           event_sids):
```
mtydykov: Need to add parameters/return values to docstring for completeness.

mtydykov:
Need to update some docstrings; otherwise, looks good.

PR #1275, opened 2016-06-13
Schedule function rules refactoring

lianga888:
- refactor the week rules to eliminate unnecessary type coercion of dates
- `next_scheduled_day` and `previous_scheduled_day` will raise `NoFurtherDataError` if there is no such day available in the calendar, instead of returning None. This is what `add_scheduled_days` does when `|n| > 1`, and make the behavior of the function consistent across all values of `n`

### DIFF
--- a/zipline/utils/calendars/calendar_helpers.py
+++ b/zipline/utils/calendars/calendar_helpers.py
@@ -73,7 +73,7 @@ def next_scheduled_day(date, last_trading_day, is_scheduled_day_hook):
         dt += delta
         if is_scheduled_day_hook(dt):
             return dt
-    return None
+    raise NoFurtherDataError(msg='Cannot find next day after %s' % date)
 
 
 def previous_scheduled_day(date, first_trading_day, is_scheduled_day_hook):
@@ -97,7 +97,7 @@ def previous_scheduled_day(date, first_trading_day, is_scheduled_day_hook):
         dt += delta
         if is_scheduled_day_hook(dt):
             return dt
-    return None
+    raise NoFurtherDataError(msg='Cannot find previous day before %s' % date)
 
 
 def next_open_and_close(date, open_and_close_hook,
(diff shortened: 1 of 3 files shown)


zipline/utils/calendars/calendar_helpers.py
```diff
@@ -73,7 +73,7 @@ def next_scheduled_day(date, last_trading_day, is_scheduled_day_hook):
         dt += delta
         if is_scheduled_day_hook(dt):
             return dt
-    return None
+    raise NoFurtherDataError(msg='Cannot find next day after %s' % date)
```
richafrank: Are there any consumers besides the scheduled events that need updating to handle this error?

tests/utils/test_events.py
```diff
@@ -360,7 +360,7 @@ def test_NthTradingDayOfWeek(self, n):
     @subtest(param_range(MAX_WEEK_RANGE), 'n')
     def test_NDaysBeforeLastTradingDayOfWeek(self, n):
         cal = get_calendar('NYSE')
-        rule = NDaysBeforeLastTradingDayOfWeek(n)
+        rule = NDaysBeforeLastTradingDayOfWeek(n, True)
```
richafrank: I think `invert` is really tied to the particular class being instantiated here.  What do you think about moving the decision to invert into each subclass's `__init__`?

richafrank:
Nice cleanup - I just had a question and a suggestion.

tests/utils/test_events.py
lianga888: Yeah, that makes sense. But you had suggested that the subclasses should share an `__init__`

zipline/utils/calendars/calendar_helpers.py
lianga888: I found one that needed to have the error handled, because there is a case where it might expect `None`. There's a bunch of other ones that would crash anyway if `None` was returned. Those don't need to be handled, correct?

richafrank: Yea, that sounds right.  Unless the particular kind of crash was important.  But I'd think something like an `AttributeError` on `None` is *not* important.

PR #1276, opened 2016-06-13
ENH: add ffill checkpointing to blaze core loader

llllllllll:
Adds an optional argument to `from_blaze` for users to provide a ffill checkpoint table.

### DIFF
--- a/docs/source/whatsnew/1.0.2.txt
+++ b/docs/source/whatsnew/1.0.2.txt
@@ -1,0 +1,23 @@
+Release 1.0.2
+-------------
+
+:Release: 1.0.2
+:Date: TBD
+
+Enhancements
+~~~~~~~~~~~~
+
+- Adds forward fill checkpoint tables for the blaze core loader. This allow the
+  loader to more efficiently forward fill the data by capping the lower date it
+  must search for when querying data. The checkpoints should have novel deltas
+  applied (:issue:`1276`).
+
+Bug Fixes
+~~~~~~~~~
+
+None
+
+Documentation
+~~~~~~~~~~~~~
+
+None
--- a/zipline/pipeline/loaders/blaze/core.py
+++ b/zipline/pipeline/loaders/blaze/core.py
@@ -1,4 +1,5 @@
-"""Blaze integration with the Pipeline API.
+"""
+Blaze integration with the Pipeline API.
 
 For an overview of the blaze project, see blaze.pydata.org
 
@@ -80,6 +81,18 @@ actually -1 and that we learned on the fifth that the value on the third was
 actually 3. By pulling our data into these two tables and not silently updating
 our original table we can run our pipelines using the information we would
 have had on that day, and we can prevent lookahead bias in the pipelines.
+
+
+Another optional expression that may be provided is ``checkpoints``. The
+``checkpoints`` expression is used when doing a forward fill query to cap the
+lower date that must be searched. This expression has the same shape as the
+``baseline`` and ``deltas`` expressions but should be downsampled with novel
+deltas applied. For example, imagine we had one data point per asset per day
+for some dataset. We could dramatically speed up our queries by pre populating
+a downsampled version which has the most recently known value at the start of
+each month. Then, when we query, we only must look back at most one month
+before the start of the pipeline query to provide enough data to forward fill
+correctly.
 
 Conversion from Blaze to the Pipeline API
 -----------------------------------------
--- a/zipline/pipeline/loaders/blaze/core.py
+++ b/zipline/pipeline/loaders/blaze/core.py
@@ -140,7 +140,6 @@ from __future__ import division, absolute_import
 from abc import ABCMeta, abstractproperty
 from collections import namedtuple, defaultdict
 from copy import copy
-import datetime
 from functools import partial
 from itertools import count
 import warnings
@@ -240,7 +239,7 @@ class ExprData(namedtuple('ExprData', 'expr deltas checkpoints odo_kwargs')):
         return super(ExprData, cls).__repr__(cls(
             str(self.expr),
             str(self.deltas),
-            str(self.checkpoint),
+            str(self.checkpoints),
             self.odo_kwargs,
         ))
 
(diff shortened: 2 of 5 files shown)


docs/source/whatsnew/1.0.2.txt
```diff
@@ -0,0 +1,23 @@
+Release 1.0.2
+-------------
+
+:Release: 1.0.2
+:Date: TBD
+
+Enhancements
```
richafrank: `Performance`?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -548,13 +580,28 @@ def from_blaze(expr,
...
+        deltas,
+        no_deltas_rule,
+    )
+    checkpoints = _get_metadata(
+        'checkpoints',
+        expr,
+        checkpoints,
+        no_checkpoints_rule,
+    )
+    if {deltas, checkpoints} != {None}:
```
richafrank: Could you just explain this check?

llllllllll: oh, this was not the full check that I wanted, I need to fix this. I think it should be: `if 'auto' in (deltas, checkpoints)`

docs/source/whatsnew/1.0.2.txt
llllllllll: eh, I think it is more of a feature because you will need to do work to take advantage of it.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -411,62 +419,90 @@ def _check_datetime_field(name, measure):
...
+        ``no_metadata_rule``.
+    no_metadata_rule : {'warn', 'raise', 'ignore'}
+        How to handle the case where the metadata_expr='auto' but no expr
+        could be found.
 
     Returns
     -------
-    deltas : Expr or None
+    metadata : Expr or None
         The deltas table to use.
```
richafrank: This needs an update to be deltas or checkpoints.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -411,62 +419,90 @@ def _check_datetime_field(name, measure):
...
     """
-    if isinstance(deltas, bz.Expr) or deltas != 'auto':
-        return deltas
+    if isinstance(metadata_expr, bz.Expr) or metadata_expr != 'auto':
```
richafrank: I find this a little bit confusing.  I *think* it's equivalent to `if isinstance(metadata_expr, bz.Expr) or metadata_expr is None:`, which is a bit clearer to me.  What do you think?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -411,62 +419,90 @@ def _check_datetime_field(name, measure):
...
+
+    Returns
+    -------
+    transformed : Expr or None
+        The transformed expression or None if ``expr`` is None.
+    """
+    return (
+        None
+        if expr is None else
+        bz.transform(expr, **{TS_FIELD_NAME: expr[AD_FIELD_NAME]})
```
richafrank: Is this a mutation or a new construction?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -925,17 +1002,39 @@ def collect_expr(e):
...
+                predicate &= e[TS_FIELD_NAME] >= lower
+
+            return odo(e[predicate][colnames], pd.DataFrame, **odo_kwargs)
+
+        if checkpoints is not None:
+            ts = checkpoints[TS_FIELD_NAME]
+            checkpoints_ts = odo(ts[ts <= lower_dt].max(), pd.Timestamp)
+            if pd.isnull(checkpoints_ts):
+                materialized_checkpoints = pd.DataFrame(columns=colnames)
+                lower = lower_dt
```
richafrank: In this case, will the `materialized_expr` be missing forward-filled values at its lower bound when we `collect_expr(expr, lower)` below?

zipline/pipeline/loaders/blaze/core.py
llllllllll: All blaze operations are pure

zipline/pipeline/loaders/blaze/core.py
llllllllll: Ah, I think this case needs to be `lower = None` because this means we have no checkpoints so we should just query everything.

zipline/pipeline/loaders/blaze/core.py
richafrank: Ok great.

zipline/pipeline/loaders/blaze/core.py
llllllllll: I will add a test for this case.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -925,17 +1003,43 @@ def collect_expr(e):
...
+            lower = None
+
+        materialized_expr = collect_expr(expr, lower)
+        if materialized_checkpoints is not None:
+            materialized_expr = pd.concat(
+                (
+                    materialized_checkpoints,
+                    materialized_expr,
+                ),
+                ignore_index=True,
```
richafrank: What's the thought - are the indices not consistent here?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -411,62 +419,90 @@ def _check_datetime_field(name, measure):
...
+    no_metadata_rule : {'warn', 'raise', 'ignore'}
+        How to handle the case where the metadata_expr='auto' but no expr
+        could be found.
 
     Returns
     -------
-    deltas : Expr or None
-        The deltas table to use.
+    metadata : Expr or None
+        The the deltas or metadata table to use.
```
richafrank: typo - extra "the"

zipline/pipeline/loaders/blaze/core.py
llllllllll: they don't matter, but I would rather they be a simple index instead of a non-unique index.

richafrank: Oh, why's that?

llllllllll: Some operations are not supported by structures if the index isn't unique. Also, in later versions of pandas a simple index can be a `RangeIndex` object which will be much more memory efficient.

richafrank: Ok thanks.

PR #1279, opened 2016-06-15
MAINT: Creates deprecation warning for tradingcalendar imports

yankees714:
The `tradingcalendar` module was been replaced by the new exchange calendars and trading schedules. Sets up an import hook in `algorithm.py` to issue a ZiplineDeprecationWarning on `tradingcalendar` imports.

### DIFF
--- a/zipline/utils/tradingcalendar.py
+++ b/zipline/utils/tradingcalendar.py
@@ -14,14 +14,24 @@
 # limitations under the License.
 import pandas as pd
 import pytz
+import warnings
 
 from datetime import datetime
 from dateutil import rrule
 from functools import partial
 
+from zipline.zipline_warnings import ZiplineDeprecationWarning
+
 # IMPORTANT: This module is deprecated and is only here for temporary backwards
 # compatibility.  Look at the `zipline.utils.calendars.trading_schedule`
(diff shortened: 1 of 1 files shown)


llllllllll:
Can't we just put a `warnings.warn` at module scope with a stacklevel of 1?

yankees714:
@llllllllll good call, should have realized that, updated to do that instead.

zipline/utils/tradingcalendar.py
```diff
@@ -14,14 +14,22 @@
...
 from dateutil import rrule
 from functools import partial
 
+from zipline.zipline_warnings import ZiplineDeprecationWarning
+
 # IMPORTANT: This module is deprecated and is only here for temporary backwards
 # compatibility.  Look at the `zipline.utils.calendars.trading_schedule`
 # module, as well as the calendar definitions in `zipline.utils.calendars`.
+warnings.warn(
+    "The `tradingcalendar` module is deprecated.",
```
llllllllll: deprecation warnings should always tell the user what the new feature is that deprecates the old feature. Can we add something like, `see 'zipline.utils.calendar' for the trading calendar API`

llllllllll:
lgtm, just squash it all down when you merge

PR #1280, opened 2016-06-15
BUG: Fail fast on invalid pipeline columns

dmichalowicz:
If you try to run a pipeline with a column of `USEquityPricing.close` (or something similar) and you meant to use `USEquityPricing.close.latest`, it errors in a non-helpful/obvious way.

### DIFF
--- a/zipline/pipeline/pipeline.py
+++ b/zipline/pipeline/pipeline.py
@@ -1,6 +1,6 @@
 from zipline.utils.input_validation import expect_types, optional
 
-from .term import Term, AssetExists
+from .term import AssetExists, ComputableTerm
 from .filters import Filter
 from .graph import TermGraph
 
@@ -37,6 +37,12 @@ class Pipeline(object):
 
         if columns is None:
             columns = {}
+        for term in columns.values():
+            if not isinstance(term, ComputableTerm):
+                raise TypeError(
+                    '"{term}" is not a valid pipeline column. Did you mean '
+                    'to add ".latest"?'.format(term=term)
+                )
         self._columns = columns
         self._screen = screen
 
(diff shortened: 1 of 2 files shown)


zipline/pipeline/pipeline.py
```diff
@@ -37,6 +37,12 @@ def __init__(self, columns=None, screen=None):
 
         if columns is None:
             columns = {}
+        for term in columns.values():
```
ssanderson: Showing the name of the column here would be nice as well.

zipline/pipeline/pipeline.py
```diff
@@ -54,7 +60,7 @@ def screen(self):
         """
         return self._screen
 
-    @expect_types(term=Term, name=str)
+    @expect_types(term=ComputableTerm, name=str)
```
ssanderson: I would try to re-use the error message above rather than using `expect_types` here.  This will spit out a message telling the user that we expected a `ComputableTerm` but got a `BoundColumn`, which won't mean anything to most users.

zipline/pipeline/pipeline.py
```diff
@@ -79,6 +87,12 @@ def add(self, term, name, overwrite=False):
             else:
                 raise KeyError("Column '{}' already exists.".format(name))
 
+        if not isinstance(term, ComputableTerm):
+            raise TypeError(
+                "{term} is not a valid pipeline column. Did you mean to "
+                "append '.latest'?".format(term=term)
+            )
```
dmichalowicz: Added the column name here but not above. In this case you don't need the name as you're only adding a single term.

ssanderson:
:+1: LGTM

Issue #1228, opened 2016-05-23
yahoo bundle causes SQLiteAdjustmentWriter error - sqlite3.OperationalError: unable to open database file


cwengc:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

* Operating System: Windows 7
* Python Version: 3.4
* Python Bitness: 64
* How did you install Zipline: conda
* Python packages: 
```
_license                  1.1                      py34_0
alabaster                 0.7.3                    py34_0
anaconda                  2.3.0                np19py34_0
argcomplete               0.8.9                    py34_0
astropy                   1.0.3                np19py34_0
babel                     1.3                      py34_0
bcolz                     0.12.1               np19py34_0
beautiful-soup            4.3.2                    py34_1
beautifulsoup4            4.3.2                     <pip>
binstar                   0.11.0                   py34_0
bitarray                  0.8.1                    py34_1
blaze                     0.8.0                     <pip>
blaze-core                0.8.0                np19py34_0
blz                       0.6.2                np19py34_1
bokeh                     0.9.0                np19py34_0
boto                      2.38.0                   py34_0
bottleneck                1.0.0                np19py34_0
cachetools                1.1.6                    py34_0
certifi                   14.05.14                 py34_0
cffi                      1.1.0                    py34_0
click                     6.6                      py34_0
clyent                    0.3.4                    py34_0
colorama                  0.3.3                    py34_0
conda                     4.0.6                    py34_0
conda-build               1.14.1                   py34_0
conda-env                 2.4.5                    py34_0
configobj                 5.0.6                    py34_0
contextlib2               0.4.0                    py34_0
cryptography              0.9.1                    py34_0
cyordereddict             0.2.2                    py34_0
cython                    0.24                     py34_0
cytoolz                   0.7.3                    py34_0
datashape                 0.4.5                np19py34_0
decorator                 4.0.9                    py34_0
docutils                  0.12                     py34_1
dynd-python               0.6.5                np19py34_0
fastcache                 1.0.2                    py34_0
flask                     0.10.1                   py34_1
greenlet                  0.4.7                    py34_0
h5py                      2.5.0                np19py34_1
hdf5                      1.8.15.1                      2
idna                      2.0                      py34_0
intervaltree              2.1.0                    py34_0
ipython                   3.2.0                    py34_0
ipython-notebook          3.2.0                    py34_0
ipython-qtconsole         3.2.0                    py34_0
[...]

PR #1282, opened 2016-06-15
TST: fix bundle test discovery

llllllllll:


tests/data/bundles/test_core.py
```diff
@@ -303,7 +303,7 @@ def _(environ,
                   output_dir):
                 _wrote_to.append(output_dir)
 
-        _wrote_to.clear()
+        _wrote_to[:] = []
```
richafrank: What was this change for?  Also, is it intentional that the default value for `_wrote_to` is mutable?

richafrank:
Looks like travis is failing with a precision issue, and appveyor is failing with #1228.

tests/data/bundles/test_core.py
llllllllll: clear was added to list in python 3, it is supposed to be mutable.

linc09:
I also find this problem when ingest yahoo at windows 7, the error msg is 
"
sqlite3.OperationalError: unable to open database file
"
then I find when run
`            if overwrite and exists(conn_or_path):
                try:
                    remove(conn_or_path)
                except OSError as e:
                    if e.errno != ENOENT:
                        raise`
the 'conn_or_path' file still exist after remove.

richafrank:
(This is in service of fixing https://github.com/quantopian/zipline/issues/1228.)

zipline/utils/cache.py
```diff
@@ -298,9 +304,25 @@ def __enter__(self):
         return self
 
     def __exit__(self, *exc_info):
+        self._tmpfile.__exit__(*exc_info)
         if exc_info[0] is None:
             self._commit()
-        self._tmpfile.__exit__(*exc_info)
+
+    @property
```
llllllllll: why are we using getters and setters here instead of just naming the attribute `tmpfile`?

zipline/data/bundles/core.py
```diff
@@ -353,14 +353,19 @@ def ingest(name,
                         asset_db_path(name, timestr, environ=environ),
                     )).path,
                 )
+                wf = stack.enter_context(working_file(
+                    adjustment_db_path(name, timestr, environ=environ),
+                ))
+                # close the file to delete it on windows
```
llllllllll: I don't think this comment is still correct.

zipline/__init__.py
```diff
@@ -12,6 +12,24 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
+
+
+if os.name == 'nt':
```
llllllllll: does this still matter?

zipline/data/us_equity_pricing.py
```diff
@@ -1021,6 +1020,12 @@ def write_dividend_data(self, dividends, stock_dividends=None):
         dividend_ratios = self.calc_dividend_ratios(dividends)
         self.write_frame('dividends', dividend_ratios)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *exc_info):
```
llllllllll: where do we enter this?

zipline/utils/cache.py
```diff
@@ -288,7 +289,12 @@ def path(self):
     def _commit(self):
         """Sync the temporary file to the final path.
         """
-        copyfile(self.name, self._final_path)
+        self._tmpfile.close()
+        try:
+            move(self._name, self._final_path)
+        except OSError as e:
+            if e != errno.ENOENT:
```
llllllllll: shouldn't this fail if the file doesn't exist?

zipline/utils/cache.py
LotannaEzenwa: No reason in particular; if the tmpfile isn't necessarily a closeable file, then I wanted to be able to handle that in the future. 

In this case, the file is actually an SQLite3 Connection object. I figured to add these just in case there was some unforeseeable object type that the working file needed to use.

llllllllll: if we need to handle that in the future then we can swap to a property. I think that for now it is more clear to just expose the attribute.

zipline/utils/cache.py
LotannaEzenwa: Yeah, didn't even realize that was still in there

zipline/utils/cache.py
llllllllll: also, do we ever need to access the tmpfile directly, we forward all attrs to the tmpfile through our `__getattribute__`, so `working_file.close` is the same as `workingfile.tmpfile.close`

zipline/data/us_equity_pricing.py
LotannaEzenwa: we don't; I added it because there's an exit.

zipline/__init__.py
LotannaEzenwa: I think so; if this is making all the tmp files go to this directory, it is necessary. The tmp files have auto-delete turned off, so if they're ending up here, it's necessary for cleanup

LotannaEzenwa:
@llllllllll  These should all pass, can you take a look at the refactored code?

tests/data/bundles/test_yahoo.py
```diff
@@ -1,11 +1,13 @@
+from __future__ import division
```
llllllllll: are we doing any division in here?

zipline/data/bundles/core.py
```diff
@@ -38,39 +37,59 @@
 
 def asset_db_path(bundle_name, timestr, environ=None):
     return pth.data_path(
-        [bundle_name, timestr, 'assets-%d.sqlite' % ASSET_DB_VERSION],
-        environ=environ,
+        list(asset_db_relative(bundle_name, timestr, environ)),
```
llllllllll: do we need to call list on these tuples? Can the input be any iterable?

zipline/utils/cache.py
```diff
@@ -350,7 +373,7 @@ def getpath(self, *path_parts):
     def _commit(self):
         """Sync the temporary directory to the final path.
         """
-        copytree(self.path, self._final_path)
+        dir_util.copy_tree(self.path, self._final_path, verbose=1)
```
llllllllll: what is this change?

zipline/utils/cache.py
```diff
@@ -334,7 +353,11 @@ def mkdir(self, *path_parts):
             The parts of the path after the working directory.
         """
         path = self.getpath(*path_parts)
-        os.mkdir(path)
+        try:
```
llllllllll: I would call this function `ensure_dir` because we have changed the semantics away from `mkdir`. We could have both, where `ensure_dir` is implemented in terms of `mkdir`. What do you think?

zipline/utils/cache.py
llllllllll: we can now remove these, right?

zipline/testing/fixtures.py
```diff
@@ -651,6 +655,10 @@ def make_equity_daily_bar_data(cls):
             )
 
     @classmethod
+    def make_bcolz_daily_bar_rootdir_path(cls):
```
llllllllll: what is this change?

zipline/data/bundles/core.py
```diff
@@ -321,53 +340,56 @@ def ingest(name,
...
-                    )).path,
-                )
-                adjustment_db_writer = SQLiteAdjustmentWriter(
-                    stack.enter_context(working_file(
-                        adjustment_db_path(name, timestr, environ=environ),
-                    )).path,
-                    BcolzDailyBarReader(daily_bars_path),
-                    bundle.calendar,
-                    overwrite=True,
+                asset_db_writer = stack.enter_context(
```
llllllllll: thanks for entering these now to ensure the connection is closed!

zipline/data/bundles/core.py
```diff
@@ -321,53 +340,56 @@ def ingest(name,
...
+                    AssetDBWriter(
+                        wd.getpath(*asset_db_relative(
+                            name, timestr, environ=environ)
```
llllllllll: can we align the closing paren with the indentation level of the line of the opening paren. For example:

```python
AssetDBWriter(
    wd.getpath(*asset_db_relative(
        name, timestr, environ=environ,
    ),
    ...
)
```

tests/data/bundles/test_yahoo.py
LotannaEzenwa: Already merged with origin/master, so a few of these updates are from the recent PR's

zipline/data/bundles/core.py
LotannaEzenwa: Not necessarily, I made these a list so the underlying logic of paths.py wouldn't be changed

zipline/utils/cache.py
LotannaEzenwa: copytree() fails if the target directory already exists on py2.7 but not 3.4, dir_util.copy_tree is okay with merging on both versions of python

zipline/utils/cache.py
LotannaEzenwa: If `ensure_dir` makes subdirectories when the parent doesn't exist, then that change will work. The reason for using `makedirs` as opposed to `mkdir` is that `mkdir` fails when the parents don't exist.

zipline/utils/cache.py
llllllllll: cool, why is it verbose though?

zipline/testing/fixtures.py
LotannaEzenwa: From the merge with o/m

zipline/utils/cache.py
llllllllll: also, the class docstring needs to be updated to reflect this change

zipline/testing/fixtures.py
llllllllll: this shouldn't appear as a diff in your pr though, did you rebase?

zipline/data/us_equity_pricing.py
LotannaEzenwa: Now we enter this, as above

zipline/utils/cache.py
llllllllll: Sure, I was more concerned with the fact that this will not fail if the directory already exists. Also, we have `paths.ensure_directory` which should do this.

zipline/assets/asset_writer.py
```diff
@@ -563,3 +560,9 @@ def _load_data(self, equities, futures, exchanges, root_symbols):
             exchanges=exchanges_output,
             root_symbols=root_symbols_output,
         )
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *exc_info):
+        pass
```
richafrank: Should this be closing a connection here?

zipline/utils/cache.py
LotannaEzenwa: It was verbose for testing, but yeah it should be suppressed now.

zipline/utils/cache.py
LotannaEzenwa: Ah, I see. There's actually no possible way for the directory to exist because it's within a temp directory...I think. I could be wrong. 

But yeah you're right about it not failing, that was for catching the errors in testing.

zipline/testing/fixtures.py
LotannaEzenwa: I rebased ~35 of my commits before merging

zipline/data/bundles/core.py
LotannaEzenwa: With 2 paren, it would look like this?
```python
AssetDBWriter(
    wd.getpath(*asset_db_relative(
        name, timestr, environ=environ,
    )),
)
```

zipline/utils/cache.py
LotannaEzenwa: Yeah, working file isn't used in ingest anymore

zipline/assets/asset_writer.py
LotannaEzenwa: The only thing that the cm is capable of closing is the SqlAlchemy engine, which has a dispose method. It closes all the checked-in connections. All of the checked out connections will be left open, however. 

I'm not exactly sure if this method should try to close all connections, or just those that are not in use

LotannaEzenwa:
@llllllllll , @richafrank fixed up some last styles/comments.

zipline/assets/asset_writer.py
```diff
@@ -563,3 +560,9 @@ def _load_data(self, equities, futures, exchanges, root_symbols):
             exchanges=exchanges_output,
             root_symbols=root_symbols_output,
         )
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *exc_info):
+        pass  # Shouldn't dispose of the engine explicitly
```
richafrank: Do you think there's still a benefit to this becoming a context manager if it hasn't nothing to clean up?

tests/data/bundles/test_yahoo.py
richafrank: Just a note: before merging the PR, we should clean this up.  Ideally. we shouldn't be seeing unrelated changes from other PRs.

zipline/assets/asset_writer.py
LotannaEzenwa: Functionally, no. 

But for the code as a whole, I think it makes sense for all of the writer objects to be context managers as they control system resources. 
This asset writer is a special case because it's querying system doesn't need to be cleaned up explicitly.
If someone were to change the asset writer's underlying system for performance or to remove redundancies, it would be easy to add the exit method if needed.

richafrank: Ok, two things then:

1. If we want the asset writer to be a ctx manager, then we should update all consumers to dispose of it when they're done.  Right now, I think we've only updated the zipline data bundle.  However, given that it doesn't actually manage any resources right now that need explicit disposal, my tendency is to undo this and re-do it with all the call sites updated if/when there are resources to clean up.  @llllllllll What do you think?
1. We should also update this comment to explain why we shouldn't dispose of the engine.  For example, will it cause problems if we do, or is it just unnecessary?

llllllllll: Can you explain more why we don't need to dispose of the engine? I afree that we should either undo this change, or make it actually dispose of the engine and update the callers to use the ctx manager aspect of the object.

zipline/data/us_equity_pricing.py
```diff
@@ -210,8 +210,11 @@ class BcolzDailyBarWriter(object):
     def __init__(self, filename, calendar, start_session, end_session):
         self._filename = filename
 
-        assert calendar.is_session(start_session), "Start session is invalid!"
-        assert calendar.is_session(end_session), "End session is invalid!"
+        if start_session != end_session:
+            assert calendar.is_session(start_session),\
```
llllllllll: Can we make these a `ValueError` and show the session in the message. In general, assertions should be used to validate invariants in code that do not depend on user inputs. An assertion error should indicate that there is a bug in the program, but not that a function was called with invalid arguments.

llllllllll:
This looks great, thanks for getting this fixed! Other than the question about the context manager, we just need to clean up the commit messages and then this should be good to merge.

zipline/data/us_equity_pricing.py
LotannaEzenwa: Okay, fixing it

zipline/assets/asset_writer.py
LotannaEzenwa: Well I took another look at the overall docs for the sqlalchemy engine, and it said that the engine shouldn't be disposed until the end of the program because this prevents any future queries to the database. 
As of right now, the readers don't try to explicitly create an engine, so closing the engine would cause the readers to fail.

 I agree that the change wouldn't be necessary, as of right now

tests/data/bundles/test_yahoo.py
richafrank: @llllllllll thinks he added this to the PR intentionally to make this module compatible with both python 2 and 3.

LotannaEzenwa:
Merging

PR #1283, opened 2016-06-16
TST: Allow customization of various fixture paths.

ehebert:
To support testing configurations which need control over the full path
to the asset, adjustment, and equity bcolz directories; which is
required by some of our internal testing which exercises servers which
coordinate these files via a date slug in the full path.

Also, by allowing customization of the full path, it is now possible to
have the AssetFinder and AdjustmentReader sqlite databases be written to
disk, which is also required for our server testing setup.


zipline/testing/core.py
```diff
@@ -761,7 +763,8 @@ class tmp_assets_db(object):
     """
     _default_equities = sentinel('_default_equities')
 
-    def __init__(self, equities=_default_equities, **frames):
+    def __init__(self, url='sqlite://', equities=_default_equities, **frames):
```
ssanderson: I think you can equivalently have this be `'sqlite:///:memory:'`, which might be clearer for the reader.

ehebert: Was going off of what we had before, as well the SQLAlchemy docs, http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlite

I'm fine with changing to make the use of `:memory:` more explicit, though, will change here and in `tmp_asset_finder` and `WithAssetFinder`.

zipline/testing/core.py
```diff
@@ -809,9 +814,9 @@ class tmp_asset_finder(tmp_assets_db):
     --------
     tmp_assets_db
     """
-    def __init__(self, finder_cls=AssetFinder, **frames):
+    def __init__(self, url='sqlite://', finder_cls=AssetFinder, **frames):
```
ssanderson: ditto note from above

zipline/testing/fixtures.py
```diff
@@ -342,8 +345,13 @@ def make_equity_info(cls):
         )
 
     @classmethod
+    def make_asset_finder_db_url(cls):
+        return 'sqlite://'
```
ssanderson: ibid

zipline/testing/fixtures.py
```diff
@@ -884,9 +909,13 @@ def make_adjustment_writer_daily_bar_reader(cls):
         return cls.bcolz_daily_bar_reader
 
     @classmethod
+    def make_adjustment_db_conn_str(cls):
+        return ':memory:'
```
ssanderson: seems odd that this one is different from the others

ssanderson:
logic here all looks reasonable.  My only comment is that it might be clearer to use `'sqlite:///:memory:'` for the places where we're forwarding to sqlalchemy.create_engine.

zipline/testing/fixtures.py
ssanderson: Oh, this is going straight to sqlite3.connect instead of sqlalchemy.  Might be worth a comment.

ehebert: There is a comment in the class docstring which includes `returns the sqlite3 connection string for the database`. Should we double that up here?

ehebert:
Made the 'sqlite:///:memory:` change and squashed it in.

zipline/testing/fixtures.py
ssanderson: This is probably fine then.

ssanderson:
LGTM

PR #1287, opened 2016-06-17
DOC: Remove add_history

richafrank:
* Removed remaining mentions of `add_history`
* Removed inaccurate downloads/day badge

There's nothing we need to replace this text with, right?


llllllllll:
Thanks, lgtm

PR #1289, opened 2016-06-17
wildcard object and doctests

llllllllll:


zipline/pipeline/factors/factor.py
```diff
@@ -333,9 +333,9 @@ class Factor(RestrictedDTypeMixin, ComputableTerm):
     Factors.  For example, constructing a Factor that computes the average of
     two other Factors is simply::
 
-        >>> f1 = SomeFactor(...)
-        >>> f2 = SomeOtherFactor(...)
-        >>> average = (f1 + f2) / 2.0
+        >>> f1 = SomeFactor(...)  # doctest: +SKIP
```
ssanderson: Is there not a way to tell doctest to skip an entire docstring?

ssanderson: (I think these will show up in autogenerated docs unless sphinx is smart about these.)

llllllllll: It is smart about these (or has the ability to be). We do this in the blaze docs and they don't appear in the output.

llllllllll:
> +2.3%

nice

zipline/utils/events.py
```diff
@@ -301,8 +301,6 @@ class AfterOpen(StatelessRule):
     """
     A rule that triggers for some offset after the market opens.
     Example that triggers after 30 minutes of the market opening:
-
-    >>> AfterOpen(minutes=30)
```
ssanderson: Why did we remove this?

ssanderson: (If we do remove it, we should remove the sentence above referring to it, but it seems useful to have a usage example.)

llllllllll: what is it showing us?

ssanderson: (It looks like `minutes` is the wrong kwarg here though?)

llllllllll: Okay, I will update the docstring

PR #1290, opened 2016-06-20
MAINT: Fixes DataPortal.get_spot_value to correctly handle 'price' field

yankees714:
Querying for the price field of an equity actually looks at the close field, so we should do the same for futures. Otherwise `data.can_trade` and `data.current` of 'price' fail for futures.


ehebert:
@yankees714 and I chatted in person.

We agree that this is a good short term fix, but that long term we should change the structure to take advantage that both equities and futures now implement the same reader interface.

PR #1291, opened 2016-06-20
ENH: Allow for sid-to-sid correlations/regressions

dmichalowicz:
Give the option to pass a `Slice` _or_ another 2D factor to the correlation/regression methods of `Factor`.


dmichalowicz:
@ssanderson I believe this is all that would be necessary to support the option of passing another full-size factor to these methods. Would still need tests, but definitely seems worth adding.

dmichalowicz:
Closed in favor of https://github.com/quantopian/zipline/pull/1300

PR #1292, opened 2016-06-20
ENH: Create a method which overwrite portions of existing bcolz data

lianga888:


zipline/data/minute_bars.py
```diff
@@ -483,9 +487,9 @@ def write(self, data, show_progress=False):
         write_sid = self.write_sid
         with ctx as it:
             for e in it:
-                write_sid(*e)
+                write_sid(*e, overwrite=False)
```
yankees714: Should we be forwarding the `overwrite` arg of `write` here?

zipline/data/minute_bars.py
```diff
@@ -546,7 +553,51 @@ def write_cols(self, sid, dts, cols):
...
+            high : float64
+            low  : float64
+            close : float64
+            volume : float64|int64
+        """
+        sidpath = self.sidpath(sid)
+        if not os.path.exists(sidpath):
+            raise BcolzTableDoesNotExist('Could not find bcolz table for sid '
+                                         '%s while trying to overwrite data'
+                                         % sid)
```
yankees714: Is there a reason not to use `self._ensure_ctable` here?

ssanderson:
When do we need the ability to mutate an existing table in place?

zipline/data/minute_bars.py
lianga888: `self._ensure_ctable` will give us a blank table if one doesn't exist for the path. Although that is unlikely to happen in the context for which this method is currently used, I'd rather it just raise an error if it does happen (I don't want to be trying to overwrite parts of an empty table)

ssanderson:
@lianga888 @yankees714 is this still being worked on?

PR #1293, opened 2016-06-21
MAINT: Remove unused module.

ehebert:
Remove module, last usage was removed during lazy access pattern
rewrite.


ehebert:
cc: @richafrank @llllllllll @ssanderson 

Noticed this test failing on usage of `iloc` assignment when using more recent versions of pandas.

richafrank:
Yep, looks like it was only used by the old history container. 👍

PR #1296, opened 2016-06-21
BUG: get_first_trading_day_of_month needs to return normalized dt

lianga888:
Was previously returning a non-normalized dt when the first day of the month is a trading day. As a result, comparing the return value to the normalized value of the current dt is always False

### DIFF
--- a/zipline/utils/events.py
+++ b/zipline/utils/events.py
@@ -539,7 +539,7 @@ class NthTradingDayOfMonth(TradingDayOfMonthRule):
         self.month = dt.month
 
         dt = dt.replace(day=1)
-        first_day = (dt if self.cal.is_open_on_day(dt)
+        first_day = (normalize_date(dt) if self.cal.is_open_on_day(dt)
                      else self.cal.next_trading_day(dt))
         return first_day
 
--- a/tests/utils/test_events.py
+++ b/tests/utils/test_events.py
@@ -277,9 +277,14 @@ class TestStatelessRules(RuleTestCase):
 
         cls.nyse_cal = get_calendar('NYSE')
 
+        # First day of 09/2014 is closed whereas that for 10/2014 is open
         cls.sept_days = cls.nyse_cal.trading_days_in_range(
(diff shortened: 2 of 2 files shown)


zipline/utils/events.py
```diff
@@ -539,7 +539,7 @@ def get_first_trading_day_of_month(self, dt):
         self.month = dt.month
 
         dt = dt.replace(day=1)
-        first_day = (dt if self.cal.is_open_on_day(dt)
+        first_day = (normalize_date(dt) if self.cal.is_open_on_day(dt)
```
ssanderson: `normalize_date` will be wrong here if we're using a calendar that's not NYSE and US/Eastern. cc @jbredeche this is one of the cases we talked about yesterday.

In general, we need a method on `ExchangeCalendar` that says "give me the session date associated with the current minute", which doesn't appear to be available currently.

yankees714: I think this is the `session_date` method?

jbredeche: Yup, I saw this.  Leave it as `normalize_date` now, I don't trust the existing `session_date` method to do the right thing.  I'll go through the usages of `normalize_date` later and sort it out.

ssanderson: One thought here might be that we need a method with a signature like:

```
def find_nearest_session_date(self, dt, search_direction):
    """
    Find the nearest session date to dt.

    If dt is in a trading session, return that session's label.
    Otherwise, if search_direction == 'forward', find the label of the first session starting after dt;
    if search_direction == 'backward', find the label of the latest session ending before dt.
    """
```
    
```

jbredeche: (that method that @ssanderson described is exactly what I'm in the middle of adding to `ExchangeCalendar` right now)

ssanderson: OK, I'm fine leaving this as-is as long as we're aware that it's broken for nonstandard calendars and we're working on a fix.

PR #1300, opened 2016-06-23
Factor-to-factor correlations and regressions

dmichalowicz:
Allow correlations/regressions to be computed asset-wise between two 2D factors instead of just with a Slice.

### DIFF
--- a/tests/pipeline/test_engine.py
+++ b/tests/pipeline/test_engine.py
@@ -14,7 +14,6 @@ from numpy import (
     float32,
     float64,
     full,
-    full_like,
     log,
     nan,
     tile,
@@ -37,7 +36,6 @@ from pandas import (
 )
 from pandas.compat.chainmap import ChainMap
 from pandas.util.testing import assert_frame_equal
-from scipy.stats.stats import linregress, pearsonr, spearmanr
 from six import iteritems, itervalues
 from toolz import merge
 
@@ -82,6 +80,8 @@ from zipline.testing import (
     OpenPrice,
     parameter_space,
     product_upper_triangle,
+    run_correlations,
+    run_regressions,
 )
 from zipline.testing.fixtures import (
     WithAdjustmentReader,
@@ -90,6 +90,7 @@ from zipline.testing.fixtures import (
     ZiplineTestCase,
 )
 from zipline.utils.memoize import lazyval
+from zipline.utils.numpy_utils import bool_dtype
 
 
 class RollingSumDifference(CustomFactor):
@@ -584,7 +585,7 @@ class ConstantInputTestCase(WithTradingEnvironment, ZiplineTestCase):
         )
 
         expected_no_mask_result = full(
-            shape=(num_dates, num_assets), fill_value=True, dtype=bool,
+            shape=(num_dates, num_assets), fill_value=True, dtype=bool_dtype,
         )
 
         masks = cascading_mask, alternating_mask, NotSpecified
@@ -1236,7 +1237,7 @@ class ParameterizedFactorTestCase(WithTradingEnvironment, ZiplineTestCase):
         )
 
         expected_no_mask_result = full(
-            shape=(num_days, num_assets), fill_value=True, dtype=bool,
+            shape=(num_days, num_assets), fill_value=True, dtype=bool_dtype,
         )
 
         masks = cascading_mask, alternating_mask, NotSpecified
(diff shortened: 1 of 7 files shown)


dmichalowicz:
@ssanderson For when you have a chance next week

tests/pipeline/test_engine.py
```diff
@@ -584,7 +585,7 @@ def create_expected_results(expected_value, mask):
         )
 
         expected_no_mask_result = full(
-            shape=(num_dates, num_assets), fill_value=True, dtype=bool,
```
ssanderson: Did this break something, or was this just a style change?  (:+1: either way)

zipline/pipeline/factors/factor.py
```diff
@@ -626,7 +622,9 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
         return Rank(self, method=method, ascending=ascending, mask=mask)
 
     @expect_types(
-        target=Slice, correlation_length=int, mask=(Filter, NotSpecifiedType),
+        target=ComputableTerm,
```
ssanderson: I would probably put `(Factor, Slice)` here rather than `ComputableTerm`.  We shouldn't accept a Filter or a Classifier here.

zipline/pipeline/factors/statistical.py
```diff
@@ -66,11 +80,10 @@ class RollingPearson(_RollingCorrelation):
...
     instance of this class.
     """
-    def compute(self, today, assets, out, factor_data, slice_data):
-        slice_data_column = slice_data[:, 0]
-        for i in range(len(out)):
-            # pearsonr returns the R-value and the P-value.
-            out[i] = pearsonr(factor_data[:, i], slice_data_column)[0]
+    def compute(self, today, assets, out, factor_data, other_data):
+        super(RollingPearson, self).compute(
+            today, assets, out, factor_data, other_data, pearsonr,
```
ssanderson: I think the right way to do this is to forward up the correlation function in `__new__` rather than every time at compute.  As-is, you can construct an instance of the base class that can't actually be computed correctly.

zipline/pipeline/factors/statistical.py
```diff
@@ -27,13 +28,26 @@ def __new__(cls,
                 target_slice,
```
dmichalowicz: What could this be renamed to? `target_slice` doesn't apply if it can be another 2D factor

tests/pipeline/test_engine.py
dmichalowicz: Just a style change

zipline/testing/core.py
```diff
@@ -1470,6 +1475,188 @@ def ensure_doctest(f, name=None):
     return f
 
 
+def run_correlations(pearson_factor,
```
ssanderson: I'm nervous about two things here:

1. This is a **lot** of logic to be sharing between tests.  Do we really want to be running all of this in two places?  What useful functionality is being by running this twice?

2. There's a lot of logic in here that looks like it's making assumptions about the particular test data that's being passed in here (e.g. indexing with `int(asset) - `).

zipline/pipeline/factors/statistical.py
ssanderson: I think just `target` is fine?

dmichalowicz: I'm worried that might be confusing since this takes in 2 "targets"

ssanderson: `independent` and `dependent`?

dmichalowicz: That would be good for the regression factor. For the correlation one, I'm thinking `base_factor` and `target`

zipline/testing/core.py
dmichalowicz: The two things using this testing logic is 1) the built-in factors `RollingPearsonOfReturns`, `RollingLinear...` etc. and 2) the `Factor` methods `pearsonr`, `spearmanr` and `linear_regression` when passed two 2D factors. In both (1) and (2) we want to compute the expected results by running a `Returns` factor and calculating correlations/regressions over the `Returns` results. While cases (1) and (2) each definitely need their own test, after thinking more about it, I think that (1) and (2) are disjoint enough that attempting to write a shared function here is too convoluted. The best option is probably for each test to just re-write this logic, even though that makes for rather long tests, but there's no way to get around that anyway.

PR #1301, opened 2016-06-23
Return column vector for datasets with no sids

dmichalowicz:
Differentiate between 1D and 2D datasets given, e.g., macro datasets which have no sids.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -1574,14 +1574,8 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
 
         nassets = len(simple_asset_info)
         expected_views = keymap(pd.Timestamp, {
-            '2014-01-03': repeat_last_axis(
-                np.array([ffilled_value]),
-                nassets,
-            ),
-            '2014-01-04': repeat_last_axis(
-                np.array([1.0]),
-                nassets,
-            ),
+            '2014-01-03': np.array([[ffilled_value]]),
+            '2014-01-04': np.array([[1.0]]),
         })
 
         with tmp_asset_finder(equities=simple_asset_info) as finder:
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -12,7 +12,7 @@ import blaze as bz
 from datashape import dshape, var, Record
 from nose_parameterized import parameterized
 import numpy as np
-from numpy.testing.utils import assert_array_almost_equal
+from numpy.testing.utils import assert_array_almost_equal, assert_equal
 from odo import odo
 import pandas as pd
 from pandas.util.testing import assert_frame_equal
@@ -20,6 +20,7 @@ from toolz import keymap, valmap, concatv
 from toolz.curried import operator as op
 
 from zipline.assets.synthetic import make_simple_equity_info
+from zipline.errors import UnsupportedPipelineOutput
 from zipline.pipeline import Pipeline, CustomFactor
 from zipline.pipeline.data import DataSet, BoundColumn, Column
 from zipline.pipeline.engine import SimplePipelineEngine
@@ -44,6 +45,7 @@ from zipline.utils.numpy_utils import (
     repeat_last_axis,
 )
 from zipline.testing.predicates import assert_equal, assert_isidentical
+
 
 nameof = op.attrgetter('name')
 dtypeof = op.attrgetter('dtype')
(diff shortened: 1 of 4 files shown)


zipline/pipeline/data/dataset.py
```diff
@@ -115,14 +118,15 @@ class BoundColumn(LoadableTerm):
     inputs = ()
     window_safe = True
 
-    def __new__(cls, dtype, missing_value, dataset, name):
+    def __new__(cls, dtype, missing_value, dataset, name, ndim):
```
llllllllll: how do you feel about having boundcolumn have an `ndim` property that returns the ndim of `self.dataset`? This would ensure it is always the same.

dmichalowicz: @llllllllll I tried adding the `ndim` property after our discussion, but `ndim` is being set in `_init` of the `Term` base class, so I wasn't sure how to work around setting an attribute defined as a property

llllllllll: hmm, maybe then we should just forward `dataset.ndim` on the line below. It seems strange to be able to pass an `ndim` that doesn't match the `ndim` of the dataset. Idk how this plays with the caching here. ping @ssanderson

dmichalowicz: Agreed on forwarding `dataset.ndim`. Not sure what my rationale on passing `ndim`all the way through was, but that seems much cleaner

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -337,7 +333,7 @@ def new_dataset(expr, deltas, missing_values):
     the same type.
     """
     missing_values = dict(missing_values)
-    columns = {}
+    columns = {'ndim': 2 if SID_FIELD_NAME in expr.fields else 1}
```
llllllllll: could we rename this to `class_dict` or something?

tests/pipeline/test_blaze.py
```diff
@@ -778,6 +776,44 @@ def _test_id(self, df, dshape, expected, finder, add):
...
+
+        class UsesMacroInputs(CustomFactor):
+            inputs = macro_inputs
+            window_length = 1
+
+            def compute(self, today, assets, out, *inputs):
+                e = expected.loc[today]
+                for i, input_ in enumerate(inputs):
+                    # Each macro input should only have one column.
+                    assert input_.shape == (1, 1)
```
llllllllll: can we use `window_length` here to be clear why it is `(1, 1)`?

tests/pipeline/test_blaze.py
```diff
@@ -778,6 +776,44 @@ def _test_id(self, df, dshape, expected, finder, add):
...
+                    assert_equal(input_[0][0], e[i])
```
llllllllll: `input_[0, 0]`

tests/pipeline/test_blaze.py
```diff
@@ -972,27 +1008,19 @@ def test_id_macro_dataset(self):
...
-        2014-01-01 Equity(65 [A])      0
-                   Equity(66 [B])      0
-                   Equity(67 [C])      0
-        2014-01-02 Equity(65 [A])      1
-                   Equity(66 [B])      1
-                   Equity(67 [C])      1
-        2014-01-03 Equity(65 [A])      2
-                   Equity(66 [B])      2
-                   Equity(67 [C])      2
+                    value
```
llllllllll: thank you for remembering to update these!

tests/pipeline/test_blaze.py
```diff
@@ -1185,32 +1183,18 @@ def test_id_take_last_in_group_macro(self):
...
+            index=self.dates,
         )
-        self._test_id(
+        self._test_id_macro(
             df,
             var * Record(fields),
             expected,
             self.asset_finder,
-            ('value', 'other'),
+            ('other', 'value'),
```
llllllllll: why did we flip the order here?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1142,19 +1138,13 @@ def last_in_date_group(df, reindex, have_sids=have_sids):
...
-                # array which uses a buffer to represent `len(assets)` columns.
-                # The engine puts nans at the indicies for which we do not have
-                # sid information so that the nan-aware reductions still work.
-                # A future change to the engine would be to add first class
-                # support for macro econimic datasets.
-                copy,
-                partial(repeat_last_axis, count=len(assets)),
-            )
+            # If we do not have sids, use the column view to make a single
+            # column vector which is unassociated with any assets.
```
llllllllll: this comment should be above the `column_view` function

llllllllll:
how is the ndim getting set on columns now?

tests/pipeline/test_blaze.py
dmichalowicz: Yup

tests/pipeline/test_blaze.py
dmichalowicz: This order had to align with the order of the columns of `expected`, as this order matters when adding these columns as factor inputs in `_test_id_macro` (previously I don't think order mattered when just doing `pipeline.add`).

tests/pipeline/test_blaze.py
```diff
@@ -778,6 +776,44 @@ def _test_id(self, df, dshape, expected, finder, add):
...
+        ds = from_blaze(
+            expr,
+            loader=loader,
+            no_deltas_rule='ignore',
+            missing_values=self.missing_values,
+        )
+
+        p = Pipeline()
+        macro_inputs = []
+        for column_name in add:
```
dmichalowicz: Regarding comment below on ordering; maybe just scrap `add` and do `for column_name in expected.columns` instead?

zipline/pipeline/loaders/blaze/core.py
dmichalowicz: Yes, meant to do that

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -966,7 +962,7 @@ def _load_dataset(self, dates, assets, mask, columns):
             raise AssertionError('all columns must come from the same dataset')
 
         expr, deltas, checkpoints, odo_kwargs = self[dataset]
-        have_sids = SID_FIELD_NAME in expr.fields
+        have_sids = (dataset.ndim == 2)
```
dmichalowicz: Updated this

PR #1302, opened 2016-06-25
Point in time asset db

llllllllll:
Adds the ability to model when an equity changes symbol. This uses an im memory cache for by-symbol lookups so there is no more assetfindercachedequities.

### DIFF
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -35,7 +35,7 @@ from zipline.assets.asset_db_schema import (
     version_info,
 )
 
-from zipline.utils.range import from_tuple, intersecting_ranges, range
+from zipline.utils.range import from_tuple, intersecting_ranges
 
 # Define a namedtuple for use with the load_data and _load_data methods
 AssetData = namedtuple(
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -23,15 +23,39 @@ from toolz import first
 
 from zipline.errors import AssetDBVersionError
 from zipline.assets.asset_db_schema import (
-    generate_asset_db_metadata,
-    asset_db_table_names,
     ASSET_DB_VERSION,
+    asset_db_table_names,
+    asset_router,
+    equities as equities_table,
+    equity_symbol_mappings,
+    futures_contracts as futures_contracts_table,
+    futures_exchanges,
+    futures_root_symbols,
+    metadata,
+    version_info,
 )
 
+from zipline.utils.range import from_tuple, intersecting_ranges, range
+
 # Define a namedtuple for use with the load_data and _load_data methods
-AssetData = namedtuple('AssetData', 'equities futures exchanges root_symbols')
+AssetData = namedtuple(
+    'AssetData', (
+        'equities',
+        'equities_mappings',
+        'futures',
+        'exchanges',
+        'root_symbols',
+    ),
+)
 
 SQLITE_MAX_VARIABLE_NUMBER = 999
+
+symbol_columns = frozenset({
+    'symbol',
+    'company_symbol',
+    'share_class_symbol',
+})
+mapping_columns = symbol_columns | {'start_date', 'end_date'}
 
 # Default values for the equities DataFrame
 _equities_defaults = {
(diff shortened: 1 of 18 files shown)


zipline/assets/asset_writer.py
```diff
@@ -91,16 +115,22 @@ def split_delimited_symbol(symbol):
...
-    split_list = re.split(pattern=_delimited_symbol_delimiter_regex,
-                          string=symbol,
-                          maxsplit=1)
+    symbol = symbol.upper()
+
+    split_list = re.split(
+        pattern=_delimited_symbol_delimiters_regex,
+        string=symbol,
+        maxsplit=1,
+    )
```
yankees714: Since we're compiling the regex now this could be
```
    split_list = _delimited_symbol_delimiters_regex.split(
        string=symbol,
        maxsplit=1,
    )
```

zipline/assets/asset_writer.py
```diff
@@ -151,19 +176,70 @@ def _generate_output_dataframe(data_subset, defaults):
...
+    return data_subset
+
+
+def _check_asset_group(group):
+    for colname in set(group.columns) - mapping_columns:
+        col = group[colname]
+        if len(col.unique()) != 1:
+            raise ValueError(
+                'All values must be the same for the %s column' % colname,
+            )
```
yankees714: Can you add a comment for this block?

zipline/assets/assets.py
```diff
@@ -138,6 +184,72 @@ def _reset_caches(self):
...
+                    a.start,
+                    b.start,
+                    a.sid,
+                    a.symbol,
+                ) for a, b in sliding_window(
+                    2,
+                    concatv(
+                        sorted(v),
+                        # concat with a fake ownership object to make the last
+                        # end date be max timestamp
```
yankees714: Why do we need that?

zipline/assets/assets.py
```diff
@@ -354,181 +510,179 @@ def _retrieve_assets(self, sids, asset_tbl, asset_type):
...
+                company_symbol,
                 share_class_symbol,
-                self.equities.c.start_date <= ad_value),
-        ).order_by(
-            self.equities.c.end_date.desc(),
-        ).execute().fetchall()
-        return candidates
+            ]
+            if not owners:
+                # we have an entry but it is empty
```
yankees714: When does this case occur?

llllllllll: It doesn't , this can just be an `assert owners`

zipline/assets/assets.py
```diff
@@ -354,181 +510,179 @@ def _retrieve_assets(self, sids, asset_tbl, asset_type):
...
+                raise KeyError('goto error')
+        except KeyError:
+            # no equity has ever held a symbol matching the fuzzy symbol
+            raise SymbolNotFound(symbol=symbol)
+
+        if not as_of_date:
+            if not owners:
+                # no equity held the fuzzy symbol ever
+                raise SymbolNotFound(symbol=symbol)
```
yankees714: Seems like this isn't necessary either

llllllllll: Yeah, I will also swap this one out with an assert

zipline/assets/asset_writer.py
llllllllll: I think the current impl makes it more clear that it is not `str.split`, but I don't feel strongly about this.

zipline/assets/asset_writer.py
llllllllll: If the error message is not enough, maybe we can update the error message to provide the context. What information do you think would help here?

yankees714: Just on the initial read I wasn't clear what these columns would be and why they should be the same

zipline/assets/assets.py
llllllllll: We are trying to answer the question: "Who was the last equity to hold this ticker?" so we need to ffill the dates. Here we ffill the last date out to the max time. Is there a way I can reword the comment to make this more clear?

yankees714: Right, I guess what I mean to ask is why do we want the max time instead of the end_date of the last equity to hold the ticker?

PR #1305, opened 2016-06-28
MAINT: Disable deprecation warning for tradingcalendar

yankees714:
The new calendar API is currently in flux, so disabling the deprecation warning for `tradingcalendar`. To be re-enabled once the new API is stabilized.


jbredeche:
LGTM

PR #1307, opened 2016-06-28
Factor-to-factor correlations and regressions

dmichalowicz:
Opening in favor of https://github.com/quantopian/zipline/pull/1300
- Allow correlations/regressions to be computed between two 2D factors by doing computations asset-wise
- Moves all correlation/regression related tests from `test_engine.py` to `test_statistical.py`

### DIFF
--- a/tests/pipeline/test_statistical.py
+++ b/tests/pipeline/test_statistical.py
@@ -1,0 +1,780 @@
+"""
+Tests for statistical pipeline terms.
+"""
+from numpy import (
+    arange,
+    full,
+    full_like,
+    nan,
+    where,
+)
+from pandas import (
+    DataFrame,
+    date_range,
+    Int64Index,
+    Timestamp,
+)
+from pandas.util.testing import assert_frame_equal
+from scipy.stats import linregress, pearsonr, spearmanr
+
+from zipline.assets import Equity
+from zipline.errors import IncompatibleTerms, NonExistentAssetInTimeFrame
+from zipline.pipeline import CustomFactor, Pipeline
+from zipline.pipeline.data import USEquityPricing
+from zipline.pipeline.data.testing import TestingDataSet
+from zipline.pipeline.engine import SimplePipelineEngine
+from zipline.pipeline.factors import (
+    Returns,
+    RollingLinearRegressionOfReturns,
+    RollingPearsonOfReturns,
+    RollingSpearmanOfReturns,
+)
+from zipline.pipeline.loaders.frame import DataFrameLoader
+from zipline.pipeline.sentinels import NotSpecified
+from zipline.testing import (
+    AssetID,
+    AssetIDPlusDay,
+    check_arrays,
+    make_alternating_boolean_array,
+    make_cascading_boolean_array,
+    parameter_space,
+)
+from zipline.testing.fixtures import (
+    WithSeededRandomPipelineEngine,
+    WithTradingEnvironment,
+    ZiplineTestCase,
+)
+from zipline.utils.numpy_utils import (
+    bool_dtype,
+    datetime64ns_dtype,
+    float64_dtype,
+)
+
+
+class StatisticalBuiltInsTestCase(WithTradingEnvironment, ZiplineTestCase):
+    sids = ASSET_FINDER_EQUITY_SIDS = Int64Index([1, 2, 3])
+    START_DATE = Timestamp('2015-01-31', tz='UTC')
+    END_DATE = Timestamp('2015-03-01', tz='UTC')
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(StatisticalBuiltInsTestCase, cls).init_class_fixtures()
+
+        day = cls.trading_calendar.day
+        cls.dates = dates = date_range(
+            '2015-02-01', '2015-02-28', freq=day, tz='UTC',
+        )
+
+        # Using these start and end dates because they are a contigous span of
+        # 5 days (Monday - Friday) and they allow for plenty of days to look
+        # back on when computing correlations and regressions.
+        cls.start_date_index = start_date_index = 14
+        cls.end_date_index = end_date_index = 18
+        cls.pipeline_start_date = dates[start_date_index]
+        cls.pipeline_end_date = dates[end_date_index]
+        cls.num_days = num_days = end_date_index - start_date_index + 1
+
+        sids = cls.sids
+        cls.assets = assets = cls.asset_finder.retrieve_all(sids)
+        cls.my_asset_column = my_asset_column = 0
+        cls.my_asset = assets[my_asset_column]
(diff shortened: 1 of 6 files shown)


zipline/pipeline/factors/factor.py
```diff
@@ -626,7 +632,9 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
         return Rank(self, method=method, ascending=ascending, mask=mask)
 
     @expect_types(
-        target=Slice, correlation_length=int, mask=(Filter, NotSpecifiedType),
+        target=(FactorProxy, LoadableTerm, Slice),
```
dmichalowicz: Note to fix: If the wrong type of term is given here, `expect_types` says something like "this parameter was expected to be of type FactorProxy or..." which is confusing since we actually mean `Factor`

dmichalowicz: Also, thinking this should be `BoundColumn` instead of `LoadableTerm`?

zipline/pipeline/factors/factor.py
```diff
@@ -625,9 +626,7 @@ def rank(self, method='ordinal', ascending=True, mask=NotSpecified):
         """
         return Rank(self, method=method, ascending=ascending, mask=mask)
 
-    @expect_types(
-        target=Slice, correlation_length=int, mask=(Filter, NotSpecifiedType),
-    )
+    @expect_types(correlation_length=int, mask=(Filter, NotSpecifiedType))
```
dmichalowicz: Omitting the type check here for `target` and instead relying on the type checking done on `target` in `RollingPearson`

zipline/pipeline/factors/statistical.py
```diff
@@ -104,31 +118,43 @@ class RollingSpearman(_RollingCorrelation):
     Most users should call Factor.spearmanr rather than directly construct an
     instance of this class.
     """
-    def compute(self, today, assets, out, factor_data, slice_data):
-        slice_data_column = slice_data[:, 0]
-        for i in range(len(out)):
-            # spearmanr returns the R-value and the P-value.
-            out[i] = spearmanr(factor_data[:, i], slice_data_column)[0]
+    def compute(self, today, assets, out, base_data, target_data):
+        if target_data.shape[1] > 1:
```
ssanderson: Cleaner here might be:

```
target_date = np.broadcast_to(target_date, base_data.shape)
for i in range(len(out)):
    ...
```

The same pattern could be used pretty much everywhere in this file.

zipline/pipeline/factors/statistical.py
```diff
@@ -66,32 +72,40 @@ class RollingPearson(_RollingCorrelation):
     Most users should call Factor.pearsonr rather than directly construct an
     instance of this class.
     """
-    def compute(self, today, assets, out, factor_data, slice_data):
-        slice_data_column = slice_data[:, 0]
-        for i in range(len(out)):
-            # pearsonr returns the R-value and the P-value.
-            out[i] = pearsonr(factor_data[:, i], slice_data_column)[0]
+    def compute(self, today, assets, out, base_data, target_data):
+        if target_data.shape[1] > 1:
```
ssanderson: See note below re: cleaning this up via broacasting.

zipline/pipeline/factors/factor.py
```diff
@@ -765,6 +759,10 @@ def linear_regression(self, target, regression_length, mask=NotSpecified):
 
         Parameters
         ----------
+        target : zipline.pipeline.Term with a numeric dtype
```
ssanderson: `target` appears twice in this docstring

zipline/pipeline/factors/factor.py
ssanderson: You might want to throw an `expect_types` of Term for target here?

tests/pipeline/test_statistical.py
```diff
@@ -0,0 +1,759 @@
...
+    ZiplineTestCase,
+)
+from zipline.utils.numpy_utils import (
+    bool_dtype,
+    datetime64ns_dtype,
+    float64_dtype,
+)
+
+
+class StatisticalBuiltInsTestCase(WithTradingEnvironment, ZiplineTestCase):
```
dmichalowicz: This whole test suite has just been moved from `test_engine`, there is no logic change here.

tests/pipeline/test_statistical.py
```diff
@@ -0,0 +1,759 @@
...
+        }
+
+        results = run_pipeline(Pipeline(columns=columns), start_date, end_date)
+        regression_results = results['regression'].unstack()
+        expected_regression_results = results['expected_regression'].unstack()
+
+        assert_frame_equal(regression_results, expected_regression_results)
+
+        # Make sure we cannot call the linear regression method on factors or
+        # slices of dtype `datetime64[ns]`.
```
dmichalowicz: Note to self: make this a separate test?

tests/pipeline/test_statistical.py
```diff
@@ -0,0 +1,759 @@
...
+            date_factor.linear_regression(
+                target=returns_slice, regression_length=regression_length,
+            )
+        with self.assertRaises(TypeError):
+            returns.linear_regression(
+                target=date_factor_slice, regression_length=regression_length,
+            )
+
+    @parameter_space(correlation_length=[1, 2, 3, 4])
+    def test_factor_correlation_methods_two_factors(self, correlation_length):
```
dmichalowicz: This test and `test_factor_regression_method_two_factors` below are the only new tests for the factor-to-factor computations. @ssanderson Previously I had tried to factor out some of the shared code here with `test_correlation_factors` above, but that proved to be too confusing for what it was worth, so this test contains similar logic to the pre-existing tests. It might be worth a glance, but there's not much new here.

ssanderson: This looks fine to me.  I agree that it'd be nice to clean up a bit, but I don't think that's the most efficient use of your time.

zipline/pipeline/factors/statistical.py
```diff
@@ -66,32 +73,38 @@ class RollingPearson(_RollingCorrelation):
...
     instance of this class.
     """
-    def compute(self, today, assets, out, factor_data, slice_data):
-        slice_data_column = slice_data[:, 0]
+    def compute(self, today, assets, out, base_data, target_data):
+        # If `target_data` is a Slice or single column of data, broadcast it
+        # out to the same shape as `base_data`, then compute column-wise. This
+        # is efficient because each column of the broadcasted array only refers
+        # to a single memory location.
+        target_data = broadcast_arrays(target_data, base_data)[0]
```
dmichalowicz: @ssanderson I'm using `broadcast_arrays` here because `broadcast_to` was not introduced until numpy version 1.10.0

PR #1309, opened 2016-06-28
BUG: Correct AverageDollarVolume NaN handling

nathanwolfe:
`AverageDollarVolume` used `nanmean`, which discards NaNs before
averaging, giving an ADV which is too high for any equities that have
any NaNs.

Changing the method to `nansum` divided by window length so that the
denominator is the same no matter whether there are NaNs or not.

### DIFF
--- a/tests/pipeline/test_engine.py
+++ b/tests/pipeline/test_engine.py
@@ -1056,7 +1056,12 @@ class ParameterizedFactorTestCase(WithTradingEnvironment, ZiplineTestCase):
             index=dates,
             columns=cls.asset_finder.retrieve_all(sids),
         )
+        cls.raw_data_with_nans = cls.raw_data.where(cls.raw_data % 3 != 0)
 
+        open_loader = DataFrameLoader(
+            USEquityPricing.open,
+            cls.raw_data_with_nans,
+        )
         close_loader = DataFrameLoader(USEquityPricing.close, cls.raw_data)
         volume_loader = DataFrameLoader(
             USEquityPricing.volume,
@@ -1065,6 +1070,7 @@ class ParameterizedFactorTestCase(WithTradingEnvironment, ZiplineTestCase):
 
         cls.engine = SimplePipelineEngine(
             {
+                USEquityPricing.open: open_loader,
                 USEquityPricing.close: close_loader,
                 USEquityPricing.volume: volume_loader,
             }.__getitem__,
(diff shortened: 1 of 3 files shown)


richafrank:
Nice!  Should we add tests for this case?

nathanwolfe:
Added tests. Old `AverageDollarVolume` fails, new one passes.

tests/pipeline/test_engine.py
```diff
@@ -1056,7 +1056,12 @@ def init_class_fixtures(cls):
             index=dates,
             columns=cls.asset_finder.retrieve_all(sids),
         )
+        cls.raw_data_with_nans = cls.raw_data.where(cls.raw_data % 3 != 0)
```
dmichalowicz: I would probably write this as `((cls.raw_data % 3) != 0)` for clarity

dmichalowicz: Actually, I would also use `% 2` here, because otherwise this data is just all NaNs for the first asset and all filled values for the others. This doesn't test the case of having some NaNs and some values.

tests/pipeline/test_engine.py
```diff
@@ -1207,6 +1221,15 @@ def test_dollar_volume(self):
         expected_5 = rolling_mean((self.raw_data ** 2) * 2, window=5)[5:]
         assert_frame_equal(results['dv5'].unstack(), expected_5)
 
+        expected_1_nan = (self.raw_data_with_nans[5:]
+                          * self.raw_data[5:] * 2).fillna(0)
```
dmichalowicz: You can still just do `(self.raw_data_with_nans ** 2) * 2` here. Multiplying by `self.raw_data` will just fill with NaNs in the same places.

tests/pipeline/test_engine.py
```diff
@@ -1207,6 +1221,15 @@ def test_dollar_volume(self):
...
+        assert_frame_equal(results['dv1_nan'].unstack(), expected_1_nan)
+
+        expected_5_nan = rolling_mean((self.raw_data_with_nans
+                                       * self.raw_data * 2).fillna(0),
```
dmichalowicz: Same as above

tests/pipeline/test_engine.py
nathanwolfe: Is there any reason to make that change? Not squaring reflects that `USEquityPricing.open` and `USEquityPricing.volume` don't both have the NaNs.

dmichalowicz: Good point, I would just comment that in then

dmichalowicz:
Looks good to me, unless @llllllllll has anything

richafrank:
This fix also seems worthy of the release notes - would you add an entry there?

dmichalowicz:
@richafrank Is that appveyor failure intermittent? I'm not seeing any logs

richafrank:
> @richafrank Is that appveyor failure intermittent? I'm not seeing any logs

That's my guess.  We can see on the next update...

dmichalowicz:
> We can see on the next update...

Looks like it was

Issue #1268, opened 2016-06-08
Zipline does not work with 1 minute data


nalon99:
# Environment

* Operating System: Windows 8.1 Pro 64 bit
* Python Version: 2.7.11 (Anaconda 4.0.0 64 bit)
* Python Bitness: 64
* How did you install Zipline: 'conda'
* Python packages:

```
alabaster                 0.7.7                    py27_0    defaults
anaconda                  4.0.0               np110py27_0    defaults
anaconda-client           1.4.0                    py27_0    defaults
anaconda-navigator        1.1.0                    py27_0    defaults
argcomplete               1.0.0                    py27_1    defaults
astropy                   1.1.2               np110py27_0    defaults
awscli                    1.6.2                     <pip>
babel                     2.2.0                    py27_0    defaults
backports-abc             0.4                       <pip>
backports.ssl-match-hostname 3.5.0.1                   <pip>
backports_abc             0.4                      py27_0    defaults
bcdoc                     0.12.2                    <pip>
bcolz                     0.12.1              np110py27_0    Quantopian
beautifulsoup4            4.4.1                    py27_0    defaults
bitarray                  0.8.1                    py27_1    defaults
blaze                     0.9.1                    py27_0    defaults
bokeh                     0.11.1                   py27_0    defaults
boto                      2.39.0                   py27_0    defaults
botocore                  0.73.0                    <pip>
bottleneck                1.0.0               np110py27_0    defaults
bzip2                     1.0.6                     vc9_2  [vc9]  defaults
cachetools                1.1.6                    py27_0    Quantopian
cdecimal                  2.3                      py27_1    defaults
certifi                   2015.4.28                 <pip>
cffi                      1.5.2                    py27_0    defaults
chest                     0.2.3                    py27_0    defaults
click                     6.6                      py27_0    defaults
cloudpickle               0.1.1                    py27_0    defaults
clyent                    1.2.1                    py27_0    defaults
colorama                  0.3.7                    py27_0    defaults
comtypes                  1.1.2                    py27_0    defaults
conda                     4.0.6                    py27_0    defaults
conda-build               1.20.0                   py27_0    defaults
conda-env                 2.4.5                    py27_0    defaults
conda-manager             0.3.1                    py27_0    defaults
configobj                 5.0.6                    py27_0    defaults
console_shortcut          0.1.1                    py27_1    defaults
contextlib2               0.4.0                    py27_0    defaults
cryptography              1.3                      py27_0    defaults
[...]

PR #1313, opened 2016-06-30
BUG: Add support for Panel data in accordance with documentation

nathanwolfe:
`TradingAlgorithm.run` didn't work using Panel minute data (https://github.com/quantopian/zipline/issues/1268). This was because no `PanelMinuteBarReader` class existed, so I wrote the class, basically copying from `PanelDailyBarReader`. *Edit: The two are now combined into `PanelBarReader`.*

Another important addition to `TradingAlgorithm.run` is Panel data frequency inference, if the argument `overwrite_sim_params` is True (which it is by default). `run` infers daily bars if every single Timestamp in the DatetimeIndex has the same time of day, and minute bars if any of those times of day are different. Let me know if this inference should be done differently.

An additional issue was that `zipline.run_algorithm` didn't work with Panel data; it only worked with bundles. That's fixed too.

I've added a test. It runs both methods (`TradingAlgorithm.run` and `run_algorithm`) on both data frequencies and verifies that `data.current` and `data.history` get the correct price value at every bar.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -33,7 +33,10 @@ import numpy as np
 import pandas as pd
 import pytz
 
-from zipline import TradingAlgorithm
+from zipline import (
+    run_algorithm,
+    TradingAlgorithm,
+)
 from zipline.api import FixedSlippage
 from zipline.assets import Equity, Future
 from zipline.assets.synthetic import (
@@ -161,6 +164,7 @@ from zipline.test_algorithms import (
     no_handle_data,
 )
 from zipline.utils.api_support import ZiplineAPI, set_algo_instance
+from zipline.utils.calendars import get_calendar
 from zipline.utils.context_tricks import CallbackManager
 from zipline.utils.control_flow import nullctx
 import zipline.utils.events
@@ -4110,3 +4114,66 @@ class AlgoInputValidationTestCase(ZiplineTestCase):
                     script=script,
                     **{method: lambda *args, **kwargs: None}
                 )
+
+
+class TestPanelData(ZiplineTestCase):
+
+    @parameterized.expand([
+        ('daily',
+         pd.Timestamp('2015-12-23', tz='UTC'),
+         pd.Timestamp('2016-01-05', tz='UTC'),),
+        ('minute',
+         pd.Timestamp('2015-12-23', tz='UTC'),
+         pd.Timestamp('2015-12-24', tz='UTC'),),
+    ])
+    def test_panel_data(self, data_frequency, start_dt, end_dt):
+        if data_frequency == 'daily':
+            history_freq = '1d'
+            df = create_daily_df_for_asset(get_calendar('NYSE'),
+                                           start_dt, end_dt)
+        elif data_frequency == 'minute':
+            history_freq = '1m'
+            df = create_minute_df_for_asset(get_calendar('NYSE'),
+                                            start_dt, end_dt)
+
+        panel = pd.Panel({1: df})
+
+        price_record = pd.DataFrame(columns=['current', 'previous'])
+
+        def initialize(algo):
+            algo.first_bar = True
+
+        def handle_data(algo, data):
+            price_record.loc[algo.get_datetime(), 'current'] = (
+                data.current(algo.sid(1), 'price')
+            )
+            if algo.first_bar:
+                algo.first_bar = False
+            else:
+                price_record.loc[algo.get_datetime(), 'previous'] = (
+                    data.history(algo.sid(1), 'price', 2, history_freq)[0]
+                )
+
+        trading_algo = TradingAlgorithm(initialize=initialize,
+                                        handle_data=handle_data)
+        trading_algo.run(data=panel)
+        np.testing.assert_array_equal(
+            np.array(price_record.transpose(), dtype='float64'),
+            np.array([df['close'], df['close'].shift(1)], dtype='float64')
+        )
+
+        price_record.drop(price_record.index)
+
+        run_algorithm(
+            start=start_dt,
+            end=end_dt,
+            capital_base=1,
+            initialize=initialize,
+            handle_data=handle_data,
+            data_frequency=data_frequency,
+            data=panel
+        )
+        np.testing.assert_array_equal(
+            np.array(price_record.transpose(), dtype='float64'),
+            np.array([df['close'], df['close'].shift(1)], dtype='float64')
+        )
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -4127,42 +4127,62 @@ class TestPanelData(ZiplineTestCase):
          pd.Timestamp('2015-12-24', tz='UTC'),),
     ])
     def test_panel_data(self, data_frequency, start_dt, end_dt):
+        trading_calendar = get_calendar('NYSE')
         if data_frequency == 'daily':
             history_freq = '1d'
-            df = create_daily_df_for_asset(get_calendar('NYSE'),
-                                           start_dt, end_dt)
+            create_df_for_asset = create_daily_df_for_asset
+            dt_transform = trading_calendar.minute_to_session_label
         elif data_frequency == 'minute':
             history_freq = '1m'
-            df = create_minute_df_for_asset(get_calendar('NYSE'),
-                                            start_dt, end_dt)
+            create_df_for_asset = create_minute_df_for_asset
 
-        panel = pd.Panel({1: df})
+            def dt_transform(dt):
+                return dt
 
-        price_record = pd.DataFrame(columns=['current', 'previous'])
+        sids = range(1, 3)
+        dfs = {}
+        for sid in sids:
+            dfs[sid] = create_df_for_asset(trading_calendar,
+                                           start_dt, end_dt, interval=sid)
+            dfs[sid]['prev_close'] = dfs[sid]['close'].shift(1)
+        panel = pd.Panel(dfs)
+
+        price_record = pd.Panel(items=sids,
+                                major_axis=panel.major_axis,
+                                minor_axis=['current', 'previous'])
 
         def initialize(algo):
             algo.first_bar = True
+            algo.equities = []
+            for sid in sids:
+                algo.equities.append(algo.sid(sid))
 
         def handle_data(algo, data):
-            price_record.loc[algo.get_datetime(), 'current'] = (
-                data.current(algo.sid(1), 'price')
+            price_record.loc[:, dt_transform(algo.get_datetime()),
+                             'current'] = (
+                data.current(algo.equities, 'price')
             )
             if algo.first_bar:
                 algo.first_bar = False
             else:
-                price_record.loc[algo.get_datetime(), 'previous'] = (
-                    data.history(algo.sid(1), 'price', 2, history_freq)[0]
+                price_record.loc[:, dt_transform(algo.get_datetime()),
+                                 'previous'] = (
+                    data.history(algo.equities, 'price',
+                                 2, history_freq).iloc[0]
                 )
+
+        def check_panels():
+            np.testing.assert_array_equal(
+                price_record.values.astype('float64'),
+                panel.loc[:, :, ['close',
+                                 'prev_close']].values.astype('float64')
+            )
 
         trading_algo = TradingAlgorithm(initialize=initialize,
                                         handle_data=handle_data)
         trading_algo.run(data=panel)
-        np.testing.assert_array_equal(
-            np.array(price_record.transpose(), dtype='float64'),
-            np.array([df['close'], df['close'].shift(1)], dtype='float64')
-        )
-
-        price_record.drop(price_record.index)
+        check_panels()
+        price_record.loc[:] = np.nan
 
         run_algorithm(
             start=start_dt,
@@ -4173,7 +4193,4 @@ class TestPanelData(ZiplineTestCase):
             data_frequency=data_frequency,
             data=panel
         )
-        np.testing.assert_array_equal(
-            np.array(price_record.transpose(), dtype='float64'),
-            np.array([df['close'], df['close'].shift(1)], dtype='float64')
-        )
+        check_panels()
--- a/tests/test_panel_daily_bar_reader.py
+++ b/tests/test_panel_daily_bar_reader.py
@@ -22,13 +22,12 @@ from zipline.data.us_equity_pricing import PanelBarReader
 from zipline.testing import ExplodingObject
 from zipline.testing.fixtures import (
     WithAssetFinder,
-    WithNYSETradingDays,
     ZiplineTestCase,
 )
+from zipline.utils.calendars import get_calendar
 
 
 class TestPanelDailyBarReader(WithAssetFinder,
-                              WithNYSETradingDays,
                               ZiplineTestCase):
 
     START_DATE = pd.Timestamp('2006-01-03', tz='utc')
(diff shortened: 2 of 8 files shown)


zipline/algorithm.py
```diff
@@ -636,16 +649,31 @@ def run(self, data=None, overwrite_sim_params=True):
...
-                    copy_panel,
-                )
-                self.data_portal = DataPortal(
-                    self.asset_finder,
-                    self.trading_schedule,
-                    first_trading_day=equity_daily_reader.first_trading_day,
-                    equity_daily_reader=equity_daily_reader,
-                )
+
+                if self.sim_params.data_frequency == 'daily':
```
llllllllll: since the only difference here is the single constructor, we can just conditionally call `PanelDailyBarReader` or `PanelMinuteBarReader` like:

```python
equity_minute_reader = (
    PanelMinuteBarReader
    if self.sim_params.data_frequency == 'minute' else
    PanelDailyBarReader
)(
    self.trading_schedule.all_execution_days,
    copy_panel,
)
self.data_portal = DataPortal(
    self.asset_finder,
    self.trading_schedule,
    first_trading_day=equity_minute_reader
        .first_trading_day,
    equity_minute_reader=equity_minute_reader,
)
```

nathanwolfe: Definitely. Will change.

zipline/data/minute_bars.py
```diff
@@ -950,3 +957,107 @@ def load_raw_arrays(self, fields, start_dt, end_dt, sids):
...
+            The asset identifier.
+        dt : datetime64-like
+            Midnight of the day for which data is requested.
+
+        Returns
+        -------
+        pd.Timestamp : The last known dt for the asset and dt;
+                       NaT if no trade is found before the given dt.
+        """
+        for ts in self.panel.major_axis[self.panel.major_axis
```
llllllllll: I think this function is the same as:
```python
self.panel.loc[sid, :dt, 'close'].last_valid_index()
```

zipline/utils/run_algo.py
```diff
@@ -148,14 +149,21 @@ def choose_loader(column):
...
+        choose_loader = None
 
     perf = TradingAlgorithm(
         namespace=namespace,
         capital_base=capital_base,
-        start=start,
-        end=end,
         env=env,
         get_pipeline_loader=choose_loader,
+        sim_params=create_simulation_parameters(
```
llllllllll: we should pass the env to the sim params as well.

zipline/data/minute_bars.py
```diff
@@ -950,3 +957,107 @@ def load_raw_arrays(self, fields, start_dt, end_dt, sids):
...
+
+        self.panel = panel
+
+        self._ohlc_inverse = 1. / OHLC_RATIO
+
+    @property
+    def last_available_dt(self):
+        return self.panel.major_axis[-1]
+
+    def load_raw_arrays(self, columns, start_dt, end_dt, assets):
```
llllllllll: I think we can vectorize this operation:
```python
return list(self.panel.loc[list(assets), start_dt:end_dt, list(columns)].values.T)
```

This works by slicing the panel down by all the constraints and then asking for the data as an ndarray. This will give us a `(sids, dt, columns)` array. We actually need to get the a `(dt, sid)` per column, so by taking the transpose we get a `(columns, dt, sids)` array. We can then iterate over this to take slices along the column axis to give us `len(columns)` `(dt, sid)` arrays.

llllllllll: also, I think can return an iterator so we may not need to call `list` on the result array.

zipline/algorithm.py
```diff
@@ -611,16 +612,28 @@ def run(self, data=None, overwrite_sim_params=True):
...
                     # Changing period_start and period_close might require
                     # updating of first_open and last_close.
                     self.sim_params.update_internal_from_trading_schedule(
                         trading_schedule=self.trading_schedule
                     )
 
+                    # Assume data is daily if timestamp times are
+                    # standardized, otherwise assume minute bars.
+                    times = data.major_axis.time
+                    if np.all(times == times[0]):
```
llllllllll:
```
if pd.get_period_alias(pd.infer_freq(data.major_axis.time)) == 'D'
```

zipline/algorithm.py
```diff
@@ -611,16 +612,28 @@ def run(self, data=None, overwrite_sim_params=True):
 
             if isinstance(data, pd.Panel):
                 # For compatibility with existing examples allow start/end
-                # to be inferred.
+                # to be inferred. Start and end have to be midnight UTC.
                 if overwrite_sim_params:
-                    self.sim_params.period_start = data.major_axis[0]
-                    self.sim_params.period_end = data.major_axis[-1]
+                    self.sim_params.period_start = normalize_date(
+                        data.major_axis[0].tz_convert('UTC')
```
llllllllll: should we guard against a tz-naiive index?

tests/test_algorithm.py
```diff
@@ -3982,3 +3986,66 @@ def analyze(context, results):
...
+        if data_frequency == 'daily':
+            history_freq = '1d'
+            df = create_daily_df_for_asset(default_nyse_schedule,
+                                           start_dt, end_dt)
+        elif data_frequency == 'minute':
+            history_freq = '1m'
+            df = create_minute_df_for_asset(default_nyse_schedule,
+                                            start_dt, end_dt)
+
+        panel = pd.Panel({1: df})
```
llllllllll: should we test with more than 1 sid to ensure the indexing logic is correct?

zipline/data/minute_bars.py
llllllllll: is this method tested anywhere?

zipline/algorithm.py
nathanwolfe: Sure. Not sure what the best way is. Could call `normalize_date` and then set the `tz` property of the result to `'UTC'`.

A tz-naive index breaks other things as well, though. May be best to automatically convert index to UTC midnight if it's daily mode.

zipline/data/minute_bars.py
nathanwolfe: Pretty sure `data.history` uses `load_raw_arrays`, and `data.history` gets tested.

zipline/utils/run_algo.py
nathanwolfe: `create_simulation_parameters` doesn't take an env argument.

llllllllll: then we should just create it manually.

nathanwolfe: I don't think `SimulationParameters` has an env attribute. Correct me if I'm wrong.

zipline/algorithm.py
llllllllll: if it is naive, I think we can assume it is utc by tz_localizing to utc iff it is tz-naive

zipline/data/minute_bars.py
llllllllll: awesome, just checking

zipline/utils/run_algo.py
llllllllll: You are correct, I think this is a recent change.

zipline/algorithm.py
nathanwolfe: Actually, the two cases call for two different keyword arguments: `equity_minute_reader` and `equity_daily_reader`. Any way to resolve that?

zipline/data/minute_bars.py
nathanwolfe: should `load_raw_arrays` be changed in `PanelDailyBarReader` too? https://github.com/quantopian/zipline/blob/master/zipline/data/us_equity_pricing.py#L729

llllllllll: sure

nathanwolfe: I suppose we can have `PanelMinuteBarReader` inherit from `PanelDailyBarReader`.

llllllllll: are there actually any differences? could it just be a `PanelBarReader`?

nathanwolfe: `PanelMinuteBarReader` alone has `self._ohlc_inverse`, but it can be removed if [this line](https://github.com/quantopian/zipline/blob/65a1e3465cf74f108be7ac9d881e265c277b390a/zipline/data/data_portal.py#L556) in `DataPortal` can be removed. That line assigns `self.MINUTE_PRICE_ADJUSTMENT_FACTOR`, which is [never used](https://github.com/quantopian/zipline/search?utf8=%E2%9C%93&q=MINUTE_PRICE_ADJUSTMENT_FACTOR&type=Code).

Everything else is roughly equivalent and can be consolidated. I suppose I'll put the `PanelBarReader` class in zipline/data/us_equity_pricing.py. Should I make it a subclass of `DailyBarReader`?

llllllllll: Sure, it is just an abc so it won't hurt.

nathanwolfe:
AppVeyor failed on its own side.

zipline/algorithm.py
```diff
@@ -603,14 +603,26 @@ def run(self, data=None, overwrite_sim_params=True):
...
+                    data.major_axis = data.major_axis.tz_localize('UTC')
+
                 # For compatibility with existing examples allow start/end
-                # to be inferred.
+                # to be inferred. Start and end have to be midnight UTC.
                 if overwrite_sim_params:
                     self.sim_params = self.sim_params.create_new(
-                        data.major_axis[0],
-                        data.major_axis[-1]
+                        normalize_date(data.major_axis[0].tz_convert('UTC')),
```
richafrank: Should we be normalizing to midnight nowadays, or finding the session for the particular minute according to the trading calendar?

zipline/data/us_equity_pricing.py
```diff
@@ -767,35 +767,38 @@ def __init__(self, calendar, panel):
             # Fake volume if it does not exist.
             panel.loc[:, :, 'volume'] = int(1e9)
 
-        self.first_trading_day = panel.major_axis[0]
-        self._calendar = calendar
+        self.first_trading_day = normalize_date(panel.major_axis[0])
```
richafrank: Same question about normalizing here

nathanwolfe: `PanelBarReader` isn't constructed with a trading calendar; it's given `calendar`, but that's a list of execution datetimes.

nathanwolfe: Should I amend `PanelBarReader` to be constructed with a `trading_calendar` argument?

zipline/algorithm.py
nathanwolfe: Changing to `self.trading_calendar.minute_to_session_label` in the next commit.

zipline/data/us_equity_pricing.py
richafrank: Yea, I assume the caller constructed the list of dts from a trading calendar, so probably they could pass the whole thing along, and we can make use of more of its methods here.  (So whatever the name of the arg, yes, I agree we should pass a trading calendar instance instead of just its sessions.)

zipline/data/data_portal.py
```diff
@@ -554,8 +554,6 @@ def __init__(self,
                 self._equity_minute_reader,
                 self._adjustment_reader
             )
-            self.MINUTE_PRICE_ADJUSTMENT_FACTOR = \
```
llllllllll: was this just dead code?

zipline/data/us_equity_pricing.py
```diff
@@ -760,15 +759,33 @@ class PanelDailyBarReader(DailyBarReader):
...
+
+        self._sessions = trading_calendar.sessions_in_range(
+            self.first_trading_day,
+            last_trading_day
+        )
+
+        if data_frequency == 'daily':
+            self._calendar = self._sessions
+        elif data_frequency == 'minute':
+            self._calendar = (trading_calendar.minutes_for_sessions_in_range(
```
llllllllll: the parens around the whole expr are not needed

zipline/data/us_equity_pricing.py
```diff
@@ -760,15 +759,33 @@ class PanelDailyBarReader(DailyBarReader):
...
+        )
+
+        self._sessions = trading_calendar.sessions_in_range(
+            self.first_trading_day,
+            last_trading_day
+        )
+
+        if data_frequency == 'daily':
+            self._calendar = self._sessions
+        elif data_frequency == 'minute':
```
llllllllll: Can we add an `ensure_element(data_frequency={'daily', 'minute'})` so that users cannot pass invalid frequencies?

zipline/data/us_equity_pricing.py
```diff
@@ -760,15 +759,33 @@ class PanelDailyBarReader(DailyBarReader):
         The first trading day in the dataset.
     """
     @preprocess(panel=call(verify_indices_all_unique))
-    def __init__(self, calendar, panel):
+    def __init__(self, trading_calendar, panel, data_frequency='daily'):
```
llllllllll: why does `data_frequency` have a default argument?

zipline/data/data_portal.py
nathanwolfe: It's never used within Zipline: https://github.com/quantopian/zipline/search?utf8=%E2%9C%93&q=MINUTE_PRICE_ADJUSTMENT_FACTOR

zipline/data/us_equity_pricing.py
```diff
@@ -778,24 +795,17 @@ def last_available_dt(self):
 
     @property
     def trading_calendar(self):
-        return None
+        return self._trading_calendar
```
llllllllll: is there a reason this is wrapped in a property instead of used directly?

nathanwolfe: I made `PanelBarReader` a subclass of `DailyBarReader`, which has`trading_calendar` as an abstract property: https://github.com/quantopian/zipline/blob/master/zipline/data/us_equity_pricing.py#L405

llllllllll: Okay, to get around this we could just have `trading_calendar = None` in the class body and then in `__init__` have `self.trading_calendar = calendar`. I think either way is fine. If we leave the proprty as it is can we have a comment so people don't try to remove it later?

zipline/data/us_equity_pricing.py
nathanwolfe: No real reason. I can remove the default.

zipline/data/us_equity_pricing.py
nathanwolfe: I'll change it to what you're suggesting.

zipline/data/us_equity_pricing.py
nathanwolfe: added here: https://github.com/quantopian/zipline/pull/1313/files#diff-bd2429cc65fd41a1669bbdf31c6af6a6R762

nathanwolfe:
I guess I'll expand `tests/test_panel_daily_bar_reader.py` to test minute data as well. Otherwise `sessions` isn't covered for minute frequency.

tests/test_panel_daily_bar_reader.py
```diff
@@ -27,24 +27,26 @@
...
+        if cls.FREQUENCY == 'daily':
+            major_axis = trading_calendar.sessions_in_range(
+                cls.START_DATE,
+                cls.END_DATE
+            )
+        elif cls.FREQUENCY == 'minute':
+            major_axis = trading_calendar.minutes_for_sessions_in_range(
+                cls.START_DATE,
+                cls.END_DATE
+            )
```
nathanwolfe: Realized this could be condensed. I'll fix it in a few minutes.

PR #1315, opened 2016-07-05
change str to string_types to avoid errors

FayolChang:
when in python2.7, and unicode_literals is imported
type check will raise error because 'type' is not str but unicode
use six.string_types to avoid this error.


richafrank:
Thanks for the improvement @FayolChang !

It looks like the tests are failing on your branch.  I know `isinstance` handles an iterable of types instead of a single type, so I'm guessing the `check_parameters` decorator does not.  Is that something you can fix?

richafrank:
Thanks @FayolChang !  Merged your commit in with #1467.

PR #1317, opened 2016-07-07
TST: Speed up TestMiscellaneousAPI

ehebert:
Limit the date range to one day, instead of a year.

On my local machine the following goes from taking 12 seconds to 0.6 seconds.

`$ nosetests -x tests/test_algorithm.py:TestMiscellaneousAPI`


ehebert:
cc: @llllllllll

ehebert:
@richafrank tests have been updated to do 2 days instead of 1, to cover any issues with rolling over a day. Still less than 0.7 seconds.

richafrank:
👍 from me!

PR #1319, opened 2016-07-10
NamedTemp manual deletion for windows users.

LotannaEzenwa:
Solution to the problem on PR #1282

### DIFF
--- a/zipline/utils/cache.py
+++ b/zipline/utils/cache.py
@@ -41,7 +41,6 @@ class CachedObject(namedtuple("_CachedObject", "value expires")):
     >>> obj.unwrap(expires)
     1
     >>> obj.unwrap(expires + Timedelta('1 minute'))
-    ... # doctest: +IGNORE_EXCEPTION_DETAIL
     Traceback (most recent call last):
         ...
     Expired: 2014-01-01 00:00:00+00:00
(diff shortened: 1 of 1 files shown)


LotannaEzenwa:
@llllllllll @richafrank, Could we go over this tomorrow?

zipline/utils/cache.py
```diff
@@ -41,7 +41,6 @@ class CachedObject(namedtuple("_CachedObject", "value expires")):
     >>> obj.unwrap(expires)
     1
     >>> obj.unwrap(expires + Timedelta('1 minute'))
-    ... # doctest: +IGNORE_EXCEPTION_DETAIL
```
llllllllll: We still need this line, this is what is causing the tests to fail in py3

zipline/utils/cache.py
```diff
@@ -302,6 +304,7 @@ def __exit__(self, *exc_info):
         if exc_info[0] is None:
             self._commit()
         self._tmpfile.__exit__(*exc_info)
+        os.remove(self._tmpfile.name)
```
llllllllll: won't this try to remove twice, once in the commit and then here?

LotannaEzenwa: Yeah, that's true. Changing it.

PR #1320, opened 2016-07-11
Added manual deletion to temp files

LotannaEzenwa:
Because windows doesn't allow multiple processes on a file

@llllllllll , can you merge this onto your branch?


LotannaEzenwa:
@llllllllll

Issue #855, opened 2015-11-18
Unify risk metrics across zipline and pyfolio


twiecki:
Discussion started over here: #853 

With @llllllllll  saying "The main reason for moving this to a new repo would be to make all the risk metrics that we report consistent. We can share implementation and tests for zipline and pyfolio which is great for avoiding duplication and introducing subtle differences. I would want this to be a standalone project because I don't think it makes sense for zipline to be a dependency of pyfolio or for pyfolio to be a dependency of zipline."

One thing to consider is that the data passed into zipline and pyfolio risk metrics is quite different. pyfolio requires arrays and zipline passes in e.g. mean returns. If we rely on separate code paths we might just move that problem somewhere else.

ehebert:
What are the shape/contents of the arrays passed by pyfolio?

It should not be too much work to change Zipline over to using a different array structure since risk modules are already keeping track of the original returns as well.

twiecki:
It takes an array of daily returns and daily treasury returns.

twiecki:
I ported the risk metrics to https://github.com/quantopian/pyrisk so zipline is free to move over.

twiecki:
CC @llllllllll @richafrank

ahgnaw:
Closed by https://github.com/quantopian/zipline/pull/1322 and https://github.com/quantopian/pyfolio/pull/320

PR #1322, opened 2016-07-13
Move risk calculations

ahgnaw:
Centralize the way metrics are calculated in zipline by adding qrisk as a dependency and replacing the current functions.

Testing of actual calculated values within classes has been removed and replaced by a test of existence and type. Any test for metric correctness will be done in the qrisk repository from now on.

Treasury curves are no longer used as the risk-free rate when calculating the sharpe ratio.


jbredeche:
sweet!

tests/risk/AnswerKeyAnnotations.ipynb
```diff
@@ -1,87 +0,0 @@
-{
```
ehebert: :+1:

I can not recall if this powers a link on the forums discussing risk metrics.
I think it is ok to remove, since if that link does break, we can change the link to point to a commitish where this file does exist (if it doesn't already.)

tests/risk/test_risk_period.py
```diff
@@ -18,459 +18,341 @@
...
-            year=2006,
-            month=1,
-            day=1,
-            hour=0,
-            minute=0,
-            tzinfo=pytz.utc)
-        end_date = datetime.datetime(
-            year=2006, month=12, day=31, tzinfo=pytz.utc)
-
+        self.start_date = datetime.datetime(2006, 1, 1, tzinfo=pytz.utc)
```
ehebert: Since this test had been written, we have more or less standardized on using `pd.Timestamp` instead of `datetime.datetime`.

abhijeetkalyan:
@ahgnaw @ehebert what do you think of keeping the existing methods in places like `risk.py`, but just making them forward directly to the qrisk functions? This is similar to what we did in pyfolio, where we kept the existing functions and made them forward directly to qrisk. To make it clear we'd change this in the future, we added deprecation warnings.

etc/requirements.txt
```diff
@@ -62,3 +62,6 @@ sortedcontainers==1.4.4
 intervaltree==2.1.0
 
 cachetools==1.1.5
+
+# For financial risk calculations
+qrisk>=0.1.4
```
richafrank: We should peg the version to exactly 0.1.4.  Currently we use this like a constraints file for our environments, and setup.py does some replacement to make pip installing zipline require `>=` the versions specified here.

zipline/finance/risk/period.py
```diff
@@ -208,38 +231,7 @@ def calculate_period_returns(self, returns):
...
-
-    def calculate_information(self):
-        """
-        http://en.wikipedia.org/wiki/Information_ratio
-        """
-        return information_ratio(self.algorithm_returns,
-                                 self.benchmark_returns)
-
-    def calculate_beta(self):
+    def calculate_covariance(self):
```
ehebert: I believe that we were only calculating covariance as an input to pass into the beta calculation.
With `pyrisk` now handling the calculation, we may be able to remove the covariance calculation from Zipline completely.

(I checked the `to_dict` methods in `risk.cumulative` and `risk.period` and it does not appear there, meaning it is not exposed in risk packets.)

ehebert:
@abhijeetkalyan `risk.py` was intended to share common functions between `period.py` and `cumulative.py`. An advantage of the current PR is fewer moving pieces, with `qrisk` serving the role formerly done by `risk.py`.

ehebert:
@ahgnaw this looks good to me. (Glad to have this old code replaced!)

Two notes:
- Looks like some stray `from . import answer_key` are still present, breaking the build.
- Left a comment about removing `covariance`.

conda/qrisk/meta.yaml
```diff
@@ -0,0 +1,27 @@
+package:
```
llllllllll: this should probably live in the qrisk package.

tests/risk/test_risk_cumulative.py
```diff
@@ -38,86 +44,100 @@ def init_instance_fixtures(self):
...
+                BENCHMARK_BASE,
+                0.0
+            )
+
+    def test_algorithm_volatility(self):
+        np.testing.assert_equal(
+            len(self.algo_returns),
+            len(self.cumulative_metrics.algorithm_volatility)
+        )
+        np.testing.assert_equal(
```
ssanderson: If all we want to test is that an array contains floats, the right thing to do is `assert array.dtype == np.dtype('float64')`.  Separately, is this test even worth having if it's making no assertions about the output content?  Ditto most of the rest of the tests in this suite.

tests/risk/test_risk_cumulative.py
```diff
@@ -38,86 +44,100 @@ def init_instance_fixtures(self):
...
+            all(isinstance(x, float)
+                for x in self.cumulative_metrics.sharpe),
+            True)
+
+    def test_downside_risk(self):
+        np.testing.assert_equal(
+            len(self.algo_returns),
+            len(self.cumulative_metrics.downside_risk)
+        )
+        np.testing.assert_equal(
```
ssanderson: `assert_equal(expr, True)` is just `assert expr`.

tests/risk/test_risk_cumulative.py
ahgnaw: The output content is being tested in qrisk. We could have multiple locations that test the correctness of the underlying function, but that might get confusing down the line. We want to still test the class properties.

twiecki:
This is my favorite PR.

richafrank:
@ahgnaw Would you add a bullet to the latest [whatsnew] (https://github.com/quantopian/zipline/blob/e09fadb7e7d68ba1ca12bd542c84482039bbef5d/docs/source/whatsnew/1.0.2.txt) that describes this change?

ahgnaw:
@richafrank Can do

docs/source/whatsnew/1.0.2.txt
```diff
@@ -36,6 +36,12 @@ Enhancements
   produced a True on N or more days in the previous ``window_length``
   days (:issue:`1367`).
 
+- Remove risk-free rate from sharpe ratio calculation. (:issue:`853`).
```
richafrank: Seems like this fixed other issues with the sharpe ratio (mean versus cumulative returns, etc)?  Should we mention those and/or move this to Bug Fixes?

docs/source/whatsnew/1.0.2.txt
```diff
@@ -36,6 +36,12 @@ Enhancements
...
+
+- Use external library empyrical for risk calculations. Empyrical unifies risk
+  metric calculations between pyfolio and zipline. Empyrical adds custom
+  annualization options for returns of custom frequencies.
```
richafrank: Looks like a good place to reference https://github.com/quantopian/zipline/issues/855.

etc/requirements.txt
```diff
@@ -62,3 +62,6 @@ sortedcontainers==1.4.4
 intervaltree==2.1.0
 
 cachetools==1.1.5
+
+# For financial risk calculations
+empyrical>=0.1.9
```
richafrank: Can we pin this to an exact version, like all the others?

richafrank: Our setup.py should already convert this to a `>=`, so you can pip install zipline with a newer version of empyrical.

docs/source/whatsnew/1.0.2.txt
ahgnaw: Good point. I'll include changes to other metrics as well.

ahgnaw:
https://github.com/quantopian/qexec/issues/8627

PR #1326, opened 2016-07-14
MAINT: Remove unused serialization code.

ehebert:
Remove code and modules left over from larger removal of serialization.


jbredeche:
@ehebert should we merge this?

ehebert:
Closing this to prevent accidental merge.

I believe we should still remove this from Zipline, but an internal library depends on it.

PR #1328, opened 2016-07-14
PERF: Speed up event utils test case.

ssanderson:
Just take a sample of all 5000 permutations of 7 dates.


mtydykov:
Nice, huge performance improvement on my machine. LGTM

PR #1330, opened 2016-07-15
Fix daily bar writer

jbredeche:
Make daily bar readers take advantage of the `TradingCalendar`.

Make `USEquityPricingLoader` use the same calendar.

Both the reader and writer were updated, but the reader maintains backwards compatibility.  The test examples have *not* had their data updated, to ensure that we can read old data.

### DIFF
--- a/zipline/data/us_equity_pricing.py
+++ b/zipline/data/us_equity_pricing.py
@@ -40,12 +40,11 @@ from numpy import (
 )
 from pandas import (
     DataFrame,
-    DatetimeIndex,
     read_csv,
     Timestamp,
     NaT,
     isnull,
-)
+    DatetimeIndex)
 from pandas.tslib import iNaT
 from six import (
     iteritems,
@@ -53,6 +52,7 @@ from six import (
     viewkeys,
 )
 
+from zipline.utils.calendars import get_calendar
 from zipline.utils.functional import apply
 from zipline.utils.preprocess import call
 from zipline.utils.input_validation import (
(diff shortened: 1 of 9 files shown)


jbredeche:
cc @ehebert @yankees714

zipline/data/us_equity_pricing.py
```diff
@@ -204,9 +189,34 @@ class BcolzDailyBarWriter(object):
         'volume': float64,
     }
 
-    def __init__(self, filename, sessions, calendar):
+    def __init__(self, filename, calendar, start_session, end_session):
+        """
```
ssanderson: Why did we move this docstring?  The convention pretty much everywhere I've seen is to put the docs for `__init__` at the top of the class.

zipline/data/us_equity_pricing.py
```diff
@@ -363,10 +375,13 @@ def iterator(iterator=iterator, assets=set(assets)):
         full_table.attrs['first_trading_day'] = (
             earliest_date if earliest_date is not None else iNaT
         )
+
         full_table.attrs['first_row'] = first_row
         full_table.attrs['last_row'] = last_row
         full_table.attrs['calendar_offset'] = calendar_offset
-        full_table.attrs['calendar'] = sessions.asi8.tolist()
+        full_table.attrs['calendar_str'] = self._calendar.name
```
ssanderson: I might call this `'calendar_name'`?

zipline/data/us_equity_pricing.py
jbredeche: Including the parameters to `__init__`?  Seems strange to have them separate from the method that uses them.

jbredeche: I'll move it back.

zipline/data/us_equity_pricing.py
```diff
@@ -474,7 +497,20 @@ def _table(self):
 
     @lazyval
     def _calendar(self):
```
ssanderson: Should we rename this to `sessions` or something like it?  It's confusing to have both `_calendar` and `trading_calendar`.

zipline/data/us_equity_pricing.py
```diff
@@ -387,6 +402,10 @@ def spot_price(self, sid, day, colname):
     def last_available_dt(self):
         pass
 
+    @abstractproperty
+    def trading_calendar(self):
```
ssanderson: Can we document what's expected to be provided by subclasses here?

jbredeche: yup, did so

PR #1333, opened 2016-07-18
BUG: Fixes issue with offsetting days in days_at_time calendar helper

yankees714:
Offsetting days in the local timezone can cause issues when there are discontinuities, such as at the start of DST. So we instead do this offset in UTC, then localize to the local timezone once dealing with times.

### DIFF
--- a/zipline/utils/calendars/trading_calendar.py
+++ b/zipline/utils/calendars/trading_calendar.py
@@ -21,7 +21,7 @@ from pandas import (
     DataFrame,
     date_range,
     DatetimeIndex,
-    DateOffset
+    Timedelta
 )
 from pandas.tseries.offsets import CustomBusinessDay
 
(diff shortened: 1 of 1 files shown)


zipline/utils/calendars/trading_calendar.py
```diff
@@ -677,8 +677,13 @@ def days_at_time(days, t, tz, day_offset=0):
     day_offset : int
         The number of days we want to offset @days by
     """
-    days = DatetimeIndex(days).tz_localize(None).tz_localize(tz)
+
+    # Offset days in UTC to avoid discontinuities due to DST, before
+    # localizing to tz.
+    days = DatetimeIndex(days).tz_localize(None).tz_localize('UTC')
```
ssanderson: What bug is this fixing?  Do we need a regression test for this?

jbredeche: Yes, I'm adding the test on my branch (building the test framework for the CME calendar (and others) there).

It's fixing this problem:
```
(Pdb++) calendar.schedule.loc['2004-04-05']
market_open    2004-04-03 23:01:00
market_close   2004-04-05 22:00:00
Name: 2004-04-05 00:00:00+00:00, dtype: datetime64[ns]
```

yankees714: Which with this change is corrected to:
```
(Pdb++) calendar.schedule.loc['2004-04-05']
market_open    2004-04-04 22:01:00
market_close   2004-04-05 22:00:00
Name: 2004-04-05 00:00:00+00:00, dtype: datetime64[ns]
```

ssanderson: Is that a bug in this function, or are we working around a pandas bug?  I don't understand how the changes here fix that problem.

yankees714: Looks like prior to https://github.com/pydata/pandas/pull/10744, there were issues with using `DateOffset` in this context. `Timedelta` seems to be a better choice.

PR #1335, opened 2016-07-19
ENH: ICE calendar, more calendar tests, cleaned up CME calendar

jbredeche:
Added tests for CME calendar.

Added ICE calendar (and tests).


zipline/utils/calendars/exchange_calendar_ice.py
```diff
@@ -0,0 +1,78 @@
...
+                USLaborDay,
+                USThanksgivingDay
+            ]))
+        ]
+
+    @property
+    def adhoc_holidays(self):
+        # ICE was only closed on the first day of the Hurricane Sandy closings
+        # (was not closed on 2012-10-30)
+        return [Timestamp('2012-10-29', tz='UTC')]
```
yankees714: Just want to verify, CME includes `USNationalDaysofMourning` here, but ICE doesn't?

zipline/utils/calendars/trading_calendar.py
```diff
@@ -176,7 +256,7 @@ def next_open(self, dt):
             The UTC timestamp of the next open.
         """
         idx = next_divider_idx(self.market_opens_nanos, dt.value)
-        return self.schedule.market_open[idx].tz_localize('UTC')
+        return pd.Timestamp(self.market_opens_nanos[idx], tz='UTC')
```
yankees714: What's this change for?

zipline/utils/calendars/trading_calendar.py
```diff
@@ -645,15 +725,15 @@ def _special_dates(self, calendars, ad_hoc_dates, start_date, end_date):
 
     def _special_opens(self, start, end):
         return self._special_dates(
-            self.special_opens_calendars,
+            self.special_opens,
```
yankees714: Seems a little weird that we're using `self.special_opens` within `self._special_opens`, maybe there's a better name for one of these?

tests/calendars/test_ice_calendar.py
```diff
@@ -0,0 +1,80 @@
...
+            next_day = pd.Timestamp(date_pair[1], tz='UTC')
+            the_open = self.calendar.schedule.loc[next_day].market_open
+
+            localized_open = the_open.tz_localize("UTC").tz_convert(
+                self.calendar.tz
+            )
+
+            self.assertEqual(
+                date_pair[0],
+                (localized_open.year, localized_open.month, localized_open.day)
```
yankees714: Probably worth checking the time here too

zipline/utils/calendars/trading_calendar.py
jbredeche: sorry, I should have commented, or put it in a separate commit - this is a performance improvement.  The new line runs ~8x faster than the old line.

zipline/utils/calendars/exchange_calendar_ice.py
jbredeche: good catch.  I can't find an explicit mention online of ICE being closed for Reagan and Ford's deaths, but I found enough hints of "other futures markets will do the same" that it's a safe assumption that ICE was closed.

zipline/utils/calendars/trading_calendar.py
jbredeche: I renamed `_special_opens` to `_calculate_special_opens`, etc.  Hope that helps.

tests/calendars/test_ice_calendar.py
jbredeche: Agreed.  Added hour/minute check, and then moved this test to the base test suite so it's hit in all the calendar tests.

PR #1337, opened 2016-07-20
Capital Changes Refactoring

lianga888:
- Allow capital changes to be defined by a target value, in addition to the currently support input of deltas. The delta will be calculated on the portfolio value at the dt of the capital change
- Refactor the checking, calculating and processing of capital changes to better support upstream processes

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -408,7 +408,7 @@ class TradingAlgorithm(object):
 
         self.benchmark_sid = kwargs.pop('benchmark_sid', None)
 
-        # A dictionary of capital change values keyed by timestamp
+        # A dictionary of capital changes keyed by timestamp
         self.capital_changes = kwargs.pop('capital_changes', {})
 
     def init_engine(self, get_loader):
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -408,8 +408,12 @@ class TradingAlgorithm(object):
 
         self.benchmark_sid = kwargs.pop('benchmark_sid', None)
 
-        # A dictionary of capital changes keyed by timestamp
+        # A dictionary of capital changes, keyed by timestamp, indicating the
+        # target/delta of the capital changes, along with values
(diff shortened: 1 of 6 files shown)


zipline/finance/performance/period.py
```diff
@@ -550,8 +548,8 @@ def as_account(self):
             getattr(self, 'day_trades_remaining', float('inf'))
         account.leverage = getattr(self, 'leverage',
                                    period_stats.gross_leverage)
-        account.net_leverage = period_stats.net_leverage
-
+        account.net_leverage = getattr(self, 'net_leverage',
```
ssanderson: When is this not set?

zipline/algorithm.py
```diff
@@ -785,6 +789,70 @@ def _create_daily_stats(self, perfs):
...
+
+            log.info('Processing capital change to target %s at %s. Capital '
+                     'change delta is %s' % (capital_change['value'], dt,
+                                             capital_change_amount))
+        elif capital_change['type'] == 'delta':
+            capital_change_amount = capital_change['value']
+            log.info('Processing capital change of delta %s at %s'
+                     % (capital_change_amount, dt))
+        else:
+            log.error("Capital change %s does not indicate a valid type "
```
jdricklefs: What's your thinking on ignoring errors here? This seems pretty serious to me.

Or alternatively, what do you think of moving this validation of the given capital changes earlier in the algorithm's lifecycle so the error condition is recognized before any simulation processing?

zipline/finance/performance/period.py
lianga888: I'm not sure I follow, but previously we always calculated `net_leverage` from the positions zipline records in the position tracker. This change allows us to get a value from the broker, just like the other account attributes

zipline/algorithm.py
lianga888: There doesn't seem to be much validation for similar inputs done in zipline itself. In this case, upstream processes do any necessary validation before the simulation

PR #1340, opened 2016-07-21
ENH: Change daily mode to use last minute of session instead of session label

jbredeche:
Before this change, when a simulation runs with `data_frequency="daily"`, the algorithm's clock would emit session labels (midnight utc).  In other words, if the algorithm ever used `TradingAlgorithm.get_datetime()`, it would be midnight UTC.

This is misleading because at midnight UTC, the algorithm can get the daily bar for an asset that represents (on a NYSE calendar) data that hasn't technically happened yet (since the trading hours haven't started for that day yet).

As we're moving towards supporting various trading calendars, this problem has become trickier.  For example, `data.can_trade` is going to be extended to check if the asset's exchange is open at the moment of the order.  This is a lot easier to answer and reason about if the order's `dt` is a trading minute, not a trading session.

With this PR, daily backtests now run at the last minute of each trading session (for NYSE calendar, that means 4PM Eastern on most trading days).  All orders are timestamped at 4pm (and start getting filled the next day, just like before).

Users don't need to change anything about the data they provide (or via the Quandl, etc bundles).  Daily data is still timestamped midnight internally, and now we convert to the session label before fetching the data when we're in daily mode.

@richafrank this is the design we discussed.

### DIFF
--- a/tests/pipeline/test_pipeline_algo.py
+++ b/tests/pipeline/test_pipeline_algo.py
@@ -26,6 +26,7 @@ from pandas import (
     Series,
     Timestamp,
 )
+from pandas.tseries.tools import normalize_date
 from six import iteritems, itervalues
 
 from zipline.algorithm import TradingAlgorithm
@@ -512,7 +513,7 @@ class PipelineAlgorithmTestCase(WithBcolzEquityDailyBarReaderFromCSVs,
             attach_pipeline(pipeline, 'test')
 
         def handle_data(context, data):
-            today = get_datetime()
+            today = normalize_date(get_datetime())
             results = pipeline_output('test')
             expect_over_300 = {
                 AAPL: today < self.AAPL_split_date,
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -3723,12 +3723,18 @@ class TestEquityAutoClose(WithTmpDir, WithTradingCalendar, ZiplineTestCase):
         transactions = output['transactions']
         initial_fills = transactions.iloc[1]
         self.assertEqual(len(initial_fills), len(assets))
+
+        last_minute_of_session = \
+            self.trading_calendar.open_and_close_for_session(
+                self.test_days[1]
+            )[1]
+
         for sid, txn in zip(sids, initial_fills):
             self.assertDictContainsSubset(
                 {
                     'amount': order_size,
                     'commission': None,
-                    'dt': self.test_days[1],
+                    'dt': last_minute_of_session,
                     'price': initial_fill_prices[sid],
                     'sid': sid,
                 },
@@ -3795,15 +3801,17 @@ class TestEquityAutoClose(WithTmpDir, WithTradingCalendar, ZiplineTestCase):
                 context.portfolio.cash == context.portfolio.starting_cash
             )
 
-            now = context.get_datetime()
+            today_session = self.trading_calendar.minute_to_session_label(
+                context.get_datetime()
+            )
 
-            if now == first_asset_end_date:
+            if today_session == first_asset_end_date:
                 # Equity 0 will no longer exist tomorrow, so this order will
                 # never be filled.
                 assert len(context.get_open_orders()) == 0
                 context.order(context.sid(0), 10)
                 assert len(context.get_open_orders()) == 1
-            elif now == first_asset_auto_close_date:
+            elif today_session == first_asset_auto_close_date:
                 assert len(context.get_open_orders()) == 0
 
         algo = TradingAlgorithm(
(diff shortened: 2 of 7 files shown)


tests/test_algorithm.py
```diff
@@ -3665,12 +3666,18 @@ def test_daily_delisted_equities(self,
         transactions = output['transactions']
         initial_fills = transactions.iloc[1]
         self.assertEqual(len(initial_fills), len(assets))
+
+        last_minute_of_session = \
+            self.trading_calendar.open_and_close_for_session(
+                self.test_days[1]
+            )[1]
```
ssanderson: I've noticed that we often want just the open or just the close, and we end up calling this to get both and then ditching one.  Maybe we need `get_session_open` and `get_session_close`?

jbredeche: yeah, @richafrank suggested the same.  Seems reasonable to me.

tests/pipeline/test_pipeline_algo.py
```diff
@@ -26,6 +26,7 @@
     Series,
     Timestamp,
 )
+from pandas.tseries.tools import normalize_date
```
richafrank: Was a surprise to see this import, but it makes sense for backward compat.

tests/test_algorithm.py
```diff
@@ -3737,15 +3744,15 @@ def handle_data(context, data):
                 context.portfolio.cash == context.portfolio.starting_cash
             )
 
-            now = context.get_datetime()
+            today_session = normalize_date(context.get_datetime())
```
richafrank: Should this be a session label or a normalized date?

zipline/_protocol.pyx
```diff
@@ -226,6 +229,11 @@ cdef class BarData:
             dt = \
                 self.data_portal.trading_calendar.previous_minute(dt)
 
+        if self._daily_mode:
+            # if we're in daily mode, take the given dt (which is the last
+            # minute of the session) and get the session label for it.
+            dt = self.data_portal.trading_calendar.minute_to_session_label(dt)
```
richafrank: Hard for me to know if this is the right place for this translation.  Would you mind adding some doco to this method to explain who the consumers are and what the expected return values are?

zipline/algorithm.py
```diff
@@ -492,25 +489,32 @@ def _create_clock(self):
         """
         If the clock property is not set, then create one based on frequency.
         """
+        trading_o_and_c = self.trading_calendar.schedule.ix[
+            self.sim_params.sessions]
+        market_closes = trading_o_and_c['market_close'].values.astype(
+            'datetime64[ns]').astype(np.int64)
```
richafrank: Does the first `astype` do something useful, if we immediately `astype` the result?

richafrank: (Same question below)

zipline/data/data_portal.py
```diff
@@ -773,19 +773,20 @@ def get_spot_value(self, asset, field, dt, data_frequency):
         if field not in BASE_FIELDS:
             raise KeyError("Invalid column: " + str(field))
 
+        session_label = self.trading_calendar.minute_to_session_label(dt)
```
richafrank: I was hoping we wouldn't have to do this translation in multiple places, but to abstract it away in one location.  Any chance that's possible?

zipline/data/data_portal.py
```diff
@@ -773,19 +773,20 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
+
         if dt < asset.start_date or \
-                (data_frequency == "daily" and dt > asset.end_date) or \
+                (data_frequency == "daily" and
+                    session_label > asset.end_date) or \
```
richafrank: Why do we convert this dt to a label?  Seems like it was using the actual minute before?

jbredeche: Asset start/end dates are always session labels.  It's safer/better to compare session labels with session labels, I think.

zipline/_protocol.pyx
jbredeche: yup, will add.

zipline/algorithm.py
jbredeche: removed the intermediate `astype`, good catch.

zipline/data/data_portal.py
jbredeche: Not sure we can, easily.  We were already normalizing the date below (in minute mode check), and we should be getting the session label anyway.

tests/test_algorithm.py
jbredeche: Both.  that test is using the NYSE calendar, so they're equivalent.

zipline/data/data_portal.py
richafrank: Ah ok, wasn't sure the start/end dates were labels.

tests/test_algorithm.py
richafrank: Should we try to be explicit by using the calendar to get the label instead of normalizing?  Then we could even parameterize the test over various calendars.

jbredeche: sure, I made it more explicit.

tests/test_algorithm.py
```diff
@@ -3795,15 +3802,17 @@ def handle_data(context, data):
                 context.portfolio.cash == context.portfolio.starting_cash
             )
 
-            now = context.get_datetime()
+            today_session = get_calendar("NYSE").minute_to_session_label(
```
richafrank: Should this use `self.trading_calendar`?

jbredeche: yup.  just made the change.

PR #1341, opened 2016-07-21
ENH: Removes dependence of BcolzMinuteBarReader on 390 minutes per day

yankees714:
Adds a `minutes_per_day` field to `BcolzMinuteBarMetadata`, so that the `minutes_per_day` value passed to the `BcolzMinuteBarWriter` is written as a part of the minute bar metadata. This allows the reader to share this value, instead of always assuming 390.

Also rebuilds the example data to incorporate this format change.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -131,19 +131,28 @@ class BcolzMinuteBarMetadata(object):
                                            unit='m',
                                            utc=True)
             ohlc_ratio = raw_data['ohlc_ratio']
-            return cls(first_trading_day,
-                       market_opens,
-                       market_closes,
-                       ohlc_ratio)
+            minutes_per_day = raw_data['minutes_per_day']
+            return cls(
+                first_trading_day,
+                market_opens,
+                market_closes,
+                ohlc_ratio,
+                minutes_per_day,
+            )
 
-    def __init__(self, first_trading_day,
-                 market_opens,
-                 market_closes,
-                 ohlc_ratio):
+    def __init__(
+        self,
+        first_trading_day,
+        market_opens,
+        market_closes,
+        ohlc_ratio,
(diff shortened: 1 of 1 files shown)


richafrank:
If I'm a user of (standalone) zipline, and I've previously downloaded a bunch of pricing data to bcolz files, will upgrading the version of zipline mean I can't read those pricing files anymore?

ehebert:
@richafrank, it would. What do you think about adding a fallback value of `390` with a deprecation warning saying something akin to `No minutes_per_day found, defaulting to 390.`

richafrank:
@ehebert That makes sense.  If we have a way to add the metadata to the files in place, then the warning could also tell users to do that.

ehebert:
Also, this seems like a case where we should have a file format upgrade path?

richafrank:
Yup, exactly.

ehebert:
@richafrank how do you feel about not building a full upgrade system for the file format for the benefit of this PR, but setting version `1` for incoming format, and then adding a check that looks for the `version` field in the metadata and if it is not found version `0` is found.

Then if version `1` is present look for `minute_per_day` else default to 390 (Equities).

richafrank:
@ehebert Sounds good to me.  We can build more later.

ehebert:
@richafrank https://github.com/quantopian/zipline/pull/1341/commits/d0704fe3f569555b3de108818b739b09cc76d735 is a simple implementation of the fallback to version 0 and providing a default for minutes per day. Mind taking a look?

richafrank:
Makes sense to me!  Thanks @ehebert

ehebert:
@yankees714 the updated example data had an encoding issue with Python 3. With the addition of having a fallback value for the data, the existing data is usable, so I edited your patch to remove change to the sample data.

PR #1342, opened 2016-07-21
ENH: Deal with bad parameters to `get_datetime`

jbredeche:


zipline/algorithm.py
```diff
@@ -1497,9 +1499,15 @@ def get_datetime(self, tz=None):
             # Convert to the given timezone passed as a string or tzinfo.
             if isinstance(tz, string_types):
                 tz = pytz.timezone(tz)
+            elif not isinstance(tz, tzinfo):
```
ssanderson: Can we not just make this an error?

PR #1345, opened 2016-07-22
ENH: Add NotNullFilter.

ssanderson:


llllllllll:
What is wrong with `~self.isnull()`?

ssanderson:
> What is wrong with ~self.isnull()?

Downstream @mtydykov wanted to write a Filter that was `NotNullFilter` with default inputs, and implementing it here and using it for `notnull` saves an intermediate array in the term graph.

llllllllll:
Okay. Do we have the ability to do a graph rewrite rule for `~IsNull` -> `NotNull`?

ssanderson:
> Okay. Do we have the ability to do a graph rewrite rule for ~IsNull -> NotNull?

We're not doing any graph rewriting right now.

PR #1346, opened 2016-07-24
ENH: time/calendar business logic improvements

jbredeche:
I pulled in 03f120d3524434e6eb04759e65f9f394ce14b67b because I needed it on this branch, so once that PR is in, I can rebase that commit out.

The goal of this PR was to augment `data.can_trade` to also check whether the asset's exchange is open.  To do that, I also needed to move `before_trading_start` to fire at 8:45am (ie, possibly an exchange minute) instead of midnight UTC (a session label), so that got put into this PR too.


tests/pipeline/test_pipeline_algo.py
```diff
@@ -512,7 +513,7 @@ def initialize(context):
             attach_pipeline(pipeline, 'test')
 
         def handle_data(context, data):
-            today = get_datetime()
+            today = normalize_date(get_datetime())
```
ehebert: Does this mean that existing pipeline algos will change results without normalizing the date?

zipline/assets/_assets.pyx
```diff
@@ -187,37 +201,58 @@ cdef class Asset:
...
-        cdef int64_t dt_value
         cdef int64_t ref_start
         cdef int64_t ref_end
 
-        if not normalized:
-            dt_value = normalize_date(dt).value
-        else:
-            dt_value = dt.value
-
         ref_start = self.start_date.value
```
richafrank: What calendar's sessions do these boundary values correspond to?  Will futures contracts have start/end dates (session labels) that mean they expire at the close of the exchange they trade on?

zipline/assets/_assets.pyx
```diff
@@ -187,37 +201,58 @@ cdef class Asset:
...
+            The minute to check.
+
+        Returns
+        -------
+        boolean: whether the asset's exchange is open at the given minute.
+        """
+        calendar = self._exchange_trading_calendar_for_asset()
+        return calendar.is_open_on_minute(dt_minute)
+
+    def _exchange_trading_calendar_for_asset(self):
```
richafrank: Naming-wise, `Asset._exchange_trading_calendar_for_asset` strikes me as redundant.

zipline/assets/_assets.pyx
```diff
@@ -42,6 +44,16 @@ from pandas.tslib import normalize_date
...
 CACHE_FILE_TEMPLATE = '/tmp/.%s-%s.v6.cache'
 
+FUTURE_EXCHANGE_MAPPING = {
+    "CFE": "CFE",
+    "CBOT": "CME",
+    "CME": "CME",
+    "COMEX": "CME",
+    "NYMEX": "CME",
+    "ICEUS": "ICE",
+    "NYFE": "ICE"
```
richafrank: Do you think this mapping belongs in zipline, or should the creator of the data have mapped it already?

zipline/data/daily_history_aggregator.py
```diff
@@ -0,0 +1,416 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+import pandas as pd
+
+
+class DailyHistoryAggregator(object):
```
richafrank: Are there changes in here?

jbredeche: no, just moved the file.

zipline/gens/sim_engine.pyx
```diff
@@ -59,28 +64,50 @@ cdef class MinuteSimulationClock:
...
-        for day in self.trading_days:
-            yield day, DAY_START
+        for idx, session in enumerate(self.sessions):
+            yield session, SESSION_START
+
+            bts_minute = self.before_trading_start_minutes[idx]
+            regular_minutes = self.minutes_by_session[session]
+
+            # we have to search anew every session, because there is no
+            # guarantee that any two session start on the same minute
```
richafrank: Could you clarify this?

zipline/gens/sim_engine.pyx
```diff
@@ -59,28 +64,50 @@ cdef class MinuteSimulationClock:
...
 
-            for minute in minutes:
-                yield minute, BAR
-                if minute_emission:
-                    yield minute, MINUTE_END
+                # emit all the minutes after bts_minute
+                for minute in regular_minutes[bts_idx:]:
+                    yield minute, BAR
+                    if minute_emission:
+                        yield minute, MINUTE_END
```
richafrank: Instead of repeating this 3 times over different iterables, would it make sense to factor out a function?

zipline/gens/tradesimulation.py
```diff
@@ -216,15 +214,19 @@ def calculate_minute_capital_changes(dt):
...
+                elif action == SESSION_END:
+                    # End of the session.
+                    if algo.perf_tracker.emission_rate == 'daily':
                         handle_benchmark(normalize_date(dt))
                     execute_order_cancellation_policy()
 
                     yield self._get_daily_message(dt, algo, algo.perf_tracker)
+                elif action == BEFORE_TRADING_START_BAR:
+                    # call before trading start
+                    algo.on_dt_changed(dt)
```
richafrank: If the sim engine might emit bars within a session both before and after BTS, might the dt not change here?

zipline/gens/sim_engine.pyx
```diff
@@ -59,28 +64,50 @@ cdef class MinuteSimulationClock:
...
+            bts_minute = self.before_trading_start_minutes[idx]
+            regular_minutes = self.minutes_by_session[session]
+
+            # we have to search anew every session, because there is no
+            # guarantee that any two session start on the same minute
+            bts_idx = regular_minutes.searchsorted(bts_minute)
+
+            if bts_idx == len(regular_minutes):
+                # before_trading_start is after the last close, so don't emit
+                # it
```
richafrank: If a session runs from BTS time to BTS time the next day, would we not run BTS?

zipline/assets/_assets.pyx
```diff
@@ -187,37 +201,58 @@ cdef class Asset:
...
+
+        Raises KeyError if the asset's exchange calendar cannot be found.
+
+        Returns
+        -------
+        The asset's exchange's trading calendar.
+        """
+        if isinstance(self, Equity):
+            # FIXME: probably too Quantopian-specific
+            return get_calendar("NYSE")
```
richafrank: Are we planning to move the calendars into the asset db + materialized instances?  My expectation was that we would look up the exchange on the asset.

zipline/algorithm.py
```diff
@@ -497,28 +497,33 @@ def _create_clock(self):
         trading_o_and_c = self.trading_calendar.schedule.ix[
             self.sim_params.sessions]
         market_closes = trading_o_and_c['market_close'].values.astype(np.int64)
+        minutely_emission = False
```
richafrank: Looks like this value is only used in the `else`.  Could move it there for parallel constructions between the `if` and the `else`, but not a big deal.

zipline/_protocol.pyx
```diff
@@ -460,15 +462,25 @@ cdef class BarData:
...
     cdef bool _can_trade_for_asset(self, asset, dt, adjusted_dt, data_portal):
-        if asset._is_alive(dt, False):
-            # is there a last price?
-            return not np.isnan(
-                data_portal.get_spot_value(
-                    asset, "price", adjusted_dt, self.data_frequency
-                )
-            )
+        session_label = normalize_date(dt) # FIXME
+        if not asset._is_alive_for_session(session_label):
```
richafrank: Naming-wise, looks like these underscore methods of `Asset` are actually public, in that consumers outside `Asset` call them?

zipline/assets/_assets.pyx
```diff
@@ -187,37 +201,58 @@ cdef class Asset:
         """
         return cls(**dict_)
 
-    def _is_alive(self, dt, bool normalized):
+    def _is_alive_for_session(self, session_label):
```
richafrank: Looks like each call site for this function calls `session_for_minute` immediately before.  Should we move that into here?

zipline/assets/_assets.pyx
```diff
@@ -89,6 +101,8 @@ cdef class Asset:
         self.first_traded = first_traded
         self.auto_close_date = auto_close_date
 
+
+
```
richafrank: I guess we don't have flake8 for cython, but probably don't want these extra lines.

tests/warnings_catcher.py
```diff
@@ -0,0 +1,32 @@
+from warnings import catch_warnings, WarningMessage
+
+
+class WarningsCatcher(catch_warnings):
+    """
+    Subclass of warnings.catch_warnings that takes a list of warning types to
+    ignore.
```
richafrank: Should we use `warnings`'s filters to manage this instead?

tests/warnings_catcher.py
```diff
@@ -0,0 +1,32 @@
...
+            raise RuntimeError("Cannot enter %r twice" % self)
+        self._entered = True
+        self._filters = self._module.filters
+        self._module.filters = self._filters[:]
+        self._showwarning = self._module.showwarning
+        if self._record:
+            log = []
+
+            def showwarning(*args, **kwargs):
+                if args[1] in self._types_to_ignore:
```
richafrank: Looks like this is the only intended difference from the base class, but there are other differences from the python3 version.  Since the intended difference is governed by `self._record`, which is always `True` for the subclass, I recommend actually setting it to `False`, and using:

```python
def __enter__(self):
    super(WarningsCatcher, self).__enter__()

    log = []
    def showwarning(*args, **kwargs):
        if args[1] in self._types_to_ignore:
            return
        log.append(WarningMessage(*args, **kwargs))

    self._module.showwarning = showwarning
    return log
```
which isolates the differences from the base class.

zipline/assets/_assets.pyx
jbredeche: Yes.  asset lifetimes are defined as a start and end session.  and futures contracts will expire at the end of a session.

zipline/assets/_assets.pyx
jbredeche: agreed, updated.

zipline/gens/sim_engine.pyx
jbredeche: Sure.  A trading calendar can have late opens, which means that for two trading sessions, the index of the `bts_minute` in the sessions' minutes can be different.

richafrank: Ah ok cool.  Could you include that in the comment? - just wasn't clear to me.  (Also typo "two session start")

zipline/gens/sim_engine.pyx
jbredeche: done

zipline/gens/tradesimulation.py
jbredeche: yes.  this just ensures that the algo dt is right, even if it's the same dt.

this ensures that the portfolio and account info are up to date if they are being accessed in `before_trading_start`.

zipline/assets/_assets.pyx
jbredeche: indeed.

zipline/gens/sim_engine.pyx
jbredeche: I made this more explicit, by checking if `bts_minute > regular_minutes[-1]`

zipline/_protocol.pyx
jbredeche: good call, they are public, fixed the naming.

zipline/assets/_assets.pyx
jbredeche: I prefer keeping this method simple - it's responsible for taking a session and returning whether the asset was alive for that session.

tests/warnings_catcher.py
jbredeche: good call, done

zipline/data/daily_history_aggregator.py
ehebert: Can we do this in a separate PR or commit?

ehebert: There is a change, `normalize_date` was changed to using `session_label`, working on a quick PR now to break out the move to a sep module so that difference is shown.

ehebert: https://github.com/quantopian/zipline/pull/135

zipline/assets/_assets.pyx
jbredeche: yup.  The asset db now has the canonical exchange name ("NASDAQ", "CME", etc) and we use that to look up the calendar.

PR #1347, opened 2016-07-25
BUG: Implement `sessions` property for `PanelDailyBarReader`

jbredeche:
Also, renamed it from `_sessions_` to `sessions` and defined an
abstractproperty in `DailyBarReader`.


ssanderson:
LGTM.  Longer term we probably need explicit tests for `USEquityDailyHistoryLoader` using `PanelDailyBarReader`, but for a quick fix this is good I think.

llllllllll:
ping @nathanwolfe

PR #1350, opened 2016-07-26
BUG: run_algorithm errors if market data passed as DataFrame

mannau:
- line 318, in run_algorithm:  ValueError: The truth value of a
DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or
a.all().
- line 159, in _run_algorithm: UnboundLocalError: local variable 'env'
referenced before assignment

The following code reproduces the error(s):

```python
from zipline import run_algorithm
from datetime import datetime
from pytz import UTC
from zipline.data.loader import load_from_yahoo
from zipline.api import order, symbol, record

def initialize(context):
    pass

    order(symbol('AAPL'), 10)

if __name__ == '__main__':
    start = datetime(2000, 1, 1, tzinfo=UTC)
    end = datetime(2016, 1, 1, tzinfo=UTC)
    capital_base = 100000
    
    data = load_from_yahoo(stocks=['AAPL'], indexes={}, adjusted=True)
    
    results = run_algorithm(
                  start,
                  end,
                  initialize,
                  capital_base,
                  data=data,
                  bundle=None,
                  default_extension=False
                  )
```

### DIFF
--- a/zipline/utils/run_algo.py
+++ b/zipline/utils/run_algo.py
@@ -12,7 +12,7 @@ try:
     PYGMENTS = True
 except:
     PYGMENTS = False
-from toolz import valfilter, concatv
+from toolz import concatv
 
 from zipline.algorithm import TradingAlgorithm
 from zipline.data.bundles.core import load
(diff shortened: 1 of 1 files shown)


llllllllll:
Thanks for the bugfix! https://github.com/quantopian/zipline/pull/1313 fixes this issue to support passing panels. The implementations are slightly different here. I think think that we do need to create the sim_params directly to properly support the dataframe/panel case. We should be merging the other pr soon which should clean this up.

mannau:
ok, thanks for the reference!

mannau:
Already fixed by #1313

PR #1352, opened 2016-07-27
MAINT: Move daily aggregator to own module.

ehebert:
Break out the daily history aggregator into its own module, instead of
being collocated with DataPortal.


richafrank:
@ehebert Is this purely a move?  If so, then 👍

jbredeche:
lgtm, thanks

PR #1353, opened 2016-07-27
ENH: Yield capital change information

lianga888:
Yield packets for capital changes so downstream processes can consume then

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -847,14 +847,16 @@ class TradingAlgorithm(object):
         self.perf_tracker.prepare_capital_change(is_interday)
 
         if capital_change['type'] == 'target':
-            capital_change_amount = capital_change['value'] - \
+            target = capital_change['value']
+            capital_change_amount = target - \
                 self.updated_portfolio().portfolio_value
             self.portfolio_needs_update = True
 
             log.info('Processing capital change to target %s at %s. Capital '
-                     'change delta is %s' % (capital_change['value'], dt,
+                     'change delta is %s' % (target, dt,
                                              capital_change_amount))
         elif capital_change['type'] == 'delta':
+            target = None
             capital_change_amount = capital_change['value']
             log.info('Processing capital change of delta %s at %s'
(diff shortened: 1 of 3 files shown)


zipline/gens/tradesimulation.py
```diff
@@ -215,8 +209,17 @@ def calculate_minute_capital_changes(dt):
 
             for dt, action in self.clock:
                 if action == BAR:
+                    cc = calculate_minute_capital_changes(dt)
```
ssanderson: Rather than making this a separate step, could we just make `every_bar` return an iterator of packets to yield?  Same with `once_a_day` below.

zipline/algorithm.py
```diff
@@ -853,10 +850,20 @@ def calculate_capital_changes(self, dt, emission_rate, is_interday):
...
-        self.capital_change_deltas.update({dt: capital_change_amount})
         self.perf_tracker.process_capital_change(capital_change_amount,
                                                  is_interday)
 
+        packet = {
+            'capital_change': [
+                {'date': dt,
+                 'type': 'cash',
+                 'target': capital_change['value'] if
+                 capital_change['type'] == 'target' else None,
```
richafrank: Since we already check these cases above, would it make sense to construct `packet` there instead of doing the same check for 'type' here?

zipline/algorithm.py
```diff
@@ -410,9 +410,6 @@ def noop(*args, **kwargs):
         # target/delta of the capital changes, along with values
         self.capital_changes = kwargs.pop('capital_changes', {})
 
-        # A dictionary of the actual capital change deltas, keyed by timestamp
-        self.capital_change_deltas = {}
```
richafrank: Are these the same values that are now being yielded out, so we don't track them here anymore?

lianga888: Correct, I had previously saved the deltas as I thought we might use them, but this attribute doesn't actually get referenced anywhere

richafrank: Ok cool.

PR #1358, opened 2016-08-01
ENH: added smoothing to zipline

thewassermann:
smoothing filter added to zipline

### DIFF
--- a/tests/pipeline/test_filter.py
+++ b/tests/pipeline/test_filter.py
@@ -409,7 +409,7 @@ class FilterTestCase(BasePipelineTestCase):
             window_length = 0
 
         strictly_true_filter = StrictlyTrueFilter(
-            inputs=[InputFilter()],
+            inputs=(InputFilter(), ),
             window_length=self.default_shape[0]
         )
 
--- a/tests/pipeline/test_filter.py
+++ b/tests/pipeline/test_filter.py
@@ -21,6 +21,7 @@ from numpy import (
     ones,
     ones_like,
     putmask,
+    reshape,
     rot90,
     sum as np_sum
 )
@@ -398,10 +399,10 @@ class FilterTestCase(BasePipelineTestCase):
     def test_strictly_true_filter(self):
         from zipline.pipeline.filters import StrictlyTrueFilter
 
-        data = full(self.default_shape, True, dtype=bool)
-        # one column all false
-        data[0, 0] = False
-        data[1, 1] = False
+        data = ~eye(N=self.default_shape[0],
+                    M=self.default_shape[1],
+                    k=1,
+                    dtype=bool)
 
         class InputFilter(Filter):
             inputs = ()
@@ -417,11 +418,10 @@ class FilterTestCase(BasePipelineTestCase):
             initial_workspace={InputFilter(): data}
         )
 
-        expected_result = full(self.default_shape[1], True, dtype=bool)
-        expected_result[0] = False
-        expected_result[1] = False
+        expected_result = full(self.default_shape[1], False, dtype=bool)
+        expected_result[0] = True
         check_arrays(
-            results['Filter'].flatten(),
+            reshape(results['Filter'], expected_result.shape[0]),
             expected_result,
         )
 
--- a/tests/pipeline/test_filter.py
+++ b/tests/pipeline/test_filter.py
@@ -410,7 +410,7 @@ class FilterTestCase(BasePipelineTestCase):
 
         strictly_true_filter = StrictlyTrueFilter(
             inputs=(InputFilter(), ),
-            window_length=self.default_shape[0]
+            window_length=(self.default_shape[0]-1)
         )
 
         results = self.run_graph(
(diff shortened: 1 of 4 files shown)


thewassermann:
making docs now. needed to submit pr first to get issue number

zipline/pipeline/filters/filter.py
```diff
@@ -488,3 +489,18 @@ def _compute(self, arrays, dates, assets, mask):
                 asset=self._asset, start_date=dates[0], end_date=dates[-1],
             )
         return out
+
+
+class SmoothingFilter(CustomFilter):
+    """
+    A Filter that requires its inputs to have
+    been True for the last `window_length` days.
+    An integral part of the Q500US methodology
```
ssanderson: This note doesn't belong here.  A reader of Zipline won't necessarily know what Q500US is, and nothing about the behavior here is specific to that project.

zipline/pipeline/filters/filter.py
```diff
@@ -488,3 +489,18 @@ def _compute(self, arrays, dates, assets, mask):
...
+
+    **Default Inputs**: None
+
+    **Default Window Length**: None
+    """
+
+    def compute(self, today, assets, out, arg):
+        out[:] = (sum_(arg, axis=0) == self.window_length)
```
ssanderson: I would probably just use `arg.sum` here instead of the top-level np.sum.

zipline/pipeline/filters/filter.py
```diff
@@ -488,3 +489,18 @@ def _compute(self, arrays, dates, assets, mask):
                 asset=self._asset, start_date=dates[0], end_date=dates[-1],
             )
         return out
+
+
+class SmoothingFilter(CustomFilter):
```
ssanderson: I think we need a more specific name for this.  There are lots of ways we can implement smoothing.  This is one specific way of doing it.

thewassermann: Here are some ideas:
AllTrueInWindowFilter
AllTrueSmoothingFilter
AllTrueFilter
StrictSmoothingFilter
StrictlyTrueFilter

Any of these work/inspire workable ideas?

tests/pipeline/test_filter.py
```diff
@@ -395,6 +395,36 @@ def test_isfinite(self):
         )
         check_arrays(results['isfinite'], isfinite(data))
 
+    def test_smoothing_filter(self):
+        from zipline.pipeline.filters import SmoothingFilter
+
+        data = full(self.default_shape, True, dtype=bool)
+        # one column all false
```
ssanderson: This comment isn't true, you have an array with the first two diagonal values False.

ssanderson: A good test-case for this is `~np.eye(self.default_shape, dtype=bool)`, which gives you `True` values everywhere except the diagonal.

thewassermann: i updated the code but not the comment apparently, ill use np.eye

tests/pipeline/test_filter.py
```diff
@@ -395,6 +395,36 @@ def test_isfinite(self):
...
+        results = self.run_graph(
+            TermGraph({'smoothing': smoothing_filter}),
+            initial_workspace={InputFilter(): data}
+        )
+
+        expected_result = full(self.default_shape[1], True, dtype=bool)
+        expected_result[0] = False
+        expected_result[1] = False
+        check_arrays(
+            results['smoothing'].flatten(),
```
ssanderson: This is losing information.  If `expected_result` was 4 x 4, and the actual output came back as 2 x 8, this wouldn't catch the error.  Instead of flattening here, you should `reshape` expected_result to the shape that's actually expected.

zipline/pipeline/filters/filter.py
```diff
@@ -488,3 +488,17 @@ def _compute(self, arrays, dates, assets, mask):
...
+    A Filter that requires its inputs to have
+    been True for the last `window_length` days.
+
+    **Default Inputs**: None
+
+    **Default Window Length**: None
+    """
+
+    def compute(self, today, assets, out, arg):
+        out[:] = (sum(arg) == self.window_length)
```
ssanderson: This is badly incorrect:
```
In [1]: data = np.arange(25).reshape(5, 5)

In [2]: data
Out[2]:
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14],
       [15, 16, 17, 18, 19],
       [20, 21, 22, 23, 24]])

In [3]: sum(data)
Out[3]: 300
```

If tests are passing here, then we're missing coverage in a serious way.

thewassermann:
unsure why this is failing the build. have checked all tests and flake8...

thewassermann:
have had a look and run the tests on my machine. looks amazing. ready to merge when the ci tests finish/you give the word

ssanderson:
LGTM

PR #1361, opened 2016-08-01
Point in time assets db again

llllllllll:
We are doing it this time!

There are 2 commits after the revert that should be reviewed now.

### DIFF
--- a/zipline/utils/range.py
+++ b/zipline/utils/range.py
@@ -1,0 +1,284 @@
+import operator as op
+
+from six import PY2
+from toolz import peek
+
+from zipline.utils.functional import foldr
+
+
+if PY2:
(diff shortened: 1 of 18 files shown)


zipline/utils/range.py
```diff
@@ -64,6 +65,15 @@ def __repr__(self):
...
                 (', ' + str(self.step)) if self.step != 1 else '',
             )
+
+        def __hash__(self):
+            return hash((type(self), self.start, self.stop, self.step))
+
+        def __eq__(self, other):
+            return reduce(
+                getattr(self, attr) == getattr(other, attr)
+                for attr in self.__slots__
```
llllllllll: nope, I will add a test for this.

PR #1362, opened 2016-08-02
PERF: lru-dict is a much faster LRU cache

jbredeche:
Just opening for discussion.

On my machine, a simulation that buys a share of AAPL every minute takes ~102 seconds for a year.  This PR makes it take 77 seconds.

In that relatively artificial scenario, a ton of time is being spent in the LRUCache in `minute_bars._open_minute_file`.  

I switched to `lru-dict`, a C implementation of an LRU cache, and things speed up nicely.

EDIT: I added a second commit that puts that fast LRU cache in front of `trading_calendar.minute_to_session_label`, and the same simulation then went from 77 to 70 seconds.

### DIFF
--- a/zipline/utils/calendars/trading_calendar.py
+++ b/zipline/utils/calendars/trading_calendar.py
@@ -674,7 +674,6 @@ class TradingCalendar(with_metaclass(ABCMeta)):
 
         return DatetimeIndex(all_minutes).tz_localize("UTC")
 
-    @preprocess(dt=coerce(pd.Timestamp, attrgetter('value')))
     def minute_to_session_label(self, dt, direction="next"):
         """
         Given a minute, get the label of its containing session.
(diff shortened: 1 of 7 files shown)


ehebert:
https://github.com/quantopian/zipline/pull/1362/commits/df8419b07604634e24e312d29f41d25d156a194c looks like a nice drop-in replacement, so no objections from me.

ssanderson:
Are there conda packages for lru-dict on Windows?  If not, I think we'd need to write a recipe and put it in our channel?

ssanderson:
AppVeyor build failing because of aforemention conda issue: https://ci.appveyor.com/project/Quantopian/zipline/build/1.0.1634/job/s2eybxdkkduw3jq4#L996

richafrank:
We have many cache implementations either in the code or in requirements.  Should we try to consolidate before adding another?

zipline/utils/calendars/trading_calendar.py
```diff
@@ -683,6 +692,22 @@ def minute_to_session_label(self, dt, direction="next"):
         pd.Timestamp (midnight UTC)
             The label of the containing session.
         """
+        if direction == "next":
```
richafrank: The check later for `elif direction != "next":` seems redundant now.

PR #1364, opened 2016-08-02
update readme

sagivo:
remove symbols, print head


ssanderson:
@sagivo thanks for the tweaks!

PR #1366, opened 2016-08-02
TST: Make room for multiple calendars in tests.

ehebert:
When adding fixtures for futures data, there will be a need for multiple
calendars in the fixture ecosystem. e.g. a test that includes both
equities and futures would need an overall calendar which encompasses
both equities and futures; however, the test data for equities should
still still be limited to the bounds set by the NYSE calendar.

Make the fixtures that setup trading calendars and values dervied from
the trading calendar (e.g. trading sessions) accept an iterable of
calendars which need to be created, then populate those values into a
dict keyed by the calendar name.

Change `WithNYSETradingDays` to include sessions in the name,
since we are moving to session as the name for the 'day' unit.

Provide `trading_days` which is really "NYSE trading sessions` on
`WithTradingSessions` for backwards compatibility.

### DIFF
--- a/zipline/testing/fixtures.py
+++ b/zipline/testing/fixtures.py
@@ -3,7 +3,7 @@ from unittest import TestCase
 
 from contextlib2 import ExitStack
 from logbook import NullHandler, Logger
-from six import with_metaclass
+from six import with_metaclass, iteritems
 from toolz import flip
 import pandas as pd
 import responses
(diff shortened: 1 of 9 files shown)


zipline/testing/fixtures.py
```diff
@@ -365,23 +365,44 @@ def init_class_fixtures(cls):
...
     Attributes
     ----------
-    TRADING_CALENDAR_STR : str
-        The identifier of the calendar to use.
+    TRADING_CALENDAR_STRS : iterable
+        iterable of identifiers of the calendars to use.
+    TRADING_CALENDAR_MASTER_STR : str
+        The trading calendar to assign to `trading_calendar`.
+    TRADING_CALENDAR_FOR_ASSET_TYPE : dict
+        A dictionay which maps asset type names to the calendar associated
```
yankees714: dictionary

zipline/testing/fixtures.py
```diff
@@ -365,23 +365,44 @@ def init_class_fixtures(cls):
 
 class WithTradingCalendar(object):
```
yankees714: Might make sense to rename this `WithTradingCalendars`?

zipline/testing/fixtures.py
```diff
@@ -365,23 +365,44 @@ def init_class_fixtures(cls):
...
     """
-    ZiplineTestCase mixing providing cls.trading_calendar as a class-level
-    fixture.
+    ZiplineTestCase mixing providing cls.trading_calendar,
```
yankees714: mixin

yankees714:
Discussed with @ehebert offline, these changes do feel a bit awkward, but the current hierarchy of fixtures doesn't really account for multiple calendars and asset classes. This will allow us to build out some tests of that sort, and then adapt the fixtures to how they're being used by the tests.

Otherwise, just found a couple typos and had a naming question.

ehebert:
Talked to @yankees714 a bit more offline, and I agree that it is awkward but necessary. To reduce some of the awkwardness, going to change from using the dicts `all_trading_calendars` and `all_trading_sessions`, and instead assign `futures_calendar`, `futures_sessions`, `equities_calendar` and `equities_sessions`.

PR #1368, opened 2016-08-02
Lots of symbols

llllllllll:
ran into this using our actual asset data


zipline/utils/range.py
```diff
@@ -32,11 +32,37 @@ def __init__(self, stop, *args):
                 except IndexError:
                     self.step = 1
 
+            if self.step == 0:
+                raise ValueError('range step must not be zero')
+
+        def _high_low_step(self):
```
richafrank: Is this unused?

llllllllll: yeah, sorry

zipline/utils/range.py
```diff
@@ -57,6 +83,38 @@ def __contains__(self, other, _ops=_ops):
...
+                high = self.stop
+            else:
+                low = self.stop
+                high = self.start
+                step = -step
+
+            if low >= high:
+                return 0
+
+            return (high - low - 1) // step + 1
```
richafrank: Nice consolidation of the cases

llllllllll: I should probably comment that the algo was taken from CPython

richafrank:
LGTM!

zipline/utils/range.py
richafrank: That was my answer key as well.

Issue #1369, opened 2016-08-03
Limit orders do not always fill at limit price


Mamady:
Limit orders currently fill at the market price when the limit price has been crossed. This is not live trading behavior - instead when the limit price has been crossed, the order should fill at the limit price (not market price).

Mamady:
Further information with a backtest showing the issue can be seen here: https://www.quantopian.com/posts/simulation-of-non-marketable-limit-orders

luca-s:
#1376 This patch should fix the issue

PR #1370, opened 2016-08-03
Limit orders filled at limit price

Mamady:
Made orders will fill at limit price, not at the market price of the current bar

This patch addresses #1369


Mamady:
This patch does not correctly address the issue - closing.

PR #1372, opened 2016-08-03
TST: Use sum for volume on daily data resample.

ehebert:
Change the mock minute data to no longer use an increasing arange, so
that a days worth of minute data can be summed and fit inside of a
uint32.

This change was required because of working on new test data that looked
like [0, 100, 200, 0, ] which was resulting in a daily rollup of 0 data,
when the coverage needed a non-0 value.

### DIFF
--- a/zipline/data/resample.py
+++ b/zipline/data/resample.py
@@ -1,0 +1,48 @@
+# Copyright 2016 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+_MINUTE_TO_SESSION_OHCLV_HOW = {
+    'open': 'first',
+    'high': 'max',
+    'low': 'min',
+    'close': 'last',
+    'volume': 'sum'
+}
+
+
+def minute_to_session(minute_frame, calendar):
+    """
+    Resample a DataFrame with minute data into the frame expected by a
+    BcolzDailyBarWriter.
+
+    Parameters
+    ----------
+    minute_frame : pd.DataFrame
+        A DataFrame with the columns `open`, `high`, `low`, `close`, `volume`,
+        and `dt` (minute dts)
+    calendar : zipline.utils.calendars.trading_calendar.TradingCalendar
+        A TradingCalendar on which session labels to resample from minute
+        to session.
+
+    Return
+    ------
+    session_frame : pd.DataFrame
+        A DataFrame with the columns `open`, `high`, `low`, `close`, `volume`,
+        and `day` (datetime-like).
+    """
+    # Group minutes into their respective days. Note that this will
+    # create groups for all trading days in the desired range,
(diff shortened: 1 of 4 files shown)


ehebert:
This is required for further development on this branch https://github.com/quantopian/zipline/tree/data-portal-coverage

zipline/testing/fixtures.py
```diff
@@ -614,12 +614,7 @@ def _make_equity_daily_bar_from_minute(cls):
             'high': 'max',
             'low': 'min',
             'close': 'last',
-            # TODO: Change test data so that large minute volumes are not used,
-            # so that 'sum' can be used without going over the uint limit.
-            # When that data is changed, this function can and should be moved
-            # to the `data` module so that loaders and tests can use the same
-            # source from minute logic.
-            'volume': 'last'
```
yankees714: Anything to do here re moving this function?
> this function can and should be moved to the `data` module

ehebert: Thanks for spotting that.
Yes, I will take our internal resampling and put in a module in data, then move our internal code to use that new function.

ehebert: @yankees714 that refactoring has been squashed in, the new function is here https://github.com/quantopian/zipline/pull/1372/files#diff-64eeb1b15a7a2ce1ec06af6c9a2f4b47R24

zipline/data/resample.py
```diff
@@ -0,0 +1,48 @@
...
+    'high': 'max',
+    'low': 'min',
+    'close': 'last',
+    'volume': 'sum'
+}
+
+
+def minute_to_session(minute_frame, calendar):
+    """
+    Resample a DataFrame with minute data and resample into the format
```
yankees714: I think the wording here got mixed up?

zipline/data/resample.py
```diff
@@ -0,0 +1,48 @@
...
+    minute_frame : pd.DataFrame
+        A DataFrame with the columns `open`, `high`, `low`, `close`, `volume`,
+        and `dt` (minute dts)
+    calendar : zipline.utils.calendars.trading_calendar.TradingCalendar
+        A TradingCalendar on which session labels to resample from minute
+        to session.
+
+    Return
+    ------
+    daily_frame : pd.DataFrame
```
yankees714: Should this be `session_frame`?

zipline/testing/fixtures.py
```diff
@@ -611,34 +612,10 @@ class WithEquityDailyBarData(WithTradingEnvironment):
...
-            last_minute = m_closes.loc[asset.end_date]
-            asset_df = minute_data[asset]
-            slicer = asset_df.index.slice_indexer(first_minute, last_minute)
-            asset_df = asset_df[slicer]
-            minutes = mm[mm.slice_indexer(start=first_minute,
-                                          end=last_minute)]
-            asset_df = asset_df.reindex(minutes)
-            yield asset.sid, asset_df.resample('1d', how=ohclv_how).dropna()
+            yield asset.sid, minute_to_session(minute_data[asset.sid],
+                                               cls.trading_calendar)
```
yankees714: Do we still want the `dropna()`?

zipline/data/resample.py
ehebert: Yes, thank you. Had renamed the function while writing it.

zipline/testing/fixtures.py
ehebert: No, it was required by the resample when using `1d` because that resampling used every day calendar, including days where there were no sessions. By passing the `calendar.day` instead, the resampling is already restricted to the trading calendar.

zipline/data/resample.py
ehebert: Yes, will remove the extra `and resample`.

PR #1376, opened 2016-08-05
BUG: when limit price is crossed, fill at limit price (not market price)

luca-s:
For "non-marketable" limit orders (limit price has been crossed)
the final price must be the limit price while for "marketable" limit orders
the market price must be used (as before)
To disinguish between marketable and non-marketable limit order we can use
the following check:
if both open and close price are below/above (buy/sell) the limit price
the order is markettable
if open price is above (or below for sell) limit price and close price is
below (or above for sell) limit price, then the order is non-marketable

This patch doesn't cover the scenario where the crossing of the limit
price happens between the close price of the previous minute and the
open price of the next minute. But in my understanding the close price
of any minute must be almost identical to the open price of the following
minute, so I don't believe this is a real issue.

### DIFF
--- a/tests/finance/test_slippage.py
+++ b/tests/finance/test_slippage.py
@@ -240,7 +240,7 @@ class SlippageTestCase(WithSimParams, WithDataPortal, ZiplineTestCase):
         txn = orders_txns[0][1]
 
         expected_txn = {
-            'price': float(3.50021875),
+            'price': float(3.6),
             'dt': datetime.datetime(
                 2006, 1, 5, 14, 34, tzinfo=pytz.utc),
             # we ordered 100 shares, but default volume slippage only allows
@@ -326,7 +326,7 @@ class SlippageTestCase(WithSimParams, WithDataPortal, ZiplineTestCase):
         _, txn = orders_txns[0]
 
         expected_txn = {
-            'price': float(3.49978125),
+            'price': float(3.4),
             'dt': datetime.datetime(
                 2006, 1, 5, 14, 32, tzinfo=pytz.utc),
             'amount': int(-50),
@@ -638,7 +638,7 @@ class SlippageTestCase(WithSimParams, WithDataPortal, ZiplineTestCase):
         _, txn = orders_txns[0]
 
         expected_txn = {
-            'price': float(3.50021875),
+            'price': float(3.6),
             'dt': datetime.datetime(
                 2006, 1, 5, 14, 34, tzinfo=pytz.utc),
             'amount': int(50),
@@ -762,7 +762,7 @@ class SlippageTestCase(WithSimParams, WithDataPortal, ZiplineTestCase):
         _, txn = orders_txns[0]
 
         expected_txn = {
-            'price': float(3.49978125),
+            'price': float(3.4),
             'dt': datetime.datetime(
                 2006, 1, 5, 14, 32, tzinfo=pytz.utc),
             'amount': int(-50),
--- a/zipline/finance/slippage.py
+++ b/zipline/finance/slippage.py
@@ -177,6 +177,24 @@ class VolumeShareSlippage(SlippageModel):
                     (order.direction < 0 and impacted_price < order.limit):
                 return None, None
 
+            # For "non-marketable" limit orders (limit price has been crossed)
+            # the final price must be the limit price.
+            # To disinguish between marketable and non-marketable limit
+            # order we can use the following check:
+            # if both open and close price are below/above (buy/sell)
+            # the limit price the order is markettable
+            # if open price is above (or below for sell) limit price
+            # and close price is below (or above for sell) limit price,
+            # then the order is non-marketable
+
+            open_price = data.current(order.asset, "open")
+            # Note: no need to check for close price if we are here
+            non_marketable = (order.direction > 0 and open_price > order.limit) or \
+                             (order.direction < 0 and open_price < order.limit)
+
+            if non_marketable:
+                impacted_price = order.limit
+
         return (
             impacted_price,
             math.copysign(cur_volume, order.direction)


luca-s:
#1370

LotannaEzenwa:
@luca-s thanks for the fix.

Can you elaborate on the difference in the testing values and the final value?

From `/tests/finance/test_slippage`:

```
expected_txn = {
            **'price': float(3.50021875),**
            'dt': datetime.datetime(
                2006, 1, 5, 14, 34, tzinfo=pytz.utc),
            # we ordered 100 shares, but default volume slippage only allows
            # for 2.5% of the volume.  2.5% * 2000 = 50 shares
            'amount': int(50),
            'sid': int(133),
            'order_id': open_orders[0].id
        }
```

luca-s:
That test must fail after the fix, that's the purpose of the patch (I am sorry I didn't update the expected test results too in my patch). 

In that particular test the limit price is 3.6 while the minute bar has open price 4.0 and close price 3.5. That means the limit price is crossed and the fill price must be 3.6 (and not 3.5 + slippage).

You will find another test that fails too (and it should): 
`
        # short, does trade
        open_orders = [
            Order(**{
                'dt': datetime.datetime(2006, 1, 5, 14, 30, tzinfo=pytz.utc),
                'amount': -100,
                'filled': 0,
                'sid': self.ASSET133,
                'limit': 3.4})
        ]

        expected_txn = {
            'price': float(3.49978125),
            'dt': datetime.datetime(
                2006, 1, 5, 14, 32, tzinfo=pytz.utc),
            'amount': int(-50),
            'sid': int(133)
        }
`

In this test the limit price is 3.4 and the minute bar open price is 3.0 and close price 3.5. Again, the limit price is crossed so the fill price must be 3.4 (not 3.5 + slippage).

volconst:
I think that the fill price of the order should not depend on open price, but on the high or low depending on the direction. The gap between previous close and open also should not matter because, if there is such, it can be viewed as a slippage, i.e. there was a gap because no one was offering shares at better price than open. The current code also has a problems because the order.check_triggers(price, dt) is calculated against close price, i.e. the decision whether we will execute at open price depends on the close price, which introduces hindsight bias.
The right implementation for me is to pass to order.check_triggers(self, price, dt) not price, but BarData and let check_triggers () use the appropriate price field depending on the direction and limit of the order. Determining the price field is not a slippage consideration.

luca-s:
@volconst 

I believe I understand what you say but I carefully decided to make a simple patch to avoid entering the realm of  slippage model (it hasn't changed for ages even if it is a poor simulation of reality and so I don't believe they want to change it any soon). 

What we are trying to address here is  #1369 issue: the fill price for a limit order must be the fill price when the price is crossed (because the current implementation is too optimistic and gives better results than reality), while keeping the current slippage adjusted fill price when the limit price is marketable.

The  patch provides the minimum changes to avoid over optimistic results, that means it keeps current behaviour for all limit orders except those that are crossed (those will have worse, more realistic, results). I had to use open price to make a decision between marketable and non marketable orders.

Using open, low, high prices for fill price decision (limit orders or not) is an interesting topic, but for a different issue.

volconst:
Sorry, I have to correct myself. High or low price should only be used to determine if there is execution, and not for the price. I now believe that execution price should be based on open price or be equal to limit price. Market orders should also execute based at open price. In simulation, an order is placed after we have access to the complete bar, i.e. at the end of the bar and as market orders are immediate, they should execute at open price of the next bar.
Moving from 'close' to 'open' seems like a big change and probably will not be integrated in the default slippage.
According to this PR i only find it strange to check for marketability the open and close price and to execute at close price.
I intend to use limit orders for daily simulations and maybe I will create two slippage models - one that uses only limit price (conservative) and another that uses the better of limit and open (aggressive).
I consider the marketable logic of limit orders as rarely used and more like an insurance against typing errors, and as we do not have bid/ask in zipline we do not lose much if we do not use it.

luca-s:
Just to be informative...

The slippage model has used close price (of the next minute bar after the order is placed) since forever. The close price is used to determine if an order is executed and the fill/execution price. Slippage model uses close price only because that is considered the most conservative approach. There are other options to model the slippage, for example using open, low, high price information in the decision logic and price filling. I agree with that but, as stated before, this is not the issue I want to address with this patch.

This patch concerns a bug in the fill price for certain limit orders. When a limit order is placed there can be two scenarios: the current price is better than the limit price or it is worse. if close price is better than limit price the slippage model applies the slippage to the close price and if the resulting price is still better than limit price an order is filled with that. When close price is worse than limit price the order is not filled. After some time it might happens that the close price becomes better than the limit price and at that time the order is filled. The bug is that the fill price in this latter case is the close price + slippage (as in the former scenario). That's not what happens in reality, when a limit price is crossed the fill price is equal to the limit price. The current implementation gives better price than reality all the time. 

Considering that current slippage model is way too conservative (worse than reality) by design, it doesn't make sense to have a scenario, like the one above, where it performs better than reality.

volconst:
If you have order buy at $3 and OHLC = (2.9, 3.5, 2.8, 3.1), no execution will occur. It will be more realistic to buy at $2.9 (open price, aggressive), or at $3 (limit price, conservative).

luca-s:
I agree with you and It would make sense to improve the current behaviour (in another patch/issue).

volconst:
You assume that open price is a good predictor for market price. The simulation uses close price for market price. So close price does better job than open price, and if we use it instead, we arrive at the official implementation. I did not expect this. Do I err somewhere?
So, if we just want to be conservative we should execute always at limit price if the order is marketable at close price.

luca-s:
The patch uses open price only to check if the limit order was marketable or not, but it uses close price for the fill/execution price (or limit price in the case order was not marketable).

Imagine a buy limit order with limit at 10. Suppose the close price is 9.5 and open price 9. The order is filled at 9.5 + slippage. That is correct and it would be a bug to fill the order at limit price.  Now suppose the close price is 9.5 and open price 10.2, that means when the order was entered it wasn't marketable but it became marketable at close price (the limit price has been crossed).  In reality you would get a fill price that is the limit price i.e. 10. Unfortunately, zipline gives you a better price, that is 9.5 + slippage. 

To fix that behaviour the patch checks the open price to determine if the limit was crossed or not. if it is crossed it applies the limit price, otherwise it uses the old behaviour of close + slippage. 

From my point of view the patch fixes the bug without any additional side effects. Can you see any issue I didn't see?

volconst:
Try to implement a real trading account algorithm that behaves as this PR.
If the order is not marketable at open we wait until we cross the limit and execute.
If the order is marketable at open price we have to execute at close price. But what is the close price? We can't see the future, so we have to wait till close. But at close the order could be unmarketable and we end without position that would be profitable. The simulation shows profits, which we cannot achieve even if the exact same bars happen. If we execute immediately after inspecting open or close price, we will achieve similar results as simulation (less slippage, commissions errors)

luca-s:
@LotannaEzenwa any updates on this? I can improve the patch if required. Or are you planning to fix the issue #1369 in a different way?

luca-s:
@Mantainers. Could you please review this patch and state what's wrong? As a contributor I cannot do much if you don't review the code and accept it or ask for improvements. There is a bug #1369 and I would like to see the issue solved and I am giving my time, but I need maintainers collaboration to address the problem.

jbredeche:
hi @luca-s, sorry for the radio silence.  We are discussing internally and will get back to you.

volconst:
This PR has several problems:
 - it is not reproducible in reality because of forward bias even if we have all the ticks of the bar, see my last comment from Aug 18.
 - if an order is marketable at market price (close), it should be executed at market price. This PR may execute at limit price even if the order is marketable at market price, which contradicts the definition of limit order.
 - if an order is not marketable at market price, it should execute at limit if the limit is within the range [low..high]. In this PR the range [open..close] is used. In reality limit orders never depend on open price.
 - In reality, market price shows up and is checked before limit price. In this PR market price (close) is after limit price.
I think this PR should be rejected. I have another pull request with alternative solution, but it failed to verify and will post it here later.

volconst:
Alternative implementation #1512 which can use OHL, HLC or only Close prices.

luca-s:
I understand this is not a big issue and there is room for big improvements (considering high and low prices in the slippage model for example) but while a better slippage model is no available, why not accepting this simple bug fix to the current broken model?

luca-s:
Closing this as maintainers seem not interested in fixing the issue

PR #1377, opened 2016-08-05
MAINT: Isolate global calendar state.

ssanderson:
Encapsulate the shared global calendar map in an object.

This allows consumers that don't want to participate in custom
registration to pass around a calendar dispatcher, and would make it
easier to support contextual management of the global calendar map if we
want to do that in the future.

As a bonus, we now only create one instance of each calendar, instead of
one per alias.


jbredeche:
LGTM

zipline/utils/calendars/calendar_utils.py
```diff
@@ -13,92 +13,132 @@
 _static_calendars = {}
```
richafrank: Is this still used?

ssanderson: Removed.

PR #1379, opened 2016-08-08
Prepare data portal tests for covering Futures assets types.

ehebert:
Add direct coverage of `get_last_traded_dt` in `test_data_portal` for `Equity` asset types, to be followed by a test case for `Future` assets.

In support of routing between `Equity` and `Future` asset types, add and use a `_pricing_readers` dict keyed by asset type and data frequency.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -147,6 +147,17 @@ class DataPortal(object):
         self._future_daily_reader = future_daily_reader
         self._future_minute_reader = future_minute_reader
 
+        self._pricing_readers = {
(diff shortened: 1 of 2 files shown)


zipline/data/data_portal.py
```diff
@@ -309,17 +320,18 @@ def _get_ctable(self, asset):
 
         return bcolz.open(path, mode='r')
 
+    def _get_pricing_reader(self, asset, data_frequency):
+        return self._pricing_readers[type(asset)][data_frequency]
```
yankees714: 👍

PR #1383, opened 2016-08-09
BUG: Fix HolidayCalendar init

jbredeche:


zipline/utils/calendars/trading_calendar.py
```diff
@@ -824,4 +824,4 @@ def _overwrite_special_dates(midnight_utcs,
 
 class HolidayCalendar(AbstractHolidayCalendar):
     def __init__(self, rules):
-        super(HolidayCalendar, self).__init__(self, rules=rules)
+        super(HolidayCalendar, self).__init__(rules=rules)
```
richafrank: This looks correct in terms of types now.  

In terms of the semantics, it looks like pandas uses `AbstractHolidayCalendar` like we do `TradingCalendar`, where each type has a name with which it is registered, so you can get it via `pandas.tseries.holiday.get_calendar(name)`.  However, we're using many instances of a single `HolidayCalendar` type here, and that type is registered with no rules.  It does look like pandas provides essentially the same functionality as our concrete `HolidayCalendar` not via a base class, but instead via a `pandas.tseries.holiday.HolidayCalendarFactory` factory function.  Do you think we should use this framework instead?  e.g. instead of
```python
    @property
    def regular_holidays(self):
        return HolidayCalendar([
            USNewYearsDay,
            USMartinLutherKingJrAfter1998,
            USPresidentsDay,
            GoodFriday,
            USMemorialDay,
            USIndependenceDay,
            USLaborDay,
            USThanksgivingDay,
            Christmas,
        ])
```

we'd have:
```python
    @lazyval  # lazyval so we aren't making tons of types
    def regular_holidays(self):
        return HolidayCalendarFactory(
            'NYSE Holidays', 
            AbstractHolidayCalendar, [
                USNewYearsDay,
                USMartinLutherKingJrAfter1998,
                USPresidentsDay,
                GoodFriday,
                USMemorialDay,
                USIndependenceDay,
                USLaborDay,
                USThanksgivingDay,
                Christmas,
        ])
```

jbredeche:
We definitely could, but out of scope for this PR.  I just wanted to fix that obvious mistake.

PR #1384, opened 2016-08-09
BUG/TEST: Fix stochastic oscillator test.

ssanderson:
- Don't create unnecessary extra data (requires passing fastd_period=1
  to TA-Lib or else it fills the FastK with NaNs even though it must
  have already computed them...

- Use random_sample instead of random_integers so that we're not
  dependent on integer arithmetic.

- Pass array_decimal to assert_equal so that we do almost equal checking
  on results.


ehebert:
The consistent seed looks good to me. :+1:

PR #1387, opened 2016-08-10
v5 downgrade

llllllllll:
Selects the most recent ticker for each sid.


jdricklefs:
Thanks for the test case. LGTM pending Travis

PR #1391, opened 2016-08-15
MAINT: Define minute bar reader interface.

ehebert:
Define the minute bar interface and make the `BcolzMinuteBarReader`
implement that interface.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from abc import ABCMeta, abstractmethod, abstractproperty
 import json
 import os
 from os.path import join
@@ -22,6 +23,7 @@ from bcolz import ctable
 from intervaltree import IntervalTree
 import numpy as np
 import pandas as pd
+from six import with_metaclass
 
 from zipline.data._minute_bar_internal import (
     minute_value,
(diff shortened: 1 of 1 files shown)


ehebert:
This spun out of work on resampling session data from minute data. That branch had originally added a `DataFrameMinuteBarReader`, where having the interface to implemented assisted development. The branch with the resampling work has since moved to using the existing `Bcolz*` readers so that only one novel part was being worked on at a given time, but may add the `DataFrameMinuteBarReader` back. Either way, having the interface defined with comments is beneficial, especially when needing a reminder on the return type of `load_raw_arrays`.

zipline/data/minute_bars.py
```diff
@@ -48,6 +50,99 @@ class BcolzMinuteWriterColumnMismatch(Exception):
...
+        """
+        Returns
+        -------
+        dt : pd.Timestamp
+            The last minute for which the reader can provide data.
+        """
+        pass
+
+    @abstractproperty
+    def first_trading_day(self):
```
ssanderson: Any reason not to call this `first_trading_session`, given that the docstring here takes pains to clarify the intent?

zipline/data/minute_bars.py
```diff
@@ -48,6 +50,99 @@ class BcolzMinuteWriterColumnMismatch(Exception):
...
+
+        Returns
+        -------
+        value : float|int
+            The value at the given coordinates, `float` for OHLC, `int` for
+            'volume'.
+        """
+
+    @abstractmethod
+    def get_last_traded_dt(self, asset, dt):
```
ssanderson: Is there a reason this is a separate entry from `get_value`?  I would think, conceptually, that this would just delegate to `get_value(field='last_traded')`.

zipline/data/minute_bars.py
```diff
@@ -48,6 +50,99 @@ class BcolzMinuteWriterColumnMismatch(Exception):
     pass
 
 
+class MinuteBarReader(with_metaclass(ABCMeta)):
+
+    @abstractproperty
+    def last_available_dt(self):
```
ssanderson: Why does the interface only require the last available minute, but requires the first available trading date?  The asymmetry there is surprising.

zipline/data/minute_bars.py
```diff
@@ -48,6 +50,99 @@ class BcolzMinuteWriterColumnMismatch(Exception):
...
+        """
+        Retrieve the most recently traded minute for the given asset,
+        when viewing the data from the input minute. The last traded minute is
+        the provided minute if a trade occured on that minute.
+
+        Parameters
+        ----------
+        asset : zipline.asset.Asset
+            The asset for which to get the last traded minute.
+        dt : datetime64-like
```
ssanderson: Does this accept values other than `pd.Timestamp`?  I wouldn't expect both `Timestamp` and `np.datetime64` to work, since timestamps here would have to be tz-aware.

zipline/data/minute_bars.py
```diff
@@ -48,6 +50,99 @@ class BcolzMinuteWriterColumnMismatch(Exception):
...
+        last_traded : pd.Timestamp
+            The minute of the last trade for the given asset, using the input
+            dt as a vantage point.
+        """
+
+    @abstractmethod
+    def load_raw_arrays(self, fields, start_dt, end_dt, sids):
+        """
+        Retrieve the arrays of pricing data for the given coordinates of
+        `fields` (OHLCV), minute range (`start_dt` -> `end_dt`) and sids.
```
ssanderson: Might be worth clarifying that we're inclusive of both start_dt and end_dt here?

zipline/data/minute_bars.py
```diff
@@ -48,6 +50,99 @@ class BcolzMinuteWriterColumnMismatch(Exception):
...
+    @abstractproperty
+    def first_trading_day(self):
+        """
+        Returns
+        -------
+        dt : pd.Timestamp
+            The first trading day (session) for which the reader can provide
+            data.
+        """
+        pass
```
ssanderson: Style nit: some of the methods here have a `pass` and some don't.  I don't care strongly about one vs the other, but being consistent is nice.

zipline/data/minute_bars.py
ehebert: I agree that name is better/more in line with current standards. I will attempt a patch on top of this one to do the rename.

zipline/data/minute_bars.py
ehebert: I believe the original distinction was that `get_value` was returning the pricing values, i.e. the trade bar floats/ints, and this function was a helper to find the right dt to seek into when doing forward filling; with the distinction that this returns the location of data as opposed to the value.

I am not opposed to merging `last_traded` into `get_value`, but there is a slight semantic difference. Also, some care may need to be taken when changing the code since this method is used internally by the data portal when returning values for `price`.

zipline/data/minute_bars.py
ehebert: I was also surprised that they were different.

After looking at the data portal code which uses `first_trading_day`; `first_trading_day` is used to also get the `first_trading_minute`, as well. Which is then used to determine whether history windows are over valid ranges.

This code will need to be changed soon to support both equities and futures, which may have different start minutes, if not starting sessions.

`last_available_dt` is used in the `us_equity_loader` module to get the range of minutes supported by the reader; which could be its own method supported by the minute bar reader.
`last_available_dt` is also used by own of our internal data loaders to get the last session provided by the reader.

If we picked one of session/date or minute on which to standardize, I would lean towards supporting standardizing on minute, since in the live trading case the last minute available is not derivable from the session label.
We may want to support both date and minute, so that consumers are not responsible for maintaining a calendar associated with the reader to map back and forth between the `last_available_dt` and `last_available_session` values, if only one of the two were provided.

zipline/data/minute_bars.py
```diff
@@ -48,6 +50,99 @@ class BcolzMinuteWriterColumnMismatch(Exception):
...
+        value : float|int
+            The value at the given coordinates, `float` for OHLC, `int` for
+            'volume'.
+        """
+
+    @abstractmethod
+    def get_last_traded_dt(self, asset, dt):
+        """
+        Retrieve the most recently traded minute for the given asset,
+        when viewing the data from the input minute. The last traded minute is
```
ehebert: Sounds good, will edit. And the ``dt`` when not found is `pd.NaT`. (At least that is the current implementation.)

zipline/data/minute_bars.py
ehebert: Will change to `pd.Timestamp`.

zipline/data/minute_bars.py
ehebert: Would "minute range [`start_date`, `end_dt`]", suffice?

zipline/data/minute_bars.py
ehebert: It was the intent for all to have `pass`; looks like the got clipped while putting in the docstrings. Will make them all the same.

zipline/data/minute_bars.py
ehebert: There are a few too many touch points for `first_trading_day` to attach to this PR. Making a note to attempt on another branch.

PR #1394, opened 2016-08-16
Add Generic Downsampling to Pipeline

ssanderson:
Adds a new ``downsample`` method to all computable terms.  Computable
terms (Filters, Factors, and Classifiers) can be downsampled to yearly,
quarterly, monthly, or weekly frequency.

The result of ``term.downsample`` is a new term of the same
family (Filter/Factor/Classifier) as ``term``.  The downsampled term
computes by delegating to the original term; repeatedly calling its
``compute`` method with length-1 date ranges.

Downsampled terms take advantage of a new ``compute_extra_rows`` Term
method, which allows terms to dynamically request that additional extra
rows of themselves be computed based on the dates for which they're
being computed.  This ensures, for example, that a monthly-downsampled
term always computes at the start of a month, even when a
naively-calculated pipeline window would end in the middle of the month.

### DIFF
--- a/zipline/pipeline/engine.py
+++ b/zipline/pipeline/engine.py
@@ -211,10 +211,10 @@ class SimplePipelineEngine(object):
         start_idx, end_idx = self._calendar.slice_locs(start_date, end_date)
         if start_idx < extra_rows:
             raise NoFurtherDataError(
-                msg="Insufficient data to compute Pipeline mask: "
-                "start date was %s, "
-                "earliest known date was %s, "
-                "and %d extra rows were requested." % (
+                msg="Insufficient data to compute Pipeline.\n\n"
+                "start date was %s\n"
+                "earliest known date was %s\n"
+                "%d extra rows were required" % (
                     start_date, calendar[0], extra_rows,
                 ),
             )
--- a/zipline/pipeline/engine.py
+++ b/zipline/pipeline/engine.py
@@ -21,7 +21,7 @@ from zipline.errors import NoFurtherDataError
 from zipline.utils.numpy_utils import repeat_first_axis, repeat_last_axis
 from zipline.utils.pandas_utils import explode
 
-from .term import AssetExists, LoadableTerm
+from .term import AssetExists, InputDates, LoadableTerm
 
 
 class PipelineEngine(with_metaclass(ABCMeta)):
@@ -98,6 +98,7 @@ class SimplePipelineEngine(object):
         '_calendar',
         '_finder',
         '_root_mask_term',
+        '_root_mask_dates_term',
         '__weakref__',
     )
 
@@ -105,7 +106,9 @@ class SimplePipelineEngine(object):
         self._get_loader = get_loader
         self._calendar = calendar
         self._finder = asset_finder
+
         self._root_mask_term = AssetExists()
+        self._root_mask_dates_term = InputDates()
 
     def run_pipeline(self, pipeline, start_date, end_date):
         """
@@ -161,7 +164,13 @@ class SimplePipelineEngine(object):
             )
 
         screen_name = uuid4().hex
-        graph = pipeline.to_graph(screen_name, self._root_mask_term)
+        graph = pipeline.to_execution_plan(
+            screen_name,
+            self._root_mask_term,
+            self._calendar,
+            start_date,
+            end_date,
+        )
         extra_rows = graph.extra_rows[self._root_mask_term]
         root_mask = self._compute_root_mask(start_date, end_date, extra_rows)
         dates, assets, root_mask_values = explode(root_mask)
@@ -170,7 +179,10 @@ class SimplePipelineEngine(object):
             graph,
             dates,
             assets,
-            initial_workspace={self._root_mask_term: root_mask_values},
+            initial_workspace={
+                self._root_mask_term: root_mask_values,
+                self._root_mask_dates_term: dates.values[:, None],
+            },
         )
 
         return self._to_narrow(
--- a/zipline/pipeline/engine.py
+++ b/zipline/pipeline/engine.py
@@ -222,13 +222,11 @@ class SimplePipelineEngine(object):
         finder = self._finder
         start_idx, end_idx = self._calendar.slice_locs(start_date, end_date)
         if start_idx < extra_rows:
-            raise NoFurtherDataError(
-                msg="Insufficient data to compute Pipeline.\n\n"
-                "start date was %s\n"
-                "earliest known date was %s\n"
-                "%d extra rows were required" % (
-                    start_date, calendar[0], extra_rows,
-                ),
+            raise NoFurtherDataError.from_lookback_window(
+                initial_message="Insufficient data to compute Pipeline:",
+                first_date=calendar[0],
+                lookback_start=start_date,
+                lookback_length=extra_rows,
             )
 
         # Build lifetimes matrix reaching back to `extra_rows` days before
--- a/zipline/pipeline/engine.py
+++ b/zipline/pipeline/engine.py
@@ -18,7 +18,11 @@ from toolz.curried.operator import getitem
 
 from zipline.lib.adjusted_array import ensure_adjusted_array, ensure_ndarray
 from zipline.errors import NoFurtherDataError
-from zipline.utils.numpy_utils import repeat_first_axis, repeat_last_axis
+from zipline.utils.numpy_utils import (
+    as_column,
+    repeat_first_axis,
+    repeat_last_axis,
+)
 from zipline.utils.pandas_utils import explode
 
 from .term import AssetExists, InputDates, LoadableTerm
@@ -181,7 +185,7 @@ class SimplePipelineEngine(object):
             assets,
             initial_workspace={
                 self._root_mask_term: root_mask_values,
-                self._root_mask_dates_term: dates.values[:, None],
+                self._root_mask_dates_term: as_column(dates.values)
             },
         )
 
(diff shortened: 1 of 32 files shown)


zipline/pipeline/graph.py
```diff
@@ -34,21 +33,23 @@ class TermGraph(DiGraph):
...
     """
     def __init__(self, terms):
-        super(TermGraph, self).__init__(self)
+        super(TermGraph, self).__init__()
 
         self._frozen = False
         parents = set()
-        for term in itervalues(terms):
-            self._add_to_graph(term, parents, extra_rows=0)
+        for name, term in iteritems(terms):
```
twiecki: Seems like `name` doesn't get used.

zipline/pipeline/term.py
```diff
@@ -516,6 +585,31 @@ def postprocess(self, data):
...
+
+    def downsample(self, frequency):
+        """
+        Make a term that computes from ``self`` at lower-than-daily frequency.
+
+        Parameters
+        ----------
+        frequency : str, {'Y', 'Q', 'M', 'W'}
+            A string indicating the desired sampling rate.
+            'Y' -> sample on the first trading day of each calendar year
```
twiecki: Often you also want the last trading day of a YQMW. Not saying it has to be part of this PR but preferably the syntax does not exclude that from getting implemented.

llllllllll: we could use pandas strings like `M` for month end and `MS` for month start

twiecki: +1 for using pandas' standard.

zipline/testing/fixtures.py
```diff
@@ -527,7 +527,7 @@ def init_class_fixtures(cls):
         cls.sim_params = cls.make_simparams()
 
 
-class WithTradingSessions(WithTradingCalendars):
+class WithTradingSessions(WithTradingCalendars, WithDefaultDateBounds):
```
dmichalowicz: Looks like the order of these parent classes needs to be switched

zipline/pipeline/graph.py
```diff
@@ -34,21 +33,23 @@ class TermGraph(DiGraph):
...
     ordered()
         Return a topologically-sorted iterator over the terms in self.
+
+    See Also
+    --------
+    ExecutionPlan
     """
     def __init__(self, terms):
-        super(TermGraph, self).__init__(self)
+        super(TermGraph, self).__init__()
```
dmichalowicz: Why this change?

zipline/pipeline/factors/factor.py
```diff
@@ -1468,7 +1473,9 @@ def __iter__(self):
 
 
 class RecarrayField(SingleInputMixin, Factor):
-
+    """
+    A single field from a multi-output factor.
+    """
```
dmichalowicz: 🎉

zipline/pipeline/graph.py
```diff
@@ -58,6 +59,149 @@ def __init__(self, terms):
...
+    often needed when a term is an input to a rolling window computation.  For
+    example, if we compute a 30 day moving average of price from day X to day
+    Y, we need to load price data for the range from day (X - 29) to day Y.
+
+    Parameters
+    ----------
+    terms : dict
+        A dict mapping names to final output terms.
+    all_dates : pd.DatetimeIndex
+        The dates fo
```
dmichalowicz: Incomplete sentence ;)

zipline/testing/fixtures.py
ssanderson: The order here shouldn't matter; both `WithTradingCalendars` and `WithDefaultDateBounds` are direct subclasses of `object`.  The tests are currently failing because another class is manually inheriting from both `WithTradingSessions` and `WithDefaultDateBounds`.  The right fix is to remove the duplication at that use-site.

zipline/pipeline/term.py
ssanderson: My only objection to this is that pandas inexplicably users `A` and `AS` for yearly instead of `Y` and `YS`, which I have never once successfully remembered.

zipline/pipeline/graph.py
ssanderson: This was just a bug.  We were passing `self` twice to the parent constructor, which happened to work because the constructor of `DiGraph` takes another graph as its first argument, which is copied.  In this case we were forwarding an empty graph to the parent.

zipline/pipeline/term.py
ssanderson: @twiecki @llllllllll my other concern with using the pandas conventions is that I think calculating at the start of a period will be the more common desired outcome, but the more obvious period names of `'M'`, `'Y'` and `'Q'` mean all mean the end of the period.  Pandas also doesn't have a start/end distinction for weekly frequency.  Given that I think relatively few of our users are going to know the pandas symbols by heart, I don't think there's a ton to be gained by copying their conventions.  At the same time, there are probably enough users who are familiar with pandas that they'd be confused if we re-used `'M'` and `'Q'` but had them mean different things.

In light of all of the above, what do you think about just using slightly-more-verbose strings like `'month_start'`, `'year_start'`, `'quarter_end'`?

llllllllll: I agree that we should not make it close but slightly different to pandas. If you really don't like the pandas version then I think the verbose names are fine. I think people will look this up when they need to use it anyways so following the pandas convention can help the few people that do know them.

zipline/pipeline/pipeline.py
```diff
@@ -148,24 +148,57 @@ def set_screen(self, screen, overwrite=False):
...
+        Compile into a simple TermGraph with no extra row metadata.
+
+        Parameters
+        ----------
+        screen_name : str
+            Name to supply for self.screen.
+        default_screen : zipline.pipeline.term.Term
+            Term to use as a screen if self.screen is None.
+        """
+        return TermGraph(self._prepare_graph_terms())
```
dmichalowicz: I think this is missing its arguments

ssanderson: Good catch.  The codepath that calls this isn't tested because it relies on having the `dot` program installed.  I'll see if I can at least throw in a smoketest for this.

zipline/pipeline/engine.py
```diff
@@ -170,7 +179,10 @@ def run_pipeline(self, pipeline, start_date, end_date):
             graph,
             dates,
             assets,
-            initial_workspace={self._root_mask_term: root_mask_values},
+            initial_workspace={
+                self._root_mask_term: root_mask_values,
+                self._root_mask_dates_term: dates.values[:, None],
```
dmichalowicz: Maybe use `as_column` here?

zipline/utils/numpy_utils.py
```diff
@@ -364,3 +367,66 @@ def vectorized_is_element(array, choices):
...
+    """
+    if a.ndim != 1:
+        raise ValueError(
+            "as_column expected an 1-dimensional array, "
+            "but got an array of shape %s" % a.shape
+        )
+    return a[:, None]
+
+
+def changed_locations(a, include_first):
```
dmichalowicz: Do you think this is worth adding a short test for?

dmichalowicz:
@ssanderson Done with a first review. I only skimmed through most of the tests though, so if there were certain parts of the tests you think I should look at more in depth look just let me know.

zipline/utils/numpy_utils.py
ssanderson: The example here is run automatically by doctest.  Do you think it needs something beyond that?

zipline/pipeline/engine.py
ssanderson: I think this is what I actually wrote `as_column` for and then forgot.  Will update.

zipline/utils/numpy_utils.py
dmichalowicz: No I think that is sufficient (I did not know that)

zipline/pipeline/term.py
twiecki: I only knew the `BMonthStart` and `BMonthEnd` date offsets to begin with. The verbose names sound fine to me as well.

zipline/pipeline/mixins.py
```diff
@@ -400,14 +400,35 @@ def _compute(self, windows, dates, assets, mask):
...
+                    next(w)
+        else:
+            # If we're not windowed, inputs are just ndarrays.  We need to
+            # slice off one row when forwarding to real_compute, but we don't
+            # need to do anything to skip an input.
+            def prepare_inputs():
+                # i is the loop iteration variable below.
+                return [a[[i]] for a in inputs]
+
+            def skip_this_input():
```
dmichalowicz: Would this be more aptly named `skip_this_date`?

ssanderson: I would think that that would encompass the whole action that happens in the loop iteration, whereas this just applies to the part that interacts with `inputs`.

zipline/pipeline/factors/factor.py
```diff
@@ -1072,9 +1073,9 @@ def isfinite(self):
         """
         return (-inf < self) & (self < inf)
 
-    @property
+    @classlazyval
     def _downsampled_type(self):
-        return DownsampledFactor
+        return DownsampledMixin.make_downsampled_type(Factor)
```
dmichalowicz: Can you remove `DownsampledFactor` now?

zipline/pipeline/mixins.py
```diff
@@ -218,3 +238,224 @@ def _validate(self):
...
+        Factory for making Downsampled{Filter,Factor,Classifier}.
+        """
+        docstring = dedent(
+            """
+            A {t} that defers to another {t} at lower-than-daily frequency.
+
+            Parameters
+            ----------
+            term : {t}
+            {{frequency}}
```
dmichalowicz: Does this require the `templated_docstring` decorator?

dmichalowicz: Ah never mind, didn't see below

dmichalowicz:
Other than the `to_simple_graph` fix, looks good to me

PR #1396, opened 2016-08-16
Add estimates quarter loader to pipeline

mtydykov:
This PR adds an estimates quarter loader to be used with Pipeline. 

Some open questions are:
- What other tests do we need to add?
- How do we ensure that the `num_quarters` attribute has been added to the dynamically generated dataset by the time `load_adjusted_array` is called?


zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,140 @@
...
+    The expression should have a tabular dshape of::
+
+       Dim * {{
+           {SID_FIELD_NAME}: int64,
+           {TS_FIELD_NAME}: datetime,
+       }}
+
+    And other dataset-specific fields, where each row of the table is a
+    record including the sid to identify the company, the timestamp where we
+    learned about the announcement, and the date when the earnings will be z
```
llllllllll: extra `z`?

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,140 @@
...
+    announced.
+
+    If the '{TS_FIELD_NAME}' field is not included it is assumed that we
```
llllllllll: I think that these format strings probably came from another docstring where we had `__doc__ = """....""".format(SID_FIELD_NAME=SID_FIELD_NAME, TS_FIELD_NAME=TS_FIELD_NAME)` which we should preserve.

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,140 @@
...
+            dates,
+            assets,
+            mask,
+        )
+
+
+class BlazeNextEstimatesLoader(BlazeEstimatesLoader):
+    loader = NextQuartersEstimatesLoader
+
+    def __init__(self,
```
llllllllll: Is this `__init__` doing anything other than forwarding everything to the super? If so we can just remove it.

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,140 @@
...
+            odo_kwargs,
+            data_query_time,
+            data_query_tz,
+        )
+
+
+class BlazePreviousEstimatesLoader(BlazeEstimatesLoader):
+    loader = PreviousQuartersEstimatesLoader
+
+    def __init__(self,
```
llllllllll: same here

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,140 @@
...
+
+from .core import (
+    bind_expression_to_resources,
+    load_raw_data,
+)
+from zipline.pipeline.loaders.base import PipelineLoader
+from zipline.pipeline.loaders.quarter_estimates import (
+    NextQuartersEstimatesLoader,
+    PreviousQuartersEstimatesLoader,
+    required_estimates_fields)
```
llllllllll: Can you close the paren on a new line?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1220,6 +1220,36 @@ def bind_expression_to_resources(expr, resources):
     })
 
 
+def load_raw_data(assets, dates, data_query_time, data_query_tz, expr,
```
llllllllll: This feels more like a utility than something that should be in core since it is not used by the core loader.

Also, could you add a docstring to document the use of this function?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,287 @@
...
+    FISCAL_QUARTER_FIELD_NAME,
+    FISCAL_YEAR_FIELD_NAME,
+    SID_FIELD_NAME,
+    TS_FIELD_NAME,
+)
+from zipline.pipeline.loaders.base import PipelineLoader
+from zipline.pipeline.loaders.frame import DataFrameLoader
+
+ALL_DATES = 'dates'
+FISCAL_QUARTER = 'fiscal_quarter'
```
llllllllll: these two are in `common.py`

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+def calc_forward_shift(yrs, qtrs, num_qtrs_shift):
+    """
+    Calculate the new years and quarters based on on shifting the specified
+    number of quarters forward.
+
+    Parameters
+    ----------
+    yrs : np.Series
```
llllllllll: can we just call these `years` and `quarters`?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,287 @@
...
+                missing=sorted(missing),
+                received=sorted(received),
+                required=sorted(required),
+            )
+        )
+
+
+def cross_product(df1, df2):
+    df1['key'] = 1
+    df2['key'] = 1
```
llllllllll: `key` seems a little too likely to collide with existing keys. Could we use a uuid4 or something? Also, could we move this to pandas_utils?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,287 @@
...
+
+        self.estimates = estimates[
+            estimates[EVENT_DATE_FIELD_NAME].notnull() &
+            estimates[FISCAL_QUARTER_FIELD_NAME].notnull() &
+            estimates[FISCAL_YEAR_FIELD_NAME].notnull()
+        ]
+
+        self.base_column_name_map = base_column_name_map
+
+    def load_quarters(self, num_quarters, dates_sids, final_releases_per_qtr):
```
llllllllll: should this be an abstract class and abstract method?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,287 @@
+import pandas as pd
+from six import viewvalues
+from toolz import groupby
+from zipline.pipeline.common import (
```
llllllllll: there should be one line of whitespace between the third party imports and internal imports.

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,287 @@
...
+    Parameters
+    ----------
+    yrs : np.Series
+        The starting years.
+    qtrs : np.Series
+        The starting quarters.
+    num_qtrs_shift : int
+        The number of quarters to shift forward.
+
+
```
llllllllll: extra whitespace

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,287 @@
...
+    Returns
+    -------
+    result_years : pd.Series
+        A series contains the new years.
+    result_qtrs : pd.Series
+        A series that contains the new quarters.
+    """
+
+    if num_qtrs_shift < 0:
+        raise AssertionError("Must pass a number of quarters >= 0")
```
llllllllll: this should be a `ValueError` since it is an exception based on user input to a function

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,291 @@
...
+            for c in columns:
+                column_name = name_map[c]
+                # Pivot to get a DataFrame with dates as the index and
+                # sids as the columns.
+                loader = DataFrameLoader(
+                    c,
+                    result.pivot(index=SIMULTATION_DATES,
+                                 columns=SID_FIELD_NAME,
+                                 values=column_name),
+                    adjustments=None
```
llllllllll: No deltas?

mtydykov: Not yet. We don't apply them in the events loader either, so I'd imagine that we'd want to implement both in the future.

zipline/pipeline/loaders/blaze/core.py
mtydykov: Ok, I can create a new `blaze/utils.py` and put it there.

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: Yes, I'll add the appropriate decorator.

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,291 @@
...
+FISCAL_QUARTER = 'fiscal_quarter'
+FISCAL_YEAR = 'fiscal_year'
+NEXT_FISCAL_QUARTER = 'next_fiscal_quarter'
+NEXT_FISCAL_YEAR = 'next_fiscal_year'
+PREVIOUS_FISCAL_QUARTER = 'previous_fiscal_quarter'
+PREVIOUS_FISCAL_YEAR = 'previous_fiscal_year'
+SIMULTATION_DATES = 'dates'
+
+
+def calc_forward_shift(yrs, qtrs, num_qtrs_shift):
```
mtydykov: `divmod` doesn't work on Series.

mtydykov: Modified it a bit and it works great. Thanks!

zipline/lib/adjustment.pyx
```diff
@@ -364,6 +364,150 @@ cdef class Float64Overwrite(Float64Adjustment):
...
+
+    Subclasses should inherit and provide a `values` attribute and a `mutate`
+    method.
+    """
+    def __init__(self,
+                 int64_t first_row,
+                 int64_t last_row,
+                 int64_t first_col,
+                 int64_t last_col):
+        super(ArrayAdjustment, self).__init__(
```
llllllllll: this is just a nop `__init__`

zipline/lib/adjustment.pyx
```diff
@@ -364,6 +364,150 @@ cdef class Float64Overwrite(Float64Adjustment):
...
+                 int64_t first_col,
+                 int64_t last_col,
+                 float64_t[:] values):
+        super(Float641DArrayOverwrite, self).__init__(
+            first_row=first_row,
+            last_row=last_row,
+            first_col=first_col,
+            last_col=last_col,
+        )
+        assert (last_row + 1 - first_row) == len(values)
```
llllllllll: don't use an assert to check input to a function, this should raise a value error if not correct.

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -175,9 +175,10 @@
 from zipline.pipeline.data.dataset import DataSet, Column
 from zipline.pipeline.loaders.utils import (
     check_data_query_args,
+    last_in_date_group,
     normalize_data_query_bounds,
     normalize_timestamp_to_query_time,
-)
+    ffill_across_cols)
```
llllllllll: close the paren on a new line

zipline/pipeline/loaders/utils.py
```diff
@@ -272,3 +274,113 @@ def check_data_query_args(data_query_time, data_query_tz):
...
+
+    last_in_group = df.drop(TS_FIELD_NAME, axis=1).groupby(
+        idx,
+        sort=False,
+    ).last()
+
+    # For the number of things that we're grouping by (except TS), unstack
+    # the df
+    for _ in range(len(idx) - 1):
+        last_in_group = last_in_group.unstack()
```
llllllllll: `last_in_group.unstack([-1, -2])`

zipline/pipeline/loaders/utils.py
```diff
@@ -272,3 +274,113 @@ def check_data_query_args(data_query_time, data_query_tz):
...
+    last_in_group : pd.DataFrame
+        A DataFrame with dates as the index and fields used in the groupby as
+        levels of a multiindex of columns.
+
+    """
+    idx = dates[dates.searchsorted(
+        df[TS_FIELD_NAME].values.astype('datetime64[D]')
+    )]
+    if have_sids:
+        idx = [idx, SID_FIELD_NAME] + extra_groupers
```
llllllllll: we should add the extra groupers regardless.

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,266 @@
...
+        groups = groupby(lambda x: x.dataset.num_quarters, columns)
+        groups_columns = dict(groups)
+        if (pd.Series(groups_columns.keys()) < 0).any():
+            raise ValueError("Must pass a number of quarters >= 0")
+        out = {}
+        self.estimates[NORMALIZED_QUARTERS] = normalize_quarters(
+            self.estimates[FISCAL_YEAR_FIELD_NAME],
+            self.estimates[FISCAL_QUARTER_FIELD_NAME],
+        ).astype(float)
+        for num_quarters, columns in groups_columns.iteritems():
```
llllllllll: .items

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,266 @@
...
+        # attribute, given that they're created dynamically?
+        groups = groupby(lambda x: x.dataset.num_quarters, columns)
+        groups_columns = dict(groups)
+        if (pd.Series(groups_columns.keys()) < 0).any():
+            raise ValueError("Must pass a number of quarters >= 0")
+        out = {}
+        self.estimates[NORMALIZED_QUARTERS] = normalize_quarters(
+            self.estimates[FISCAL_YEAR_FIELD_NAME],
+            self.estimates[FISCAL_QUARTER_FIELD_NAME],
+        ).astype(float)
```
llllllllll: why is this a float?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,266 @@
...
+            # on each date in the index for each sid and quarter.
+            last_per_qtr = last_in_date_group(
+                self.estimates, True, dates, assets,
+                extra_groupers=[NORMALIZED_QUARTERS]
+            )
+
+            # Forward fill values for each quarter.
+            ffill_across_cols(last_per_qtr, columns)
+            # Stack quarter and sid into the index
+            stacked_last_per_qtr = last_per_qtr.stack(
```
llllllllll: why are we restacking the data? also this can be `.stack([NORMALIZED_QUARTERS, SID_FIELD_NAME])`

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,266 @@
...
+                SID_FIELD_NAME
+            ) == sid]
+            # Determine where we think quarters are changing for this sid.
+            qtr_shifts = sid_result[
+                sid_result[SHIFTED_NORMALIZED_QTRS] !=
+                sid_result[SHIFTED_NORMALIZED_QTRS].shift(1)
+            ]
+            # Iterate backwards making adjustments. No adjustment for 1st
+            # quarter.
+            for row_indexer in list(reversed(qtr_shifts.index))[:-1]:
```
llllllllll: why are we reversing this twice?

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: Left over from a previous iteration. I'll remove.

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: I couldn't figure out any other way to perform the filtering/grouping operations in `load_quarters`.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+    NextQuartersEstimatesLoader,
+    PreviousQuartersEstimatesLoader,
+    split_normalized_quarters, normalize_quarters)
+from zipline.testing import ZiplineTestCase
+from zipline.testing.fixtures import WithAssetFinder, WithTradingSessions
+from zipline.testing.predicates import assert_equal
+from zipline.utils.numpy_utils import datetime64ns_dtype, float64_dtype
+
+
+class Estimates(DataSet):
```
llllllllll: why are these in the tests and not in the actual package?

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+    TS_FIELD_NAME: [pd.Timestamp('2015-01-05'), pd.Timestamp('2015-01-07'),
+                    pd.Timestamp('2015-01-05'), pd.Timestamp('2015-01-17'),
+                    pd.Timestamp('2015-01-05'), pd.Timestamp('2015-01-17'),
+                    pd.Timestamp('2015-01-22'), pd.Timestamp('2015-02-02')],
+    EVENT_DATE_FIELD_NAME:
+        [pd.Timestamp('2015-01-10'), pd.Timestamp('2015-01-10'),
+         pd.Timestamp('2015-01-20'), pd.Timestamp('2015-01-20'),
+         pd.Timestamp('2015-02-10'), pd.Timestamp('2015-02-10'),
+         pd.Timestamp('2015-02-10'), pd.Timestamp('2015-02-10')],
+    'estimate': [1.]*2 + [2.] * 2 + [4.] * 4,
```
llllllllll: spaces around operators.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+    FISCAL_QUARTER_FIELD_NAME: [1]*2 + [2] * 2 + [4] * 4,
+    FISCAL_YEAR_FIELD_NAME: [2015]*8,
+    SID_FIELD_NAME: [0]*8
```
llllllllll: you can just put scalars here

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+    'estimate': [.1, .2, .3, .4],
+    FISCAL_QUARTER_FIELD_NAME: [1.0, 1.0, 2.0, 2.0],
+    FISCAL_YEAR_FIELD_NAME: [2015.0, 2015.0, 2015.0, 2015.0]
+})
+
+
+def gen_estimates():
+    sid_estimates = []
+    sid_releases = []
+    for sid, (q1e1, q1e2, q2e1, q2e2) in enumerate(
```
llllllllll: just store the iterator like:
```python
it = enumerate(...)
for sid, (...) in it:
    ...
```

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+            itertools.permutations(q1_knowledge_dates + q2_knowledge_dates,
+                                   4)
+    ):
+        # We're assuming that estimates must come before the relevant release.
+        if (q1e1 < q1e2 and
+                q2e1 < q2e2 and
+                q1e1 < q1_release_dates[0] and
+                q1e2 < q1_release_dates[1]):
+            sid_estimate = estimates.copy(True)
```
llllllllll: could we just make a factory function that builds this dataframe

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+            sid_estimate[TS_FIELD_NAME] = [q1e1, q1e2, q2e1, q2e2]
+            sid_estimate[SID_FIELD_NAME] = sid
+            sid_estimates += [sid_estimate]
+            sid_release = releases.copy(True)
+            sid_release[SID_FIELD_NAME] = sid_estimate[SID_FIELD_NAME]
+            sid_releases += [sid_release]
```
llllllllll: append

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+    return pd.concat(sid_estimates + sid_releases).reset_index(drop=True)
+
+
+class EstimateTestCase(object):
```
llllllllll: This is a fixture, not a test case

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+
+        cls.ASSET_FINDER_EQUITY_SIDS = list(
+            cls.events[SID_FIELD_NAME].unique()
+        )
+        cls.ASSET_FINDER_EQUITY_SYMBOLS = [
+            's' + str(n) for n in cls.ASSET_FINDER_EQUITY_SIDS
+        ]
+        super(EstimateTestCase, cls).init_class_fixtures()
+
+    def _test_wrong_num_quarters_passed(self):
```
llllllllll: we can drop the leading underscore and then tests that mix this in will inherit the test automatically.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+        Tests that we overwrite values with the correct quarter's estimate at
+        the correct dates.
+        """
+        dataset = QuartersEstimates(num_quarters_out)
+
+        class SomeFactor(CustomFactor):
+            inputs = [dataset.estimate]
+            window_length = window_len
+
+            def compute(self, today, assets, out, *inputs):
```
llllllllll: we know what the inputs are, it is ust `estimate`

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,747 @@
...
+                unique_inputs = np.unique(inputs).tolist()
```
llllllllll: I find this test extremely confusing. Could we just store the window for in a dictionary for each day and then walk through the days and assert that the windows are equal to some hard coded expectations? Also, we are only testing with constants for each earnings quarter so it may be misaligned.

llllllllll: we also should test with more than 1 sid

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,391 @@
...
+            estimates[FISCAL_QUARTER_FIELD_NAME].notnull() &
+            estimates[FISCAL_YEAR_FIELD_NAME].notnull()
+        ]
+
+        self.base_column_name_map = base_column_name_map
+
+    @abstractmethod
+    def load_quarters(self, num_quarters, last, dates):
+        pass
+
```
llllllllll: `raise NotImplementedError('load_quarters')`

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,391 @@
...
+        Returns
+        --------
+        requested_qtr_data : pd.DataFrame
+            The DataFrame with final values for the requested quarter for all
+            columns; `dates` are the index and columns are a MultiIndex with
+            sids at the top level and the dataset columns on the bottom.
+        """
+        requested_qtr_data = stacked_last_per_qtr.loc[idx]
+        # We no longer need this in the index, but we do need it as a column
```
llllllllll: `iloc`?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,391 @@
...
+            split_normalized_quarters(
+                requested_qtr_data[SHIFTED_NORMALIZED_QTRS]
+            )
+        # Move sids into the columns. Once we're left with just dates
+        # as the index, we can reindex by all dates so that we have a
+        # value for each calendar date.
+        requested_qtr_data = requested_qtr_data.unstack(
+            SID_FIELD_NAME
+        ).reindex(dates)
+        return requested_qtr_data
```
llllllllll: should this be `reindex(method='ffill')`? What if the starting index is not a subset of `dates`?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,391 @@
...
+        else:
+            overwrite = Float641DArrayOverwrite
+            missing_value = np.NaN
+        for sid_idx, sid in enumerate(assets):
+            zero_qtr_sid_data = zero_qtr_data[
+                zero_qtr_data.index.get_level_values(SID_FIELD_NAME) == sid
+            ]
+            # Determine where quarters are changing for this sid.
+            qtr_shifts = zero_qtr_sid_data[
+                zero_qtr_sid_data[NORMALIZED_QUARTERS] !=
```
llllllllll: `.diff()`

zipline/pipeline/loaders/utils.py
```diff
@@ -272,3 +274,116 @@ def check_data_query_args(data_query_time, data_query_tz):
...
+                    names=cols.names,
+                ),
+            )
+        else:
+            last_in_group = last_in_group.reindex(dates)
+
+    return last_in_group
+
+
+def ffill_across_cols(df, columns, name_map):
```
llllllllll: cant the name map be built from the columns argument

tests/pipeline/test_quarters_estimates.py
mtydykov: This is just a dummy `DataSet` for testing.

tests/pipeline/test_quarters_estimates.py
mtydykov: My thinking was that this way it's easier to see how the data corresponds to the cases and that it's more concise.

tests/pipeline/test_quarters_estimates.py
mtydykov: Did you mean that we should move this function to `zipline/utils/factory.py`?

llllllllll: I just mean we should move this logic into its own function in this file

tests/pipeline/test_quarters_estimates.py
llllllllll: idk how `[0] * 8` is more concise than `0`

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: No, should be `loc` because `idx` is a `MultiIndex`.

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: The starting index is built from `dates` and from that point dates are only removed; why would it not be a subset?

zipline/pipeline/loaders/utils.py
mtydykov: Not in all cases. This would work for the events loader, but for the estimates loader, the datasets of the columns passed in are built dynamically, and so we need to look at the parent to get the right name mapping.

tests/pipeline/test_quarters_estimates.py
mtydykov: Ah, forgot you can do that when creating a DataFrame; I thought you meant writing out the list of scalars.

tests/pipeline/test_quarters_estimates.py
llllllllll: why do we not have a rea dataset?

mtydykov: Do we need one in zipline? We don't have one for events.

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def load_adjusted_array(self, columns, dates, assets, mask):
+        # TODO: how can we enforce that datasets have the num_quarters
+        # attribute, given that they're created dynamically?
+        groups = groupby(lambda x: x.dataset.num_quarters, columns)
+        groups_columns = dict(groups)
+        if (np.array(list(groups_columns.keys())) < 0).any():
+            raise ValueError("Must pass a number of quarters >= 0")
+        out = {}
+        self.estimates[NORMALIZED_QUARTERS] = normalize_quarters(
```
llllllllll: we don't need to do this each time we call `load_adjusted_array`, could we move this to the constructor?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,391 @@
...
+        qtr_crossover_point :
+            Whether we should use the 'right' or 'left' side when doing
+            searchsorted on the dates for quarter boundaries.
+
+        Returns
+        -------
+        adjusted_array : AdjustedArray
+            The array of data and overwrites for the given column.
+        """
+        adjustments = defaultdict(list)
```
llllllllll: why is this a defaultdict?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,391 @@
...
+                    # values going up to the starting index of that quarter
+                    # with the missing value for this column.
+                    else:
+                        adjustments[qtr_start_idx] = [
+                            overwrite(
+                                0,
+                                qtr_start_idx - 1,
+                                sid_idx,
+                                sid_idx,
+                                np.array(
```
llllllllll: `np.full`

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,391 @@
...
+        zero_qtr_data = stacked_last_per_qtr.loc[zero_qtr_idx]
+        # We no longer need this in the index, but we do need it as a column
+        # to calculate adjustments.
+        zero_qtr_data = zero_qtr_data.reset_index(NORMALIZED_QUARTERS)
+        if column.dtype == datetime64ns_dtype:
+            overwrite = Datetime641DArrayOverwrite
+            missing_value = np.datetime64('NaT', 'ns')
+        else:
+            overwrite = Float641DArrayOverwrite
+            missing_value = np.NaN
```
llllllllll: `column.missing_value`

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: I guess it doesn't have to be, since we're not extending the list for any key at any point.

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: Makes sense

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: How should I use it? 

I want:

```
(Pdb++) zero_qtr_sid_data[zero_qtr_sid_data[NORMALIZED_QUARTERS] != zero_qtr_sid_data[NORMALIZED_QUARTERS].shift(1)]
                               normalized_quarters  estimate event_date  \
dates                     sid
2015-01-05 00:00:00+00:00 0                   8060        10 2015-01-10
2015-01-12 00:00:00+00:00 0                   8061        20 2015-01-20

                               fiscal_quarter  fiscal_year
dates                     sid
2015-01-05 00:00:00+00:00 0                 1         2015
2015-01-12 00:00:00+00:00 0                 2         2015

```

I get: 


```
(Pdb++) zero_qtr_sid_data[NORMALIZED_QUARTERS].diff(1)
dates                      sid
2015-01-05 00:00:00+00:00  0     NaN
2015-01-06 00:00:00+00:00  0       0
2015-01-07 00:00:00+00:00  0       0
2015-01-08 00:00:00+00:00  0       0
2015-01-09 00:00:00+00:00  0       0
2015-01-12 00:00:00+00:00  0       1
2015-01-13 00:00:00+00:00  0       0
2015-01-14 00:00:00+00:00  0       0
2015-01-15 00:00:00+00:00  0       0
2015-01-16 00:00:00+00:00  0       0
2015-01-20 00:00:00+00:00  0       0
```

tests/pipeline/test_adjusted_array.py
```diff
@@ -442,6 +548,22 @@ def test_multiplicative_adjustments(self,
                     datetime64ns_dtype,
                 ),
             ),
+            _gen_overwrite_1d_array_adjustment_case(
+                'float',
+                make_input=as_dtype(float64_dtype),
```
llllllllll: could we just pass the dtype in, the rest of the arguments can be infered from that.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,891 @@
...
+
+class WithEstimates(WithTradingSessions, WithAssetFinder):
+    """
+    ZiplineTestCase mixin providing cls.loader and cls.events as class
+    level fixtures.
+
+
+    Methods
+    -------
+    make_loader() -> PipelineLoader
```
llllllllll: this takes two arguments.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,891 @@
...
+        Method which returns the loader to be used throughout tests.
+    """
+
+    # Short window defined in order for test to run faster.
+    START_DATE = pd.Timestamp('2014-12-28')
+    END_DATE = pd.Timestamp('2015-02-03')
+
+    @classmethod
+    def make_loader(cls, events, columns):
+        return None
```
llllllllll: maybe change this to: `raise NotImplementedError('make_loader')` so users know they must implement this.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,891 @@
...
+                          columns=[SID_FIELD_NAME,
+                                   TS_FIELD_NAME,
+                                   EVENT_DATE_FIELD_NAME,
+                                   FISCAL_QUARTER_FIELD_NAME,
+                                   FISCAL_YEAR_FIELD_NAME,
+                                   'estimate'],
+                          index=[0])
+
+    def test_wrong_num_quarters_passed(self):
+        with self.assertRaises(ValueError):
```
llllllllll: can we lift all but the expected failure out of the `with self.assertRaises`? This is wrapping too much with a fairly general error.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,891 @@
...
+    """
+    Tests that next quarter loader correctly breaks if an incorrect
+    number of quarters is passed.
+    """
+    @classmethod
+    def make_loader(cls, events, columns):
+        return NextQuartersEstimatesLoader(events, columns)
+
+
+class WithEstimatesT0(WithEstimates):
```
llllllllll: what does `TO` stand for?

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,109 @@
...
+        The expression representing the data to load.
+    resources : dict, optional
+        Mapping from the loadable terms of ``expr`` to actual data resources.
+    odo_kwargs : dict, optional
+        Extra keyword arguments to pass to odo when executing the expression.
+    data_query_time : time, optional
+        The time to use for the data query cutoff.
+    data_query_tz : tzinfo or str
+        The timezeone to use for the data query cutoff.
+    dataset : DataSet
```
llllllllll: This doesn't accept a `dataset` argument

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,109 @@
...
+    announced.
+
+    If the '{TS_FIELD_NAME}' field is not included it is assumed that we
+    start the backtest with knowledge of all announcements.
+    """
+
+    @preprocess(data_query_tz=optionally(ensure_timezone))
+    def __init__(self,
+                 expr,
+                 columns,
```
llllllllll: this is not in the docstring

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,109 @@
...
+    Notes
+    -----
+    The expression should have a tabular dshape of::
+
+       Dim * {{
+           {SID_FIELD_NAME}: int64,
+           {TS_FIELD_NAME}: datetime,
+       }}
+
+    And other dataset-specific fields, where each row of the table is a
```
llllllllll: we know the other fields are, we should include them in the docstring.

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,109 @@
...
+    record including the sid to identify the company, the timestamp where we
+    learned about the announcement, and the date when the earnings will be
+    announced.
+
+    If the '{TS_FIELD_NAME}' field is not included it is assumed that we
+    start the backtest with knowledge of all announcements.
+    """
```
llllllllll: This docstring has string format args but is not being formatted.

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,109 @@
...
+
+       Dim * {{
+           {SID_FIELD_NAME}: int64,
+           {TS_FIELD_NAME}: datetime,
+       }}
+
+    And other dataset-specific fields, where each row of the table is a
+    record including the sid to identify the company, the timestamp where we
+    learned about the announcement, and the date when the earnings will be
+    announced.
```
llllllllll: we also need fiscal year/quarter.

zipline/pipeline/loaders/blaze/events.py
```diff
@@ -90,34 +84,8 @@ def __init__(self,
...
-        )
-        if data_query_time is not None:
-            normalize_timestamp_to_query_time(
-                raw,
-                data_query_time,
-                data_query_tz,
-                inplace=True,
-                ts_field=TS_FIELD_NAME,
-            )
+        raw = load_raw_data(assets, dates, self._data_query_time,
```
llllllllll: can you put each arg on its own line?

zipline/pipeline/loaders/blaze/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    dates : pd.datetimeindex
+        the simulation dates to load data for.
+    data_query_time : datetime.time
+        the time used as cutoff for new information.
+    data_query_tz : tzinfo
+        the timezone to normalize your dates to before comparing against
+        `time`.
+    expr : expr
+        the expression representing the data to load.
+    odo_kwargs : dict, optional
```
llllllllll: there is no default for this so it is not optional.

zipline/pipeline/loaders/blaze/utils.py
```diff
@@ -0,0 +1,61 @@
...
+        data_query_time,
+        data_query_tz,
+    )
+    raw = ffill_query_in_range(
+        expr,
+        lower_dt,
+        upper_dt,
+        odo_kwargs,
+    )
+    sids = raw.loc[:, SID_FIELD_NAME]
```
llllllllll: is this different from `raw[SID_FIELD_NAME]`?

zipline/pipeline/loaders/blaze/utils.py
```diff
@@ -0,0 +1,61 @@
...
+        `time`.
+    expr : expr
+        the expression representing the data to load.
+    odo_kwargs : dict, optional
+        extra keyword arguments to pass to odo when executing the expression.
+
+    returns
+    -------
+    raw : pd.dataframe
+        the data symbolized by `expr` materialized in a dataframe.
```
llllllllll: Could we say: 'The result of computing expr and materializing the result as a dataframe'.

tests/pipeline/test_quarters_estimates.py
mtydykov: Time 0 - the immediate next/previous quarter on which all other calculations should be based.

zipline/pipeline/loaders/blaze/estimates.py
mtydykov: Yeah, we should include fiscal quarter/year and event date. There can be other dataset-specific fields that we don't know about, though.

zipline/pipeline/loaders/blaze/estimates.py
mtydykov: Good point; looking at it now, I'm not sure what the original intention was (`blaze/events.py` has a similar docstring format). How should we format these?

zipline/pipeline/loaders/blaze/utils.py
mtydykov: Not that I can think of; that's just how it was done originally in the blaze events loader. Happy to change it - any reason you can think of not to? Just want to make sure I don't break any edge cases.

tests/pipeline/test_quarters_estimates.py
llllllllll: can we call this `TimeZero` then? This looked like the word `to`

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,898 @@
...
+            FISCAL_QUARTER_FIELD_NAME: [1.0, 1.0, 2.0, 2.0],
+            FISCAL_YEAR_FIELD_NAME: [2015.0, 2015.0, 2015.0, 2015.0],
+            TS_FIELD_NAME: [q1e1, q1e2, q2e1, q2e2],
+            SID_FIELD_NAME: sid,
+        })
+
+    @classmethod
+    def init_class_fixtures(cls):
+        # Must be generated before call to super since super uses `events`.
+        cls.events = cls.gen_estimates()
```
llllllllll: Why does super use events? Shouldn't super just cache this? Could we make it a classlazyval maybe? In general you should only add attributes after the call to super.

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,517 @@
...
+                np.full(
+                    len(
+                        last_per_qtr.index[:next_qtr_start_idx]
+                    ),
+                    column.missing_value,
+                    dtype=column.dtype
+                ))]
+
+    def load_adjusted_array(self, columns, dates, assets, mask):
+        groups = groupby(lambda x: x.dataset.num_quarters, columns)
```
llllllllll: could we catch this attribute error and reraise something that tells the user what that attribute is supposed to represent

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,898 @@
...
+    def test_wrong_num_quarters_passed(self):
+        dataset = QuartersEstimates(-1)
+        engine = SimplePipelineEngine(
+            lambda x: self.loader,
+            self.trading_days,
+            self.asset_finder,
+        )
+        p = Pipeline({c.name: c.latest for c in dataset.columns})
+
+        with self.assertRaises(ValueError):
```
llllllllll: since this is only a valueerror could we check the message? Also, there is some logic around the formatting that we want to test.

llllllllll: we can test the message with `zipline.testing.predicates.assert_raises_regex`

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,898 @@
...
+    events = pd.DataFrame({SID_FIELD_NAME: 0},
+                          columns=[SID_FIELD_NAME,
+                                   TS_FIELD_NAME,
+                                   EVENT_DATE_FIELD_NAME,
+                                   FISCAL_QUARTER_FIELD_NAME,
+                                   FISCAL_YEAR_FIELD_NAME,
+                                   'estimate'],
+                          index=[0])
+
+    def test_wrong_num_quarters_passed(self):
```
llllllllll: should we test a dataset with no `num_quarters` attribute?

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,898 @@
...
+                    q2_knowledge = ts_eligible_estimates[
+                        ts_eligible_estimates[FISCAL_QUARTER_FIELD_NAME] == 2
+                    ]
+                    expected_estimate = self.get_expected_estimate(
+                        q1_knowledge,
+                        q2_knowledge,
+                        comparable_date,
+                    )
+                    if not expected_estimate.empty:
+                        for colname in sid_estimates.columns:
```
llllllllll: this loop could be simplified to:
```python
assert_equal(sid_estimates.iloc[i], expected_estimate[sid_estimates.columns]]
```

though, there are extra columns in the expected data which seems strange. If we resolve the extra columns then we can not do the projection on the rhs there.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,898 @@
...
+                            expected_value = expected_estimate[colname]
+                            computed_value = sid_estimates.iloc[i][colname]
+                            assert_equal(expected_value, computed_value)
+                    else:
+                        # There are no eligible 'next' estimates on this day;
```
llllllllll: or previous, right?

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,898 @@
...
+
+    def get_expected_estimate(self,
+                              q1_knowledge,
+                              q2_knowledge,
+                              comparable_date):
+        # If our latest knowledge of q1 is that the release is
+        # happening on this simulation date or later, then that's
+        # the estimate we want to use.
+        if (not q1_knowledge.empty and
+            q1_knowledge.iloc[-1][EVENT_DATE_FIELD_NAME] >=
```
llllllllll: doing the select before the iloc is faster because it shuffles less data.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,898 @@
...
+                comparable_date):
+            return q1_knowledge.iloc[-1]
+        # If q1 has already happened or we don't know about it
+        # yet and our latest knowledge indicates that q2 hasn't
+        # happened yet, then that's the estimate we want to use.
+        elif (not q2_knowledge.empty and
+              q2_knowledge.iloc[-1][EVENT_DATE_FIELD_NAME] >=
+                comparable_date):
+            return q2_knowledge.iloc[-1]
+        return pd.DataFrame()
```
llllllllll: instead of checking for an empty frame, could we just return `None` and check for that?

tests/pipeline/test_quarters_estimates.py
mtydykov: Super uses `events` during construction to assign `sids`, `ASSET_FINDER_EQUITY_SIDS`, `ASSET_FINDER_EQUITY_SIDS`,  and to make the loader. Don't see how a classlazyval would work in this case, but I could add an abstract method to super, the purpose of which will be to create `events`, and subclasses can define it.

tests/pipeline/test_quarters_estimates.py
mtydykov: There are extra columns there because it's the raw data, which includes sid and timestamp.

zipline/pipeline/loaders/blaze/estimates.py
mtydykov: Fixed

tests/pipeline/test_quarters_estimates.py
mtydykov: Changed

mtydykov:
Should we filter out estimates that were made after the release happened in the loader? Or should we assume that this doesn't happen in the raw data the loader receives?

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,955 @@
...
+                        # There are no eligible 'next'/'previous' estimates on
+                        # this day; everything should be null.
+                        assert_true(sid_estimates.iloc[i].isnull().all())
+                else:
+                    # We don't know about any estimates on this day;
+                    # everything should be null.
+                    assert_true(sid_estimates.iloc[i].isnull().all())
+
+
+class NextEstimate(WithEstimatesTimeZero, ZiplineTestCase):
```
llllllllll: This test is emitting a pandas warning about assigning into a copy:
```
test_estimates (tests.pipeline.test_quarters_estimates.NextEstimate) ... ok (4.5815s)
/home/yui/.virtualenvs/qexec/lib/python2.7/site-packages/pandas/core/indexing.py:118: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  self._setitem_with_indexer(indexer, value)
```

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,955 @@
...
+    def init_class_fixtures(cls):
+        super(WithEstimatesTimeZero, cls).init_class_fixtures()
+
+    def get_expected_estimate(self,
+                              q1_knowledge,
+                              q2_knowledge,
+                              comparable_date):
+        return pd.DataFrame()
+
+    def test_estimates(self):
```
llllllllll: we spend almost half the time in this test inside `dataframe.__getitem__`. Is there anything we can do to speed this up a bit?

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -0,0 +1,117 @@
...
+    -----
+    The expression should have a tabular dshape of::
+
+       Dim * {{
+           {SID_FIELD_NAME}: int64,
+           {TS_FIELD_NAME}: datetime,
+           {FISCAL_YEAR_FIELD_NAME}: float64,
+           {FISCAL_QUARTER_FIELD_NAME}: float64,
+           {EVENT_DATE_FIELD_NAME}: datetime,
+       }}
```
mtydykov: I'm not sure what it should look like (same question for `events.py`).

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
+from collections import defaultdict
```
mtydykov: Is it really only for earnings estimates, though? Can't this be for estimates of other values associated with a quarterly event?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+                # Find the starting index of the quarter that comes right
+                # after this row. This isn't the starting index of the
+                # requested quarter, but simply the date we cross over into a
+                # new quarter.
+                next_qtr_start_idx = dates.searchsorted(
+                    zero_qtr_data.loc[
+                        row_indexer
+                    ][EVENT_DATE_FIELD_NAME],
+                    side='left'
+                    if isinstance(self, PreviousQuartersEstimatesLoader)
```
mtydykov: Definitely, that's a good idea

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+            )
+            zero_qtr_data = stacked_last_per_qtr.loc[zeroth_quarter_idx]
+            # Doing it this way because creating a MultiIndex from scratch
+            # results in being unable to unstack sids because of duplicate
+            # values, even though the MultiIndex is created with the same
+            # exact values as below - possible pandas bug.
+            requested_qtr_idx = zero_qtr_data.reset_index(
+                NORMALIZED_QUARTERS
+            ).set_index(
+                pd.Series(self.get_shifted_qtrs(
```
mtydykov: Yeah, that's a good point.

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+                                                      requested_qtr_data,
+                                                      last_per_qtr,
+                                                      dates,
+                                                      assets_with_data,
+                                                      columns)
+            for col in columns:
+                column_name = self.name_map[col.name]
+                # We may have dropped assets if they never have any data for
+                # the requested quarter.
+                df = pd.DataFrame(data=requested_qtr_data[column_name],
```
mtydykov: I needed to put back all the missing assets. What's a better way of doing it?

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+            index values, sorted by event date.
+
+        Returns
+        -------
+        previous_releases_per_date_index : pd.MultiIndex
+            An index of calendar dates, sid, and normalized quarters, for only
+            the rows that have a previous event.
+        """
+
+        # We reset the index here because in pandas3, a groupby on the index
```
mtydykov: I probably meant some particular version of pandas, but unsure of what it is anymore. At any rate, `as_index=False` works.

zipline/testing/core.py
```diff
@@ -49,7 +49,9 @@
 from zipline.utils import security_list
 from zipline.utils.calendars import get_calendar
 from zipline.utils.input_validation import expect_dimensions
-from zipline.utils.numpy_utils import as_column
+from zipline.utils.numpy_utils import (
```
mtydykov: `fixtures.py` I'll change back - there were leftover changes to the imports from when I thought I'd be moving things there. I don't see how we can get away with not editing `core.py`, however; there are lots of changes further down in the file where we refactor certain logic to reuse it in other places.

tests/pipeline/test_quarters_estimates.py
mtydykov: I can't think of a way to change the result-checking algorithm, but I could try narrowing the quarter range down even more.

zipline/pipeline/loaders/blaze/utils.py
```diff
@@ -0,0 +1,69 @@
...
+                  data_query_time,
+                  data_query_tz,
+                  expr,
+                  odo_kwargs,
+                  checkpoints=None):
+    """
+    given an expression representing data to load, perform normalization and
+    forward-filling and return the data, materialized.
+
+    parameters
```
mtydykov: Done

zipline/pipeline/loaders/blaze/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    data_query_tz : tzinfo
+        the timezone to normalize your dates to before comparing against
+        `time`.
+    expr : expr
+        the expression representing the data to load.
+    odo_kwargs : dict
+        extra keyword arguments to pass to odo when executing the expression.
+    checkpoints : expr, optional
+        the expression representing the checkpointed data for `expr`.
+    returns
```
mtydykov: Done

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+            "Got Columns: {received}\n"
+            "Expected Columns: {required}".format(
+                missing=sorted(missing),
+                received=sorted(received),
+                required=sorted(required),
+            )
+        )
+
+
+class QuarterEstimatesLoader(PipelineLoader):
```
mtydykov: Done

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: Done

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+                                     sid,
+                                     sid_idx,
+                                     columns):
+        """
+        Add entries to the dictionary of columns to adjustments for the given
+        sid and the given quarter.
+
+        Parameters
+        ----------
+        col_to_overwrites : dict [column_name -> list of ArrayAdjustment]
```
mtydykov: Fixed

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+                        sid_idx
+                    )]
+
+    def overwrite_with_null(self,
+                            column,
+                            dates,
+                            next_qtr_start_idx,
+                            overwrite,
+                            sid_idx):
+        return overwrite(
```
mtydykov: Fixed

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+            raise ValueError(
+                INVALID_NUM_QTRS_MESSAGE % ','.join(
+                    str(qtr) for qtr in groups if qtr < 0
+                )
+
+            )
+        out = {}
+        # To optimize performance, only work below on assets that are
+        # actually in the raw data.
+        assets_with_data = set.intersection(
```
mtydykov: Fixed

zipline/pipeline/loaders/quarter_estimates.py
```diff
@@ -0,0 +1,582 @@
...
+        """
+        # Get a DataFrame indexed by date with a MultiIndex of columns of [
+        # self.estimates.columns, normalized_quarters, sid], where each cell
+        # contains the latest data for that day.
+        last_per_qtr = last_in_date_group(
+            self.estimates, dates, assets_with_data, reindex=True,
+            extra_groupers=[NORMALIZED_QUARTERS]
+        )
+        # Forward fill values for each quarter/sid/dataset column.
+        ffill_across_cols(last_per_qtr, columns, self.name_map)
```
mtydykov: Fixed this in `ffill_across_cols` by calling `unstack` in a loop. The problem was that calling `unstack` once with a list of levels as an argument was creating a DF with object columns due to a bug in pandas.

zipline/pipeline/loaders/blaze/estimates.py
llllllllll: you need to add `__doc__ = __doc__.format(...)`

tests/pipeline/test_quarters_estimates.py
llllllllll: Can we vectorize some checks or lift some of the selections out of the loops?

zipline/pipeline/loaders/quarter_estimates.py
mtydykov: Ok, modified.

tests/pipeline/test_quarters_estimates.py
mtydykov: I just spent some time fiddling around with the test; tried to do both. This didn't improve the speed, but it did simplify the test a fair amount. If you see any further optimizations happy to incorporate those.

tests/pipeline/test_quarters_estimates.py
mtydykov: Hmm, not seeing that one on my machine.

mtydykov: Hmm, not seeing that one my machine. Thought it could be because I've changed that test around.

zipline/pipeline/loaders/blaze/events.py
```diff
@@ -48,17 +46,21 @@ class BlazeEventsLoader(PipelineLoader):
...
     record including the sid to identify the company, the timestamp where we
-    learned about the announcement, and the date when the earnings will be z
-    announced.
+    learned about the announcement, and the event date.
 
     If the '{TS_FIELD_NAME}' field is not included it is assumed that we
     start the backtest with knowledge of all announcements.
     """
 
+    __doc__ == __doc__.format(SID_FIELD_NAME=SID_FIELD_NAME,
```
mtydykov: Gah, wow. My mistake. Good catch!

tests/pipeline/test_quarters_estimates.py
```diff
@@ -0,0 +1,943 @@
...
+        }
+
+
+class QuarterShiftTestCase(ZiplineTestCase):
+    """
+    This tests, in isolation, quarter calculation logic for shifting quarters
+    backwards/forwards from a starting point.
+    """
+    def test_quarter_normalization(self):
+        input_yrs = pd.Series([0] * 4, dtype=np.int64)
```
mtydykov: By 'not a constant', do you just mean not all the same year?

PR #1397, opened 2016-08-16
MAINT: Use TradingCalendar objects for bundles

yankees714:
Instead of trading days, opens, and closes, register now takes a TradingCalendar object, along with a start_session and end_session. The ingest function is now passed these values instead as well.

### DIFF
--- a/docs/source/appendix.rst
+++ b/docs/source/appendix.rst
@@ -261,6 +261,24 @@ Asset Metadata
 .. autoclass:: zipline.assets.AssetConvertible
    :members:
 
+
+Trading Calendar API
+~~~~~~~~~~~~~~~~~~~~
+
+.. autofunction:: zipline.utils.calendars.get_calendar
+
+.. autoclass:: zipline.utils.calendars.TradingCalendar
+   :members:
+
+.. autofunction:: zipline.utils.calendars.register_calendar
+
+.. autofunction:: zipline.utils.calendars.register_calendar_type
+
+.. autofunction:: zipline.utils.calendars.deregister_calendar
+
+.. autofunction:: zipline.utils.calendars.clear_calendars
+
+
 Data API
 ~~~~~~~~
 
@@ -316,6 +334,7 @@ Bundles
    :func:`~zipline.data.bundles.unregister`.
 
 .. autofunction:: zipline.data.bundles.yahoo_equities
+
 
 
 Utilities
--- a/docs/source/bundles.rst
+++ b/docs/source/bundles.rst
@@ -302,20 +302,21 @@ have.
 ``calendar``
 ````````````
 
-``calendar`` is an instance of ``zipline.utils.calendars.TradingCalendar``. The
-calendar is provided to help some bundles generate queries for the days needed.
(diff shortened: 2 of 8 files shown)


richafrank:
cc @llllllllll

zipline/data/bundles/core.py
```diff
@@ -210,9 +208,9 @@ def _make_bundle_core():
     @curry
     def register(name,
                  f,
-                 calendar=trading_days,
-                 opens=open_and_closes['market_open'],
-                 closes=open_and_closes['market_close'],
+                 calendar=nyse_cal,
```
llllllllll: could this also take the calendar name, like `'nyse'` so that we don't need to make the calendar at module scope?

llllllllll: we could also make `get_calendar` idempotent so we can just call `get_calendar` on the input.

yankees714: Yep, I think it'll be better to put the calendar_name in the payload instead of the entire calendar as well

llllllllll: I think that users should still have the option to pass the actual calendar though

llllllllll:
Can you update: https://github.com/quantopian/zipline/blob/master/docs/source/bundles.rst#calendar to describe the new format for the calendar?

docs/source/bundles.rst
```diff
@@ -300,9 +302,20 @@ have.
 ``calendar``
 ````````````
 
-``calendar`` is a ``pandas.DatetimeIndex`` object holding all of the trading
-days that the bundle should load data for. The calendar is provided to help some
-bundles generate queries for the days needed.
+``calendar`` is an instance of ``zipline.utils.calendars.TradingCalendar``. The
```
llllllllll: We should use:
```rst
:class:`zipline.utils.calendars.TradingCalendar`
```
and make sure it is in the appendix.

docs/source/bundles.rst
```diff
@@ -300,9 +302,20 @@ have.
...
+calendar is provided to help some bundles generate queries for the days needed.
+
+``start_session``
+````````````
+
+``start_session`` is a ``pandas.Timestamp`` object indicating the first day
```
llllllllll:
```rst
:class:`pandas.Timestamp`
```

llllllllll: also, do the start and end need to be in the `calendar` or will it align them?

yankees714: Do you think we should force them within the bounds of the calendar?

llllllllll: Idk, I think either is fine but whichever it is should be documented.

docs/source/appendix.rst
```diff
@@ -261,6 +261,24 @@ Asset Metadata
 .. autoclass:: zipline.assets.AssetConvertible
    :members:
 
+
+Trading Calendar API
+~~~~~~~~~~~~~~~~~~~~
```
yankees714: @jbredeche Needed to add TradingCalendar to the API reference, does this look sufficient to you?

Issue #1402, opened 2016-08-17
Data bundle `clean()` method doesn't actually remove old files


irlevesque:
# Environment

* Operating System: `Linux 4.4.11-23.53.amzn1.x86_64 #1 SMP x86_64 GNU/Linux`
* Python Version: `Python 2.7.10`
* Python Bitness: `64`
* How did you install Zipline: `pip`
* Python packages: ```bcolz==0.12.1
boto==2.32.1
Bottleneck==1.1.0
cachetools==1.1.6
click==3.3
contextlib2==0.4.0
cyordereddict==1.0.0
Cython==0.24
decorator==4.0.10
dnspython==1.11.1
DynectDNS==0.3
ecdsa==0.11
enum34==1.0
Fabric==1.10.0
flake8==2.4.1
intervaltree==2.1.0
IPy==0.81
ipython==2.2.0
ipython-genutils==0.1.0
iso8601==0.1.10
Logbook==0.12.5
mccabe==0.3.1
multipledispatch==0.4.8
networkx==1.11
numexpr==2.6.0
numpy==1.9.3
pandas==0.16.1
pandas-datareader==0.2.1
paramiko==1.14.1
patsy==0.4.1
pep8==1.7.0
pycrypto==2.6.1
pyflakes==0.8.1
PyGithub==1.25.0
pymongo==2.8.1
python-dateutil==2.5.3
pytz==2016.4
PyYAML==3.11
requests==2.5.1
requests-file==1.4
schematics==1.1.1
scipy==0.17.1
shovel==0.3.0
singledispatch==3.4.0.3
six==1.9.0
sortedcontainers==1.5.3
SQLAlchemy==1.0.8
statsmodels==0.6.1
straitlets==0.2.6
toolz==0.8.0
traitlets==4.2.2
zipline==1.0.1```

# Description of Issue

* What did you expect to happen?
Call the `clean()` method in `zipline.data.bundles` to remove old ingested data.

* What happened instead?
An empty set is returned, and none of the old data is removed.

Here is how you can reproduce this issue on your machine:

### Ingest data
1. `from zipline.data.bundles import ingest`
2. `environ = { 'ZIPLINE_ROOT': '/tmp/zipline-root' , 'QUANDL_API_KEY': api_key }`
3. `ingest('quandl', environ=environ, show_progress=False)`
```
$ ls /tmp/zipline-root/data/quandl/2016-08-16T15;29;49.497618
[...]

PR #1403, opened 2016-08-17
BUG: Fixes should_clean for keep_last=0

richafrank:
For https://github.com/quantopian/zipline/issues/1402

### DIFF
--- a/zipline/data/bundles/core.py
+++ b/zipline/data/bundles/core.py
@@ -7,7 +7,7 @@ import warnings
 from contextlib2 import ExitStack
 import click
 import pandas as pd
-from toolz import curry, complement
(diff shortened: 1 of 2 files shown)


zipline/data/bundles/core.py
```diff
@@ -563,11 +563,16 @@ def should_clean(name):
                     (after is not None and dt > after)
                 )
 
-        else:
+        elif keep_last > 0:
             last_n_dts = set(all_runs[-keep_last:])
```
llllllllll: we could use `toolz.take(keep_last, reversed(all_runs))` and then check >= 0 and cut the special case on `0`

richafrank: I like that.

PR #1404, opened 2016-08-17
MAINT: Use session bar reader interface.

ehebert:
Replace `DailyBarReader` with `SessionBarReader`.

This was intended to go with the patch that added the `SessionBarReader`
abstract base class.

Also, added `trading_calendar` property decorator.


jbredeche:
LGTM

PR #1405, opened 2016-08-17
ENH: Session bar reader resampled from minute data

ehebert:
Implement a `SessionBarReader` which uses a minute bar reader as a
backing source, resampling the minute bars into the box around the
corresponding session data.

Also, add future/CME test cases to resample suite.

This PR also includes https://github.com/quantopian/zipline/pull/1404

This PR is a step towards providing history windows for Future pricing data via the data portal, where the next step will be using the resampled reader in a `SessionHistoryLoader` (which will be a refactored version of `USEquityLoader`.

Resampling is used instead of the `BcolzDailyBarReader` for the following reasons.
1) Provide more flexibility on the boundaries of the window, e.g. using `CME` on all futures, or using Future specific calendars.
2) Remove need for also generating the bcolz daily/session bars for Futures data. The data set is small enough that the need for packing the data in a pipeline friendly structure is not required.

### DIFF
--- a/zipline/data/resample.py
+++ b/zipline/data/resample.py
@@ -15,6 +15,10 @@ from collections import OrderedDict
 
 import numpy as np
 import pandas as pd
+from pandas import DataFrame
+
+from zipline.data.session_bars import SessionBarReader
+from zipline.utils.memoize import lazyval
 
 _MINUTE_TO_SESSION_OHCLV_HOW = OrderedDict((
     ('open', 'first'),
@@ -45,11 +49,10 @@ def minute_to_session(minute_frame, calendar):
         A DataFrame with the columns `open`, `high`, `low`, `close`, `volume`,
         and `day` (datetime-like).
     """
-    # Group minutes into their respective days. Note that this will
-    # create groups for all trading days in the desired range,
-    # including days with no minute data.
-    return minute_frame.resample(calendar.day,
-                                 how=_MINUTE_TO_SESSION_OHCLV_HOW)
+    how = OrderedDict((c, _MINUTE_TO_SESSION_OHCLV_HOW[c])
+                      for c in minute_frame.columns)
+    return minute_frame.groupby(calendar.minute_to_session_label).agg(
+        how)
 
 
 class DailyHistoryAggregator(object):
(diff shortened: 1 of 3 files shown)


zipline/data/resample.py
```diff
@@ -450,3 +453,48 @@ def volumes(self, assets, dt):
...
+
+
+class MinuteResampleSessionBarReader(SessionBarReader):
+
+    def __init__(self, calendar, minute_bar_reader):
+        self._calendar = calendar
+        self._minute_bar_reader = minute_bar_reader
+
+    def _get_resampled(self, columns, start_dt, end_dt, assets):
+        minute_data = self._minute_bar_reader.load_raw_arrays(
```
jbredeche: I can't remember - what type is `minute_data`?

Also, I was surprised that the minute_bar_reader doesn't need to know about the calendar.  Or was that already handled by whoever instantiated the minute_bar_reader?

zipline/data/resample.py
```diff
@@ -45,11 +49,10 @@ def minute_to_session(minute_frame, calendar):
...
         and `day` (datetime-like).
     """
-    # Group minutes into their respective days. Note that this will
-    # create groups for all trading days in the desired range,
-    # including days with no minute data.
-    return minute_frame.resample(calendar.day,
-                                 how=_MINUTE_TO_SESSION_OHCLV_HOW)
+    how = OrderedDict((c, _MINUTE_TO_SESSION_OHCLV_HOW[c])
+                      for c in minute_frame.columns)
+    return minute_frame.dropna().groupby(calendar.minute_to_session_label).agg(
```
jbredeche: this will likely benefit from @nathanwolfe's speedup here: https://github.com/quantopian/zipline/pull/1407

zipline/data/resample.py
ehebert:
```
Signature: zipline.data.minute_bars.MinuteBarReader.load_raw_arrays(self, fields, start_dt, end_dt, sids)
Docstring:
Retrieve the arrays of pricing data for the given coordinates of
``fields`` (OHLCV), minute range [``start_dt``, ``end_dt``] and sids.

Parameters
----------
fields : iterable of str
    The OHLCV fields ('open', 'high', 'low', 'close', 'volume') for
    which to read data.
start_dt : pd.Timestamp
    The first minute of the date range for which to read data.
end_dt : pd.Timestamp
    The last minute of the date range for which to read data.
sids : iterable of int
    The sid identifiers for which to retrieve data.

Returns
-------
raw_arrays : list of ndarray
    A list where each item corresponds with the fields in the order
    the fields are given.
    Each item is a 2D array with a shape of (minutes_in_range, sids)
    The OHLC arrays are floats; the 'volume' array is ints.
```

Already handled as instantiation of the `minute_bar_reader`, though adding a calendar argument may be a way to not use the `CME` minutes for all futures when using `load_raw_arrays`.

zipline/data/resample.py
ehebert: :+1:, a next step was going to be to vectorize this. Making to use the incoming method, when merged.

PR #1406, opened 2016-08-17
ENH: Pass calendar instance to BcolzMinuteBarWriter

yankees714:
Updates `BcolzMinuteBarWriter` to take a `TradingCalendar`, start session, and end session. Previously used "first trading day" and `DatetimeIndex` objects of market opens and closes.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -198,16 +198,13 @@ class BcolzMinuteBarMetadata(object):
     ----------
     first_trading_day : datetime-like
         UTC midnight of the first day available in the dataset.
-    minute_index : pd.DatetimeIndex
-        The minutes which act as an index into the corresponding values
-        written into each sid's ctable.
-    market_opens : pd.DatetimeIndex
-        The market opens for each day in the data set. (Not yet required.)
-    market_closes : pd.DatetimeIndex
-        The market closes for each day in the data set. (Not yet required.)
     ohlc_ratio : int
          The factor by which the pricing data is multiplied so that the
          float data can be stored as an integer.
+    calendar :  zipline.utils.calendars.trading_calendar.TradingCalendar
+        The TradingCalendar on which the minute bars are based.
+    minutes_per_day : int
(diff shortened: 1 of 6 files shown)


zipline/data/minute_bars.py
```diff
@@ -229,41 +229,39 @@ def read(cls, rootdir):
...
             ohlc_ratio = raw_data['ohlc_ratio']
 
             if version == 0:
                 # version 0 always assumed US equities.
                 minutes_per_day = US_EQUITIES_MINUTES_PER_DAY
+                # version 0 always assumed NYSE.
+                calendar = get_calendar('NYSE')
             else:
                 minutes_per_day = raw_data['minutes_per_day']
+                calendar = get_calendar(raw_data['calendar_name'])
```
ehebert: This may need a check for `version < 2` (and an increment of the version.), so that Zipline can be updated in places where the file format is read, ahead of the new key being generated on the data set for the first time.

zipline/data/minute_bars.py
```diff
@@ -272,27 +270,58 @@ def write(self, rootdir):
...
+        calendar_name : str
+            The name of the TradingCalendar on which the minute bars are
+            based.
+        start_session : datetime
+            'YYYY-MM-DD' formatted representation of the first trading
+            session in the data set.
+        end_session : datetime
+            'YYYY-MM-DD' formatted representation of the last trading
+            session in the data set.
+        market_opens : list
```
ehebert: Could add a note here about these being deprecated.

zipline/data/minute_bars.py
```diff
@@ -809,11 +832,12 @@ def _get_metadata(self):
 
     @lazyval
     def last_available_dt(self):
-        return self._market_closes[-1]
+        _, close = self.calendar.open_and_close_for_session(self._end_session)
+        return close
 
     @property
     def first_trading_day(self):
```
ehebert: We could push this further and also rename the property, but that has more touch points, so may be better for a different PR.

Issue #1356, opened 2016-07-30
BUG: data.history with frequency of "1d" does not work for panel minute data 


mannau:
With minute panel data now supported thanks to 9103516 I tried a simple example algo which fetches daily data through data.history.
However, this gives me the following error:

_data_portal.py, line 1094, in _get_daily_window_for_sids_
**AttributeError: 'DataPortal' object has no attribute '_equity_history_loader'**

Below the code to reproduce this issue:

``` python
from zipline import run_algorithm
from datetime import datetime
from pytz import UTC
from zipline.data.loader import load_from_yahoo
from zipline.api import order, symbol, record, sid
import pandas as pd
import numpy as np
from zipline.utils.calendars import get_calendar

# Taken from zipline/tests/test_algorithm
def create_minute_df_for_asset(trading_calendar,
                               start_dt,
                               end_dt,
                               interval=1,
                               start_val=1,
                               minute_blacklist=None):

    asset_minutes = trading_calendar.minutes_for_sessions_in_range(
        start_dt, end_dt
    )
    minutes_count = len(asset_minutes)
    minutes_arr = np.array(range(start_val, start_val + minutes_count))

    df = pd.DataFrame(
        {
            "open": minutes_arr + 1,
            "high": minutes_arr + 2,
            "low": minutes_arr - 1,
            "close": minutes_arr,
            "volume": 100 * minutes_arr,
        },
        index=asset_minutes,
    )

    if interval > 1:
        counter = 0
        while counter < len(minutes_arr):
            df[counter:(counter + interval - 1)] = 0
            counter += interval

    if minute_blacklist is not None:
        for minute in minute_blacklist:
            df.loc[minute] = 0

    return df

def initialize(context):
    context.security = sid(1)

def handle_data(context, data):
    price_history = data.history(context.security, "price", 200, "1d")
    order(sid(1), 10)

if __name__ == '__main__':
    start = datetime(2009, 1, 1, tzinfo=UTC)
    end = datetime(2016, 1, 1, tzinfo=UTC)
    capital_base = 100000

    trading_calendar = get_calendar('NYSE')
    dt_transform = trading_calendar.minute_to_session_label

    data = create_minute_df_for_asset(get_calendar('NYSE'), datetime(1999, 1, 4), datetime(2016, 7, 29))
    dfs = {1 : data }
    panel = pd.Panel(dfs)

    results = run_algorithm(
                  start,
                  end,
                  initialize,
                  capital_base,
[...]

PR #1407, opened 2016-08-17
ENH: Add fast "vectorized" `minute_to_session_label` for DatetimeIndex

nathanwolfe:
~~*Incomplete; "vectorized" `minute_to_session_label` is done so far*~~

The new TradingCalendar method is called `minute_index_to_session_labels`.
It takes a DatetimeIndex of in-order market minutes and returns a
DatetimeIndex of the corresponding sessions.

The new method is approximately 100x faster than mapping
`minute_to_session_label` over a large DatetimeIndex.

### DIFF
--- a/zipline/utils/calendars/_calendar_helpers.pyx
+++ b/zipline/utils/calendars/_calendar_helpers.pyx
@@ -1,13 +1,13 @@
-from numpy cimport ndarray, long_t
-from numpy import searchsorted
+from numpy cimport ndarray, int64_t
+from numpy import empty, searchsorted, int64
 cimport cython
 
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def next_divider_idx(ndarray[long_t, ndim=1] dividers, long_t minute_val):
+cpdef int next_divider_idx(ndarray[int64_t, ndim=1] dividers, int64_t minute_val):
     cdef int divider_idx
-    cdef long target
+    cdef int64_t target
 
     divider_idx = searchsorted(dividers, minute_val, side="right")
     target = dividers[divider_idx]
@@ -20,8 +20,8 @@ def next_divider_idx(ndarray[long_t, ndim=1] dividers, long_t minute_val):
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def previous_divider_idx(ndarray[long_t, ndim=1] dividers,
-                      long_t minute_val):
+def previous_divider_idx(ndarray[int64_t, ndim=1] dividers,
+                      int64_t minute_val):
     cdef int divider_idx
 
     divider_idx = searchsorted(dividers, minute_val)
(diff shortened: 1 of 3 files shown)


zipline/utils/calendars/trading_calendar.py
```diff
@@ -704,6 +710,22 @@ def minute_to_session_label(self, dt, direction="next"):
 
         return current_or_next_session
 
+    def minute_index_to_session_labels(self, index):
+        """
+        Given a DatetimeIndex of market minutes, return a DatetimeIndex
+        of the corresponding session labels.
+
+        The market minutes must be in order.
```
jbredeche: can you put numpy-style docstrings with Parameters, Returns, etc sections? (like the other public-facing methods in this class)

jbredeche:
looks great to me, left a minor comment about docstring style.

looks like there's an issue in Python3.

nathanwolfe:
Fixed the docstring and the Python 3 issue.

PR #1405 (session reader based on minute reader) is actually most of what's needed to fix #1356, so I'll refocus this PR just to be the new `minute_index_to_session_labels` method, and create a new PR that uses #1405 to fix the bug.

zipline/utils/calendars/trading_calendar.py
```diff
@@ -684,17 +685,22 @@ def minute_to_session_label(self, dt, direction="next"):
             The label of the containing session.
         """
 
-        idx = searchsorted(self.market_closes_nanos, dt.value)
+        if type(dt) is pd.Timestamp:
```
richafrank: Unless subclasses aren't allowed, what do you think of `isinstance` instead?

Could be a good case for something like `@preprocess(dt=coerce(pd.Timestamp, attrgetter('value')))`.

nathanwolfe: Switching to something else would be fine. Do you recommend `isinstance`? I don't know how the `@preprocess` stuff works, so would that take care of it as well? Would it be preferable to do it that way?

richafrank: Yea, I recommend `isinstance` one way or the other.  Then it's more flexible with subclasses.  

Take a look at the `coerce` preprocessor - it uses `isinstance` [internally] (https://github.com/quantopian/zipline/blob/bfa3e6f153698df8dd04334012aeacde9f3912c9/zipline/utils/input_validation.py#L575-L575) to decide whether to run the coercion function or pass through the original arg.

richafrank:
> I think there's a platform-dependent error now... that sucks

FWIW Usually it's a 32-bit issue.

ssanderson:
> I think there's a platform-dependent error now... that sucks

@nathanwolfe where you're using `long` in Cython, you probably want `np.int64_t`.

zipline/utils/calendars/_calendar_helpers.pyx
```diff
@@ -1,11 +1,11 @@
-from numpy cimport ndarray, long_t
-from numpy import searchsorted
+from numpy cimport ndarray, long, long_t
```
ssanderson: This probably wants to be `int64_t` for anything that's converting to/from a nanos value.  datetimes/timestamps are always 64-bit in numpy, but the size of `long` is system-dependent.

Issue #1395, opened 2016-08-16
AttributeError: 'module' object has no attribute 'client'


yuanhuanglin:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment
* Operating System: Linux ubuntu-new-install 4.4.0-34-generic #53-Ubuntu SMP Wed Jul 27 16:06:39 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
* Python Version: Python 3.4.5
* Python Bitness:  64
* How did you install Zipline: pip
* Python packages: 

> backports-abc==0.4
> bcolz==0.12.1
> Bottleneck==1.1.0
> cachetools==1.1.6
> click==6.6
> contextlib2==0.5.4
> cycler==0.10.0
> cyordereddict==1.0.0
> Cython==0.24.1
> decorator==4.0.10
> entrypoints==0.2.2
> intervaltree==2.1.0
> ipykernel==4.4.1
> ipython==5.1.0
> ipython-genutils==0.1.0
> Jinja2==2.8
> jsonschema==2.5.1
> jupyter-client==4.3.0
> jupyter-core==4.1.1
> Logbook==1.0.0
> lxml==3.6.1
> MarkupSafe==0.23
> matplotlib==1.5.1
> mistune==0.7.3
> multipledispatch==0.4.8
> nbconvert==4.2.0
> nbformat==4.1.0
> networkx==1.11
> notebook==4.2.2
> numexpr==2.6.1
> numpy==1.11.1
> pandas==0.18.1
> pandas-datareader==0.2.1
> patsy==0.4.1
> pexpect==4.2.0
> pickleshare==0.7.4
> prompt-toolkit==1.0.6
> ptyprocess==0.5.1
> Pygments==2.1.3
> pyparsing==2.1.8
> python-dateutil==2.5.3
> pytz==2016.6.1
> pyzmq==15.4.0
> requests==2.11.0
> requests-file==1.4
> scipy==0.18.0
> simplegeneric==0.8.1
> six==1.10.0
> sortedcontainers==1.5.3
> SQLAlchemy==1.0.14
> statsmodels==0.6.1
> TA-Lib==0.4.9
> terminado==0.6
> toolz==0.8.0
> tornado==4.4.1
> traitlets==4.2.2
> tushare==0.5.0
> wcwidth==0.1.7
> zipline==1.0.1

Now that you know a little about me, let me tell you about the issue I am
having:

# Description of Issue
[...]

PR #1409, opened 2016-08-18
DOC: Replace run_algo.py in docs with zipline run

richafrank:
Per https://github.com/quantopian/zipline/issues/1395

### DIFF
--- a/docs/notebooks/tutorial.ipynb
+++ b/docs/notebooks/tutorial.ipynb
@@ -102,33 +102,39 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "usage: run_algo.py [-h] [-c FILE] [--algofile ALGOFILE] [--data-frequency {minute,daily}] [--start START] [--end END]\r\n",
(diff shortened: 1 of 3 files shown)


docs/notebooks/tutorial.ipynb
```diff
@@ -161,7 +167,7 @@
     }
    ],
    "source": [
-    "!run_algo.py -f ../../zipline/examples/buyapple.py --start 2000-1-1 --end 2014-1-1 -o buyapple_out.pickle"
+    "!zipline run -f ../../zipline/examples/buyapple.py --start 2000-1-1 --end 2014-1-1 -o buyapple_out.pickle"
```
llllllllll: should this be `%zipline`?

richafrank: Not sure - I think the intent is to shell out to run that file, which is the case I'm translating.  Do you think we should change this notebook to use the magic here instead?

llllllllll: I guess since we don;t want the results in memory this is fine, you can pass `-f` and `-o` to the magic though

PR #1420, opened 2016-08-22
MAINT: Add script to clean out compiled files.

ssanderson:


richafrank:
Should we update our docker- and vagrantfiles to use this, or not worth it?

ssanderson:
@richafrank do you mean you'd want to automatically call this at the start of the docker/vagrant setup?  I see us doing `find /vagrant/ -type f -name '*.c' -exec rm {} +` in the vagrant file, but it doesn't look like there's an analogous call for docker.  Do you think there should be?

At any rate, I don't feel super strongly about using this in more places either way.  The command is trivial enough that inlining it elsewhere doesn't seem super important.  If this grows to include more substantive logic, then I'd be more inclined to ensure consistency elsewhere.

richafrank:
I believe it's in Dockerfile-dev, but I'm ok with inlining it.

ssanderson:
I think the appveyor build here is spurious.  Re-running.

llllllllll:
is this good to merge?

ssanderson:
@llllllllll I just updated this with what I've been using most recently.  When downgrading from pandas 16 to pandas 18, I've found that I needed to clear the .so files and the .c files.  Not totally sure why, but it's easy enough to clear both.

llllllllll:
lgtm

PR #1421, opened 2016-08-22
ENH: Add new parameter to schedule_function that accepts a trading calendar

jbredeche:
Also, split up the events tests to explicitly test two different calendars.


zipline/utils/events.py
```diff
@@ -448,7 +448,7 @@ def __init__(self, n, invert):
 
     def should_trigger(self, dt):
         # is this market minute's period in the list of execution periods?
-        return self.cal.minute_to_session_label(dt, direction="none") in \
```
richafrank: @jbredeche Why did we change the behavior for these?

jbredeche: because it's possible, now, for `dt` not to be in `self.cal` (if the simulation is on the CME cal, but the event is on the NYSE cal, for example).  We don't want to blow up in that case.

richafrank: Gotcha

PR #1422, opened 2016-08-22
BUG: Add kwargs to ewma constructors

LotannaEzenwa:
When using the `from_x`-like constructors for EWMA, no mask can be passed to the EWMA constructor. Added a kwargs variable so that these don't break if/when the factor base class is updated


llllllllll:
this looks good, just need to fix the syntax error.

LotannaEzenwa:
@llllllllll Is this okay to merge?

LotannaEzenwa:
@llllllllll merging

PR #1423, opened 2016-08-22
Fix sharedoc

ssanderson:


zipline/utils/sharedoc.py
```diff
@@ -11,17 +11,24 @@
...
+def _gen_for_pad_lines_after_first(prefix, s):
+    lines = iter(s.splitlines())
+    yield next(lines)
+    for line in lines:
+        yield prefix + line
+
+
+def pad_lines_after_first(prefix, s):
+    """Apply a prefix to each line in s after the first."""
+    return '\n'.join(_gen_for_pad_lines_after_first(prefix, s))
```
llllllllll: is this just `('\n' + prefix).join(...)`?

ssanderson: It's `'\n'.join([lines[0], ('\n' + prefix).join(lines[1:]))`

ssanderson: (I only want to apply the prefix to lines 1-> N, because the first line is going already have the padding from the original format call.)

llllllllll: Ah, that makes sense. thanks

llllllllll: actually, are those not the same? the prefix isn't added to the first line in the join case.

ssanderson: Actually, I was wrong, `('\n' + prefix).join(...)` is what I wanted to begin with.  (What I wrote above doesn't apply the prefix to the second line.  Your expression correctly applies to everything except the first.

PR #1424, opened 2016-08-22
DOC: Updated more docs to remove --symbols, which was removed

richafrank:


llllllllll:
lgtm. I also grepped for any other usages and it is just the whatsnew for 0.7.0.

richafrank:
Yep, agreed, I did the same!

PR #1425, opened 2016-08-22
BUG: Fixes BcolzMinuteBarMetadata to read the version correctly

yankees714:
We were mistakenly using the `minute_per_day field`.

We now expose from the metadata object the version from which the metadata was read. This allows a new test that verifies the version is read correctly.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -260,7 +260,7 @@ class BcolzMinuteBarMetadata(object):
                 start_session,
                 end_session,
                 minutes_per_day,
-                version,
+                version=version,
             )
 
     def __init__(
@@ -270,7 +270,7 @@ class BcolzMinuteBarMetadata(object):
         start_session,
         end_session,
         minutes_per_day,
-        version,
+        version=FORMAT_VERSION,
     ):
         self.calendar = calendar
         self.start_session = start_session
@@ -458,7 +458,6 @@ class BcolzMinuteBarWriter(object):
             self._start_session,
             self._end_session,
             self._minutes_per_day,
-            BcolzMinuteBarMetadata.FORMAT_VERSION,
         )
         metadata.write(self._rootdir)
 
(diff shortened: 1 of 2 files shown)


zipline/data/minute_bars.py
```diff
@@ -223,7 +223,7 @@ def read(cls, rootdir):
             raw_data = json.load(fp)
 
             try:
-                version = raw_data['minutes_per_day']
+                version = raw_data['version']
```
richafrank: 👍

zipline/data/minute_bars.py
```diff
@@ -269,12 +270,14 @@ def __init__(
...
         end_session,
         minutes_per_day,
+        version,
     ):
         self.calendar = calendar
         self.start_session = start_session
         self.end_session = end_session
         self.ohlc_ratio = ohlc_ratio
         self.minutes_per_day = minutes_per_day
+        self.version = version
```
richafrank: Can you remind me the lifecycle here?  We instantiate a `BcolzMinuteBarMetadata`, which will have version from what was last written to disk.  When we write this instance to disk, it will have that previous version?

yankees714: So when we write, we'll always be using the most recent version, i.e. `FORMAT_VERSION` (https://github.com/quantopian/zipline/pull/1425/files#diff-ad0e2225f40bfedeadd927fb0d674d7bR461).

When reading, the version could potentially be older, so we look at the version to decide how to interpret everything else.

richafrank: Aha, I see - the writer specifies the latest.  That makes sense now, though it was surprising to me that the caller passes that attribute of `BcolzMinuteBarMetadata` to its own constructor - maybe a default value would make sense.

richafrank:
LGTM, just had one suggestion.

zipline/data/minute_bars.py
yankees714: I had considered that too, I'll go ahead and make that change.

PR #1428, opened 2016-08-23
MAINT: Modifies minute bars to use a dict of OHLC ratios

yankees714:
Instead of using a single ratio to scale up pricing for each sid before writing for bcolz, the writer now accepts a dict mapping each sid to the ratio to use. This arg is still optional, and we use the existing
`OHLC_RATIO` (1000) for every sid if no dict is passed.

This allows better handling of futures pricing data, where the required precision across root symbols is not consistent.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -24,6 +24,7 @@ from intervaltree import IntervalTree
 import numpy as np
 import pandas as pd
 from six import with_metaclass
+from toolz import keymap, valmap
 
 from zipline.data._minute_bar_internal import (
     minute_value,
@@ -208,7 +209,7 @@ class BcolzMinuteBarMetadata(object):
     minutes_per_day : int
         The number of minutes per each period.
     """
-    FORMAT_VERSION = 2
+    FORMAT_VERSION = 3
 
     METADATA_FILENAME = 'metadata.json'
 
@@ -229,7 +230,7 @@ class BcolzMinuteBarMetadata(object):
                 # if version does not match.
                 version = 0
 
-            ohlc_ratio = raw_data['ohlc_ratio']
+            default_ohlc_ratio = raw_data['ohlc_ratio']
 
             if version >= 1:
                 minutes_per_day = raw_data['minutes_per_day']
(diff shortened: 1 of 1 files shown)


zipline/data/minute_bars.py
```diff
@@ -265,7 +274,8 @@ def read(cls, rootdir):
 
     def __init__(
         self,
-        ohlc_ratio,
+        default_ohlc_ratio,
```
ehebert: Not 100% sure we need to put the prefix of `default` here, but not going to block on it.

zipline/data/minute_bars.py
```diff
@@ -465,6 +489,17 @@ def __init__(self,
     def first_trading_day(self):
         return self._start_session
 
+    def ohlc_ratio_for_sid(self, sid):
+        if self._ohlc_ratios_per_sid is not None:
+            try:
+                return self._ohlc_ratios_per_sid[sid]
+            except KeyError:
```
ehebert: If the `_default_ohlc_ratio` is being used as a fallback/default, should it be returned here?

zipline/data/minute_bars.py
```diff
@@ -856,6 +897,17 @@ def last_available_dt(self):
     def first_trading_day(self):
         return self._start_session
 
+    def _ohlcv_ratio_inverse_for_sid(self, sid):
+        if self._ohlc_inverses_per_sid is not None:
+            try:
+                return self._ohlc_inverses_per_sid[sid]
+            except KeyError:
+                pass
```
ehebert: Same as with `ohclv_ratio_for_sid`, should the `_default_ohlc_inverse` be returned here?

zipline/data/minute_bars.py
yankees714: I opted here to share the return below for both the case where the sid is not present in the dict, and when no dict was passed to begin with. Do you think separate returns would be better?

zipline/data/minute_bars.py
yankees714: Was trying to make the distinction from `ohlc_ratios_per_sid` clear in the naming, but if it's overkill I'd be happy to simplify it.

zipline/data/minute_bars.py
```diff
@@ -1111,11 +1164,15 @@ def load_raw_arrays(self, fields, start_dt, end_dt, sids):
...
                             excl_start - start_idx:excl_stop - start_idx + 1]
                         values = np.delete(values, excl_slice)
+
                 where = values != 0
                 # first slice down to len(where) because we might not have
                 # written data for all the minutes requested
-                out[:len(where), i][where] = values[where]
-            if field != 'volume':
-                out *= self._ohlc_inverse
+                if field != 'volume':
```
ehebert: Why the change on how the data is written to the index?
I lean towards the original, where the assignment is done in one place, to minimize drift between OHLC and Volume with regard to the write indices.

ehebert: I understand now, since the ratio is per sid, the assignment had to be changed to within the loop.

zipline/data/minute_bars.py
ehebert: Never mind my comment, the `pass` does fall through correctly. It did read slightly off to me at first, but I believe passing on the KeyError is the correct way to go about this.

PR #1429, opened 2016-08-23
Support loading futures data from bundles.

mannau:
Include future_minute_reader and future_daily_reader in DataPortal.


llllllllll:
Do you think we should pass futures loaders to the `ingest` functions so that users may create bundles with both futures and equities data?

mannau:
I already created a bundle with futures and equity data and needed a way to ingest it into quantopian. So - yes, I would highly appreciate that feature!
Cheers,
m

brian-from-quantrocket:
What's the status of this pull request? Like @mannau I've ingested futures data and found I needed to provide `future_minute_reader` and `future_daily_reader` to `DataPortal` to get the algorithm to run. Was about to submit an identical pull request and found this one already exists.

Not sure if the holdup is related to @llllllllll 's suggestion of passing futures loaders to the ingest function? That seems like a different issue which shouldn't hold up this one. Plus, ingesting futures isn't broken as far as I can tell (I was able to do it and apparently @mannau too); the problem is then running the algo because currently `DataPortal` only has readers for equities, not futures.

richafrank:
> Not sure if the holdup is related to @llllllllll 's suggestion of passing futures loaders to the ingest function?

I believe that's right, @brian-from-quantrocket . I think @llllllllll 's issue is with this implicitly mixing equities and futures data, but https://github.com/quantopian/zipline/pull/1980 (which is in progress this week) attempts to solve that design concern.

brian-from-quantrocket:
Fantastic, didn't see #1980. Better solution.

PR #1430, opened 2016-08-23
Ingest versions

richafrank:
`zipline ingest` can downgrade the assets db

This lets us publish an "old" db version for the most recent release of zipline, using the latest code base.


tests/data/bundles/test_core.py
```diff
@@ -253,6 +264,66 @@ def bundle_ingest(environ,
...
+    def test_ingest_assets_versions(self):
+        versions = (1, 2)
+
+        called = [False]
+
+        @self.register('bundle', create_writers=False)
+        def bundle_ingest_no_create_writers(*args, **kwargs):
+            called[0] = True
+
+        with self.assertRaises(ValueError):
```
llllllllll: wgy does the first ingest fail? Can we use `assert_raises_regex` to make this more clear?

zipline/utils/cache.py
```diff
@@ -290,11 +290,7 @@ def path(self):
     def _commit(self):
         """Sync the temporary file to the final path.
         """
-        self._tmpfile.close()
-        move(self._name, self._final_path)
-
-    def __getattr__(self, attr):
```
llllllllll: why did we remove the `__getattr__`?

zipline/utils/sqlite_utils.py
```diff
@@ -12,13 +12,23 @@
...
 SQLITE_MAX_VARIABLE_NUMBER = 998
 
 
 def group_into_chunks(items, chunk_size=SQLITE_MAX_VARIABLE_NUMBER):
     items = list(items)
     return [items[x:x+chunk_size]
             for x in range(0, len(items), chunk_size)]
+
+
+coerce_string_to_conn = coerce_string(sqlite3.connect)
```
llllllllll: This is awesome, thanks!

zipline/utils/cache.py
richafrank: It was unused, so I removed it.  This class itself was actually unused and had acquired some bitrot.

tests/data/bundles/test_core.py
richafrank: Will add.

PR #1432, opened 2016-08-24
ENH: Add a reader base which reindexes results.

ehebert:
Working towards history results which contain mixed asset types, add
a reader which makes `load_raw_arrays` return results indexed on the
session/minute ranges specified by the specified `trading_calendar`
instead of the calendar of the backing reader.

This reader will be used to make Equity readers align with Future
readers. It is intended for use as part of another reader (which will
dispatch queries based on asset type and then recombined results) which
will be passed to the `[Minute|Session]HistoryLoaders in the data portal.


tests/data/test_resample.py
```diff
@@ -527,3 +530,124 @@ def test_get_value(self):
...
+            OHLCV, m_open, m_close, [1, 2])
+
+        opens = DataFrame(data=result[0], index=outer_minutes,
+                          columns=[1, 2])
+        opens_with_price = opens.dropna()
+
+        self.assertEqual(
+            1440,
+            len(opens),
+            "The result should have 1440 bars, the number of minutes in a "
```
ssanderson: I :heart: these error messages.

jbredeche:
Nice, LGTM

PR #1437, opened 2016-08-27
PERF: Reduce calculations performed by empyrical

jdricklefs:
Principally by ...

1. Passing things already computed (e.g. `beta`) into methods that require them (e.g. `alpha`) so they don't re-compute them.
2. Risk-adjust the return stream only once so that empyrical doesn't have to do it separately in each method.

(Opening PR to validate Travis; we need to release a new empyrical to pypi after https://github.com/quantopian/empyrical/pull/27 is merged)

### DIFF
--- a/conda/empyrical/meta.yaml
+++ b/conda/empyrical/meta.yaml
@@ -1,27 +1,29 @@
 package:
   name: empyrical
-  version: "0.1.9"
+  version: "0.1.10"
 
 source:
-  fn: empyrical-0.1.9.tar.gz
-  url: https://pypi.python.org/packages/15/6b/de1d277d4342d2cecc8134f4935248853f32299cdd9b01728b0ec420c350/empyrical-0.1.9.tar.gz
-  md5: 2c8b928cae192fc9cb8b7104608eec56
+  fn: empyrical-0.1.10.tar.gz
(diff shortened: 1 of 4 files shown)


etc/requirements_blaze.txt
```diff
@@ -1,3 +1,5 @@
 -e git://github.com/quantopian/datashape.git@bf06a41dc0908baf7c324aeacadba8820468ee78#egg=datashape-dev
 -e git://github.com/quantopian/odo.git@9e16310b5f2c3f05162145200db7e7908f0a866e#egg=odo-dev
 -e git://github.com/quantopian/blaze.git@7b8f70c15279c7ebacf2ac2f5f74acf08b300d71#egg=blaze-dev
+
+-e git://github.com/quantopian/empyrical.git@6c8d59c0af66bb1243277bdb2822e34480395bb0#egg=empyrical-master
```
ssanderson: Why is this in requirements blaze?  Also, I don't think we can make this a committish requirement.  There's no way to make `pip` install a requirement from github when you just `pip install blaze`.  We do this with blaze because blaze is an optional dependency that you have to manually install yourself if you want to use.

jdricklefs: It's a temp commit so I could get travis to run. It's not going to live here.

See comment on 0980bfaf98e2a52a6de351d222891c9b9314e2c3

ssanderson: :+1:

PR #1442, opened 2016-08-29
ENH: Let event offsets be up to 12 hours.

jbredeche:
We used to cap at 6.5 hours, but that's not enough for longer trading sessions like the CME.

### DIFF
--- a/tests/events/test_events_cme.py
+++ b/tests/events/test_events_cme.py
@@ -15,7 +15,9 @@
 from unittest import TestCase
 import pandas as pd
 
-from test_events import StatefulRulesTests, StatelessRulesTests
+from test_events import StatefulRulesTests, StatelessRulesTests, \
+    minutes_for_days
+from zipline.utils.events import AfterOpen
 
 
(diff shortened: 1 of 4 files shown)


tests/events/test_events_cme.py
```diff
@@ -24,6 +26,18 @@ class TestStatelessRulesCME(StatelessRulesTests, TestCase):
     HALF_SESSION = pd.Timestamp("2014-07-04", tz='UTC')
     FULL_SESSION = pd.Timestamp("2014-09-24", tz='UTC')
 
+    def test_far_after_open(self):
+        minute_groups = minutes_for_days(self.cal, ordered_days=True)
```
ehebert: Does this mean we are testing every day in the calendar?

ehebert:
LGTM implementation wise; however had a question about the use of `minutes_for_days` in tests, with an eye on test duration.

tests/events/test_events_cme.py
jbredeche: yes.

PR #1444, opened 2016-08-29
BUG: Use session label instead of date for 1d.

ehebert:
`1d` history calls were failing on key errors when using the
`us_futures` calendar, because of timestamps occuring before a midnight
would present the wrong midnight (i.e. the midnight before the session,
instead of the following midnight, which is the label for the current
session.)

Tests will follow when bringing up coverage on resample and data portal
modules.


jbredeche:
lgtm

PR #1449, opened 2016-08-30
MAINT: remove __getitem__ as alias of __getattr__

llllllllll:


ssanderson:
LGTM

ssanderson:
@llllllllll anything still blocking merge on this?

llllllllll:
yeah, still waiting on downstream

zipline/protocol.py
```diff
@@ -88,8 +80,52 @@ def to_series(self, index=None):
...
+    attrs : iterable[str]
+        The set of allowed attributes.
+
+    Returns
+    -------
+    __getitem__ : callable[any, str]
+        The ``__getitem__`` method to put in the class dict.
+    """
+    attrs = frozenset(attrs)
+    msg = "'{0}[attr]' is deprecated, please use '{0}.attr' instead".format(
```
ssanderson: This might be a nicer error actually formatting in the string itself (with repr on the left and str on the right)

llllllllll: I think the repr will make it really hard to read since it just prints the dict

zipline/protocol.py
```diff
@@ -105,12 +141,26 @@ def __init__(self):
...
-        return self.__dict__[key]
-
     def __repr__(self):
         return "Portfolio({0})".format(self.__dict__)
 
+    # If you are adding new attributes, don't update this set. This method
+    # is deprecated to normal attribute access so we don't want to encourage
+    # new usages.
+    __getitem__ = _deprecated_getitem_method(
+        'portfolio', {
```
ssanderson: Any reason not to make this and the one above a `frozenset`?

ssanderson:
Two small comments.  Otherwise still LGTM.

zipline/protocol.py
llllllllll: I freeze it in the function.

zipline/protocol.py
ssanderson: I meant that we should include the actual attr that was requested (deferring the creation of the error string until `__getitem__` was called) so that the user can more easily find where they need to update their code.

llllllllll: Oh, I see. I can update that

PR #1451, opened 2016-08-30
ENH: Avoid unnecessary work with missing data.

jbredeche:


zipline/data/minute_bars.py
```diff
@@ -891,6 +891,14 @@ def __init__(self, rootdir, sid_cache_size=1000):
         self._last_get_value_dt_position = None
         self._last_get_value_dt_value = None
 
+        # This is to avoid any bad data or other performance-killing situation
+        # where there a consecutive streak of 0 (no volume) starting at an
+        # asset's start date.
+        # if asset 1 started on 2015-01-03 but its first trade is 2015-01-06
+        # 10:31 AM US/Eastern, this dict would store {1: 1420558260000000000},
+        # which is the nanosecond epoch of that date.
+        self._known_zero_volume_dict = {}
```
ehebert: I would recommend storing these as the minute since epoch, since that is what the frame of reference of the other values you are comparing against is.
This would avoid an extra `/ NANOS_IN_MINUTE` call every time it is used.

zipline/data/minute_bars.py
```diff
@@ -1078,21 +1086,42 @@ def get_last_traded_dt(self, asset, dt):
 
     def _find_last_traded_position(self, asset, dt):
         volumes = self._open_minute_file('volume', asset)
-        start_date_minutes = asset.start_date.value / NANOS_IN_MINUTE
-        dt_minutes = dt.value / NANOS_IN_MINUTE
+        start_date_minute = asset.start_date.value / NANOS_IN_MINUTE
+        dt_minute = dt.value / NANOS_IN_MINUTE
+
+        if asset.sid in self._known_zero_volume_dict:
```
ehebert: Using `try`/`KeyError` is preferable. Avoids an extra hash and lookup into dictionary.

e.g.

```
try:
    earliest_dt_to_search = self._known_zero_volume_dict[asset.sid] / NANOS_IN_MINUTE
except KeyError:
    earliest_dt_to_search = start_date_minute
```

zipline/data/minute_bars.py
```diff
@@ -1078,21 +1086,42 @@ def get_last_traded_dt(self, asset, dt):
...
+            dt_minute,
+            earliest_dt_to_search,
             volumes,
             self._minutes_per_day,
         )
 
+        if pos == -1:
+            # if we didn't find any volume before this dt, save it to avoid
+            # work in the future.
+            if asset.sid not in self._known_zero_volume_dict:
```
ehebert:
```
try:
    self._known_zero_volume_dict[asset.sid] = max(
        dt.value,
        self._known_zero_volume_dict[asset.sid]
   )
except KeyError:
    self._known_zero_volume_dict[asset.sid] = dt.value
```

PR #1453, opened 2016-08-30
ENH: Allow to fill orders at limit price or configurable price field

volconst:
Generalize SlippageModel by adding properties market_field and
limits_enhance and subclasses that extend it and configure the
properties.
Thus the slippage model can be configured whether to execute at open
or close price and whether to trigger limit and stop orders early at
limit, stop price (if they fall within [low..high] range) or at the
configured market_price.
Default settings are backward compatible with respect to execution
price (at least in the tests). To get the new behavior set the
slippage model to a subclass.
E.g.:
def initialize(context):
	context.set_slippage(ConservativeCloseSlippage())
Once configured, the slippage model either always executes limit
orders at limit price or always at market_price. It is not designed
to use limit price and market_price for different orders in the same
simulation.
A related PR #1376 contains some discussion about limit prices.

### DIFF
--- a/tests/finance/test_slippage.py
+++ b/tests/finance/test_slippage.py
@@ -24,13 +24,15 @@ from nose_parameterized import parameterized
 
 import pandas as pd
 from pandas.tslib import normalize_date
+from numpy import arange
 
-from zipline.finance.slippage import VolumeShareSlippage
+from zipline.finance.slippage import VolumeShareSlippage,\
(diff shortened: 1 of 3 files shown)


volconst:
I am closing this PR, because ConservativeCloseSlippage has a bug with stop orders and the open price slippage models does not need enhance_limits, but should switch dynamically depending on marketability. I may create another PR if I implement new ideas.

PR #1455, opened 2016-08-31
ENH: Put a cache in front of future chain lookups

jbredeche:
Cache the last 100 rootsymbol/session pairs, since future chains never
change inside a session.


ehebert:
LGTM

PR #1462, opened 2016-09-02
Symbol lookup raises

kglowinski:
- Correctly raises SymbolNotFound in fuzzy symbol lookup.
- Adds checking against NoneType (useful in case someone vectorizes across a series, one of which is a None)
- If asset file has two symbols which map to the same sid (SYM_B, and SYMB), should not raise a MultipleSymbolError, should just return the correct equity object for the sid.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -720,6 +720,43 @@ class AssetFinderTestCase(WithTradingCalendars, ZiplineTestCase):
                 self.assertEqual(results, expected)
                 self.assertEqual(missing, [])
 
+    def test_lookup_none_raises(self):
+        """
+        If lookup_symbol is vectorized across multiple symbols, and one of them
+        is None, want to raise a TypeError.
+        """
+
+        with self.assertRaises(TypeError):
+            self.asset_finder.lookup_symbol(None, pd.Timestamp('2013-01-01'))
+
+    def test_lookup_mult_are_one(self):
+        """
+        Ensure that multiple symbols that return the same sid are collapsed to
+        a single returned asset.
+        """
+
+        date = pd.Timestamp('2013-01-01', tz='UTC')
+
+        df = pd.DataFrame.from_records(
+            [
+                {
+                    'sid': 1,
+                    'symbol': symbol,
+                    'start_date': date.value,
+                    'end_date': (date + timedelta(days=30)).value,
+                    'exchange': 'NYSE',
+                }
+                for symbol in ('FOOB', 'FOO_B')
+            ]
+        )
+        self.write_assets(equities=df)
+        finder = self.asset_finder
+
+        # If we are able to resolve this with any result, means that we did not
+        # raise a MultipleSymbolError.
+        result = finder.lookup_symbol('FOO/B', date + timedelta(1), fuzzy=True)
+        self.assertEqual(result.sid, 1)
+
     def test_lookup_generic_handle_missing(self):
         data = pd.DataFrame.from_records(
             [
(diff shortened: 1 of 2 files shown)


zipline/assets/assets.py
```diff
@@ -648,9 +648,10 @@ def _lookup_symbol_fuzzy(self, symbol, as_of_date):
 
         if not options:
             # no equity owned the fuzzy symbol on the date requested
-            SymbolNotFound(symbol=symbol)
+            raise SymbolNotFound(symbol=symbol)
+
 
-        if len(options) == 1:
+        if len(options) == 1 or len(set(map(itemgetter(0), options))) == 1:
```
richafrank: What do you think about `len({opt[0] for opt in options}) == 1` for pythonic clarity?

richafrank: I notice that we're maintaining these option pairs in a list, but then here and potentially later on, we dedup them on sid.  Would it make sense to use an `OrderedDict` instead of the list, or is that much slower in most cases?

kglowinski: Why `OrderedDict` instead of a normal `dict`?

richafrank: I assumed the order mattered, but maybe not?

zipline/assets/assets.py
```diff
@@ -703,6 +704,10 @@ def lookup_symbol(self, symbol, as_of_date, fuzzy=False):
             there are multiple candidates for the given ``symbol`` on the
             ``as_of_date``.
         """
+        if symbol is None:
+            raise TypeError("Cannot lookup symbol of type NoneType for "
+                            "as of date %s" % as_of_date)
```
richafrank: Why do we include `as_of_date` here?  What about `fuzzy`?  Or could we just say "Cannot lookup None symbol"?

zipline/assets/assets.py
```diff
@@ -648,9 +648,10 @@ def _lookup_symbol_fuzzy(self, symbol, as_of_date):
 
         if not options:
             # no equity owned the fuzzy symbol on the date requested
-            SymbolNotFound(symbol=symbol)
+            raise SymbolNotFound(symbol=symbol)
```
richafrank: 👍

tests/test_assets.py
```diff
@@ -720,6 +720,43 @@ def test_lookup_generic(self):
...
+        with self.assertRaises(TypeError):
+            self.asset_finder.lookup_symbol(None, pd.Timestamp('2013-01-01'))
+
+    def test_lookup_mult_are_one(self):
+        """
+        Ensure that multiple symbols that return the same sid are collapsed to
+        a single returned asset.
+        """
+
+        date = pd.Timestamp('2013-01-01', tz='UTC')
```
richafrank: Are these dates important, or would something like `WithDefaultDateBounds.START_DATE, END_DATE` be helpful for consistency?

zipline/assets/assets.py
kglowinski: I guess my thinking was that fuzzy or not would have been passed to the method, and that would be something known by the user from the get-go. The symbol is unhelpful to pass back, since it's None, but that a user might want to identify _where_ the symbol was None in their data, the only reasonable identifier for which was the as_of_date.

richafrank:
Nice additional test coverage!

zipline/assets/assets.py
richafrank: I can buy that.  What do you think about shortening it to mention just "None" instead of "NoneType"?  I find that clearer, but maybe everyone is used to "AttributeError: Instance of NoneType has no attribute..." ?

richafrank:
LGTM!

PR #1463, opened 2016-09-02
BUG: Do not adjust returns for sharpe and sortino

ahgnaw:
Remove variable `risk_adj_returns`. Neither sharpe nor sortino require risk adjusted returns. Information ratio can create adjusted returns as necessary.


twiecki:
This looks right to me.

PR #1466, opened 2016-09-02
ENH: Dont allow length=1 regressions/correlations.

ssanderson:
They're not meaningful, and they cause warnings from numpy.

Implemented in terms of a new preprocessor, `expect_bounded`, which
takes a tuple of `upper_bound` and `lower_bound`.

### DIFF
--- a/zipline/utils/input_validation.py
+++ b/zipline/utils/input_validation.py
@@ -503,6 +503,97 @@ def expect_element(*_pos, **named):
     return preprocess(**valmap(_expect_element, named))
 
 
+def expect_bounded(**named):
+    """
+    Preprocessing decorator that verifies inputs fall between upper and lower
+    bounds.
+
+    Usage
+    -----
+    >>> @expect_bounded(x=(1, 5))
+    ... def foo(x):
+    ...    return x + 1
+    ...
+    >>> foo(1)
+    2
+    >>> foo(5)
+    6
+    >>> foo(6)  # doctest: +NORMALIZE_WHITESPACE +ELLIPSIS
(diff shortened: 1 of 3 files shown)


zipline/utils/input_validation.py
```diff
@@ -503,6 +503,97 @@ def _expect_element(collection):
     return preprocess(**valmap(_expect_element, named))
 
 
+def expect_bounded(**named):
+    """
+    Preprocessing decorator that verifies inputs fall between upper and lower
+    bounds.
```
richafrank: Could you add a comment here about the bounds being inclusive or exclusive?

zipline/utils/input_validation.py
```diff
@@ -503,6 +503,97 @@ def _expect_element(collection):
...
+                " argument '{name}', but got {bounds} instead.".format(
+                    name=name,
+                    bounds=bounds,
+                )
+            )
+
+    def _expect_bounded(bounds):
+        (lower, upper) = bounds
+        if lower is None:
+            should_fail = lambda value: value > upper
```
richafrank: 👍 for the name `should_fail` - much clearer than `pred`.

jdricklefs:
Wow, this is a fancy fix. Thanks @ssanderson

zipline/utils/input_validation.py
ssanderson: Updated.

richafrank:
LGTM

dmichalowicz:
Thanks for this Scott! Good catch on the length-1 tests

PR #1467, opened 2016-09-02
Check param string types

richafrank:
BUG: Change str to string_types to avoid errors

When in python2.7, and unicode_literals is imported
type check will raise error because 'type' is not str but unicode

Fixes up https://github.com/quantopian/zipline/pull/1315


llllllllll:
lgtm

PR #1471, opened 2016-09-05
PERF: Remove import-time calendar creations.

ssanderson:
Remove module scope invocations of `get_calendar('NYSE')`, which cuts
zipline import time in half on my machine. This make the zipline CLI
noticeably more responsive, and it reduces memory consumed at import
time from 130MB to 90MB.

All of the module-scope `get_calendar` invocations were in the bundle registration code.  There are two major changes required to make this work:

1. Added a `register_calendar_alias` method and reworked the way calendars are stored internally in `TradingCalendarDispatcher`.  There's now always one canonical name under which a calendar is stored, but possibly many "aliases" to that calendar.  This is now used for the QUANDL and YAHOO loaders to alias NYSE.

2. Removed the creation of a calendar during bundle registration.  The calendar associated with a bundle is now always registered by name.

Before:

$ time python -c 'import zipline'

real    0m1.262s
user    0m1.128s
sys     0m0.120s

After:

$ time python -c 'import zipline'

real    0m0.676s
user    0m0.536s
sys     0m0.132s

cc @jbredeche on the calendar changes and @llllllllll on the bundle setup changes.

### DIFF
--- a/tests/calendars/test_calendar_dispatcher.py
+++ b/tests/calendars/test_calendar_dispatcher.py
@@ -1,0 +1,95 @@
+"""
+Tests for TradingCalendarDispatcher.
+"""
+from zipline.errors import (
+    CalendarNameCollision,
+    CyclicCalendarAlias,
+    InvalidCalendarName,
+)
+from zipline.testing import ZiplineTestCase
+from zipline.utils.calendars.calendar_utils import TradingCalendarDispatcher
+from zipline.utils.calendars.exchange_calendar_ice import ICEExchangeCalendar
+
+
+class CalendarAliasTestCase(ZiplineTestCase):
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(CalendarAliasTestCase, cls).init_class_fixtures()
+        # Make a calendar once so that we don't spend time in every test
+        # instantiating calendars.
+        cls.dispatcher_kwargs = dict(
+            calendars={'ICE': ICEExchangeCalendar()},
+            calendar_factories={},
+            aliases={
+                'ICE_ALIAS': 'ICE',
+                'ICE_ALIAS_ALIAS': 'ICE_ALIAS',
+            },
+        )
+
+    def init_instance_fixtures(self):
+        super(CalendarAliasTestCase, self).init_instance_fixtures()
+        self.dispatcher = TradingCalendarDispatcher(
+            # Make copies here so that tests that mutate the dispatcher dicts
+            # are isolated from one another.
+            **{k: v.copy() for k, v in self.dispatcher_kwargs.items()}
+        )
+
+    def test_follow_alias_chain(self):
+        self.assertIs(
+            self.dispatcher.get_calendar('ICE'),
+            self.dispatcher.get_calendar('ICE_ALIAS'),
+        )
+        self.assertIs(
+            self.dispatcher.get_calendar('ICE'),
+            self.dispatcher.get_calendar('ICE_ALIAS_ALIAS'),
+        )
+
+    def test_add_new_aliases(self):
+        with self.assertRaises(InvalidCalendarName):
+            self.dispatcher.get_calendar('NOT_ICE')
+
+        self.dispatcher.register_calendar_alias('NOT_ICE', 'ICE')
+
+        self.assertIs(
+            self.dispatcher.get_calendar('ICE'),
+            self.dispatcher.get_calendar('NOT_ICE'),
+        )
+
+        self.dispatcher.register_calendar_alias(
+            'ICE_ALIAS_ALIAS_ALIAS',
+            'ICE_ALIAS_ALIAS'
+        )
+        self.assertIs(
+            self.dispatcher.get_calendar('ICE'),
+            self.dispatcher.get_calendar('ICE_ALIAS_ALIAS_ALIAS'),
+        )
+
+    def test_remove_aliases(self):
+        self.dispatcher.deregister_calendar('ICE_ALIAS_ALIAS')
+        with self.assertRaises(InvalidCalendarName):
+            self.dispatcher.get_calendar('ICE_ALIAS_ALIAS')
+
+    def test_reject_alias_that_already_exists(self):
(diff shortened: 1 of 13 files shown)


tests/data/bundles/test_core.py
```diff
@@ -369,7 +369,7 @@ def _empty_ingest(self, _wrote_to=[]):
         """
         if not self.bundles:
             @self.register('bundle',
-                           calendar=get_calendar('NYSE'),
+                           calendar_name=('NYSE'),
```
llllllllll: we can drop the parens

zipline/__init__.py
```diff
@@ -38,26 +38,42 @@ def cleanup_tempdir():
...
 from .utils.run_algo import run_algorithm
 from ._version import get_versions
+__version__ = get_versions()['version']
+del get_versions
+
 # These need to happen after the other imports.
 from . algorithm import TradingAlgorithm
 from . import api
 
-__version__ = get_versions()['version']
```
llllllllll: why did this move before the other imports

zipline/utils/calendars/calendar_utils.py
```diff
@@ -54,10 +42,19 @@ class TradingCalendarDispatcher(object):
 
     Methods of a global instance of this class are provided by
     zipline.utils.calendar_utils.
+
+    Parameters
+    ----------
+    calendars : dict[str -> TradingCalendar]
+        Initial set of calendars.
+    calendar_factories : dict[str -> function]
+        Factories for lazy
```
llllllllll: This seems cut off

zipline/utils/calendars/calendar_utils.py
```diff
@@ -127,21 +141,91 @@ def register_calendar_type(self, name, calendar_type, force=False):
...
+            The name of the requested calendar.
+
+        Returns
+        -------
+        canonical_name : str
+            The real name of the calendar to create/return.
+        """
+        # Use an OrderedDict as an ordered set so that we can return the order
+        # of aliases in the event of a cycle.
+        seen = OrderedDict({name: None})
```
llllllllll: this initialization will be overridden in the first pass line of the loop

zipline/utils/calendars/calendar_utils.py
```diff
@@ -127,21 +141,91 @@ def register_calendar_type(self, name, calendar_type, force=False):
...
+
+        while name in self._aliases:
+            seen[name] = None
+            name = self._aliases[name]
+
+            if name in seen:
+                cycle = seen.keys()
+                cycle.append(name)
+                raise CyclicCalendarAlias(
+                    cycle=" -> ".join([repr(k) for k in cycle])
```
llllllllll: you can drop the brackets

zipline/utils/calendars/calendar_utils.py
llllllllll: Are you just using this as an ordered set? Could we just use a list here, I can't imagine this growing very large

zipline/utils/calendars/calendar_utils.py
```diff
@@ -127,21 +141,91 @@ def register_calendar_type(self, name, calendar_type, force=False):
...
+        Returns
+        -------
+        canonical_name : str
+            The real name of the calendar to create/return.
+        """
+        # Use an OrderedDict as an ordered set so that we can return the order
+        # of aliases in the event of a cycle.
+        seen = OrderedDict({name: None})
+
+        while name in self._aliases:
```
llllllllll: This creates a really nice error, thanks

zipline/data/bundles/core.py
```diff
@@ -612,5 +620,5 @@ def should_clean(name):
 
 bundles, register, unregister, ingest, load, clean = _make_bundle_core()
 
-register_calendar("YAHOO", get_calendar("NYSE"))
-register_calendar("QUANDL", get_calendar("NYSE"))
+register_calendar_alias("YAHOO", "NYSE")
```
llllllllll: can we move these to the yahoo and quandl modules?

zipline/__init__.py
```diff
@@ -38,26 +38,42 @@ def cleanup_tempdir():
...
 
 def load_ipython_extension(ipython):
     from .__main__ import zipline_magic
     ipython.register_magic_function(zipline_magic, 'line_cell', 'zipline')
 
 
+# PERF: Fire a warning if calendars were instantiated during zipline import.
+# Having calendars doesn't break anything per-se, but it makes zipline imports
+# noticeably slower, which becomes particularly noticeable in the Zipline CLI.
+from zipline.utils.calendars.calendar_utils import global_calendar_dispatcher
```
llllllllll: +1 glad we will catch this early in the future

tests/data/bundles/test_yahoo.py
```diff
@@ -157,7 +157,7 @@ def adjustments_callback(request):
         self.register(
             'bundle',
             yahoo_equities(self.symbols),
-            calendar=self.calendar,
+            calendar_name='NYSE',
```
llllllllll: should we use `YAHOO`?

tests/calendars/test_calendar_dispatcher.py
```diff
@@ -0,0 +1,95 @@
...
+        super(CalendarAliasTestCase, self).init_instance_fixtures()
+        self.dispatcher = TradingCalendarDispatcher(
+            # Make copies here so that tests that mutate the dispatcher dicts
+            # are isolated from one another.
+            **{k: v.copy() for k, v in self.dispatcher_kwargs.items()}
+        )
+
+    def test_follow_alias_chain(self):
+        self.assertIs(
+            self.dispatcher.get_calendar('ICE'),
```
llllllllll: assertions are normally flipped like: `result, expected`

tests/data/bundles/test_yahoo.py
ssanderson: The vanilla yahoo equities loader still uses NYSE, so I think this is fine.

PR #1474, opened 2016-09-07
ENH: Don't forward fill non-ffillable fields during non-market-hours

jbredeche:
When `data.current` is called on an asset for a minute that is not in the asset's exchange's calendar, we were using the last market minute's value for every field (instead of just for `price`).

Now, OHLC return `np.nan`, volume returns `0`, and price gets you the last close.

### DIFF
--- a/tests/data/test_minute_bars.py
+++ b/tests/data/test_minute_bars.py
@@ -43,6 +43,7 @@ from zipline.data.minute_bars import (
     US_EQUITIES_MINUTES_PER_DAY,
     BcolzMinuteWriterColumnMismatch
 )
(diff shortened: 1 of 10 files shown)


ehebert:
The test here is good; however my recomendation is to change the behavior of the minute bar reader  instead of adding an extra check at the data portal layer.

The `BcolzMinuteBarReader` should be either raising a named exception, or returning a 0/`nan` value when the dt is not in the calendar. Our current behavior is that the if the minute is between market close and market open then the minute bar reader uses the last market close, which is why we appear to be forward filling when using a larger calendar.

PR #1476, opened 2016-09-08
DOC: Update testpypi install command

ehebert:
When following the release guide, installing from testpypi using the
`-i` flag failed on my, and at least one other's, development machines.
The cause of the failure appears to be that pip would look for packages,
such as `LogBook` or `pandas` on `testpypi`. However many dependencies
do not have versions that meet our version criteria. (e.g. pandas does
not have a version between 0.16.0 and 0.18.0 on testpypi.)

Instead, use `--extra-index-url` so that other packages can use `pypi`
as a fallback server, instead of being limited to `testpypi`.

### DIFF
--- a/docs/source/release-process.rst
+++ b/docs/source/release-process.rst
@@ -154,7 +154,7 @@ create a new virtualenv, ``cd`` into a clean directory and then run:
 
 .. code-block:: bash
 
-   $ pip install -i https://testpypi.python.org/pypi zipline
+   $ pip install --extra-index-url https://testpypi.python.org/pypi zipline
    $ python -c 'import zipline;print(zipline.__version__)'
 
 


richafrank:
Makes sense to me.

llllllllll:
The command should use testpypi as the main index and have normal pypi as the fallback index, otherwise installing zipline will come from real pypi and that will not test the new upload.

ehebert:
```
$ pip install --extra-index-url https://testpypi.python.org/pypi zipline
Collecting zipline
  Downloading https://testpypi.python.org/packages/65/79/c2974d2ed3ca20f45d2f28ca46155c2d01ccdf01c0d44b9823eed19bade5/zipline-1.0.2.2.tar.gz (392kB)
    100% |████████████████████████████████| 399kB 2.7MB/s
```

PR #1478, opened 2016-09-08
DOC: Updated example notebook for latest zipline cell magic

richafrank:


llllllllll:
do we need to pass `-o perf`? That is just persisting the results to a pickle file, but we have the results in memory as the result of the cell so idk if that is useful here. Related to that: should we be using `analyze`? I feel like the more natural usage would be to just do the post-run analysis in the next cell.

richafrank:
Sounds good - I removed `-o perf`.  

I don't have a strong opinion about `analyze` - it is part of the API, so I'll leave it as an example.

Issue #1374, opened 2016-08-04
run_algorithm with no data source should default to `bundle='quantopian-quandl'`


ssanderson:
We have code that's supposed to be doing this, but it's currently broken:

```
    non_none_data = valfilter(bool, {
        'data': data is not None,
        'bundle': bundle is not None,
    })
    if not non_none_data:
        # if neither data nor bundle are passed use 'quantopian-quandl'
        bundle = 'quantopian-quandl'

    if len(non_none_data) != 1:
        raise ValueError(
            'must specify one of `data`, `data_portal`, or `bundle`,'
            ' got: %r' % non_none_data,
        )
```

If hit the `if not non_none_data` check, we still raise on the next branch.

richafrank:
Should be fixed by #1479

PR #1479, opened 2016-09-11
BUG: run_algorithm with no data source should default

richafrank:
to 'quantopian-quandl' bundle

For https://github.com/quantopian/zipline/issues/1374

I considered adding a unit test.  Because of its dependence on the global `load` of a bundle, it would depend on ingesting the real quantopian-quandl bundle first, or we'd need to register a replacement bundle globally for the span of the test.  So I haven't included a unit test here.


llllllllll:
lgtm, do you think there is an easy test for this?

richafrank:
Not easy, at least. 😄

PR #1483, opened 2016-09-13
MAINT: Add additional fields to __getitem__ for Order

jdricklefs:
CC: @llllllllll 

These were previously available like the other deprecated attributes here. Not sure where the discrepancy from the list you had here vs. the list in the API docs came from - might be something else we need to update?


llllllllll:
+1, thanks

PR #1484, opened 2016-09-14
PERF: Release unneeded pipeline terms.

ssanderson:
Refcount pipeline terms during execution and release terms once they're
no longer needed.

This dramatically reduces memory usage on large pipelines.

### DIFF
--- a/zipline/pipeline/graph.py
+++ b/zipline/pipeline/graph.py
@@ -119,9 +119,14 @@ class TermGraph(DiGraph):
     def _repr_png_(self):
         return self.png.data
 
-    def initial_refcounts(self, initial_workspace):
+    def initial_refcounts(self, initial_terms):
         """
         Calculate initial refcounts for execution of this graph.
+
+        Parameters
+        ----------
+        initial_terms : iterable[Term]
+            An iterable of terms that were pre-computed before graph execution.
 
         Each node starts with a refcount equal to its outdegree, and output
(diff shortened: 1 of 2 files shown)


twiecki:
:+1: :100:

zipline/pipeline/graph.py
```diff
@@ -130,8 +130,38 @@ def initial_refcounts(self):
         refcounts = self.out_degree()
         for t in self.outputs.values():
             refcounts[t] += 1
+
+        for t in initial_workspace:
+            self.decref_dependencies(t, refcounts)
+
         return refcounts
 
+    def decref_dependencies(self, term, refcounts):
```
llllllllll: shouldn't decref release the data? It seems like this makes it easier to get the refcounts and workspace out of sync.

ssanderson: The data might not be in the actual workspace if we put `VWAP` in the initial_workspace without `close` and `volume`.

The design here, essentially, is that dependency resolution is the TermGraph's job, and computation is the PipelineEngine's job.  It's the engine's responsibility to actually do something with the terms that the graph identifies as garbage.  This is analogous to the fact that it's the Engine's job to actually compute in the order that the graph determines in `TermGraph.ordered`.

twiecki:
This is awesome. Can you let me know when this hits staging/prod?

PR #1487, opened 2016-09-14
Create in-memory restricted list

lianga888:
- Create a restricted list manager that takes in information about restricted sids and stores in memory upon instantiation
- Create a restrictions controller that aggregates and provides restrictions information from multiple restricted list managers
- Make this restrictions controller available to `BarData` so that `can_trade` can take into account these restrictions
- Register each restricted list manager as a trading control using `set_do_not_order_list` so that we may log/fail when we try to order a restricted sid


lianga888:
Some performance stats

```
The following are all 1-year backtest that calls can_trade for each 10 sids every bar:

No restrictions
88863552 function calls (88799595 primitive calls) in 138.518 seconds
983100    1.060    0.000    1.821    0.000 rl_manager.py:32(is_restricted)

Empty list of restrictions
89848085 function calls (89783898 primitive calls) in 151.495 seconds
983100    1.307    0.000    6.101    0.000 rl_manager.py:32(is_restricted)

Every sid is restricted for every minute
22692416 function calls (22627694 primitive calls) in 28.267 seconds (shorter because can_trade can just return False after checking the restricted list without checking the rest of the logic)
983100    0.880    0.000   13.295    0.000 rl_manager.py:32(is_restricted)

Every sid is restricted for every minute (don't return after checking restricted list)
97704933 function calls (97640211 primitive calls) in 162.710 seconds
983100    1.324    0.000   20.534    0.000 rl_manager.py:32(is_restricted)

50 restrictions, 10 sids are restricted for every minute
22739981 function calls (22672901 primitive calls) in 28.331 seconds
983100    0.883    0.000   13.258    0.000 rl_manager.py:32(is_restricted)

The following is 1-year backtest that calls can_trade for each 50 sids every bar:

Every sid is restricted for every minute
51301697 function calls (51235132 primitive calls) in 58.730 seconds
2850990    2.318    0.000   37.002    0.000 rl_manager.py:32(is_restricted)
```

ssanderson:
@lianga888 for the stats above, are the `can_trade` calls in a loop, or are we doing `can_trade([list, of assets])`?

ssanderson:
How do the timings above compare to an algo calling `can_trade` a comparable number of times on master?

lianga888:
@ssanderson they are in a loop. Is it more meaningful to call them all at once?

ssanderson:
> @ssanderson they are in a loop. Is it more meaningful to call them all at once?

If they're in a loop, then you're adding a lot of function call overhead to the timings.  In general, I would hope that we can process `can_trade([list, of, stocks])` faster than `for stock in list: can_trade(stock)`, because we have the potential to do work more in batches in the latter case.

lianga888:
Did a run of the no-restrictions case with `can_trade` called for all stocks at once, and it seems to be slower than the one called in loop?

lianga888:
In any case, it seems like these performance numbers are good enough?

zipline/rl_manager.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self, **kwargs):
+        pass
+
+    @abc.abstractmethod
+    def restrictions(self, sid, dt):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def is_restricted(self, sid, dt):
```
ssanderson: This should probably take an iterable of sids to allow for a future vectorized implementation.

zipline/rl_manager.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    ABC for a restricted list manager that returns information about
+    restrictions from a single source
+    """
+
+    def __init__(self, **kwargs):
+        pass
+
+    @abc.abstractmethod
+    def restrictions(self, sid, dt):
```
ssanderson: Noted below, but I don't think we need this in the interface.

zipline/rl_manager.py
```diff
@@ -0,0 +1,111 @@
...
+
+class StaticRestrictedList(RLManager):
+    """
+    Tracks a static list of sids that are restricted regardless of dt
+    """
+
+    def __init__(self, restricted_list):
+        super(StaticRestrictedList, self).__init__()
+
+        self._restricted_set = set(restricted_list)
```
ssanderson: You might make this a `frozenset`, since we don't want to mutate it in place.

zipline/utils/security_list.py
```diff
@@ -16,7 +16,7 @@
 
 class SecurityList(object):
 
-    def __init__(self, data, current_date_func, asset_finder):
```
ssanderson: :+1: on all of this.

zipline/test_algorithms.py
```diff
@@ -505,9 +505,10 @@ def initialize(self, asset=None, max_shares=None, max_notional=None):
 
 
 class SetDoNotOrderListAlgorithm(TradingAlgorithm):
-    def initialize(self, sid=None, restricted_list=None):
+    def initialize(self, sid=None, restricted_list=None,
+                   fail_on_violation=True):
         self.order_count = 0
-        self.set_do_not_order_list(restricted_list)
+        self.set_do_not_order_list(restricted_list, fail_on_violation)
```
ssanderson: This might want to be a string/enum like `on_error` with values of `'warn'` or `'error'` rather than a bool.  Having an enum makes it easier to extend the behavior here in the future.

zipline/gens/tradesimulation.py
```diff
@@ -38,7 +38,7 @@ class AlgorithmSimulator(object):
     }
 
     def __init__(self, algo, sim_params, data_portal, clock, benchmark_source,
-                 universe_func):
+                 rl_controller, universe_func):
```
ssanderson: References to `rl_controller` can probably just be `restrictions`.

zipline/finance/controls.py
```diff
@@ -137,8 +151,11 @@ def validate(self,
         """
         Fail if the asset is in the restricted_list.
         """
-        if asset in self.restricted_list:
-            self.fail(asset, amount, _algo_datetime)
+        if self.restricted_list.is_restricted(asset, _algo_datetime):
```
ssanderson: We might rename this to `restrictions` for consistency elsewehere.

ssanderson: The underscores here are intended to signify unused parameters, so this should probably no longer be prefixed with an underscore.

zipline/finance/controls.py
```diff
@@ -75,6 +78,16 @@ def fail(self, asset, amount, datetime, metadata=None):
                                       datetime=datetime,
                                       constraint=constraint)
 
+    def log(self, asset, amount, datetime, metadata=None):
+        constraint = repr(self)
+        if metadata:
+            constraint = "{constraint} (Metadata: {metadata})".format(
```
ssanderson: This is duplicated between `fail` and `log`.  We should probably refactor to something like `_format_error_message`.

zipline/finance/controls.py
```diff
@@ -75,6 +78,16 @@ def fail(self, asset, amount, datetime, metadata=None):
...
+                constraint=constraint,
+                metadata=metadata
+            )
+        log.error("Order for %s shares of %s at %s violates trading "
```
ssanderson: We don't need to format the string ahead of time here.  Instead, this wants to be:

`log.error('Order for {num_shares} of {asset} at {dt} violates ...', num_shares=amount, asset=asset, dt=datetime ...)`.

zipline/finance/controls.py
```diff
@@ -75,6 +78,16 @@ def fail(self, asset, amount, datetime, metadata=None):
                                       datetime=datetime,
                                       constraint=constraint)
 
+    def log(self, asset, amount, datetime, metadata=None):
```
ssanderson: I think rather than making a new method here, we should just push the notion of `fail_on_violation` into the base class.

zipline/_protocol.pyx
```diff
@@ -482,6 +487,9 @@ cdef class BarData:
         cdef object session_label
         cdef object dt_to_use_for_exchange_check,
 
+        if self._rl_controller.is_restricted(asset, adjusted_dt):
```
ssanderson: We might want to cache the method lookup here at the constructor.  That will save us an attribute access per invocation here.

zipline/_protocol.pyx
```diff
@@ -185,6 +189,7 @@ cdef class BarData:
         self._adjust_minutes = False
 
         self._trading_calendar = trading_calendar
+        self._rl_controller = rl_controller
```
ssanderson: See note below re: caching this.

zipline/_protocol.pyx
```diff
@@ -153,6 +153,9 @@ cdef class BarData:
     data_frequency : {'minute', 'daily'}
         The frequency of the bar data; i.e. whether the data is
         daily or minute bars
+    rl_controller : callable
```
ssanderson: `callable` isn't the right annotation here.  We just just put `Restrictions` (probably with the module path when we decide on it.)

tests/test_security_list.py
```diff
@@ -28,22 +28,22 @@
...
+        self.rl = SecurityListSet(self.asset_finder)
         self.set_do_not_order_list(self.rl.leveraged_etf_list)
         self.order_count = 0
         self.sid = self.symbol(symbol)
 
     def handle_data(self, data):
         if not self.order_count:
-            if self.sid not in \
-                    self.rl.leveraged_etf_list:
+            if not self.rl.leveraged_etf_list.is_restricted(
```
ssanderson: We probably need to check if this is going to break any existing algos on Q.

ssanderson: If it does we may need to add a backwards compat shim.

tests/finance/test_slippage.py
```diff
@@ -750,7 +774,8 @@ def test_orders_stop_limit(self):
         bar_data = BarData(self.data_portal,
                            lambda: self.minutes[1],
                            self.sim_params.data_frequency,
-                           self.trading_calendar)
```
lianga888: I've definitely refactored it and added a fixture. Not sure how you're able to comment on a piece of code that's no longer there

tests/finance/test_slippage.py
```diff
@@ -734,10 +737,10 @@ def test_orders_stop_limit(self):
                 'limit': 3.4})
         ]
 
-        bar_data = BarData(self.data_portal,
-                           lambda: self.minutes[0],
-                           self.sim_params.data_frequency,
-                           self.trading_calendar)
+        bar_data = self.create_bardata(
+            simulation_dt_func=lambda: self.minutes[0],
+            data_frequency=self.sim_params.data_frequency
```
lianga888: It's true that only `self.sim_params.data_frequency` is ever passed in this test suite. But what about the other suites that use this fixture and want to use a frequency other than `self.sim_params.data_frequency`?

tests/finance/test_slippage.py
ssanderson: I think this comment was left over from previously-staged comments.  Feel free to ignore.

tests/test_restrictions.py
```diff
@@ -0,0 +1,143 @@
...
+                self.ASSET2,
+                str_to_ts('2011-01-03'),
+                RESTRICTION_STATES.FROZEN
+            ),
+        ]
+
+        rlm = InMemoryRestrictions(restrictions)
+
+        # Check the restrictions are ordered by dt for each sid
+        self.assertEqual(
```
ssanderson: I don't think we should be making assertions about the internal representation of the data that's chosen here.  As long as the `Restrictions` class behaves as expected, it should be free to store the data however it wants.

tests/test_restrictions.py
```diff
@@ -0,0 +1,143 @@
...
+            [r.effective_date for r in
+             rlm._restrictions_by_asset[self.ASSET1]],
+            [str_to_ts(dt_str) for dt_str in
+             ('2011-01-04', '2011-01-05', '2011-01-06')]
+        )
+
+        self.assertFalse(
```
ssanderson: As a reader, it's hard for me to see the relationship between the assertions being made here and the original test data.  I would write this as something like:

```python
    def test_historical_restrictions(self):
        A1_FREEZE = str_to_ts('2011-01-04')
        A1_THAW = str_to_ts('2011-01-05')
        A1_RE_FREEZE = str_to_ts('2011-01-06')
        minute = pd.Timedelta('1 minute')

        # Build restrictions.

        def assert_is_restricted(asset, dt):
            self.assertTrue(rlm.is_restricted(asset, dt))

        def assert_not_restricted(asset, dt):
            self.assertFalse(rlm.is_restricted(asset, dt))

        # Not restricted until on or after the freeze.
        assert_not_restricted(self.ASSET1, A1_FREEZE - one_minute)
        assert_is_restricted(self.ASSET1, A1_FREEZE)
        assert_is_restricted(self.ASSET1, A1_FREEZE + one_minute)

        # Unrestricted on or after the thaw.
        assert_is_restricted(self.ASSET1, A1_THAW - one_minute)
        assert_not_restricted(self.ASSET1, A1_THAW)
        assert_not_restricted(self.ASSET1, A1_THAW + one_minute)

        # Restricted again on or after the freeze.
        assert_not_restricted(self.ASSET1, A1_RE_FREEZE - one_minute)
        assert_is_restricted(self.ASSET1, A1_RE_FREEZE)
        assert_is_restricted(self.ASSET1, A1_RE_FREEZE + one_minute)
        # Should stay restricted for the rest of time.
        assert_is_restricted(self.ASSET1, A1_RE_FREEZE + one_minute * 100000)

         # Assertions for ASSET2.
```

The above also has the benefit that we could parametrize on the freeze/thaw dates without affecting the business logic of the test.

tests/test_restrictions.py
```diff
@@ -0,0 +1,143 @@
...
+    ASSET_FINDER_EQUITY_SIDS = 1, 2, 3
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(RestrictionsTestCase, cls).init_class_fixtures()
+        cls.ASSET1 = cls.asset_finder.retrieve_asset(1)
+        cls.ASSET2 = cls.asset_finder.retrieve_asset(2)
+        cls.ASSET3 = cls.asset_finder.retrieve_asset(3)
+
+    def test_in_memory_restrictions(self):
```
ssanderson: Other cases that we should probably test:

- Multiple assets with restrictions on the same date shouldn't interact.  (Right now no assets ever have restrictions that apply on the same day.)
- The same asset with consecutive `FROZEN` or `ALLOWED` entries (this should either be an error, or we should ensure that it works.  I'm inclined to say the former, since there's a clear and obvious semantics for it.)
- Restrictions with intraday effective_dates.  (See my note below re: refactoring this to be more readable/parametrizable.
- Re-ordering of the restrictions here and asserting that there's no change in behavior.

tests/test_restrictions.py
```diff
@@ -0,0 +1,143 @@
...
+
+        assert_series_equal(
+            rlm.is_restricted([self.ASSET1, self.ASSET2],
+                              str_to_ts('2011-01-03')),
+            pd.Series(data={self.ASSET1: True, self.ASSET2: True})
+        )
+        assert_series_equal(
+            rlm.is_restricted([self.ASSET1, self.ASSET2],
+                              str_to_ts('2011-01-04')),
+            pd.Series(data={self.ASSET1: True, self.ASSET2: True})
```
ssanderson: Should the vectorized checks here include `self.ASSET3` as well?

ssanderson: One thing that would help this test from a readability perspective would be to alias the assets as something like:

```
RESTRICTED_A1 = self.ASSET1
RESTRICTED_A2 = self.ASSET2
UNRESTRICTED = self.ASSET3

`rlm = StaticRestrictions([RESTRICTED_ASSET1, RESTRICTED_ASSET2])`

# Make assertions using the above names.
```

zipline/algorithm.py
```diff
@@ -2216,20 +2227,25 @@ def set_max_order_count(self, max_count):
...
         """Set a restriction on which assets can be ordered.
 
         Parameters
         ----------
         restricted_list : container[Asset]
             The assets that cannot be ordered.
         """
-        control = RestrictedListOrder(restricted_list)
+
+        if isinstance(restricted_list, (list, tuple, set)):
```
ssanderson: We might also want something like:

```python
elif not isinstance(restricted_list, Restrictions):
    raise TypeError(...)
```

Alternatively, we could decorate this method with `@expect_types(restricted_list=(Restrictions, tuple, list, set), on_error=str)`.

ssanderson:
@lianga888 finished going through the tests.  Generally speaking :+1:.  I had a couple comments on other cases we should make sure we exercise and some notes on ways to make the test assertions clearer for the reader.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class NoRestrictions(Restrictions):
+    """
+    A no-op restrictions that contains no restrictions
+    """
+    def is_restricted(self, assets, dt):
+        if isinstance(assets, Asset):
+            return False
+        return pd.Series(index=pd.Index(assets), data=[False]*len(assets))
```
ssanderson: You can just pass `data=False` here.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+        -------
+        is_restricted : bool or pd.Series[bool] indexed by asset
+            Is the asset or assets restricted on this dt?
+
+        """
+        raise NotImplementedError('is_restricted')
+
+    def __or__(self, other_restriction):
+        """
+        Base implementation for combining two restrictions. If the right side
```
ssanderson: Info like this probably belongs in a comment, not a docstring.  In general, docstrings are meant to be externally-facing, whereas comments are for other developers reading the code.  An external consumer shouldn't need to know or care about `_UnionRestrictions`.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+        is a _UnionRestrictions, calls the overriding implementation with
+        _UnionRestrictions on the left side
+        """
+        if isinstance(other_restriction, _UnionRestrictions):
+            return _UnionRestrictions.__or__(other_restriction, self)
```
ssanderson: You can just write this as `other_restriction | self`.

ssanderson: And I'd comment here something like `# Defer to the _UnionRestrictions implementation of |, which intelligently flattens restriction lists.`

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+        return _UnionRestrictions([self, other_restriction])
+
+
+class _UnionRestrictions(Restrictions):
+    """
+    A union of a number of sub restrictions
```
ssanderson: I'd put a `Notes` section here with something like "Consumers generally shouldn't construct instances of this class directly.  Instead, they should use the `|` operator to combine restrictions."

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Parameters
+    ----------
+    sub_restrictions : iterable of Restrictions (but not _UnionRestrictions)
+        The Restrictions to be added together
+    """
+
+    def __new__(cls, sub_restrictions):
+        """
```
ssanderson: Generally we don't put docstrings on `__init__` and `__new__`, since the constructor for a type gets documented in the class docstring.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+                self.sub_restrictions + other_restriction.sub_restrictions
+        else:
+            new_sub_restrictions = self.sub_restrictions + [other_restriction]
+
+        return _UnionRestrictions(new_sub_restrictions)
+
+    def is_restricted(self, assets, dt):
+        if isinstance(assets, Asset):
+            return any(
+                r.is_restricted(assets, dt) for r in self.sub_restrictions)
```
ssanderson: Personal style note: I find it easier to read and edit expressions like this if the closing paren is on a new line by itself.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+        return pd.Series(
+            index=pd.Index(assets),
+            data=vectorized_is_element(assets, self._restricted_set)
+        )
+
+
+class HistoricalRestrictions(Restrictions):
+    """
+    Historical restrictions stored in memory with effective dates for each
+    asset
```
ssanderson: Style nit: period here.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+        for r in self._restrictions_by_asset.get(asset, ()):
+            if r.effective_date > dt:
+                break
+            state = r.state
+        return state == RESTRICTION_STATES.FROZEN
+
+
+class SecurityListRestrictions(Restrictions):
+    """
+    Restrictions based on a security list
```
ssanderson: Same note as above.  The initial sentence for a docstring should end in a period.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+
+RESTRICTION_STATES = enum(
+    'ALLOWED',
+    'FROZEN',
+)
+
+
+class Restrictions(with_metaclass(abc.ABCMeta)):
+    """
+    Abstract restricted list interface
```
ssanderson: Often for the ABC in a hierarchy, you want a sentence or two describing what an instance of that type represents abstractly.  Here, it would be something like: "ABC representing a set of stocks that an algorithm is restricted from trading.".  For bonus points, you could add a [`Methods`](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#class-docstring) section listing the names and signatures of the methods required for implementing the interface.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+
+        return reduce(
+            operator.or_,
+            (r.is_restricted(assets, dt) for r in self.sub_restrictions)
+        )
+
+
+class NoRestrictions(Restrictions):
+    """
+    A no-op restrictions that contains no restrictions
```
ssanderson: Period here.

zipline/finance/asset_restrictions.py
```diff
@@ -0,0 +1,215 @@
...
+    def is_restricted(self, assets, dt):
+        if isinstance(assets, Asset):
+            return False
+        return pd.Series(index=pd.Index(assets), data=[False]*len(assets))
+
+
+class StaticRestrictions(Restrictions):
+    """
+    Static restrictions stored in memory that are constant regardless of dt
+    for each asset
```
ssanderson: Period here.

zipline/algorithm.py
```diff
@@ -2147,27 +2161,68 @@ def set_max_order_count(self, max_count):
...
+                category=ZiplineDeprecationWarning,
+                stacklevel=2
+            )
+            restrictions = StaticRestrictions(restricted_list)
+
+        self.set_asset_restrictions(restrictions, on_error)
+
+    @api_method
+    @expect_types(
+        restrictions=(Restrictions,),
```
ssanderson: Minor note: this can just be `Restrictions` instead of a single-element tuple.

ssanderson:
@lianga888 this looks great to me.  I left a bunch of ticky-tack style notes on the docstrings for `Restrictions`, but otherwise I think this is good to merge pending downstream readiness.

tests/test_restrictions.py
```diff
@@ -0,0 +1,422 @@
...
+
+        for dt in [str_to_ts(dt_str) for dt_str in ('2011-01-03',
+                                                    '2011-01-04',
+                                                    '2020-01-04')]:
+            assert_not_restricted(self.ASSET1, dt)
+            assert_not_restricted(self.ASSET2, dt)
+            assert_not_restricted(self.ASSET3, dt)
+            assert_all_restrictions([False, False, False], dt)
+
+    def test_union_restrictions(self):
```
ssanderson: :+1:

lianga888:
```
The following are all 1-year backtest that calls can_trade for each 10 sids every bar:
-------------------------------------------------------------------------------------------------
**No restrictions, run on master**
59027945 function calls (59016856 primitive calls) in **93.299** seconds

**No restrictions**
59028121 function calls (59016996 primitive calls) in **95.662** seconds
ncalls  tottime  percall  **cumtime**  percall filename:lineno(function)
 979200    0.556    0.000    **0.829**    0.000 asset_restrictions.py:123(is_restricted)

**Empty list of restrictions**
59029588 function calls (59018233 primitive calls) in **91.450** seconds
ncalls  tottime  percall  **cumtime**  percall filename:lineno(function)
979200    0.763    0.000    **1.025**    0.000 asset_restrictions.py:143(is_restricted)

**Every sid is restricted for every minute**(by-passes remaining can_trade logic because restricted) 
12805905 function calls (12794098 primitive calls) in **14.291** seconds
ncalls  tottime  percall  **cumtime**  percall filename:lineno(function)
979200    0.689    0.000    **2.551**    0.000 asset_restrictions.py:177(is_restricted)

**10 restricted sids that are not the 10 sids checked**(runs remaining can_trade logic)
61013160 function calls (61000358 primitive calls) in **101.846** seconds
ncalls  tottime  percall  **cumtime**  percall filename:lineno(function)
979200    1.010    0.000    **2.882**    0.000 asset_restrictions.py:177(is_restricted)

**30 restrictions, 10 sids are restricted for every minute**(by-passes remaining can_trade logic because restricted) 
12852158 function calls (12838031 primitive calls) in **14.053** seconds
ncalls  tottime  percall  **cumtime**  percall filename:lineno(function)
979200    0.670    0.000    **2.484**    0.000 asset_restrictions.py:177(is_restricted)

The following are all 1-year backtest that calls can_trade for each 30 sids every bar:
-------------------------------------------------------------------------------------------------
**Every sid is restricted for every minute**(by-passes remaining can_trade logic because restricted)
25871908 function calls (25858296 primitive calls) in **21.991** seconds
ncalls  tottime  percall  **cumtime**  percall filename:lineno(function)
2839680    1.734    0.000    **6.550**    0.000 asset_restrictions.py:177(is_restricted)

lianga888:
screen shot 2016-10-05 at 9 47 50 am

PR #1488, opened 2016-09-15
ENH: Make reader.get_value raise NoDataOnDate if the date is not in the calendar.

jbredeche:
Previously, at least the minute bar reader was forward filling (using the previous market minute).

DataPortal now catches the NoDataOnDate exception and returns nan for OHLC and 0 for V.

Price is still forward filled, unchanged.

### DIFF
--- a/zipline/data/_minute_bar_internal.pyx
+++ b/zipline/data/_minute_bar_internal.pyx
@@ -38,7 +38,8 @@ def minute_value(ndarray[long_t, ndim=1] market_opens,
 def find_position_of_minute(ndarray[long_t, ndim=1] market_opens,
                             ndarray[long_t, ndim=1] market_closes,
                             long_t minute_val,
-                            short minutes_per_day):
+                            short minutes_per_day,
+                            bool forward_fill):
     """
     Finds the position of a given minute in the given array of market opens.
     If not a market minute, adjusts to the last market minute.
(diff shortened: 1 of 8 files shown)


zipline/data/_minute_bar_internal.pyx
```diff
@@ -112,7 +128,7 @@ def find_last_traded_position_internal(
 
     minute_pos = int_min(
         find_position_of_minute(market_opens, market_closes, end_minute,
```
ehebert: I see how avoiding the extra `searchsorted` is helpful, but I worry about making the behavior of `find_position_of_minute_minute` adjustable for `ffill` or not; leaving the function as simple as possible.

Another option would be to put the handling of the `NoDataOnDate` in this scope (which is focused on finding the last date.)

Something like:
```
try:
    minute_pos = int_min(
         find_position_of_minute(market_opens, market_closes, end_minute,
                                                minutes_per_day)
except NoDataOnDate:
    prev_close_loc = searchsorted(market_closes, end_minute)
    minute_pos = (prev_close_loc + 1)  * minute_per_day) - 1
```

jbredeche: hm, true. But then you're essentially replicating `find_position_of_minute`'s logic here (the searchsorted call, plus the math).  

I think I prefer it the way it is.

PR #1490, opened 2016-09-15
MAiNT: Use load_adjustments for history.

ehebert:
Instead of `HistoryLoader` containing separate adjustment calculation
logic, use `SQLiteAdjustmentReader.load_adjustments`.

This change required the addition of two offset parameters to
`load_adjustments` since the perspective on the data from within
`schedule_function` is skewed from how Pipeline looks at historical
data.

This is working towards creating an `AdjustmentReader` abc which
`SQLiteAdjustmentReader` and a upcoming continuous future adjustment
reader will share.

### DIFF
--- a/zipline/data/_adjustments.pyx
+++ b/zipline/data/_adjustments.pyx
@@ -147,7 +147,9 @@ cdef _adjustments(object adjustments_db,
 cpdef load_adjustments_from_sqlite(object adjustments_db,  # sqlite3.Connection
                                    list columns,
                                    DatetimeIndex_t dates,
-                                   Int64Index_t assets):
+                                   Int64Index_t assets,
+                                   int end_offset,
+                                   int date_offset):
     """
     Load a dictionary of Adjustment objects from adjustments_db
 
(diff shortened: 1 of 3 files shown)


zipline/data/_adjustments.pyx
```diff
@@ -228,26 +235,30 @@ cpdef load_adjustments_from_sqlite(object adjustments_db,  # sqlite3.Connection
 
         date_loc = _lookup_dt(date_ixs, eff_date, _dates_seconds)
 
+        end_loc = date_loc + end_offset
```
jbredeche: can you `cdef` both `end_loc` and `out_loc`?

zipline/data/history_loader.py
```diff
@@ -141,79 +142,10 @@ def _get_adjustments_in_range(self, asset, dts, field,
...
-                                       0,
-                                       ratio)
-                try:
-                    adjs[adj_loc].append(mult)
-                except KeyError:
-                    adjs[adj_loc] = [mult]
-        return adjs
+        end_offset = -1
+        date_offset = -1 if is_perspective_after else 0
+        return self._adjustments_reader.load_adjustments(
```
jbredeche: is there a way to pass multiple fields or assets now, instead of doing them one at a time?

zipline/data/_adjustments.pyx
ehebert: Yes, good catch.

zipline/data/history_loader.py
ehebert: Yes, I anticipate making that change on the path to supporting continuous futures.

zipline/data/_adjustments.pyx
ehebert: Squashed that in.

PR #1491, opened 2016-09-16
ENH: Switching from cachetools.LRUCache to lru-dict's LRU cache.

jbredeche:
Overall speedup is not huge, but all the time spent inside LRUCache's machinery is gone.  Most visible in `BcolzMinuteBarReader`.


jbredeche:
(@richafrank signed off on it)

PR #1492, opened 2016-09-16
PERF: Save up to 80% of the calls to minute_to_session_label

jbredeche:
One year NYSE test that buys a lot triggers 492,963 calls to
minute_to_session_label.  Only 98924 (~= (390 * 252)) make it past the
cache and trigger the heavier computation.

### DIFF
--- a/zipline/utils/calendars/trading_calendar.py
+++ b/zipline/utils/calendars/trading_calendar.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from abc import ABCMeta, abstractproperty
+from lru import LRU
 
 from pandas.tseries.holiday import AbstractHolidayCalendar
 from six import with_metaclass
@@ -94,6 +95,12 @@ class TradingCalendar(with_metaclass(ABCMeta)):
             },
             dtype='datetime64[ns]',
         )
+
+        # Simple cache to avoid recalculating the same minute -> session in
+        # "next" mode. Analysis of current zipline code paths show that
+        # `minute_to_session_label` is often called consecutively with the same
+        # inputs.
+        self._minute_to_session_label_cache = LRU(1)
 
         self.market_opens_nanos = self.schedule.market_open.values.\
             astype(np.int64)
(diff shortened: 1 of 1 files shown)


richafrank:
I'm excited!  Can you paste the before and after?

jbredeche:
reference is one year NYSE backtest that buys 1 share of apple every minute.

before, the zipline sim took 64 seconds.  afterwards, 56 seconds. (on my machine)

zipline/utils/calendars/trading_calendar.py
```diff
@@ -690,11 +699,19 @@ def minute_to_session_label(self, dt, direction="next"):
         pd.Timestamp (midnight UTC)
             The label of the containing session.
         """
+        if direction == "next":
```
ssanderson: How does this compare to just decorating this method with `cache_tools.lru_cache(maxsize=1)`?

jbredeche: We no longer have `cache_tools` in zipline, replaced it with `lru-dict`, which is faster.

I don't think it can be used as a decorator out of the box, but that could definitely be added.

zipline/utils/calendars/trading_calendar.py
```diff
@@ -95,6 +96,14 @@ def __init__(self, start=start_default, end=end_default):
             dtype='datetime64[ns]',
         )
 
+        # Simple cache to avoid recalculating the same minute -> session in
+        # "next" mode. Analysis of current zipline code paths show that
+        # `minute_to_session_label` is often called consecutively with the same
+        # inputs.
+        self._minute_to_session_label_cache = LRU(1)
+
+        self._session_nanos = self.schedule.index.values.astype(np.int64)
```
jbredeche: oops, this is cruft, will remove

zipline/utils/calendars/trading_calendar.py
```diff
@@ -690,11 +697,19 @@ def minute_to_session_label(self, dt, direction="next"):
...
+                return self._minute_to_session_label_cache[dt]
+            except KeyError:
+                pass
 
         idx = searchsorted(self.market_closes_nanos, dt)
         current_or_next_session = self.schedule.index[idx]
+        self._minute_to_session_label_cache[dt] = current_or_next_session
 
-        if direction == "previous":
+        if direction == "next":
```
richafrank: 👍 on putting the most common case first

zipline/utils/calendars/trading_calendar.py
obsidia4: When I import zipline ,it returns  'ImportError: No module named 'lru''. 
My system is win10-64bit,and environment is python 3.5. Would you please tell me what should I do to fix it?

PR #1493, opened 2016-09-16
Enumpycal

richafrank:
PERF: Use empyrical with ndarrays instead of Series

Depends on https://github.com/quantopian/empyrical/pull/32


jdricklefs:
:+1: (except for the tmp commit, of course)

PR #1494, opened 2016-09-16
MAINT: Remove unused method.

ehebert:
Dividends are read from AdjustmentReader.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -1631,13 +1631,6 @@ class TradingAlgorithm(object):
             dt = dt.astimezone(tz)
         return dt
 
-    def update_dividends(self, dividend_frame):
-        """
-        Set DataFrame used to process dividends.  DataFrame columns should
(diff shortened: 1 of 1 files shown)


zipline/algorithm.py
```diff
@@ -1631,13 +1631,6 @@ def get_datetime(self, tz=None):
             dt = dt.astimezone(tz)
         return dt
 
-    def update_dividends(self, dividend_frame):
-        """
-        Set DataFrame used to process dividends.  DataFrame columns should
-        contain at least the entries in zp.DIVIDEND_FIELDS.
-        """
-        self.perf_tracker.update_dividends(dividend_frame)
```
ssanderson: Can we also remove `PerformanceTracker.update_dividends`?

ehebert: We already had removed that method. (This method was referring to a non-existent `update_dividends` in `PerformanceTracker`.

PR #1497, opened 2016-09-19
PERF: Be smarter about counting the number of minutes across a contiguous bunch of sessions.

jbredeche:
Much, much, much faster this way.

### DIFF
--- a/zipline/utils/calendars/trading_calendar.py
+++ b/zipline/utils/calendars/trading_calendar.py
@@ -149,6 +149,29 @@ class TradingCalendar(with_metaclass(ABCMeta)):
     def close_offset(self):
         return 0
 
+    @lazyval
+    def _minutes_per_session(self):
+        diff = self.schedule.market_close - self.schedule.market_open
+        diff = diff.astype('timedelta64[m]')
+        return diff + 1
+
+    def minutes_count_for_sessions_in_range(self, start_session, end_session):
+        """
+        Parameters
+        ----------
+        start_session: pd.Timestamp
+            The first session.
+
+        end_session: pd.Timestamp
+            The last session.
+
+        Returns
+        -------
+        int: The total number of minutes for the contiguous chunk of sessions.
+             between start_session and end_session, inclusive.
+        """
+        return int(self._minutes_per_session[start_session:end_session].sum())
+
     @property
     def regular_holidays(self):
         """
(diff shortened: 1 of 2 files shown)


richafrank:
Looking at the repr of NYSE.minutes_per_session, I see:

```
1990-01-02 00:00:00+00:00    389
1990-01-03 00:00:00+00:00    389
1990-01-04 00:00:00+00:00    389
1990-01-05 00:00:00+00:00    389
1990-01-08 00:00:00+00:00    389
1990-01-09 00:00:00+00:00    389
1990-01-10 00:00:00+00:00    389
1990-01-11 00:00:00+00:00    389
1990-01-12 00:00:00+00:00    389
1990-01-15 00:00:00+00:00    389
1990-01-16 00:00:00+00:00    389
1990-01-17 00:00:00+00:00    389
...
```
Is that expected?

zipline/utils/calendars/trading_calendar.py
```diff
@@ -149,6 +149,16 @@ def open_offset(self):
...
 
+    @lazyval
+    def minutes_per_session(self):
+        """
+        Returns
+        -------
+        df: A dataframe containing the number of minutes per session.
+        """
+        diff = self.schedule.market_close - self.schedule.market_open
+        return diff.astype('timedelta64[m]')
```
richafrank: Instead of the calendar exposing a dataframe to consumers, so now we're not in the managed land of sessions and labels anymore, would it make sense to expose a function that takes a list of sessions and does the slicing, i.e. move the call site code here?

jbredeche: agreed, will make it so.

zipline/utils/calendars/trading_calendar.py
```diff
@@ -149,6 +149,21 @@ def open_offset(self):
...
+        return diff + 1
+
+    def minutes_count_for_sessions_in_range(self, start_session, end_session):
+        """
+        Returns
+        -------
+        int: The total number of minutes for the contiguous chunk of sessions.
+             between start_session and end_session, inclusive.
+        """
+        return sum(self._minutes_per_session[start_session:end_session])
```
richafrank: Is there a vectorized sum that we should use here?

jbredeche: yup, done

PR #1500, opened 2016-09-19
MAINT: Adds option for minute bar writer to not write metadata

yankees714:
With the addition of the truncate function, there are cases where we'll want to construct a `BcolzMinuteBarWriter` to call truncate, without gathering all the metadata. This commit adds a write_metadata arg to its init, which is True by default. If False is specified, no metadata is written.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -317,6 +317,10 @@ class BcolzMinuteBarWriter(object):
 
         Defaults to supporting 15 years of NYSE equity market data.
         see: http://bcolz.blosc.org/opt-tips.html#informing-about-the-length-of-your-carrays # noqa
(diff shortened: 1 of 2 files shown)


zipline/data/minute_bars.py
```diff
@@ -391,15 +396,16 @@ def __init__(self,
...
-            self._minutes_per_day,
-        )
-        metadata.write(self._rootdir)
+        if write_metadata:
+            metadata = BcolzMinuteBarMetadata(
+                self._default_ohlc_ratio,
+                self._ohlc_ratios_per_sid,
+                self._calendar,
+                self._start_session,
+                self._end_session,
```
ehebert: In the `truncate` case `_end_session` will need to be updated. Does `truncate` make the metadata change?

PR #1501, opened 2016-09-20
PERF: vectorize earnings estimates

llllllllll:
big performance improvement on larger data sets. This also fixes what I think is a bug where the data is not sorted on event_date in `get_requested_quarter_data` but I haven't written a test for that yet.

### DIFF
--- a/zipline/pipeline/loaders/earnings_estimates.py
+++ b/zipline/pipeline/loaders/earnings_estimates.py
@@ -1,5 +1,7 @@
-from collections import defaultdict
 from abc import abstractmethod, abstractproperty
+from collections import defaultdict
+
+import numpy as np
 import pandas as pd
 from six import viewvalues
 from toolz import groupby
@@ -136,11 +138,15 @@ class EarningsEstimatesLoader(PipelineLoader):
             self.estimates[FISCAL_YEAR_FIELD_NAME],
             self.estimates[FISCAL_QUARTER_FIELD_NAME],
         )
-        self.array_overwrites_dict = {datetime64ns_dtype:
-                                      Datetime641DArrayOverwrite,
-                                      float64_dtype: Float641DArrayOverwrite}
-        self.scalar_overwrites_dict = {datetime64ns_dtype: Datetime64Overwrite,
-                                       float64_dtype: Float64Overwrite}
+
+        self.array_overwrites_dict = {
+            datetime64ns_dtype: Datetime641DArrayOverwrite,
+            float64_dtype: Float641DArrayOverwrite,
+        }
+        self.scalar_overwrites_dict = {
+            datetime64ns_dtype: Datetime64Overwrite,
+            float64_dtype: Float64Overwrite,
+        }
 
         self.name_map = name_map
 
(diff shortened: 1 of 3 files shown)


zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -244,54 +246,40 @@ def get_adjustments(self,
...
-                zero_qtr_sid_data[NORMALIZED_QUARTERS] !=
-                zero_qtr_sid_data[NORMALIZED_QUARTERS].shift(1)
-            ]
-            # On dates where we don't have any information about quarters,
-            # we will get nulls, and each of these will be interpreted as
-            # quarter shifts. We need to remove these here.
-            qtr_shifts = qtr_shifts[
-                qtr_shifts[NORMALIZED_QUARTERS].notnull()
+        zero_qtr_data = zero_qtr_data.reset_index(level=NORMALIZED_QUARTERS)
+        zero_qtr_data.sort_index(inplace=True)
```
mtydykov: Why do we need to sort the index here?

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -244,54 +246,40 @@ def get_adjustments(self,
...
+
+        quarter_shifts = zero_qtr_data.loc[
```
mtydykov: I think the original version is clearer

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -403,34 +393,40 @@ def load_adjusted_array(self, columns, dates, assets, mask):
...
-                self.get_shifted_qtrs(
-                    zeroth_quarter_idx.get_level_values(NORMALIZED_QUARTERS),
-                    num_quarters
-                ),
-                append=True
-            ).index
-            requested_qtr_idx = requested_qtr_idx.rename(
-                SHIFTED_NORMALIZED_QTRS, -1
+            zero_qtr_data_idx = zero_qtr_data.index
+            requested_qtr_idx = pd.MultiIndex.from_arrays(
```
mtydykov: Thinking that maybe we should move this into `get_requested_quarter_data()`, since it's not needed anywhere else, to simplify `load_adjusted_array` a bit further.

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -614,6 +624,7 @@ def get_zeroth_quarter_idx(self, num_quarters, stacked_last_per_qtr):
             stacked_last_per_qtr[EVENT_DATE_FIELD_NAME] <=
             stacked_last_per_qtr.index.get_level_values(SIMULTATION_DATES)
         ].groupby(
-            level=[SIMULTATION_DATES, SID_FIELD_NAME], as_index=False
+            level=[SIMULTATION_DATES, SID_FIELD_NAME],
```
ssanderson: typo: SIMULTATION

zipline/pipeline/loaders/earnings_estimates.py
mtydykov: Although, looking at this more, it's not that unclear. Maybe a comment would help.

mtydykov:
The data is sorted on event date in `get_last_data_per_qtr`.

llllllllll:
I saw that we had a sort there, but some of our reshaping had removed that sorting by the time we got there.

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -530,22 +542,21 @@ def create_overwrite_for_estimate(self,
             last_per_qtr[
                 column_name,
                 requested_quarter,
-                sid
-            ][:next_qtr_start_idx].values)
+                sid,
+            ].values[:next_qtr_start_idx],
```
llllllllll: integer indexing into ndarrays is faster than indexing pandas objects.

zipline/pipeline/loaders/earnings_estimates.py
llllllllll: I am not sure what you mean, the original version was looping over the sids and doing this indexing.

zipline/pipeline/loaders/earnings_estimates.py
llllllllll: because we are doing a `diff`/`shift` so the order is very important. Without this we are not comparing the correct days.

zipline/pipeline/loaders/earnings_estimates.py
mtydykov: I just found `diff()` and then `notzero()` to be confusing at first.

llllllllll: Okay, I find it a bit more natural. It is also faster.

zipline/pipeline/loaders/earnings_estimates.py
mtydykov: Ah, and we had it sorted by event_date before, so that would have screwed up the order of the days

PR #1502, opened 2016-09-20
MAINT: Remove `future_chain` API method.

ehebert:
`future_chain` will be replaced by the as yet to be implemented method,
`data.current_chain`

Also removing `FutureChain` which will be replaced by another version
which only supports indexing and iteration.


tests/test_assets.py
```diff
@@ -1218,179 +1124,6 @@ def test_error_message_plurality(self,
...
-            6: 'M',   # June
-            7: 'N',   # July
-            8: 'Q',   # August
-            9: 'U',   # September
-            10: 'V',  # October
-            11: 'X',  # November
-            12: 'Z',  # December
-        }
-        for key in codes:
-            self.assertEqual(codes[key], month_to_cme_code(key))
```
yankees714: Since we're not removing `cme_code_to_month` or `month_to_cme_code`, should we maybe relocate these two tests?

zipline/errors.py
```diff
@@ -295,15 +295,6 @@ class SymbolNotFound(ZiplineError):
 """.strip()
 
 
-class RootSymbolNotFound(ZiplineError):
```
yankees714: We'll probably still need something like this, right?

tests/test_assets.py
ehebert: Open to adding them back in. Had removed them because it was essentially testing whether a `dict` is behaving as expected. (And the unused functions may be an unnecessary step. Following the pattern in other places where we refer to definitions, the definitions could be used as `CME_CODE_TO_MONTH[code]` and `MONTH_TO_CME_CODE[code]`.

zipline/errors.py
ehebert: Probably, can change patch so that it remains.

tests/test_assets.py
yankees714: +1 on removing the functions.

Issue #1503, opened 2016-09-21
Dramatic Memory Usage Increases After Pandas 18 Merge


ssanderson:
Prior to the merge of https://github.com/quantopian/zipline/pull/1339, peak memory usage of the test suite was around 800MB.  Post merge, we peak at over 2GB on my machine.

ssanderson:
Root cause appears to be https://github.com/quantopian/zipline/pull/1504.

FreddieV4:
This should be fixed now, right?

PR #1504, opened 2016-09-21
PERF: Remove or defer calls to get_loc on large indices.

ssanderson:
Mitigation for https://github.com/quantopian/zipline/issues/1503.

This was debugged primarily by adding the following lines to `zipline/tests/__init__.py`:

```python
+import os
+from unittest import TestCase
+
+import psutil
+import humanize
+import __builtin__
+
+pid = os.getpid()
+proc = psutil.Process(pid)
+
+def get_mem_usage():
+    return humanize.naturalsize(proc.memory_full_info().uss)
+__builtin__.gmem = get_mem_usage
+
+
+real_doCleanups = TestCase.doCleanups
+
+OUTPUT_FILE = open('/home/ssanderson/pandas18_memory_usage.txt', 'w')
+
+
+def patched_doCleanups(self, *args, **kwargs):
+    OUTPUT_FILE.write("{} {}\n".format(get_mem_usage(), self))
+    OUTPUT_FILE.flush()
+    return real_doCleanups(self, *args, **kwargs)
+
+TestCase.doCleanups = patched_doCleanups
```

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -213,12 +213,6 @@ class DataPortal(object):
             self.trading_calendar.all_sessions.get_loc(self._first_trading_day)
             if self._first_trading_day is not None else None
         )
-        self._first_trading_minute_loc = (
-            self.trading_calendar.all_minutes.get_loc(
-                self._first_trading_minute
-            )
-            if self._first_trading_minute is not None else None
-        )
 
     def _ensure_reader_aligned(self, reader):
         if reader is None:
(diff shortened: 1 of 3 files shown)


zipline/data/data_portal.py
```diff
@@ -703,10 +697,17 @@ def _get_history_daily_window_data(
 
             return daily_data
 
-    def _handle_history_out_of_bounds(self, bar_count):
+    def _handle_minute_history_out_of_bounds(self, bar_count):
+        first_trading_minute_loc = (
```
jbredeche: Can we save this calculated value somewhere so that we don't have to do it again later in the simulation?

jbredeche: maybe move this into a `@lazyval`?

ssanderson: I thought about caching this like we were doing before, but we're only calculating this offset so that we can give a useful error message in an out of bounds exception.  We don't expect this to be needed again.  The only case where this would help performance would be someone who's making invalid history calls in a loop.  Do you think it's still worth having the lazyval?

jbredeche: I'm not certain enough to say we should do it now.  LGTM.  Can always tweak it later.

PR #1507, opened 2016-09-22
BUG: Catch all missing data exceptions when computing dividend ratios

yankees714:


ehebert:
This works, seems like there should be a common base for those exceptions though.
e.g. `NoDataAfterDate` could subclass `NoDataOnDate`.

PR #1510, opened 2016-09-22
PERF: Replace get_loc calls in calc_dividend_ratios with get_indexer

yankees714:
We can make a single vectorized call outside of the loop, instead of repeatedly calling `get_loc` inside it.

### DIFF
--- a/zipline/data/us_equity_pricing.py
+++ b/zipline/data/us_equity_pricing.py
@@ -959,7 +959,7 @@ class SQLiteAdjustmentWriter(object):
             - effective_date, the date in seconds on which to apply the ratio.
             - ratio, the ratio to apply to backwards looking pricing data.
         """
-        if dividends is None:
+        if dividends is None or dividends.empty:
             return DataFrame(np.array(
                 [],
                 dtype=[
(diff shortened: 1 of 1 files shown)


zipline/data/us_equity_pricing.py
```diff
@@ -978,12 +978,21 @@ def calc_dividend_ratios(self, dividends):
         equity_daily_bar_reader = self._equity_daily_bar_reader
 
         effective_dates = full(len(amounts), -1, dtype=int64)
+
         calendar = self._calendar
+
+        # Calculate locs against a tz-naive cal, as the ex_dates are tz-
+        # naive.
+        tz_naive_calendar = calendar.tz_localize(None)
```
ssanderson: Rather than localizing the calendar to naive, can we localize the ex_dates to the calendar's time zone?  They should be semantically identical, but using the same calendar prevents allocation of a new hash table.

yankees714: Attempted this at first in https://github.com/quantopian/zipline/pull/1510/commits/189b868f7f66b5b906f162121ef621f0b224c51b, but `get_indexer` appears to drop the tz info, which throws off the comparison, so going with this workaround for now.

PR #1512, opened 2016-09-23
ENH: Add slippage models to execute at limit, open prices

volconst:
Generalize SlippageModel with the idea of bar stage.            
3 bar stages are defined: Open, HighLow and Close. The behavior
until now corresponds to using only Close stage.
The first added slippage model HLCVolumeSlippage executes limit
orders at limit price instead of close price, but still using
close price for market orders.
The second slippage model OHLVolumeSlippage takes advantage of
all the OHLC prices in the bar, executing limit orders either 
at open or limit depending on marketability.
Default behaviour is not changed. To use the new functionality:
 
def initialize(context):
        context.set_slippage(HLCVolumeSlippage())

### DIFF
--- a/tests/finance/test_slippage.py
+++ b/tests/finance/test_slippage.py
@@ -27,7 +27,7 @@ from pandas.tslib import normalize_date
 from numpy import arange
 
 from zipline.finance.slippage import VolumeShareSlippage,\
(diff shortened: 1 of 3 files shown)


rclai:
What is the status of this PR? Did you guys come to a conclusion?

PR #1513, opened 2016-09-23
Propagate exceptions in loader to prevent variable reference before use

jkleint:
`data.loader.ensure_benchmark_data()` was trying to use data after an exception was raised loading it.  The code was logging and swallowing exceptions; this re-raises.

### DIFF
--- a/zipline/data/loader.py
+++ b/zipline/data/loader.py
@@ -249,6 +249,7 @@ def ensure_benchmark_data(symbol, first_date, last_date, now, trading_day):
         data.to_csv(path)
     except (OSError, IOError, HTTPError):
         logger.exception('failed to cache the new benchmark returns')
+        raise
     if not has_data_for_dates(data, first_date, last_date):
         logger.warn("Still don't have expected data after redownload!")
     return data


ssanderson:
:+1: thanks for the patch!

zipline/data/loader.py
```diff
@@ -249,6 +249,7 @@ def ensure_benchmark_data(symbol, first_date, last_date, now, trading_day):
         data.to_csv(path)
     except (OSError, IOError, HTTPError):
         logger.exception('failed to cache the new benchmark returns')
+        raise
```
richafrank: @llllllllll I thought the intent here was to catch this failure, warn, and continue?

llllllllll: it was, I think we don't want to put the `get_benchmark_returns` call in the same try except. If `data` cannot be read we need to fail but if we cannot cache it to disk for later runs we can continue.

ssanderson: Ah, I think I misread the intended logic here.  We only want to log and continue if it's `to_csv` that failed.  If the initial read fails, we don't have anything to return (hence the `UnboundLocalsError` that @jkleint ran into).

PR #1516, opened 2016-09-28
Rlist tweaks

ssanderson:


tests/test_restrictions.py
```diff
@@ -34,126 +36,130 @@ def init_class_fixtures(cls):
...
-            assert_is_restricted(asset, dts[0] + MINUTE)
+            assert_not_restricted(asset, freeze_dt - MINUTE)
+            assert_is_restricted(asset, freeze_dt)
+            assert_is_restricted(asset, freeze_dt + MINUTE)
 
-            # Unrestricted on or after the unfreeze
-            assert_is_restricted(asset, dts[1] - MINUTE)
-            assert_not_restricted(asset, dts[1])
-            assert_not_restricted(asset, dts[1] + MINUTE)
+            # Unrestricted until on or after the unfreeze
```
lianga888: I don't think 'until' here is right. It was restricted before the unfreeze, and unrestricted on or after the unfreeze. Until the unfreeze, it was still restricted

PR #1519, opened 2016-10-03
MAINT: Begin making a common adjustment interface.

ehebert:
Start making the equity adjustments calculations for the history loader
conform to the same method signature as `load_adjustments` provided by
`SQLiteAdjustmentReader, so that an `AdjustmentReader` interface can
begin to take form.

This prepares for creating a `DispatchAdjustmentReader` which will route
adjustment calculations for equities to the
`HistoryCompatibleUSEquityAdjustmentReader` and continuous futures to a
not yet implemented adjustment reader. All of these readers will share
the `load_adjustments` method.

### DIFF
--- a/zipline/data/history_loader.py
+++ b/zipline/data/history_loader.py
@@ -32,80 +32,27 @@ from zipline.utils.memoize import lazyval
 from zipline.utils.numpy_utils import float64_dtype
 
 
-class SlidingWindow(object):
-    """
-    Wrapper around an AdjustedArrayWindow which supports monotonically
-    increasing (by datetime) requests for a sized window of data.
+class HistoryCompatibleUSEquityAdjustmentReader(object):
 
-    Parameters
-    ----------
-    window : AdjustedArrayWindow
-       Window of pricing data with prefetched values beyond the current
-       simulation dt.
-    cal_start : int
-       Index in the overall calendar at which the window starts.
-    """
+    def __init__(self, adjustment_reader):
+        self._adjustments_reader = adjustment_reader
 
-    def __init__(self, window, size, cal_start, offset):
-        self.window = window
-        self.cal_start = cal_start
-        self.current = around(next(window), 3)
-        self.offset = offset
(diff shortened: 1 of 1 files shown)


zipline/data/history_loader.py
```diff
@@ -31,83 +31,30 @@
...
+            adjs = {}
+            for asset in assets:
+                adjs.update(self._get_adjustments_in_range(
+                    asset, dates, column, perspective_offset))
+            out[i] = adjs
+        return out
 
     def _get_adjustments_in_range(self, asset, dts, field,
-                                  is_perspective_after):
+                                  perspective_offset=0):
```
ehebert: Noticed arguments and docstring were not matching. Fixing now.

ehebert:
This PR takes a smaller step towards a common interface for adjustments than https://github.com/quantopian/zipline/pull/1490 did, as that previous attempt introduced temporary performance regression due to querying for all sids at once.

After the `load_adjustment` usage is used throughout and an `AdjustmentReader` interface is added. We can re-open using `SQLiteAdjustmentReader` back the equity adjustments for the history loader directly.

zipline/data/history_loader.py
```diff
@@ -31,83 +31,30 @@
...
-        self.window = window
-        self.cal_start = cal_start
-        self.current = around(next(window), 3)
-        self.offset = offset
-        self.most_recent_ix = self.cal_start + size
+    def __init__(self, adjustment_reader):
+        self._adjustments_reader = adjustment_reader
 
-    def get(self, end_ix):
+    def load_adjustments(self, columns, dts, assets, output_offset=0):
```
ehebert: @ssanderson do you have a better alternative than `output_offset` for this value which changes the values so that `AdjustedArrayWindow` pops the adjustment an index earlier? e.g. `{2: Float64Window(0, 2, 0, 0, 0.5)`} becomes `{1: Float64Window(0, 2, 0, 0, 0.5)`

PR #1521, opened 2016-10-04
Add optimization pass for initial workspace

llllllllll:
Also adds an `alias` method of terms which allows people to "name" an expression which makes it repr better.

### DIFF
--- a/tests/pipeline/test_alias.py
+++ b/tests/pipeline/test_alias.py
@@ -29,8 +29,17 @@ class WithAlias(object):
     def test_repr(self):
         assert_equal(
             repr(self.Term().alias('ayy lmao')),
-            "Aliased%s(..., name='ayy lmao')" % self.Term.__base__.__name__,
+            "Aliased%s(Term(...), name='ayy lmao')" % (
+                self.Term.__base__.__name__,
+            ),
         )
+
+    def test_short_repr(self):
+        for name in ('a', 'b'):
+            assert_equal(
+                self.Term().alias(name).short_repr(),
+                name,
+            )
 
 
 class TestFactorAlias(WithAlias, BasePipelineTestCase):
--- a/tests/pipeline/test_alias.py
+++ b/tests/pipeline/test_alias.py
@@ -1,0 +1,54 @@
+import numpy as np
+
+from zipline.testing.predicates import assert_equal
+from zipline.pipeline import Classifier, Factor, Filter
+from zipline.utils.numpy_utils import float64_dtype, int64_dtype
+
+from .base import BasePipelineTestCase
+
+
+class WithAlias(object):
+
+    def test_alias(self):
+        f = self.Term()
+        alias = f.alias('ayy lmao')
+
+        f_values = np.random.RandomState(5).randn(5, 5)
+
+        self.check_terms(
+            terms={
+                'f_alias': alias,
+            },
+            expected={
+                'f_alias': f_values,
+            },
+            initial_workspace={f: f_values},
+            mask=self.build_mask(np.ones((5, 5))),
+        )
+
+    def test_repr(self):
+        assert_equal(
+            repr(self.Term().alias('ayy lmao')),
+            "Aliased%s(..., name='ayy lmao')" % self.Term.__base__.__name__,
+        )
+
+
+class TestFactorAlias(WithAlias, BasePipelineTestCase):
+    class Term(Factor):
+        dtype = float64_dtype
+        inputs = ()
+        window_length = 0
+
+
+class TestFilterAlias(WithAlias, BasePipelineTestCase):
+    class Term(Filter):
+        inputs = ()
+        window_length = 0
+
+
+class TestClassifierAlias(WithAlias, BasePipelineTestCase):
+    class Term(Classifier):
+        dtype = int64_dtype
+        inputs = ()
+        window_length = 0
+        missing_value = -1
--- a/tests/pipeline/test_alias.py
+++ b/tests/pipeline/test_alias.py
@@ -1,3 +1,4 @@
+from nose.tools import nottest
 import numpy as np
 
 from zipline.testing.predicates import assert_equal
@@ -7,7 +8,8 @@ from zipline.utils.numpy_utils import float64_dtype, int64_dtype
 from .base import BasePipelineTestCase
 
 
-class WithAlias(object):
+@nottest
+class BaseAliasTestCase(BasePipelineTestCase):
 
     def test_alias(self):
         f = self.Term()
@@ -42,20 +44,20 @@ class WithAlias(object):
             )
 
 
-class TestFactorAlias(WithAlias, BasePipelineTestCase):
+class TestFactorAlias(BaseAliasTestCase):
     class Term(Factor):
         dtype = float64_dtype
         inputs = ()
         window_length = 0
 
 
-class TestFilterAlias(WithAlias, BasePipelineTestCase):
+class TestFilterAlias(BaseAliasTestCase):
     class Term(Filter):
         inputs = ()
         window_length = 0
 
 
-class TestClassifierAlias(WithAlias, BasePipelineTestCase):
+class TestClassifierAlias(BaseAliasTestCase):
     class Term(Classifier):
         dtype = int64_dtype
         inputs = ()
--- a/tests/pipeline/test_engine.py
+++ b/tests/pipeline/test_engine.py
@@ -33,7 +33,7 @@ from pandas import (
 from pandas.compat.chainmap import ChainMap
 from pandas.util.testing import assert_frame_equal
 from six import iteritems, itervalues
-from toolz import merge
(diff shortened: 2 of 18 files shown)


tests/pipeline/test_alias.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+
+from zipline.testing.predicates import assert_equal
+from zipline.pipeline import Classifier, Factor, Filter
+from zipline.utils.numpy_utils import float64_dtype, int64_dtype
+
+from .base import BasePipelineTestCase
+
+
+class WithAlias(object):
```
ssanderson: I would probably inherit from `BasePipelineTestCase` in `WithAlias`, since this is where we're actually defining the functionality that depends on the superclass.

llllllllll: This is not actually a test though, just a local fixture. That is why I chose the `With*` naming convention.

tests/pipeline/test_engine.py
```diff
@@ -1315,3 +1318,58 @@ def test_string_classifiers_produce_categoricals(self):
...
+                term,
+                full(
+                    (len(dates), len(assets)),
+                    precomputed_value,
+                    dtype=float64,
+                ),
+            )
+
+        def dispatcher(column):
+            if column is base_term:
```
ssanderson: `base_term` should never be loaded no matter what because it's a `Factor`, which isn't a `LoadableTerm`.  I think you want to be checking against `column` here.

tests/pipeline/test_engine.py
```diff
@@ -1315,3 +1318,58 @@ def test_string_classifiers_produce_categoricals(self):
             columns=self.asset_finder.retrieve_all(self.asset_finder.sids),
         )
         assert_frame_equal(result.c.unstack(), expected_final_result)
+
+
+class PopulateInitialWorkspaceTestCase(WithConstantInputs, ZiplineTestCase):
+    def test_populate_default_workspace(self):
+        column = USEquityPricing.low
+        base_term = column.latest
+        term = (base_term + 1).alias('term')
```
ssanderson: Can we call this something like `precomputed_term`?

tests/pipeline/test_engine.py
```diff
@@ -1315,3 +1318,58 @@ def test_string_classifiers_produce_categoricals(self):
...
+        composed_term = term + 1
```
ssanderson: Similarly, I'd call this something like `manually_computed_term`.

zipline/pipeline/engine.py
```diff
@@ -96,24 +104,39 @@ class SimplePipelineEngine(object):
     asset_finder : zipline.assets.AssetFinder
         An AssetFinder instance.  We depend on the AssetFinder to determine
         which assets are in the top-level universe at any point in time.
+    populate_initial_workspace : callable, optional
+        A function which will be used to populate the initial workspace when
+        computing a pipeline. This function will be passed the
+        initial_workspace, the root mask term, the execution_plan, the dates
+        being computed for, and the assets requested and should return a new
```
ssanderson: An easier way to document this might be to write a numpy-style docstring on `default_compute_initial_workspace`, and point the reader there for an example.

ssanderson: The docs for this should probably note that any pre-computed terms should have a shape given by `execution_plan.mask_and_dates_for_term`.

zipline/pipeline/graph.py
```diff
@@ -94,15 +93,25 @@ def outputs(self):
...
+        need to be computed.
         """
-        return iter(self._ordered)
+        return iter(topological_sort(
+            self.graph.subgraph(
+                {term for term, refcount in refcounts.items() if refcount > 0},
+            ),
+        ))
+
+    def ordered(self):
```
ssanderson: I think this is unused now.

zipline/pipeline/graph.py
```diff
@@ -132,15 +141,28 @@ def initial_refcounts(self, initial_terms):
...
 
         for t in initial_terms:
-            self.decref_dependencies(t, refcounts)
+            self._decref_recursive(t, refcounts, set())
 
         return refcounts
 
+    def _decref_recursive(self, term, refcounts, garbage):
+        """
+        Decrement terms recursivly to build the initial workspace.
```
ssanderson: This might benefit from a description of when this should be used versus `decref_dependencies`.  I think the distinction is that this should be used when computing initial refcounts on the graph from a set of precomputed terms, whereas `decref_dependencies` should be used during computation to decrement a term that was just computed.

zipline/pipeline/graph.py
```diff
@@ -132,15 +141,28 @@ def initial_refcounts(self, initial_terms):
...
         for t in self.outputs.values():
             refcounts[t] += 1
 
         for t in initial_terms:
-            self.decref_dependencies(t, refcounts)
+            self._decref_recursive(t, refcounts, set())
 
         return refcounts
 
+    def _decref_recursive(self, term, refcounts, garbage):
```
ssanderson: I'd probably name this `_decref_dependencies_recursive`, since it doesn't actually decref `term`.

zipline/pipeline/graph.py
```diff
@@ -159,14 +181,17 @@ def decref_dependencies(self, term, refcounts):
...
-        for parent, _ in self.in_edges([term]):
+        for parent, _ in self.graph.in_edges([term]):
             refcounts[parent] -= 1
             # No one else depends on this term. Remove it from the
             # workspace to conserve memory.
             if refcounts[parent] == 0:
                 garbage.add(parent)
         return garbage
 
+    def __iter__(self):
```
ssanderson: Do we depend on having an `__iter__` somewhere?

zipline/pipeline/graph.py
```diff
@@ -366,12 +391,49 @@ def extra_rows(self):
...
+        -------
+        mask : np.ndarray
+            The correct mask for this term.
+        dates : np.ndarray
+            The slice of dates for this term.
+        """
+        mask = term.mask
+        mask_offset = self.extra_rows[mask] - self.extra_rows[term]
+
+        # This offset is computed against _root_mask_term because that is what
```
ssanderson: This comment doesn't need an _ in the name anymore.

zipline/pipeline/mixins.py
```diff
@@ -20,6 +20,7 @@
 from zipline.utils.input_validation import expect_types
 from zipline.utils.sharedoc import (
     format_docstring,
+    PIPELINE_ALIAS_DOC,
```
ssanderson: Can we call this `PIPELINE_ALIAS_NAME_DOC`?  I read this name as suggesting that it holds the entire docstring for Alias, which makes it confusing when it's templated in for just the `name` parameter.

zipline/pipeline/classifiers/classifier.py
```diff
@@ -303,10 +305,29 @@ def postprocess(self, data):
...
         return data.as_categorical()
 
+    def to_workspace_value(self, result, assets):
+        """
+        Called with the result of a pipeline. This needs to return an object
+        which can be put into the workspace to continue doing computations.
+
+        This is the inverse of :func:`~zipline.pipeline.term.Term.postprocess`.
+        """
+        data = super(Classifier, self).unprocess(result, assets)
```
llllllllll: This was the old name for this method, I will add a test for this.

tests/pipeline/test_engine.py
```diff
@@ -1315,3 +1318,58 @@ def test_string_classifiers_produce_categoricals(self):
             columns=self.asset_finder.retrieve_all(self.asset_finder.sids),
         )
         assert_frame_equal(result.c.unstack(), expected_final_result)
+
+
+class PopulateInitialWorkspaceTestCase(WithConstantInputs, ZiplineTestCase):
```
ssanderson: It looks like we're only testing terms with `window_length`s of 0 right now.  Can we add a test for aliasing a windowed term?  An example of such a term would be an instance of `SimpleMovingAverage`.

tests/pipeline/test_alias.py
ssanderson: You can decorate with `nottest` to make nose not run the base class.  IMO, if you're defining tests methods in a class, you're no longer a fixture, you're a base class for other tests.

At the very least, I'd put an `init_class_fixtures` here asserting that any instantiated subclass of this is also a subclass of `BasePipelineTestCase`, since the tests here will fail in a non-obvious way if a consumer doesn't inherit from both.

ssanderson:
@llllllllll finished this round.  Mostly looks good to me.  Had a few small notes, mostly around documentation and name choices.

ssanderson:
Flake8 failing:
```
zipline/testing/predicates.py:494:1: E303 too many blank lines (3)
tests/pipeline/test_classifier.py:523:22: E225 missing whitespace around operator
tests/pipeline/test_engine.py:37:1: F401 'assoc' imported but unused
```

PR #1522, opened 2016-10-04
MAINT: Perspective offset for load adjustments.

ehebert:
Add a perspective offset to `AdjustedArrayWindow` and `AdjustedArray`,
so that `HistoryLoader` does not need to twiddle with offsets to support
viewing the data from the bar after end of the window, (Which is the
case when a '1d' history window is retrieved in minute mode, which is
explained in the docstring for `HistoryLoader.history`)

Presently, this simplifies the logic in
`HistoryLoader._get_adjustments_in_range`, and other incoming
AdjustmentReader's, (e.g. the roll based adjustment reader for continous
futures.) This patch should also make it easier for history and pipeline
to converge on a singular `load_adjustments` method.

### DIFF
--- a/tests/pipeline/test_adjusted_array.py
+++ b/tests/pipeline/test_adjusted_array.py
@@ -202,6 +202,85 @@ def _gen_multiplicative_adjustment_cases(dtype):
     )
 
 
+def _gen_multiplicative_adjustment_cases_with_perpsective_offset(dtype):
+    """
+    Generate expected moving windows on a buffer with adjustments.
+
+    We proceed by constructing, at each row, the view of the array we expect in
+    in all windows anchored on that row.
+
+    In general, if we have an adjustment to be applied once we process the row
+    at index N, should see that adjustment applied to the underlying buffer for
+    any window containing the row at index N - 1.
+
+    We then build all legal windows over these buffers.
+    """
+    adjustment_type = {
+        float64_dtype: Float64Multiply,
+    }[dtype]
+
+    nrows, ncols = 6, 3
+    adjustments = {}
+    buffer_as_of = [None] * 6
+    baseline = full((nrows, ncols), 1, dtype=dtype)
+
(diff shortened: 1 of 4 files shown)


ehebert:
cc: @ssanderson

Related to https://github.com/quantopian/zipline/pull/1519 (which depending on orders of merges would be changed to no longer subtract 1 from the key at which `Float64Adjustments` are written into the dict returned by `_get_adjustments_in_range`.

tests/pipeline/test_adjusted_array.py
```diff
@@ -202,6 +202,85 @@ def _gen_multiplicative_adjustment_cases(dtype):
     )
 
 
+def _gen_multiplicative_adjustment_cases_with_perpsective_offset(dtype):
```
ssanderson: typo: `perpsective`

zipline/lib/_windowtemplate.pxi
```diff
@@ -84,7 +86,9 @@ cdef class AdjustedArrayWindow:
         # Apply any adjustments that occured before our current anchor.
         # Equivalently, apply any adjustments known **on or before** the date
         # for which we're calculating a window.
-        while self.next_adj < anchor:
+        while (self.next_adj != -1
```
ssanderson: I think this might be more clearly expressed as `while self.next_adj < anchor + self.perspective_offset`.

ssanderson: (I think doing that would also remove the need to return -1 from `pop_next_adj` above.)

tests/pipeline/test_adjusted_array.py
ssanderson: As a reader, it's very hard to see how this is different from the other tests in the suite here.  I think it'd probably clearer and easier to just make all the tests in this suite check perspective offsets by adding that to `gen_expectations`.

PR #1523, opened 2016-10-04
MAINT: Pass through asset instead of sid.

ehebert:
When dispatching to sub readers in dispatch reader, pass along the asset
object, instead of extracting the sid.

The in development reader for continuous futures values besides `sid`
are needed from the `ContinuousFuture` object.


ehebert:
Uncovered while working on https://github.com/quantopian/zipline/tree/prototype-current-chain

PR #1525, opened 2016-10-05
Fix estimates overwrites bug

mtydykov:
Fixes a bug in the estimates loader that causes overwrites for a quarter to overwrite any other overwrites for a quarter that started on the same day. I.e., if 2 sids have quarters that start on the same day, the second sid's quarter overwrite will replace the first sid's quarter overwrite rather than appending the second overwrite to a list.

### DIFF
--- a/zipline/pipeline/loaders/earnings_estimates.py
+++ b/zipline/pipeline/loaders/earnings_estimates.py
@@ -280,6 +280,10 @@ class EarningsEstimatesLoader(PipelineLoader):
 
         sid_to_idx = dict(zip(assets, range(len(assets))))
 
+        for column in columns:
+            column_name = self.name_map[column.name]
+            col_to_overwrites[column_name] = defaultdict(list)
+
         def collect_adjustments(group):
             next_qtr_start_indices = dates.searchsorted(
                 group[EVENT_DATE_FIELD_NAME].values,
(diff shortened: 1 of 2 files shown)


zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -358,7 +362,7 @@ def create_overwrite_for_quarter(self,
             # overwrite all values going up to the starting index of
             # that quarter with estimates for that quarter.
             if requested_quarter in quarters_with_estimates_for_sid:
-                col_to_overwrites[column_name][next_qtr_start_idx] = [
+                col_to_overwrites[column_name][next_qtr_start_idx].extend([
```
llllllllll: `.append`

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -368,19 +372,19 @@ def create_overwrite_for_quarter(self,
...
                         sid_idx
                     ),
-                ]
+                ])
             # There are no estimates for the quarter. Overwrite all
             # values going up to the starting index of that quarter
             # with the missing value for this column.
             else:
-                col_to_overwrites[column_name][next_qtr_start_idx] = [
+                col_to_overwrites[column_name][next_qtr_start_idx].extend([
```
llllllllll: same here

PR #1526, opened 2016-10-05
MAINT: Use perspective_offset in more tests.

ssanderson:
- Refactor `test_adjusted_array` to test a range of perspective_offsets in
  all tests.

- Make perspective_offset a parameter to `AdjustedArray.traverse`
  instead of `AdjustedArray`.


ehebert:
Thanks @ssanderson, I am going to merge this in and then put a patch in that limits the `perspective_offset` to 1, deferring the work to make sure we include adjustments in between the end of the window and the vantage point.

PR #1527, opened 2016-10-05
BUG: Restrictions passed into HistoricalRestrictions not sorted correctly

lianga888:
Need to use toolz.groupby intead of itertools.groupby. itertools.groupby will only work if the elements to be grouped together are contiguous


ssanderson:
LGTM

PR #1529, opened 2016-10-07
ENH: Add continuous future current contract.

ehebert:
Add the ability for an algorithm to request the current contract for a
future chain via `data.current`.

e.g.:
```
data.current(ContinuousFuture('CL', offset=0, roll='calendar'),
'contract')
```

# Notes/Questions
- Since `ContinuousFuture` will also be used in `history` and other places where `Asset`s are used.
  The `ContinuousFuture` object provides `sid` and `is_exchange_open`.

   Ideally, there would be some base interface which would harden that duck typing. But would request that is something we do post making the other `ContinuousFuture` `data` methods work.

- The `sid` calculation for the `ContinuousFuture` is intended to be consistent across simulations. However, would not recommend using that `sid` as something that would never change. If and when we add more custom sids, the struct packing format may need to change to allow more flexibility.

- Choice of `OrderedContracts` structure was based on this notebook snippet on a chain of `254`.

```
In [86]
%%cython
from numpy cimport int64_t, intp_t
cimport cython
@cython.boundscheck(False)
@cython.wraparound(False)
cpdef intp_t get_auto_close_loc(int64_t[:] auto_close_dates, int64_t dt_value):
    cdef intp_t i
    cdef int64_t val
    for i in range(auto_close_dates.shape[0]):
        val = auto_close_dates[i]
        if val > dt_value:
            break
    return i

In [87]
%%timeit
get_auto_close_loc(auto_close_dates, dt_value)
The slowest run took 18.09 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 659 ns per loop

In [11]
%%timeit
np.searchsorted(start_dates, dt_value, side='right')
The slowest run took 18.61 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 910 ns per loop
```

### DIFF
--- a/zipline/assets/continuous_futures.pyx
+++ b/zipline/assets/continuous_futures.pyx
@@ -1,0 +1,296 @@
+# cython: embedsignature=True
+#
+# Copyright 2016 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Cythonized ContinuousFutures object.
+"""
+cimport cython
+from cpython.number cimport PyNumber_Index
+from cpython.object cimport (
+    Py_EQ,
+    Py_NE,
+    Py_GE,
+    Py_LE,
+    Py_GT,
+    Py_LT,
+)
+from cpython cimport bool
+
+import numpy as np
+from numpy cimport long_t, int64_t
+import warnings
+cimport numpy as np
+
+from zipline.utils.calendars import get_calendar
+
+
+cdef class ContinuousFuture:
+    """
+    Represents a specifier for a chain of future contracts, where the
+    coordinates for the chain are:
+    root_symbol : str
+        The root symbol of the contracts.
+    offset : int
+        The distance from the primary chain.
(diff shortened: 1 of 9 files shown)


zipline/data/data_portal.py
```diff
@@ -342,7 +356,9 @@ def _is_extra_source(asset, field, map):
         # at it if it's on something like palladium and not AAPL (since our
         # own price data always wins when dealing with assets).
 
-        return not (field in BASE_FIELDS and isinstance(asset, Asset))
+        return not (field in BASE_FIELDS and
+                    (isinstance(asset, Asset) or
+                     isinstance(asset, ContinuousFuture)))
```
richafrank: `isinstance(asset, (Asset, ContinuousFuture))`

zipline/assets/continuous_futures.pyx
```diff
@@ -0,0 +1,295 @@
...
+
+        Returns
+        -------
+        boolean: whether the asset's exchange is open at the given minute.
+        """
+        calendar = get_calendar(self.exchange)
+        return calendar.is_open_on_minute(dt_minute)
+
+
+cdef class OrderedContracts(object):
```
yankees714: Maybe just me, but confused me a bit since I first read this like "contracts that we have ordered"

zipline/assets/continuous_futures.pyx
```diff
@@ -0,0 +1,295 @@
...
+                                 self.roll_style,
+                                 self.exchange))
+
+    cpdef to_dict(self):
+        """
+        Convert to a python dict.
+        """
+        return {
+            'sid': self.sid,
+            'symbol': self.root_symbol,
```
ehebert: Yes, will fix.

zipline/assets/continuous_futures.pyx
ehebert: Definitely open to a better name.

ehebert: @yankees714 and I discussed `SortedContracts` as a possible candidate. May change that post merge.

PR #1530, opened 2016-10-07
ENH: Add `SpecificAssets` filter.

ssanderson:
Adds a filter that matches a set of assets.  Mainly useful for testing and debugging.

### DIFF
--- a/tests/pipeline/test_filter.py
+++ b/tests/pipeline/test_filter.py
@@ -26,11 +26,13 @@ from numpy import (
 from numpy.random import randn, seed as random_seed
 
 from zipline.errors import BadPercentileBounds
-from zipline.pipeline import Filter, Factor
+from zipline.pipeline import Filter, Factor, Pipeline
 from zipline.pipeline.classifiers import Classifier
 from zipline.pipeline.factors import CustomFactor
-from zipline.pipeline.filters import All, Any, AtLeastN
-from zipline.testing import parameter_space, permute_rows
+from zipline.pipeline.filters import All, Any, AtLeastN, SpecificAssets
+from zipline.testing import parameter_space, permute_rows, ZiplineTestCase
+from zipline.testing.fixtures import WithSeededRandomPipelineEngine
+from zipline.testing.predicates import assert_equal
 from zipline.utils.numpy_utils import float64_dtype, int64_dtype
 from .base import BasePipelineTestCase, with_default_shape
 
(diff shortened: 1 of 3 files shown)


zipline/pipeline/filters/filter.py
```diff
@@ -494,3 +496,22 @@ def _compute(self, arrays, dates, assets, mask):
...
+
+class SpecificAssets(Filter):
+    """
+    A Filter that computes True for a specific set of predetermined assets.
+    """
+    inputs = ()
+    window_length = 0
+    params = ('sids',)
+
+    @expect_types(assets=(list, tuple, np.ndarray))
```
llllllllll: how do we get an ndarray, it looks like we will coerce it to a `list` in the coerce below.

zipline/pipeline/filters/filter.py
```diff
@@ -494,3 +496,22 @@ def _compute(self, arrays, dates, assets, mask):
...
+    @coerce_types(assets=((list, np.ndarray, pd.Series), list))
+    def __new__(cls, assets):
+        sids = frozenset(asset.sid for asset in assets)
```
llllllllll: why do we need to coerce this to a list to iterate over it?

tests/pipeline/test_filter.py
```diff
@@ -820,3 +822,40 @@ def test_top_and_bottom_with_groupby_and_mask(self, dtype, seed):
             },
             mask=self.build_mask(permute(rot90(self.eye_mask(shape=shape)))),
         )
+
+
+class SpecificAssetsTestCase(WithSeededRandomPipelineEngine,
```
llllllllll: why does this need a seeded random loader, we are not loading any values, just sids. Can we just create an engine with an exploading object for a loader?

ssanderson: We could, but setting up the engine with sids and trading days is a little bit boilerplate-ey.  I tend to use `WithSeededRandomPipelineEngine` when I just need **an** engine, but I don't care about the specific values it emits.

zipline/pipeline/filters/filter.py
ssanderson: Removed the coercions here.

justnpT:
I would vote for adding docs about what specific assets are

ssanderson:
@justnpT I added another sentence to the docstring explaining when you would use `SpecificAssets`: https://github.com/quantopian/zipline/pull/1530/files#diff-9bcf584f1d4b53abe9f16f81e5ee0716R503.

PR #1533, opened 2016-10-11
ENH: Add current chain for continuous futures.

ehebert:
Add `chain`field to current, as well as supporting methods in DataPortal
and OrderedContracts.

Enables the following example:

```
from zipline.api import continuous_future

def initialize(context):
    context.primary_cl = continuous_future('CL', offset=0, roll='calendar')
    schedule_function(print_current_chain)

def print_current_chain(context, data):
    chain = data.current(context.primary_cl, 'chain')
    print 'datetime={0}'.format(get_datetime())
    print 'primary={0}'.format(chain[0])
    print 'secondary={0}'.format(chain[1])
    print 'tertiary={0}'.format(chain[2])
```

```
datetime=2015-12-23 14:31:00+00:00
primary=Future(1058201602 [CLG16])
secondary=Future(1058201603 [CLH16])
tertiary=Future(1058201604 [CLJ16])
```

Also, make return types of OrderedContracts methods compatible across
architectures. (Noticed while adding `active_chain` method.)

### DIFF
--- a/tests/test_continuous_futures.py
+++ b/tests/test_continuous_futures.py
@@ -15,10 +15,12 @@
 
 from textwrap import dedent
 
+from numpy import array, int64
 import pandas as pd
 from pandas import Timestamp, DataFrame
 
 from zipline import TradingAlgorithm
+from zipline.assets.continuous_futures import OrderedContracts
 from zipline.testing.fixtures import (
     WithCreateBarData,
     WithSimParams,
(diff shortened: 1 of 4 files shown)


tests/test_continuous_futures.py
```diff
@@ -49,27 +49,37 @@ def make_root_symbols_info(self):
     @classmethod
     def make_futures_info(self):
         return DataFrame({
-            'symbol': ['FOF', 'FOG', 'FOH'],
-            'root_symbol': ['FO', 'FO', 'FO'],
-            'asset_name': ['Foo'] * 3,
+            'symbol': ['FO16F', 'FO16G', 'FO16H', 'FO16J', 'FO22F'],
```
yankees714: Our symbols would be formatted like 'FOF16', etc., might be worth changing to avoid confusion

tests/test_continuous_futures.py
```diff
@@ -145,28 +155,132 @@ def record_current_contract(algo, data):
...
+        self.assertEqual(result.primary_len,
+                         4,
+                         'There should be only 4 contracts in the chain for '
+                         'the primary, there are 5 contracts defined in the '
+                         'fixture, but one has a start after the simulation '
+                         'date.')
+        self.assertEqual(result.secondary_len,
+                         3,
+                         'There should be only 4 contracts in the chain for '
+                         'the primary, there are 5 contracts defined in the '
```
yankees714: I believe this should say "3 contracts" for the "secondary"

zipline/assets/continuous_futures.pyx
```diff
@@ -294,3 +294,22 @@ cdef class OrderedContracts(object):
...
+        start_dates = self.start_dates
+
+        for i in range(self._size):
+            if starting_sid == sids[i]:
+                left = i
+                break
+
+        for j in range(i, self._size):
+            if start_dates[j] > dt_value:
+                right = j
```
yankees714: Does this handle the case of all start dates coming before the current session?

zipline/data/data_portal.py
```diff
@@ -1239,6 +1239,29 @@ def get_simple_transform(self, asset, transform_name, dt, data_frequency,
 
             return ret
 
+    def get_current_future_chain(self, continuous_future, dt):
+        """
+        Retrieves the future chain for the contract at the given `dt` according
+        the `continuous_future` specification.
+
+        Returns:
+        future_chain : [Future]
```
ehebert: Had meant to follow the formatting of methods like `BcolzDailyBarWriter.write_csv`, e.g. `asset_map : dict[int -> str]`, had left off the return type though.

Will change to to `list[Future]`.

tests/test_continuous_futures.py
ehebert: Will change.

tests/test_continuous_futures.py
ehebert: Correct. Fixing.

zipline/assets/continuous_futures.pyx
ehebert: It does not. With how the function is currently used, I do not think that would occur. But will add the check of `if left > right` and handle that case.

ehebert: You are right, if we are backtesting near the end of the historical range. That case would occur. Need to set the `right` value to the last `j` in that case.

zipline/data/data_portal.py
```diff
@@ -1239,6 +1239,29 @@ def get_simple_transform(self, asset, transform_name, dt, data_frequency,
...
+        """
+        rf = self._roll_finders[continuous_future.roll_style]
+        session = self.trading_calendar.minute_to_session_label(dt)
+        contract_center = rf.get_contract_center(
+            continuous_future.root_symbol, session,
+            continuous_future.offset)
+        oc = self.asset_finder.get_ordered_contracts(
+            continuous_future.root_symbol)
+        chain = oc.active_chain(contract_center, session.value)
+        result = [self.asset_finder.retrieve_asset(sid)
```
richafrank: Would `self.asset_finder.retrieve_futures_contracts(chain)` (or `self.asset_finder.retrieve_all(chain)`) do a batch query instead of one per `sid`?

ehebert: Yes, good point. Will use `retrieve_all` instead.

```
diff --git a/zipline/data/data_portal.py b/zipline/data/data_portal.py
index aa4c070..58fdc91 100644
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -1258,9 +1258,7 @@ class DataPortal(object):
         oc = self.asset_finder.get_ordered_contracts(
             continuous_future.root_symbol)
         chain = oc.active_chain(contract_center, session.value)
-        result = [self.asset_finder.retrieve_asset(sid)
-                  for sid in chain]
-        return result
+        return self.asset_finder.retrieve_all(chain)

     def _get_current_contract(self, continuous_future, dt):
         rf = self._roll_finders[continuous_future.roll_style]
```

tests/test_continuous_futures.py
```diff
@@ -145,28 +157,197 @@ def record_current_contract(algo, data):
...
+        contract_sids = array([1, 2, 3, 4], dtype=int64)
+        start_dates = pd.date_range('2015-01-01', periods=4, tz="UTC")
+        auto_close_dates = pd.date_range('2016-04-01', periods=4, tz="UTC")
+
+        oc = OrderedContracts('FO',
+                              contract_sids,
+                              start_dates.astype('int64'),
+                              auto_close_dates.astype('int64'))
+
+        # Test sid 1 as days increment, as the sessions march forward more
```
yankees714: Think the "more" is a typo

tests/test_continuous_futures.py
```diff
@@ -145,28 +157,197 @@ def record_current_contract(algo, data):
...
+        chain = oc.active_chain(1, pd.Timestamp('2015-01-04', tz='UTC').value)
+        self.assertEquals(4, len(chain),
+                          "[1, 2, 3, 4] should be the active contracts on "
+                          "01-04, this is all defined contracts in the test "
+                          "case.")
+
+        chain = oc.active_chain(1, pd.Timestamp('2015-01-05', tz='UTC').value)
+        self.assertEquals(4, len(chain),
+                          "[1, 2, 3, 4] should be the active contracts on "
+                          "01-05. This test the case where all start dates "
```
yankees714: "This tests the case"

ehebert: Fixed this and other typo, will merge once CI passes.

PR #1543, opened 2016-10-17
BUG: load extensions in IPython magic

llllllllll:
fixes: https://github.com/quantopian/zipline/issues/1542

### DIFF
--- a/zipline/__main__.py
+++ b/zipline/__main__.py
@@ -251,6 +251,12 @@ def run(ctx,
 def zipline_magic(line, cell=None):
     """The zipline IPython cell magic.
     """
+    load_extensions(
+        default=True,
+        extensions=[],
+        strict=True,
+        environ=os.environ,
+    )
     try:
         return run.main(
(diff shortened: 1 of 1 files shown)


zipline/__main__.py
```diff
@@ -251,6 +251,12 @@ def run(ctx,
...
     """The zipline IPython cell magic.
     """
+    load_extensions(
+        default=True,
+        extensions=[],
+        strict=True,
+        environ=os.environ,
+    )
     try:
         return run.main(
```
richafrank: Since it's decorated with `main.command()`, does `run` not do this for us?

llllllllll: It does not appear to get called, I think this would work if I did `main.main`, but I don't want users to need to pass the subcommands.

llllllllll: I am going to merge this as is because I don't think I have time to work out a better solution for consistently setting this up. We can always clean this up in the future but this fixes our immediate issue with extensions in the notebook

PR #1547, opened 2016-10-19
BUG: create adjustments based on ordered assets instead of set

mtydykov:
This fixes a bug where we create adjustments for estimates based on where assets happen to fall in a set rather than using ordered assets. Not sure if this necessitates a test or what kind.

### DIFF
--- a/tests/pipeline/test_quarters_estimates.py
+++ b/tests/pipeline/test_quarters_estimates.py
@@ -825,7 +825,7 @@ class WithEstimateWindows(WithEstimates):
             SID_FIELD_NAME: 0,
         })
 
-        sid_1_timeline = pd.DataFrame({
+        sid_10_timeline = pd.DataFrame({
             TS_FIELD_NAME: [pd.Timestamp('2015-01-09'),
                             pd.Timestamp('2015-01-12'),
                             pd.Timestamp('2015-01-09'),
@@ -836,12 +836,12 @@ class WithEstimateWindows(WithEstimates):
             'estimate': [110., 111.] + [310., 311.],
             FISCAL_QUARTER_FIELD_NAME: [1] * 2 + [3] * 2,
             FISCAL_YEAR_FIELD_NAME: 2015,
-            SID_FIELD_NAME: 1
+            SID_FIELD_NAME: 10
         })
 
         # Extra sid to make sure we have correct overwrites when sid quarter
         # boundaries collide.
-        sid_3_timeline = pd.DataFrame({
+        sid_20_timeline = pd.DataFrame({
             TS_FIELD_NAME: [pd.Timestamp('2015-01-05'),
                             pd.Timestamp('2015-01-07'),
                             pd.Timestamp('2015-01-05'),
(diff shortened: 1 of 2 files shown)


llllllllll:
Why does the order matter here. I think a test would help me understand how this change affects the results, and help us ensure that we don't break this later.

mtydykov:
Order matters because when we create adjustments, we do the following to figure out at what column the adjustment should be applied:
```
sid_to_idx = dict(zip(assets, range(len(assets))))
```

Because we were passing in a _set_ of assets that are used, if the resulting map happened to put the sids out of order, adjustments got misaligned with the base data later.

So actually, I don't have it right in the fix because I'm passing an ordered subset of the assets (just the assets we use). The indexes we use to create adjustments need to line up with the entire set of assets. I just updated the code with the correct fix.

I agree, I just can't think of any kind of test for this other than figuring out in what way we can create a set out of a list of sids so that they end up out of order.

tests/pipeline/test_quarters_estimates.py
```diff
@@ -854,9 +854,21 @@ def make_events(cls):
...
             FISCAL_YEAR_FIELD_NAME: 2015,
-            SID_FIELD_NAME: 2
+            SID_FIELD_NAME: 20
         })
-        return pd.concat([sid_0_timeline, sid_1_timeline, sid_3_timeline])
+        concatted = pd.concat([sid_0_timeline,
+                               sid_10_timeline,
+                               sid_20_timeline]).reset_index()
+
+        return concatted.reindex(np.random.permutation(concatted.index))
```
llllllllll: we shouldn't use unseeded random in a test

tests/pipeline/test_quarters_estimates.py
```diff
@@ -888,6 +900,12 @@ def create_expected_df(cls, tuples, end_date):
         df.index = df.index.rename('knowledge_date')
         df['at_date'] = end_date.tz_localize('utc')
         df = df.set_index(['at_date', df.index.tz_localize('utc')]).ffill()
+        new_sids = set(cls.get_sids()) - set(df.columns)
+        df = pd.concat([df, pd.DataFrame(columns=new_sids)])
+        # Have to explicitly cast to float so that nans are of type
+        # np.float64 - otherwise they won't compare equal.
+        for col in new_sids:
```
llllllllll: can you explain why we are doing this?

llllllllll: I understand the mechanics of what you are doing, I don't understand why this is needed at all.

mtydykov: Still unsure of what you mean. Do you mean why we need sids in `assets` that aren't used in the raw data?

llllllllll: I think all of this can be replaced with `df = df.reindex(df.columns.union(new_sids))`

mtydykov: I had tried doing that before, but it doesn't work for the reason I explained in the comment. Here's what happens:

```
(Pdb++) type(estimate[0][0])
<type 'numpy.float64'>
(Pdb++) type(today_timeline[timeline_start_idx:][0][0])
<type 'float'>
(Pdb++) assert_equal(estimate,today_timeline[timeline_start_idx:])
*** AssertionError:
Arrays are not equal

(mismatch 100.0%)
 x: array([[ nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,
         nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan],
       [ nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,...
 y: array([[nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
        nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,...
```

llllllllll: I don't understand why the types are off:
```
In [1]: df = pd.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]}, dtype='f8')

In [2]: df.reindex(columns=['a', 'b', 'c'])
Out[2]: 
     a    b   c
0  1.0  4.0 NaN
1  2.0  5.0 NaN
2  3.0  6.0 NaN

In [3]: _.dtypes
Out[3]: 
a    float64
b    float64
c    float64
dtype: object
```
also, in the astype below, use `float64` or `float32` to be explicit.

mtydykov: Updated the cast

PR #1549, opened 2016-10-20
PERF: Speedup minute to session sampling.

ehebert:
The minute to session sampling reading was creating two DataFrame
objects, the first to hold the minute data, and then a second returned
by the `DataFrame.groupby` to sample down to sessions.

Instead use the arrays returned by the minute readers `load_raw_arrays`
and implement sampling logic which takes advantage that the minutes
being passed start with the first minute of the first session and end
with the last minute of the last session.

On my machine this takes the tests in `test/test_continuous_futures`
from ~4.0 to about ~0.1 seconds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,6 +103,10 @@ ext_modules = [
         'zipline.utils.calendars._calendar_helpers',
         ['zipline/utils/calendars/_calendar_helpers.pyx']
     ),
+    Extension(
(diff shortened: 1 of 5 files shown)


ehebert:
This can be further improved by using Cython to get direct access to the numpy arrays. However this  this already has an order of magnitude speed up, so starting with this change.

PR #1555, opened 2016-10-25
MAINT: Return nan from daily bcolz get_value.

ehebert:
Match the behavior of the minute bar reader, now that the session and
minute bar readers share a common interface.

isnull is slightly slower than checking against -1; however, n cases
where we check against illiquid trades in a tight loop, volume is
checked which is not using nan. The change here should be marginal with
regards to performance.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -19,6 +19,7 @@ from logbook import Logger
 import numpy as np
 from numpy import float64, int64
 import pandas as pd
+from pandas import isnull
 from pandas.tslib import normalize_date
 from six import iteritems
 from six.moves import reduce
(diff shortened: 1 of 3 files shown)


zipline/data/data_portal.py
```diff
@@ -649,7 +643,7 @@ def _get_daily_data(self, asset, column, dt):
                     value = reader.get_value(
                         asset, found_dt, "close"
                     )
-                    if value != -1:
+                    if not isnull(value):
```
yankees714: There's also `pd.notnull`

yankees714:
This should take care of https://github.com/quantopian/qexec/issues/11328.

zipline/data/data_portal.py
ehebert: Ah, forgot about that one. Will switch in a followup patch.

PR #1557, opened 2016-10-26
Split adjust estimates

mtydykov:
Adds logic that split adjusts estimates. Also adds relevant tests.

### DIFF
--- a/zipline/pipeline/loaders/blaze/estimates.py
+++ b/zipline/pipeline/loaders/blaze/estimates.py
@@ -17,7 +17,8 @@ from zipline.pipeline.loaders.earnings_estimates import (
     PreviousEarningsEstimatesLoader,
     required_estimates_fields,
     metadata_columns,
-)
+    PreviousSplitAdjustedEarningsEstimatesLoader,
+    NextSplitAdjustedEarningsEstimatesLoader)
 from zipline.pipeline.loaders.utils import (
     check_data_query_args,
 )
@@ -108,6 +109,7 @@ class BlazeEstimatesLoader(PipelineLoader):
         # Only load requested columns.
         requested_column_names = [self._columns[column.name]
                                   for column in columns]
+
         raw = load_raw_data(
             assets,
             dates,
@@ -120,7 +122,7 @@ class BlazeEstimatesLoader(PipelineLoader):
 
         return self.loader(
             raw,
-            {column.name: self._columns[column.name] for column in columns}
+            {column.name: self._columns[column.name] for column in columns},
         ).load_adjusted_array(
             columns,
             dates,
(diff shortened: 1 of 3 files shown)


zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -542,6 +699,332 @@ def get_last_data_per_qtr(self, assets_with_data, columns, dates):
...
+            The index of `sid` in the adjusted array.
+        requested_split_adjusted_columns : list of str
+            The requested split adjusted columns.
+        Returns
+        -------
+        col_to_split_adjustments : dict[str -> dict[int -> list of Adjustment]]
+            The adjustments for this sid that occurred after the
+            split-asof-date.
+        """
+        col_to_split_adjustments = defaultdict(dict)
```
ehebert: My recommendation would be to not use `defaultdict`.
It is a very powerful too, but has historically been the root cause for many bugs and/or memory leaks.

Instead of:
```
for column_name in requested_split_adjusted_columns:
                            col_to_split_adjustments[
                                column_name
                            ][start_idx].append(
                                Float64Multiply(
                                    # Always apply from first day of qtr
                                    qtr_range[0],
                                    end_idx,
                                    sid_idx,
                                    sid_idx,
                                    adjustment
                                )
                            )
```
The append could be changed to the following:
```
for column_name in requested_split_adjusted_columns:
    adj = Float64Multiply(
      # Always apply from first day of qtr
      qtr_range[0],
      end_idx,
      sid_idx,
      sid_idx,
      adjustment)
   split_adjustments = col_to_split_adjustments[column_name]
   try:
        adjs_at_index[start_index].append(adj)
   except KeyError:
        adjs_at_index[start_index] = [adj]
```

By avoiding `defaultdict` a scenario is avoided where a consumer is iterating over candidate `start_idx` values and quietly creating an empty list for each value, when a KeyError should be raised.

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -82,7 +82,10 @@ def __init__(self,
                  odo_kwargs=None,
                  data_query_time=None,
                  data_query_tz=None,
-                 checkpoints=None):
+                 checkpoints=None,
+                 split_adjustments_loader=None,
```
ehebert: Thoughts on calling this `adjustment_reader`, `adjusted_column_names`, `adjusted_asof`?
Then specifying which adjustments should be applied? e.g. a vendor may also apply dividends to historical data.

zipline/pipeline/loaders/blaze/estimates.py
```diff
@@ -103,11 +106,21 @@ def __init__(self,
...
         # Only load requested columns.
         requested_column_names = [self._columns[column.name]
                                   for column in columns]
+        requested_spilt_adjusted_columns = None
+        if self._split_adjusted_column_names:
+            requested_spilt_adjusted_columns = [
+                column_name
+                for column_name in self._split_adjusted_column_names
+                if column_name in requested_column_names
+            ]
```
ehebert: What should happen if there is no `self._split_adjusted_column_names`? Is `None` a valid value for `requested_splite_adjusted_columns`?

mtydykov: Yes, it's a valid value, as long as `split_adjustments_loader` and `split_adjusted_asof` are also both `None`. That just means that no split adjustments should be applied to the given data. However, it's not valid for only a subset of those 3 fields to have a value of `None`. The method `validate_split_adjustments_args` in `pipeline.loaders.earnings_estimates.EarningsEstimatesLoader` checks for those cases.

zipline/pipeline/loaders/blaze/estimates.py
mtydykov: Definitely agreed on `reader` instead of `loader`. As for generalizing from splits to adjustments: previously, we had only talked about applying split adjustments since our plan is to use this loader for earnings estimates, and EPS columns would be affected by splits but probably not by dividends. I guess it depends on whether or not we think we might use this loader for dividend-adjusted data. @jmccorriston

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -236,13 +253,185 @@ def get_requested_quarter_data(self,
...
+        dates : pd.DatetimeIndex
+            The calendar dates for which estimates data is requested.
+        requested_qtr_data : pd.DataFrame
+            The DataFrame with the latest values for the requested quarter
+            for all columns.
+        last_per_qtr : pd.DataFrame
+            A DataFrame with a column MultiIndex of [self.estimates.columns,
+            normalized_quarters, sid] that allows easily getting the timeline
+            of estimates for a particular sid for a particular quarter.
+        sid_to_idx : dict[int -> int]
```
ehebert: Not a major issue (and wouldn't block on it) but at this level, I think passing the `idx` directly makes sense. Leaving the lookup to scope where this method is called.

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -343,52 +515,51 @@ def create_overwrite_for_quarter(self,
         quarters_with_estimates_for_sid : np.array
             An array of all quarters for which there are estimates for the
             given sid.
+        requested_quarter : float
```
ehebert: Why a `float` and not an `int` here?

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -651,3 +825,700 @@ def get_zeroth_quarter_idx(self, stacked_last_per_qtr):
...
+            The sid for which we want to retrieve adjustments.
+        split_adjusted_asof_idx : int
+            The index in `dates` as-of which the data is split adjusted.
+
+        Returns
+        -------
+        pre_adjustments : tuple(list(float), list(int), pd.DatetimeIndex)
+            The adjustment values and indexes in `dates` for
+            adjustments that happened before the split-asof-date.
+        pre_adjustments : tuple(list(float), list(int), pd.DatetimeIndex)
```
ehebert: Should this be `post_adjustments`?

zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -651,3 +825,700 @@ def get_zeroth_quarter_idx(self, stacked_last_per_qtr):
...
+            The adjustment values, indexes in `dates`, and timestamps for
+            adjustments that happened after the split-asof-date.
+        """
+        adjustments = self._split_adjustments.get_adjustments_for_sid(
+            'splits', sid
+        )
+        sorted(adjustments, key=lambda adj: adj[0])
```
ehebert: I believe `get_adjustments_for_sid` always returns in order (but may not be explicitly defined in the adjustment reader interface). However, since any given amount of splits is small, ensuring that they are in order here should be ok.

zipline/pipeline/loaders/earnings_estimates.py
mtydykov: Fixed.

zipline/pipeline/loaders/earnings_estimates.py
ehebert: I don't think it is necessary, was just surprised about the type. Thanks for explanation.

PR #1559, opened 2016-10-26
Micro optimizations

ssanderson:


zipline/_protocol.pyx
```diff
@@ -378,15 +378,15 @@ cdef class BarData:
...
-                                    field,
-                                    self._get_current_minute(),
-                                    self.data_frequency
-                               )
-                        for asset in assets
-                        }, index=assets, name=fields)
+                    ret = self.data_portal.get_history_window(
+                        assets,
+                        self._get_current_minute(),
+                        1,
```
ehebert: We could also add a corresponding check in `_ensure_sliding_windows` which skips `load_adjustments` if the window size is 1 and `is_perspective_after` is `False`.

ssanderson:
Closing in favor of https://github.com/quantopian/zipline/pull/1559, which removes the optimization around using `history(1)` for `data.current`.

PR #1560, opened 2016-10-26
DEV delete old *.c and *.so files with rebuild-cython.sh for Darwin/OSX

prsutherland:
OSX's find supports a different syntax for `-regex`. As a result the original regex doesn't match any files and successfully does nothing.

I've confirmed that this change works with and without `brew install findutils --with-default-names`


llllllllll:
lgtm, merging on pass

PR #1561, opened 2016-10-27
Micro optimizations 2

ssanderson:


zipline/lib/_windowtemplate.pxi
```diff
@@ -82,54 +85,61 @@ cdef class AdjustedArrayWindow:
...
+        cdef:
+            Py_ssize_t anchor = self.anchor
+
+        if target_anchor < anchor:
+            raise Exception('Can not access data after window has passed.')
+
+        if target_anchor == anchor:
+            return self.output
+
+        self._tick_forward(target_anchor - anchor)
```
ehebert: Should `Exhausted` be handled here? Or is the intent to have that raise for the consumer of `seek` to handle?

ssanderson: nvm, I misread my own code.  It's an error for the caller to try to seek past the end of the array.  Currently that will fail with a `StopIteration`, which is actually probably bad because that can be caught and misinterpreted if `seek` is called inside a generator.  I don't have a strong opinion on whether it's worthwhile to try to catch and re-raise `Exhausted` here with a more descriptive error.  Given that this is in performance-critical code, I'm inclined to say that this is fine as-is.

PR #1567, opened 2016-10-28
MAINT: Restore @property decorator

keang:
This will keep `opens`, `closes`, `early_closes`, etc to the
same pattern.


richafrank:
Good catch!  Thanks @keang .

PR #1568, opened 2016-10-28
Fix microoptimizations

ssanderson:


zipline/data/data_portal.py
```diff
@@ -703,24 +703,27 @@ def _get_history_daily_window(self, assets, end_dt, bar_count,
...
 
-    def _get_history_daily_window_data(
-            self, assets, days_for_window, end_dt, field_to_use):
-        ends_at_midnight = end_dt.hour == 0 and end_dt.minute == 0
+    def _get_history_daily_window_data(self,
+                                       assets,
+                                       days_for_window,
+                                       end_dt,
+                                       field_to_use):
+        ends_at_midnight = (0 == end_dt.hour == end_dt.minute)
```
llllllllll: style nit: could we move `0` to the far rhs?

zipline/testing/core.py
```diff
@@ -748,6 +748,9 @@ def get_spot_value(self, asset, field, dt, data_frequency):
         # otherwise just return a fixed value
         return int(asset)
 
+    # XXX: These aren't actually the methods that are used by the superclasses,
+    # so these don't do anything, and this class will likely produce unexpected
+    # results for history().
     def _get_daily_window_for_sid(self, asset, field, days_in_window,
```
llllllllll: why not just remove this method if it isn't used?

ssanderson: I figured the original intent here would be more helpful for whoever eventually gets bitten by this.  I just happened to notice that these aren't used because I was looking at renaming methods in data portal.

llllllllll: kk

PR #1574, opened 2016-11-01
ENH: Mark statistical terms as window_safe.

ssanderson:


ssanderson:
> That sounds reasonable.

- @dmichalowicz

PR #1577, opened 2016-11-03
MAINT: Just take a PipelineEngine in TradingAlgorithm.

ssanderson:
Instead of a function that produces a loader, just let the user pass a
PipelineEngine.


ssanderson:
This is unlikely to be merged in its current form.

PR #1582, opened 2016-11-04
BUG: Allows 'contract' in get_spot_value with daily frequency

yankees714:


ehebert:
We should add a test to `test_continuous_future` which covers using this with daily.
To do so, we may need to change the `create_bardata` method to accept an optional data frequency.

This change works, (and not blocking on this question), but I wonder if we can make all the conditionals on `data_frequency` and `field` etc. a bit cleaner with some dispatching/routing machinery.

PR #1584, opened 2016-11-07
BUG: Fix `data.history` with a single continuous future.

ehebert:
Allow `ContinuousFuture` when checking for a single "asset".

This could be further improved by:

- Defininng a tuple of the `Asset`-like types OR making
`ContinuousFuture` and `Asset` share a common type (whether that is
`ContinuousFuture` inheriting from `Asset` or making `Asset` and
`ContinuousFuture` share a common type.)

- Make a `history` test which uses `BarData` + `ContinuousFuture`,
instead of just using the history loader directly from tests.

### DIFF
--- a/zipline/_protocol.pyx
+++ b/zipline/_protocol.pyx
@@ -24,7 +24,7 @@ from six import iteritems, PY2, string_types
 from cpython cimport bool
 from collections import Iterable
 
-from zipline.assets import Asset, Future
+from zipline.assets import Asset, AssetConvertible, Future
 from zipline.assets.continuous_futures import ContinuousFuture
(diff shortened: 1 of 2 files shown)


richafrank:
Could also register with an ABC, like we do with `AssetConvertible`.

yankees714:
I like the ABC idea, looks pretty doable here.

ehebert:
Good idea. I'll patch that in.

ehebert:
Made a mistake in merging this. Had made a change locally to use an abc of `PricingDataAssociable` instead. Fixing now.

ehebert:
https://github.com/quantopian/zipline/pull/1585

PR #1587, opened 2016-11-09
BUG: Fix continuous future end dates.

ehebert:
The end date of the last contract with a sufficient start date was being
used for the continuous future overall end date; however the end date of
that contract (which is the last day for which there is data for the
contract) is not necessarily the greatest end date out of all contracts.
It is possible for the furthest out contract to have some, but very
few, trades before it is more actively traded. Which would give it a
start date within in the range of the simulation, but an end date is
earlier than the other contracts which are active during the simulation.

This bug would result in `nan`s when getting the current price because
of the `end_date` check in `get_spot_value`. When the current simulation
time was greater than the `end_date` of the last contract the condition
which guards against attempting to get data for an instrument past its
end date would return a `nan`, even when the current underlying contract
did have data for that date.

Use max end date of all contracts instead of the last one, to ensure
that the continuous future last date is always great enough to allow
access to all contracts with in the chain.

Also, use min start date to accurately mirror the end date behavior.

### DIFF
--- a/tests/test_continuous_futures.py
+++ b/tests/test_continuous_futures.py
@@ -64,42 +64,54 @@ class ContinuousFuturesTestCase(WithCreateBarData,
     @classmethod
     def make_futures_info(self):
         return DataFrame({
-            'symbol': ['FOF16', 'FOG16', 'FOH16', 'FOJ16', 'FOK16', 'FOF22'],
-            'root_symbol': ['FO'] * 6,
-            'asset_name': ['Foo'] * 6,
+            'symbol': ['FOF16', 'FOG16', 'FOH16', 'FOJ16', 'FOK16', 'FOF22',
+                       'FOG22'],
+            'root_symbol': ['FO'] * 7,
+            'asset_name': ['Foo'] * 7,
             'start_date': [Timestamp('2015-01-05', tz='UTC'),
                            Timestamp('2015-02-05', tz='UTC'),
                            Timestamp('2015-03-05', tz='UTC'),
                            Timestamp('2015-04-05', tz='UTC'),
                            Timestamp('2015-05-05', tz='UTC'),
-                           Timestamp('2021-01-05', tz='UTC')],
+                           Timestamp('2021-01-05', tz='UTC'),
+                           Timestamp('2015-01-05', tz='UTC')],
             'end_date': [Timestamp('2016-08-19', tz='UTC'),
                          Timestamp('2016-09-19', tz='UTC'),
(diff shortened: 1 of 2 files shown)


zipline/assets/assets.py
```diff
@@ -922,8 +922,9 @@ def get_ordered_contracts(self, root_symbol):
 
     def create_continuous_future(self, root_symbol, offset, roll_style):
         oc = self.get_ordered_contracts(root_symbol)
-        start_date = self.retrieve_asset(oc.contract_sids[0]).start_date
-        end_date = self.retrieve_asset(oc.contract_sids[-1]).end_date
+        start_date = min(a.end_date for a in self.retrieve_all(
+            oc.contract_sids))
+        end_date = max(a.end_date for a in self.retrieve_all(oc.contract_sids))
```
yankees714: Can we just do one `self.retrieve_all(oc.contract_sids)` call for both of these?

zipline/assets/assets.py
```diff
@@ -922,8 +922,9 @@ def get_ordered_contracts(self, root_symbol):
 
     def create_continuous_future(self, root_symbol, offset, roll_style):
         oc = self.get_ordered_contracts(root_symbol)
-        start_date = self.retrieve_asset(oc.contract_sids[0]).start_date
-        end_date = self.retrieve_asset(oc.contract_sids[-1]).end_date
+        start_date = min(a.end_date for a in self.retrieve_all(
```
yankees714: Think this should be `a.start_date`

tests/test_continuous_futures.py
```diff
@@ -270,6 +282,20 @@ def test_get_value_close_daily(self):
...
+        value = self.data_portal.get_spot_value(
+            cf_primary,
+            'close',
+            pd.Timestamp('2016-03-26', tz='UTC'),
+            'daily',
+        )
+
+        self.assertEqual(value, 135441.44,
+                         'Auto close at beginning of session so FOG16 is now '
+                         'the current contract.')
```
yankees714: Does this failure message still make sense here?

zipline/assets/assets.py
ehebert: Good call, will change.

tests/test_continuous_futures.py
ehebert: No, will make it correspond with the tested dates.

zipline/assets/assets.py
ehebert: Also, adding a test to cover this, since this shows a lack of coverage on my part.

PR #1588, opened 2016-11-09
ENH: Add MACD, MA, and AnnVol as built in factors

ahgnaw:
These factors will be used in Quabbin and might be useful to everyone in the community.

### DIFF
--- a/tests/pipeline/test_technical.py
+++ b/tests/pipeline/test_technical.py
@@ -5,6 +5,7 @@ from six.moves import range
 import numpy as np
 import pandas as pd
 import talib
+from numpy.random import random_integers
 
 from zipline.lib.adjusted_array import AdjustedArray
 from zipline.pipeline.data import USEquityPricing
@@ -16,6 +17,7 @@ from zipline.pipeline.factors import (
     LinearWeightedMovingAverage,
     RateOfChangePercentage,
     TrueRange,
+    MovingAverageConvergenceDivergence
 )
 from zipline.testing import parameter_space
 from zipline.testing.fixtures import ZiplineTestCase
@@ -403,3 +405,78 @@ class TestTrueRange(ZiplineTestCase):
 
         tr.compute(today, assets, out, highs, lows, closes)
         assert_equal(out, np.full((3,), 2.))
+
+
+class MovingAverageConvergenceDivergenceCase(ZiplineTestCase):
+    def test_MACD_window_length_generation(self):
+        signal_period = random_integers(1, 90)
+        fast_period = random_integers(signal_period+1, signal_period+100)
+        slow_period = random_integers(fast_period+1, fast_period+100)
+        ewma = MovingAverageConvergenceDivergence(
+            fast_period=fast_period,
+            slow_period=slow_period,
+            signal_period=signal_period,
+        )
+        assert_equal(
+            ewma.window_length,
+            slow_period+signal_period-1,
+        )
+
+    def test_moving_average_convergence_divergence(self):
+        fast_period = 3
+        slow_period = 8
+        signal_period = 2
+
+        macd = MovingAverageConvergenceDivergence(
+            fast_period=fast_period,
+            slow_period=slow_period,
+            signal_period=signal_period,
+        )
+
+        today = pd.Timestamp('2016', tz='utc')
+        nassets = macd.window_length
+        assets = pd.Index(np.arange(nassets))
+        days_col = np.arange(start=-.05,
+                             stop=.01*nassets-.05,
+                             step=.01)[:, np.newaxis]
+        close = np.logspace(start=.01, stop=.10, num=nassets) - 1 + days_col
+
+        dtype = [
+            ('macd', 'f8'),
+            ('signal', 'f8'),
+            ('hist', 'f8'),
+        ]
+        out = np.recarray(
+            shape=(nassets,),
+            dtype=dtype,
+            buf=np.empty(shape=(nassets,), dtype=dtype),
+        )
+        macd.compute(
+            today,
+            assets,
+            out,
+            close,
+            fast_period,
+            slow_period,
+            signal_period,
+        )
+
+        expected_macd = np.array([0.01691553] * nassets)
+        expected_signal = np.array([0.01691553] * nassets)
+        expected_hist = np.array([0] * nassets)
+
+        np.testing.assert_almost_equal(
+            out.macd,
+            expected_macd,
+            decimal=8
+        )
+        np.testing.assert_almost_equal(
+            out.signal,
+            expected_signal,
+            decimal=8
+        )
+        np.testing.assert_almost_equal(
+            out.hist,
+            expected_hist,
+            decimal=8
+        )
--- a/tests/pipeline/test_technical.py
+++ b/tests/pipeline/test_technical.py
@@ -17,13 +17,12 @@ from zipline.pipeline.factors import (
     LinearWeightedMovingAverage,
     RateOfChangePercentage,
     TrueRange,
-    MovingAverageConvergenceDivergence,
+    MovingAverageConvergenceDivergenceSignal,
     AnnualizedVolatility,
 )
 from zipline.testing import parameter_space
 from zipline.testing.fixtures import ZiplineTestCase
 from zipline.testing.predicates import assert_equal
-
 from .base import BasePipelineTestCase
 
 
@@ -409,11 +408,24 @@ class TestTrueRange(ZiplineTestCase):
 
 
 class MovingAverageConvergenceDivergenceTestCase(ZiplineTestCase):
+
+    def expected_ewma(self, data_df, window):
+        # Comment copied from `test_engine.py`:
+        # XXX: This is a comically inefficient way to compute a windowed EWMA.
+        # Don't use it outside of testing.  We're using rolling-apply of an
+        # ewma (which is itself a rolling-window function) because we only want
+        # to look at ``window_length`` rows at a time.
+        return data_df.rolling(window).apply(
+            lambda sub: pd.DataFrame(sub)
+            .ewm(span=window)
+            .mean()
+            .values[-1])
+
     def test_MACD_window_length_generation(self):
         signal_period = random_integers(1, 90)
         fast_period = random_integers(signal_period+1, signal_period+100)
         slow_period = random_integers(fast_period+1, fast_period+100)
-        ewma = MovingAverageConvergenceDivergence(
+        ewma = MovingAverageConvergenceDivergenceSignal(
             fast_period=fast_period,
             slow_period=slow_period,
             signal_period=signal_period,
@@ -424,34 +436,22 @@ class MovingAverageConvergenceDivergenceTestCase(ZiplineTestCase):
         )
 
     def test_moving_average_convergence_divergence(self):
+        nassets = 3
         fast_period = 3
         slow_period = 8
         signal_period = 2
 
-        macd = MovingAverageConvergenceDivergence(
+        macd = MovingAverageConvergenceDivergenceSignal(
             fast_period=fast_period,
             slow_period=slow_period,
             signal_period=signal_period,
         )
 
         today = pd.Timestamp('2016', tz='utc')
-        nassets = macd.window_length
         assets = pd.Index(np.arange(nassets))
-        days_col = np.arange(start=-.05,
-                             stop=.01*nassets-.05,
-                             step=.01)[:, np.newaxis]
-        close = np.logspace(start=.01, stop=.10, num=nassets) - 1 + days_col
+        out = np.empty(shape=(nassets,), dtype=np.float64)
+        close = np.random.rand(macd.window_length, nassets)
 
-        dtype = [
-            ('macd', 'f8'),
-            ('signal', 'f8'),
-            ('hist', 'f8'),
-        ]
-        out = np.recarray(
-            shape=(nassets,),
-            dtype=dtype,
-            buf=np.empty(shape=(nassets,), dtype=dtype),
-        )
         macd.compute(
             today,
             assets,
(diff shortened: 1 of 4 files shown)


zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +686,94 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+
+    The degree of variation of a series over time as measured by the standard
+    deviation of the data over the course of a year.
+
+    **Default Inputs:** :data:`zipline.pipeline.data.USEquityPricing.close`
+    """
+    inputs = [USEquityPricing.close]
+
+    def compute(self, today, assets, out, closes):
+        out[:] = nanstd(closes, ddof=1, axis=0) * (252 ** 0.5)
```
ssanderson: We run with trading calendars other than NYSE in zipline.  I don't think hard-coding 252 makes sense here.

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +686,94 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+    Moving Average
+    https://en.wikipedia.org/wiki/Moving_average
+
+    Generates a series of averages using a subset of data. It is often used to
+    smooth out short-term fluctuations and highlight longer-term trends.
+
+    **Default Inputs:** None
+    **Default Window Length:** None
+    """
+    def compute(self, today, assets, out, data):
```
richafrank: I think this exists already as `SimpleMovingAverage`!

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +680,117 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+        return super(MovingAverageConvergenceDivergence, cls).__new__(
+            cls,
+            fast_period=fast_period,
+            slow_period=slow_period,
+            signal_period=signal_period,
+            window_length=slow_period + signal_period - 1,
+            *args, **kwargs
+        )
+
+    def calculate_ewa(self, data, length):
```
dmichalowicz: Is there a reason you call it "EWA" here but "EMA" above? (assuming both of these mean exponentially weighted moving average)

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +680,117 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+    """
+    Volatility
+    https://en.wikipedia.org/wiki/Volatility_(finance)
+
+    The degree of variation of a series over time as measured by the standard
+    deviation of the data over the course of a year.
+
+    **Default Inputs:** :data:`zipline.pipeline.data.USEquityPricing.close`
+    """
+    inputs = [USEquityPricing.close]
```
dmichalowicz: According to the wikipedia link cited: "In finance, volatility is the degree of variation of a trading price series over time as measured by the standard deviation of returns." Would this default value be better of as `Returns`?

zipline/pipeline/factors/__init__.py
```diff
@@ -62,4 +64,6 @@
     'TrueRange',
     'VWAP',
     'WeightedAverageValue',
+    'MovingAverageConvergenceDivergence',
+    'MACD'
```
dmichalowicz: AnnualVolatility was not added here

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +680,117 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+
+    def compute(self, today, assets, out, close, fast_period, slow_period,
+                signal_period):
+        macd, sig, hist = zip(*map(self.calculate_macd, close.T))
+        out.macd[:] = macd
+        out.signal[:] = sig
+        out.hist[:] = hist
+
+
+class AnnualVolatility(CustomFactor):
```
dmichalowicz: AnnualizedVolatility is more clear here

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +680,117 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+            *args, **kwargs
+        )
+
+    def calculate_ewa(self, data, length):
+        decay_rate = 1.0 - (2.0 / (1.0 + length))
+        return average(data,
+                       axis=1,
+                       weights=self.weights(length, decay_rate))
+
+    def rolling_window(self, a, window):
```
dmichalowicz: this can use `rolling_window`in `zipline.utils.numpy_utils`

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +680,117 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+                    self.params['fast_period']
+                )[-self.params['signal_period']:],
+                self.params['fast_period'])
+            macd = fast_EWA - slow_EWA
+            signal_line = self.calculate_ewa(
+                macd.reshape(-1, self.params['signal_period']),
+                self.params['signal_period'])
+            hist = macd[-1] - signal_line
+            return macd[-1], signal_line[-1], hist[-1]
+        except:
```
ssanderson: We should only be catching the exceptions we expect to see here.

zipline/pipeline/factors/technical.py
ssanderson: :+1: on this question.  Calculating this as stddev of price seems incorrect to me.

tests/pipeline/test_technical.py
```diff
@@ -403,3 +406,130 @@ def test_tr_basic(self):
...
+        ann_vol = AnnualizedVolatility()
+        today = pd.Timestamp('2016', tz='utc')
+        assets = np.arange(nassets, dtype=np.float)
+        returns = np.random.normal(loc=0.001,
+                                   scale=0.01,
+                                   size=(ann_vol.window_length, nassets))
+        out = np.empty(shape=(nassets,), dtype=np.float64)
+        ann_vol.compute(today, assets, out, returns, 252)
+
+        mean = returns.sum(axis=0) / returns.shape[0]
```
dmichalowicz: This could be `np.mean(returns, axis=0)`

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+    annualization_factor :
+        The number of time units per year. Defaults to average number of NYSE
+        trading days per year, 252.
+    """
+    inputs = [Returns(window_length=2)]
+    params = {'annualization_factor': 252}
+    window_length = 252
+
+    def compute(self, today, assets, out, returns, annualization_factor):
+        out[:] = nanstd(returns, ddof=0, axis=0) * (annualization_factor ** .5)
```
dmichalowicz: `ddof` is zero by default

tests/pipeline/test_technical.py
```diff
@@ -403,3 +406,130 @@ def test_tr_basic(self):
...
+        today = pd.Timestamp('2016', tz='utc')
+        assets = np.arange(nassets, dtype=np.float)
+        returns = np.full((ann_vol.window_length, nassets),
+                          0.004,
+                          dtype=np.float64)
+        out = np.empty(shape=(nassets,), dtype=np.float64)
+
+        ann_vol.compute(today, assets, out, returns, 252)
+
+        expected_vol = np.array([0] * nassets)
```
dmichalowicz: This could be `np.zeros(nassets)`

zipline/pipeline/factors/technical.py
ahgnaw: I wanted to make it explicit since we use `ddof=1` in empyrical.

dmichalowicz:
Had a couple small things but otherwise looks good to me. @ssanderson Did you have any other comments on this?

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
             )),
             2
         )
+
+
+class MovingAverageConvergenceDivergence(_ExponentialWeightedFactor):
```
ssanderson: Why is this inheriting from `_ExponentialWeightedFactor`  I don't think it will work if you call any of the inherited classmethods.

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+            self.params['slow_period'])
+        fast_EWMA = self.calculate_ewma(
+            rolling_window(
+                col,
+                self.params['fast_period']
+            )[-self.params['signal_period']:],
+            self.params['fast_period'])
+        macd = fast_EWMA - slow_EWMA
+        signal_line = self.calculate_ewma(
+            macd.reshape(-1, self.params['signal_period']),
```
ssanderson: What does the `reshape(-1` do here?

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+            self.params['signal_period'])
+        hist = macd[-1] - signal_line
+        return macd[-1], signal_line[-1], hist[-1]
+
+    def compute(self, today, assets, out, close, fast_period, slow_period,
+                signal_period):
+        macd, sig, hist = zip(*map(self.calculate_macd, close.T))
```
ssanderson: Why are calling `calclulate_macd` separately on each column of `close`?  It seems like we should be able to do it vectorized on the whole 2d array?

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+        decay_rate = 1.0 - (2.0 / (1.0 + length))
+        return average(data,
+                       axis=1,
+                       weights=self.weights(length, decay_rate))
+
+    def calculate_macd(self, col):
+        slow_EWMA = self.calculate_ewma(
+            rolling_window(
+                col,
+                self.params['slow_period']
```
ssanderson: We should pull out all the params values once instead of multiple times per value.

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+        :data:`zipline.pipeline.factors.Returns(window_length=2)`
+
+    Parameters
+    ----------
+    annualization_factor :
+        The number of time units per year. Defaults to average number of NYSE
+        trading days per year, 252.
+    """
+    inputs = [Returns(window_length=2)]
+    params = {'annualization_factor': 252}
```
ssanderson: When would you not want `annualization_factor` to match the window length?  If we want to make the length configurable, more natural to me would just be to have:

```
class Volatility(CustomFactor):
    inputs = [Returns(...)]

    def compute(self, today, assets, out):
        out[:] =  nanstd(returns, ddof=0, axis=0) * (self.window_length ** .5)
```

ssanderson: And then, we could have an `AnnualizedVolatility` defined as:

```
class AnnualizedVolatility(Volatility):
    window_length = 252
```

zipline/pipeline/factors/__init__.py
```diff
@@ -62,4 +65,7 @@
     'TrueRange',
     'VWAP',
     'WeightedAverageValue',
+    'MovingAverageConvergenceDivergence',
```
ssanderson: The rest of the entries here were in sorted order prior to these additions.  It's nice for a reader looking for an entry to maintain that order.

zipline/pipeline/factors/__init__.py
```diff
@@ -32,6 +32,9 @@
     TrueRange,
     VWAP,
     WeightedAverageValue,
+    MovingAverageConvergenceDivergence,
```
ssanderson: Same note as below re: sorting of the imports here.

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+    slow_period : int >= 0, <= window_length
+        The window length for the "slow" EWMA. Default is 26.
+    signal_period' : int >= 0, <= slow_period
+        The window length for the signal line. Default is 9.
+
+    Returns
+    -------
+    MACD: The difference between "fast" EWMA and "slow" EWMA.
+    signal: The EWMA of the MACD line using `signal_period` as span.
+    hist: Difference between MACD and signal. (Divergence series)
```
ssanderson: Why is this called `hist`?  I don't have much intuition for that name.

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+    fast_period : int >= 0, <= window_length
+        The window length for the "fast" EWMA. Default is 12.
+    slow_period : int >= 0, <= window_length
+        The window length for the "slow" EWMA. Default is 26.
+    signal_period' : int >= 0, <= slow_period
+        The window length for the signal line. Default is 9.
+
+    Returns
+    -------
+    MACD: The difference between "fast" EWMA and "slow" EWMA.
```
ssanderson: Do we expect people to use the outputs from MACD other than `signal`?  Having multiple outputs makes it significantly easier to misuse, so I wonder if we'd be better off with a simpler API.

zipline/pipeline/factors/technical.py
```diff
@@ -683,3 +679,115 @@ def compute(self, today, assets, out, highs, lows, closes):
...
+    params = ('fast_period', 'slow_period', 'signal_period')
+    outputs = ('MACD', 'signal', 'hist')
+
+    def __new__(cls,
+                fast_period=12,
+                slow_period=26,
+                signal_period=9,
+                *args,
+                **kwargs):
+        return super(MovingAverageConvergenceDivergence, cls).__new__(
```
ssanderson: We should raise an error here if `slow_period <= fast_period`.

zipline/pipeline/factors/technical.py
ahgnaw: `self.weights()` is used for `calculate_ewma`, just realized it's a static method

ssanderson: We should just pull that into a free function.

tests/pipeline/test_technical.py
```diff
@@ -403,3 +406,130 @@ def test_tr_basic(self):
...
+            assets,
+            out,
+            close,
+            fast_period,
+            slow_period,
+            signal_period,
+        )
+
+        expected_macd = np.array([0.01691553] * nassets)
+        expected_signal = np.array([0.01691553] * nassets)
```
ahgnaw: Good point, I'll use a different (less efficient) method to calculate MACD and check against that.

zipline/pipeline/factors/technical.py
twiecki: Shouldn't we use `empyrical` for this computation in the first place?

zipline/pipeline/factors/technical.py
ahgnaw: Annualized volatility doesn't prescribe a minimum or maximum sample size. The annualization factor is derived from the time interval of the data, e.g. daily returns.

PR #1590, opened 2016-11-11
PERF: Speed up retrieval of HistoryLoader calendar.

ehebert:
The use of `slice_indexer` on all market minutes was taking about 110ms on my
development machine.

This change to getting the start and end indices changes the entire `_calendar`
method to take 10ms on the same machine.

Noticed while creating a `HistoryLoader` in a notebook context.

### DIFF
--- a/zipline/data/history_loader.py
+++ b/zipline/data/history_loader.py
@@ -573,8 +573,9 @@ class MinuteHistoryLoader(HistoryLoader):
     @lazyval
     def _calendar(self):
         mm = self.trading_calendar.all_minutes
-        return mm[mm.slice_indexer(start=self._reader.first_trading_day,
-                                   end=self._reader.last_available_dt)]
+        start = mm.searchsorted(self._reader.first_trading_day)
+        end = mm.searchsorted(self._reader.last_available_dt, side='right')
+        return mm[start:end]
 
     def _array(self, dts, assets, field):
         return self._reader.load_raw_arrays(


zipline/data/history_loader.py
```diff
@@ -573,8 +573,9 @@ def _frequency(self):
     @lazyval
     def _calendar(self):
         mm = self.trading_calendar.all_minutes
-        return mm[mm.slice_indexer(start=self._reader.first_trading_day,
-                                   end=self._reader.last_available_dt)]
+        start = mm.searchsorted(self._reader.first_trading_day)
+        end = mm.searchsorted(self._reader.last_available_dt)
```
yankees714: Could do `side='right'` here to avoid the `+ 1` in the slice

zipline/data/history_loader.py
```diff
@@ -573,8 +573,9 @@ def _frequency(self):
     @lazyval
     def _calendar(self):
         mm = self.trading_calendar.all_minutes
-        return mm[mm.slice_indexer(start=self._reader.first_trading_day,
-                                   end=self._reader.last_available_dt)]
+        start = mm.searchsorted(self._reader.first_trading_day)
```
ssanderson: We've written this algorithm in a whole bunch of places, and every time I do it I have to think nervously about the boundary conditions.  Should we just put a `slice_dates` function in `zipline.utils.pands_utils` that does this?

ehebert: I agree that we should. Will follow up with a refactoring of this and the other use of the same pattern.

Issue #1541, opened 2016-10-16
get_datetime() and using Minute data frequency


nyxynyx:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

* Operating System: Windows 10
* Python Version: Python 2.7.12 :: Anaconda custom (64-bit)
* How did you install Zipline: `conda install -c Quantopian zipline`
* Python packages: 

	_license                  1.1                      py27_1
	_nb_ext_conf              0.3.0                    py27_0
	alabaster                 0.7.9                    py27_0
	alembic                   0.7.7                    py27_0    Quantopian
	anaconda                  custom                   py27_0
	anaconda-clean            1.0.0                    py27_0
	anaconda-client           1.5.1                    py27_0
	anaconda-navigator        1.2.3                    py27_0
	argcomplete               1.0.0                    py27_1
	astroid                   1.4.7                    py27_0
	astropy                   1.0.5                np19py27_1
	babel                     2.3.4                    py27_0
	backports                 1.0                      py27_0
	backports_abc             0.4                      py27_0
	bcolz                     0.12.1               np19py27_0    Quantopian
	beautifulsoup4            4.5.1                    py27_0
	bitarray                  0.8.1                    py27_1
	blaze                     0.10.1                   py27_0
	bokeh                     0.12.2                   py27_0
	boto                      2.42.0                   py27_0
	bottleneck                1.0.0                np19py27_0
	bzip2                     1.0.6                     vc9_3  [vc9]
	cachetools                1.1.6                    py27_0    Quantopian
	cdecimal                  2.3                      py27_2
	cffi                      1.7.0                    py27_0
	chest                     0.2.3                    py27_0
	click                     6.6                      py27_0
	cloudpickle               0.2.1                    py27_0
	clyent                    1.2.2                    py27_0
	colorama                  0.3.7                    py27_0
	comtypes                  1.1.2                    py27_0
	configobj                 5.0.6                    py27_0
	configparser              3.5.0                    py27_0
	console_shortcut          0.1.1                    py27_1
	contextlib2               0.5.3                    py27_0
	cryptography              1.5                      py27_0
	curl                      7.49.0                    vc9_0  [vc9]
	cycler                    0.10.0                   py27_0
	cyordereddict             0.2.2                    py27_0    Quantopian
	cython                    0.24.1                   py27_0
	cytoolz                   0.8.0                    py27_0
	dask                      0.11.0                   py27_0
[...]

Issue #448, opened 2014-12-08
Port fetch_csv() to zipline


twiecki:
Quantopian algos can use `fetch_csv()`. It should be pretty straight froward to port it to zipline and increase portability.

richafrank:
Part of https://github.com/quantopian/zipline/pull/858

thiagof:
+1

llllllllll:
done

girmax:
fetch_csv() was ported?

richafrank:
Hi @girmax ,

It looks like you're running `zipline.api.fetch_csv` outside of a trading algorithm, which doesn't have much use, since `fetch_csv` just adds the new data to the running algo.  I think we could have a better error message here.

Instead of calling `fetch_csv` directly from module scope in your notebook, you'll want to call it from `initialize` in your algo, and then run your algo with the zipline [cell magic] (http://www.zipline.io/beginner-tutorial.html#ipython-notebook), for example.

adamluc:
I am working to download the following dataset into a zipline algorithim via fetch_csv and running into an issue. Is there an example zipline algorithm that uses fetch_csv that I can review? Not sure if I am doing something wrong.

Thank you!

First, the code I am using is the following:

System specs: Running zipline v1.0.2, Mac OS El Capitan 10.11.6.

`
from zipline.api import fetch_csv

def initialize(context):
    fetch_csv('https://www.quandl.com/api/v3/datasets/YAHOO/INDEX_VIX.csv?api_key=5aSzjTEym3ZyeocFnynn', symbol='VXST', date_column='Date' )
`
The error I get is the following:
`Traceback (most recent call last):
  File "test-5.py", line 713, in <module>
    algo.run(data, overwrite_sim_params=False)
  File "//anaconda/lib/python2.7/site-packages/zipline-0+untagged.4617.ga73aeda-py2.7-macosx-10.6-x86_64.egg/zipline/algorithm.py", line 697, in run
    for perf in self.get_generator():
  File "//anaconda/lib/python2.7/site-packages/zipline-0+untagged.4617.ga73aeda-py2.7-macosx-10.6-x86_64.egg/zipline/algorithm.py", line 596, in get_generator
    return self._create_generator(self.sim_params)
  File "//anaconda/lib/python2.7/site-packages/zipline-0+untagged.4617.ga73aeda-py2.7-macosx-10.6-x86_64.egg/zipline/algorithm.py", line 567, in _create_generator
[...]

PR #1593, opened 2016-11-11
ENH: An improved error message when calling zipline api functions

richafrank:
outside of a running simulation. Previously, an AttributeError was
raised.

The goal here is to make the situation clearer for developers, for example: https://github.com/quantopian/zipline/issues/1541 and https://github.com/quantopian/zipline/issues/448#issuecomment-249312347

What do folks think of the message?

### DIFF
--- a/zipline/utils/api_support.py
+++ b/zipline/utils/api_support.py
@@ -48,7 +48,13 @@ def api_method(f):
     @wraps(f)
     def wrapped(*args, **kwargs):
         # Get the instance and call the method
-        return getattr(get_algo_instance(), f.__name__)(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


ssanderson:
:+1: from me

ssanderson:
The only change I could maybe see here is adding a specific exception type for this (e.g. `NoSimulationRunning`) rather than `RuntimeError`.  But I don't feel super strongly about it either way.

llllllllll:
+1 also, I think the runtime error is fine because I don't expect anyone to want to catch this

richafrank:
Ok, cool, then I'm going to merge on travis passing.

PR #1596, opened 2016-11-18
Order batch

richafrank:
Adds `order_batch` to the blotter and factors out ordering helper methods, so one can generate the orders and give them to `order_batch`.

### DIFF
--- a/tests/test_blotter.py
+++ b/tests/test_blotter.py
@@ -340,22 +340,22 @@ class BlotterTestCase(WithCreateBarData,
         blotter2 = Blotter(self.sim_params.data_frequency,
                            self.asset_finder)
         for i in range(1, 4):
-            order_args = [
+            order_arg_lists = [
                 (self.asset_24, i * 100, MarketOrder()),
                 (self.asset_25, i * 100, LimitOrder(i * 100 + 1)),
             ]
 
-            order_batch_ids = blotter1.order_batch(order_args)
+            order_batch_ids = blotter1.order_batch(order_arg_lists)
             order_ids = []
-            for order_arg in order_args:
-                order_ids.append(blotter2.order(*order_arg))
+            for order_args in order_arg_lists:
+                order_ids.append(blotter2.order(*order_args))
             self.assertEqual(len(order_batch_ids), len(order_ids))
 
             self.assertEqual(len(blotter1.open_orders),
                              len(blotter2.open_orders))
 
             for (asset, _, _), order_batch_id, order_id in zip(
-                    order_args, order_batch_ids, order_ids
+                    order_arg_lists, order_batch_ids, order_ids
             ):
                 self.assertEqual(len(blotter1.open_orders[asset]),
                                  len(blotter2.open_orders[asset]))
--- a/tests/test_blotter.py
+++ b/tests/test_blotter.py
@@ -329,3 +329,37 @@ class BlotterTestCase(WithCreateBarData,
         )
 
         blotter.prune_orders([other_order])
+
+    def test_order_batch_matches_multi_order(self):
+        """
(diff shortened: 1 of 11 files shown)


tests/test_blotter.py
```diff
@@ -83,64 +89,59 @@ def CREATE_BARDATA_DATA_FREQUENCY(cls):
...
+        result = blotter.open_orders[self.asset_24][0]
 
         self.assertEqual(result.limit, expected_lmt)
         self.assertEqual(result.stop, expected_stp)
 
     def test_cancel(self):
-        blotter = Blotter('daily', self.env.asset_finder)
+        blotter = Blotter('daily', self.asset_finder)
 
-        asset_24 = blotter.asset_finder.retrieve_asset(24)
```
ssanderson: Not super important, but we might make this diff less noisy by just replacing this line with `asset_24 = self.asset_24` instead of changing every usage below.

zipline/finance/blotter.py
```diff
@@ -114,6 +114,9 @@ def order(self, sid, amount, style, order_id=None):
 
         return order.id
 
+    def order_batch(self, orders):
```
ssanderson: Do we want to document this in some way? (Or at least comment it so we don't remove it?)

ssanderson: `orders` here makes me think this is an iterable of `Order` objects.  Not sure what's a bettter name though, since `order_args` makes me think it's a single tuple of arguments, and `order_argses` just sounds like hobbit-speak.

ssanderson:
@richafrank finished this round.  No major issues.  Had a couple questions about naming/docs for `order_batch` and a style note for some of the test refactoring.

zipline/algorithm.py
```diff
@@ -1946,11 +1964,44 @@ def order_target_percent(self, asset, target,
...
+        order_ids : pd.Series[Asset -> str]
+
+        See Also
+        --------
+        :func:`zipline.api.order_target_percent`
+        """
+        order_args = OrderedDict()
+        for asset, target in iteritems(weights):
+            if self._can_order_asset(asset):
+                amount = self._calculate_order_target_percent_amount(
```
ssanderson: Should we `continue` here if `amount` is 0?

zipline/finance/blotter.py
```diff
@@ -75,12 +75,29 @@ def set_date(self, dt):
         self.current_dt = dt
 
     def order(self, sid, amount, style, order_id=None):
+        """Place an order.
```
ssanderson: :+1: :memo:

zipline/test_algorithms.py
```diff
@@ -429,9 +429,17 @@ def handle_data(self, data):
...
+        self._order(self.sid(0), .002)
         self.ordered = True
 
+    def _order(self, asset, target):
+        return self.order_target_percent(asset, target)
+
+
+class TestBatchTargetPercentAlgorithm(TestTargetPercentAlgorithm):
+    def _order(self, asset, target):
+        return self.batch_order_target_percent({asset: target})
```
ssanderson: Do we want a test for actually ordering more than one asset at a time?

zipline/algorithm.py
richafrank: I figured we shouldn't check the amount in multiple places, but what do you think? We check it currently in `Blotter.order`.

zipline/test_algorithms.py
richafrank: Yes, definitely - that's the test I still need to add.

zipline/algorithm.py
richafrank: I'll filter out the null entries before returning them to the caller.

PR #1597, opened 2016-11-18
BUG: Fix index error when volume roll occurs at end of window.

ehebert:
When a volume roll makes the back contract the primary at the end of a window,
the next auto close to use in the roll finder loop (i.e. the auto close of the
front contract on the last session of the window) is greater than the last date
of the window. This caused an index error because the auto close would
searchsorted

Fix by changing the start date when entering the loop to calculate rolls so that
it takes the floor of the auto close and the next date to consider while working
backwards.

Also, use `get_loc` instead of `searchsorted` since exact dates are used.

### DIFF
--- a/tests/test_continuous_futures.py
+++ b/tests/test_continuous_futures.py
@@ -618,6 +618,49 @@ def record_current_contract(algo, data):
                          2,
                          "Should be FOH16 on session after roll.")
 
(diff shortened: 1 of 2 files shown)


ehebert:
Closing, bounds errors were fixed via https://github.com/quantopian/zipline/pull/1600

PR #1598, opened 2016-11-18
BUG: Ensure minute OHLC values can safely be converted to uint32

yankees714:
Otherwise, we either raise an exception or filter out all unsafe values.
This addresses an issue where the BcolzMinuteBarWriter would scale up
values to convert to uint32, but the resulting values were too large,
and would be mangled.

Based on the approach we take in the BcolzDailyBarWriter. See https://github.com/quantopian/zipline/blob/master/zipline/data/us_equity_pricing.py#L116.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -35,6 +35,7 @@ from zipline.data._minute_bar_internal import (
 from zipline.gens.sim_engine import NANOS_IN_MINUTE
 
 from zipline.data.bar_reader import BarReader, NoDataOnDate
+from zipline.data.us_equity_pricing import check_uint32_safe
 from zipline.utils.calendars import get_calendar
 from zipline.utils.cli import maybe_show_progress
 from zipline.utils.memoize import lazyval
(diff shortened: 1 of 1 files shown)


zipline/data/minute_bars.py
```diff
@@ -108,6 +109,46 @@ def _sid_subdir_path(sid):
...
+            raise
+
+        if invalid_data_behavior == 'warn':
+            logger.warn(
+                'Values for sid={}, col={} contain some too large for '
+                'uint32 (max={}), filtering them out',
+                sid, col_name, max_val,
+            )
+
+        return scaled_col[scaled_col <= np.iinfo(np.uint32).max]
```
ehebert: This will make the results no longer align, because elements which do not meet the predicate will be removed.

e.g.:
```
In [37]: foo = np.random.rand(10)

In [38]: foo
Out[38]: 
array([ 0.86608194,  0.03375079,  0.76099617,  0.29876178,  0.01236926,
        0.17863316,  0.85661549,  0.96673179,  0.12907978,  0.11848878])

In [39]: foo[foo < 0.5]
Out[39]: 
array([ 0.03375079,  0.29876178,  0.01236926,  0.17863316,  0.12907978,
        0.11848878])
```

PR #1599, opened 2016-11-22
Memory savings

ssanderson:


ssanderson:
These are all relatively minor in the grand scheme of things.  The overwhelming consumer of memory in zipline simulations that touch many assets are the active blocks of minutely bcolz files.

zipline/data/history_loader.py
```diff
@@ -380,7 +380,7 @@ def _ensure_sliding_windows(self, assets, dts, field,
         assets = self._asset_finder.retrieve_all(assets)
 
         try:
-            end_ix = self._calendar.get_loc(end)
+            end_ix = self._calendar.searchsorted(end)
```
ehebert: Thanks for spotting this one. I was unaware of that bug/behavior.

On the same track, I do question whether we should pull in all minutes ever from the calendar (mea culpa), in favor of calculating the location in a similar fashion to the minute bar reader's find position of minute.
(The minute index takes 11MB at minimum for 14 years just for the datetime values.)

Does using `getitem` also trigger the creation of the hash table?
If so we should also audit usage of `trading_calendar.previous_minute`.

zipline/utils/cache.py
```diff
@@ -20,7 +20,7 @@ class Expired(Exception):
     """
 
 
-class CachedObject(namedtuple("_CachedObject", "value expires")):
```
ehebert: Why the change from using `namedtuple`.

ssanderson: @llllllllll and I had some concerns about the behavior of garbage collection on objects with empty slots.  See the discussion in https://bugs.python.org/issue24379.

The specific issue, I think, is that objects with empty `__slots__` don't participate in garbage collection, which means they can leak if they participate in a cycle.  I don't think that's happening here, but there's also not much benefit in making this a subclass of namedtuple.

PR #1600, opened 2016-11-22
BUG: Fix bounds errors in roll finder.

ehebert:
Fix common error condition which was triggered whenever the session at the end
of the prefetched history window was a session where the back contract was
active. When the back contract was the active contract, the next contract for
consideration was the front contract at the end of the window, which
definitionally always has an autoclose after the end of the window.
Instead, just start seeking backwards from the end of the window.

Also prevent lookahead bias in volume rolls, which was caused by the using the
volume for a session to determine whether that session had rolled. Information
that would not have been available at the beginning of the session.

This change makes the volume rolls overly conservative, and may be improved by
looking at vectors of the preceding volume and making the roll off of momentum.

### DIFF
--- a/zipline/assets/roll_finder.py
+++ b/zipline/assets/roll_finder.py
@@ -96,23 +96,21 @@ class RollFinder(with_metaclass(ABCMeta, object)):
             i -= 1
         else:
             i -= 2
-        auto_close_date = Timestamp(oc.auto_close_dates[i], tz='UTC')
-        while auto_close_date > start and i > -1:
-            session_loc = sessions.searchsorted(auto_close_date)
+        curr = sessions[-1]
+        while curr > start and i > -1:
+            session_loc = sessions.searchsorted(curr)
             front = oc.contract_sids[i]
             back = oc.contract_sids[i + 1]
-            while session_loc > -1:
+            while session_loc > 0:
                 session = sessions[session_loc]
-                if back != self._active_contract(oc, front, back, session):
+                prev = sessions[session_loc - 1]
+                if back != self._active_contract(oc, front, back, prev):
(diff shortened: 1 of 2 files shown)


zipline/assets/roll_finder.py
```diff
@@ -149,7 +147,15 @@ def __init__(self, trading_calendar, asset_finder, session_reader):
         self.session_reader = session_reader
 
     def _active_contract(self, oc, front, back, dt):
-        # FIXME: Possible vector for look ahead bias.
-        front_vol = self.session_reader.get_value(front, dt, 'volume')
-        back_vol = self.session_reader.get_value(back, dt, 'volume')
-        return back if back_vol > front_vol else front
+        for i, sid in enumerate(oc.contract_sids):
+            if sid == front:
+                break
```
yankees714: Can we move this into the else block below? Seems like we don't need `i` except for accessing `oc.auto_close_dates`.

ehebert:
@yankees714 I agree that `get_rolls` needs more comments (and its counter part get partitions). Will add in a follow up PR.

zipline/assets/roll_finder.py
ehebert: Yes, will move.

ehebert: Also, have been considering changing the implementation so that we are using a contract object which has some structure like:

```
class ChainedContract(object):
    def. __init__(...):
        self.future = ... # A Future object
        self.next = # Next contract in the chain
        self.prev = # Prev contract in the chain
```

Where `chained_contarct.future.sid` would be the way to get the current sid. 
But would do that after merging this.

PR #1604, opened 2016-11-29
MAINT: Use a doubly linked list for contract chain.

ehebert:
Instead of requiring the roll finder to juggle the indices into the ordered
contracts, use a doubly linked list where the nodes element is the contract
with members pointing to the previous and next contracts in the chain.

Besides improving legibility in the roll finder code, this change is on the path
to adding a predicate to exclude contracts from the chain, e.g. contracts in ED
which are not in the roll schedule.

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -15,7 +15,7 @@
 from abc import ABCMeta
 import array
 import binascii
-from collections import namedtuple
+from collections import deque, namedtuple
 from numbers import Integral
 from operator import itemgetter, attrgetter
 import struct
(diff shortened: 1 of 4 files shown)


ehebert:
This change could be pushed further by having methods such as `get_contract_center` return the `ContractNode` instead of the `sid`.

zipline/assets/continuous_futures.pyx
```diff
@@ -249,16 +277,12 @@ cdef class OrderedContracts(object):
...
-    which is why search by comparison over the range of contracts is
-    used. At this size, this is faster than using an index or np.searchsorted.
-
     Members
     -------
     root_symbol : str
         The root symbol of the future contract chain.
-    contract_sids : long[:]
-        The contract sids in sorted order of occurrence.
+    contract : deque
```
yankees714: Should be `contracts`

zipline/assets/assets.py
```diff
@@ -880,15 +880,14 @@ def lookup_expired_futures(self, start, end):
...
+    def _get_contract_sids(self, root_symbol):
         fc_cols = self.futures_contracts.c
 
-        fields = (fc_cols.sid, fc_cols.start_date, fc_cols.auto_close_date)
-
-        return list(sa.select(fields).where(
-            (fc_cols.root_symbol == root_symbol) &
-            (fc_cols.start_date != pd.NaT.value))
-            .order_by(fc_cols.auto_close_date).execute().fetchall())
+        return [r[0] for r in
```
ehebert: Will try.

PR #1607, opened 2016-12-01
ENH: Allow future chains to only use certain delivery months.

ehebert:
To support contracts such as `PL` which should roll from F->J->N->V, add the
ability to pass a predicate function to the ordered contract chain contstrution
which returns `True` if the contract is allowed in the chain.

### DIFF
--- a/tests/test_continuous_futures.py
+++ b/tests/test_continuous_futures.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from collections import deque
+from functools import partial
 from textwrap import dedent
 
 from numpy import (
@@ -28,7 +29,10 @@ import pandas as pd
 from pandas import Timestamp, DataFrame
 
 from zipline import TradingAlgorithm
-from zipline.assets.continuous_futures import OrderedContracts
+from zipline.assets.continuous_futures import (
+    OrderedContracts,
+    delivery_predicate
+)
 from zipline.data.minute_bars import FUTURES_MINUTES_PER_DAY
 from zipline.testing.fixtures import (
     WithAssetFinder,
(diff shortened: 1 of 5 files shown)


ehebert:
A graph of continuous `PL` volume without this patch applied.

pl-without-chain-predicate

This is with this patch applied. 

pl-with-chain-predicate

zipline/assets/continuous_futures.pyx
```diff
@@ -295,12 +312,13 @@ cdef class OrderedContracts(object):
...
     cdef readonly object root_symbol
-    cdef readonly object head_contract
+    cdef readonly object _head_contract
     cdef readonly dict sid_to_contract
     cdef readonly int64_t _start_date
     cdef readonly int64_t _end_date
+    cdef readonly object chain_predicate
 
-    def __init__(self, object root_symbol, object contracts):
+    def __init__(self, object root_symbol, object contracts, object chain_predicate=None):
```
dmichalowicz: Same here with the docstring

zipline/assets/assets.py
```diff
@@ -193,7 +197,7 @@ class AssetFinder(object):
     PERSISTENT_TOKEN = "<AssetFinder>"
 
     @preprocess(engine=coerce_string_to_eng)
-    def __init__(self, engine):
+    def __init__(self, engine, future_chain_predicates=CHAIN_PREDICATES):
```
dmichalowicz: Thinking more, this feels like it doesn't quite belong as a parameter to `AssetFinder`, but I'm not sure how else to refactor it. Maybe make this an optional parameter to `get_ordered_contracts`?

zipline/assets/continuous_futures.pyx
```diff
@@ -37,6 +39,21 @@ import warnings
 from zipline.utils.calendars import get_calendar
 
 
+def delivery_predicate(codes, contract):
+    # This relies on symbols that are construct following a pattern of
+    # root symbol + delivery code + year, e.g. PLF16
+    # This check would be more robust if the future contract class had
+    # a 'delivery_month' member.
+    delivery_code = contract.symbol[2]
```
dmichalowicz: Is the `delivery_code` guaranteed to be the third character of the symbol, i.e. the root symbol is always 2 characters?

tests/test_continuous_futures.py
```diff
@@ -144,7 +152,34 @@ def make_futures_info(self):
...
+                         Timestamp('2016-10-19', tz='UTC')],
+            'notice_date': [Timestamp('2016-01-11', tz='UTC'),
+                            Timestamp('2016-02-13', tz='UTC'),
+                            Timestamp('2016-03-10', tz='UTC')],
+            'expiration_date': [Timestamp('2016-01-11', tz='UTC'),
+                                Timestamp('2016-02-13', tz='UTC'),
+                                Timestamp('2016-03-10', tz='UTC')],
+            'auto_close_date': [Timestamp('2016-01-11', tz='UTC'),
+                                Timestamp('2016-02-13', tz='UTC'),
+                                Timestamp('2016-03-10', tz='UTC')],
```
dmichalowicz: Is there a particular method to contrive these dates, or are they arbitrary?

dmichalowicz:
@ehebert Done with a first round of review, mostly had a few questions but overall looks good. Also thanks for the screenshots

zipline/assets/assets.py
ehebert: Yes.

PR #1608, opened 2016-12-01
BUG: fix blaze pipeline queries for asof_date

llllllllll:


ssanderson:
Assuming the new test failed prior to the other changes, this looks good to me.

llllllllll:
It failed with the same error we saw earlier.

PR #1609, opened 2016-12-05
MAINT: Use session index freq for loop logic.

ehebert:
Instead of maintaining a separate index into the sessions index, use the `.freq`
member of the sessions index for decrementing to the current session and finding
the previous session.

### DIFF
--- a/zipline/assets/roll_finder.py
+++ b/zipline/assets/roll_finder.py
@@ -89,25 +89,25 @@ class RollFinder(with_metaclass(ABCMeta, object)):
         tc = self.trading_calendar
         sessions = tc.sessions_in_range(tc.minute_to_session_label(start),
                                         tc.minute_to_session_label(end))
+        freq = sessions.freq
         if first == front:
             curr = first_contract << 1
         else:
             curr = first_contract << 2
-        sess = sessions[-1]
-        while sess > start and curr is not None:
-            session_loc = sessions.searchsorted(sess)
+        session = sessions[-1]
+
+        while session > start and curr is not None:
             front = curr.contract.sid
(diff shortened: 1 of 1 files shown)


zipline/assets/roll_finder.py
```diff
@@ -89,22 +89,21 @@ def get_rolls(self, root_symbol, start, end, offset):
...
+            while session > start:
+                prev = session - freq
                 if back != self._active_contract(oc, front, back, prev):
                     rolls.insert(0, ((curr >> offset).contract.sid, session))
                     break
-                session_loc -= 1
+                session = prev
             curr = curr.prev
             if curr is not None:
                 sess = curr.contract.auto_close_date
```
yankees714: Should this now be `session`?

zipline/assets/roll_finder.py
```diff
@@ -89,22 +89,21 @@ def get_rolls(self, root_symbol, start, end, offset):
...
                                         tc.minute_to_session_label(end))
+        freq = sessions.freq
         if first == front:
             curr = first_contract << 1
         else:
             curr = first_contract << 2
-        sess = sessions[-1]
-        while sess > start and curr is not None:
-            session_loc = sessions.searchsorted(sess)
+        session = sessions[-1]
```
yankees714: Super nitpicky, but can we add a newline for legibility here, to separate out the "initialization" from the loop?

zipline/assets/roll_finder.py
ehebert: Yes, will fix. (Clearing up that difference *was* one the intents of this change...)

PR #1612, opened 2016-12-07
ENH: Adds support for supplementary asset mappings

yankees714:
- Adds a `supplementary_mappings` table to the assets.db, to hold point-in-time mappings of sids to arbitrary categories of values, e.g. alternative identifiers. This bumps `ASSET_DB_VERSION`.
- Adds `supplementary_map` and `supplementary_map_by_sid` to AssetFinder, caches of the underlying table that are fully populated on first access, which map the supplementary values to sids and vice versa, respectively.
- Adds `lookup_by_supplementary_mapping method`, which fronts `supplementary_map` to query for the asset last known to have held a value at a given dt.
- Add `get_supplementary_field` method, which fronts `supplementary_map_by_sid` to query for the last known value held by an asset at a given dt.


zipline/assets/assets.py
```diff
@@ -809,6 +882,79 @@ def lookup_future_symbol(self, symbol):
             raise SymbolNotFound(symbol=symbol)
         return self.retrieve_asset(data['sid'])
 
+    def lookup_by_supplementary_mapping(self, mapping_type, value, as_of_date):
```
yankees714: Probably want to have this accept a list of values

zipline/assets/assets.py
```diff
@@ -809,6 +882,79 @@ def lookup_future_symbol(self, symbol):
...
+
+        # no equity held the value on the given asof date
+        raise ValueNotFoundForMappingType(type=mapping_type, value=value)
+
+    def get_supplementary_field(
+        self,
+        asset_convertible,
+        field_name,
+        as_of_date,
+    ):
```
yankees714: And this can accept a list of asset convertibles

zipline/assets/assets.py
```diff
@@ -91,6 +95,35 @@
 })
 
 SymbolOwnership = namedtuple('SymbolOwnership', 'start end sid symbol')
+OwnershipPeriod = namedtuple('OwnershipPeriod', 'start end sid value')
```
richafrank: Do you think we need both of these, or could we use `OwnershipPeriod` for symbols too?

zipline/assets/assets.py
```diff
@@ -313,6 +346,46 @@ def fuzzy_symbol_ownership_map(self):
             fuzzy_owners.sort()
         return fuzzy_mappings
 
+    @lazyval
+    def supplementary_map(self):
```
richafrank: Also `lookup_by_supplementary_mapping` and `_lookup_symbol_strict`

zipline/assets/assets.py
richafrank: Would we have a different data structure that supports vectorized lookups, or would it be saving the caller from writing a loop themselves?

zipline/assets/asset_writer.py
```diff
@@ -432,6 +445,8 @@ def write(self,
                   A short description of this root symbol.
               exchange : str
                   The exchange where this root symbol is traded.
+        supplementary_mappings : pd.DataFrame, optional
+            Additional mappings from values of abitrary type to assets.
```
richafrank: It looked like the foreign key is to the equities table.  Do we want to support lookup to any asset class or just equities, and if it's just equities, should we name these various objects more specifically, e.g. `equity_supplementary_mappings`?

zipline/assets/asset_writer.py
```diff
@@ -379,7 +392,7 @@ def write(self,
                   The exchange where this asset is traded.
 
             The index of this dataframe should contain the sids.
-        futures : pd.Dataframe, optional
+        futures : pd.DataFrame, optional
```
richafrank: 👍

zipline/assets/assets.py
```diff
@@ -809,6 +882,79 @@ def lookup_future_symbol(self, symbol):
...
+
+        for start, end, sid, _ in owners:
+            if start <= as_of_date < end:
+                # find the equity that owned it on the given asof date
+                return self.retrieve_asset(sid)
+
+        # no equity held the value on the given asof date
+        raise ValueNotFoundForMappingType(type=mapping_type, value=value)
+
+    def get_supplementary_field(
```
richafrank: Could we be more consistent about the method names, like `lookup_by_supplementary_field` and `get_supplementary_field` for example?

tests/test_assets.py
```diff
@@ -966,6 +970,198 @@ def test_sids(self):
...
+            equities=equities,
+            supplementary_mappings=supplementary_mappings,
+        )
+
+        af = self.asset_finder
+
+        # Before sid 0 has changed ALT_ID.
+        dt = pd.Timestamp('2013-6-28', tz='UTC')
+
+        asset_0 = af.lookup_by_supplementary_mapping('ALT_ID', '100000000', dt)
```
yankees714: Yeah definitely agree, what do you think of `lookup_by_supplementary_field`, to go along with `get_supplementary_field`?

richafrank: That sounds good to me.

zipline/assets/asset_db_schema.py
```diff
@@ -162,6 +163,29 @@
     sa.Column('asset_type', sa.Text),
 )
 
+supplementary_mappings = sa.Table(
+    'supplementary_mappings',
+    metadata,
+    sa.Column(
+        'id',
```
yankees714: I included it since we also have it in the `equity_symbol_mappings` table. It's populated with the index of the dataframe of mappings that we write, which ends up just being the default dataframe index, since we don't set it to a particular column. Do you think it's unnecessary?

zipline/assets/assets.py
yankees714: Yeah I think we should consolidate these, since `OwnershipPeriod` is just a generalized version of `SymbolOwnership`.

zipline/assets/assets.py
yankees714: Oh ha, missed this before my comment above

zipline/assets/assets.py
```diff
@@ -809,6 +882,79 @@ def lookup_future_symbol(self, symbol):
             raise SymbolNotFound(symbol=symbol)
         return self.retrieve_asset(data['sid'])
 
+    def lookup_by_supplementary_mapping(self, mapping_type, value, as_of_date):
+        try:
+            owners = self.supplementary_map[
```
yankees714: Oh interesting, hadn't thought of that. I wonder if the hit we take for loading everything in is worth avoiding the added complexity? Guess it depends how many different types of mappings you have...

zipline/assets/assets.py
yankees714: I was thinking of the latter for now, like we do in `lookup_generic`. Maybe the vectorized lookups could be a follow up improvement?

zipline/assets/assets.py
yankees714: 👍

zipline/assets/asset_db_schema.py
```diff
@@ -103,7 +103,7 @@
         index=True,
     ),
     sa.Column('value', sa.Text, nullable=False),
-    sa.Column('mapping_type', sa.Text, nullable=False),
+    sa.Column('field', sa.Text, nullable=False),
```
yankees714: @richafrank I think this change makes sense given the rest of our naming

zipline/assets/assets.py
richafrank: Yea, I was also considering whether to load both directions (sid->identifier, identifier->sid) together when one direction is first requested, re-using all the query results, instead of reading the whole table again. I ended up thinking leave it as is, as I don't think it's a huge data set memory-wise, and it then becomes just a question of when to load it, which I don't have a strong opinion about. Curious if you have an opinion.

zipline/assets/asset_db_schema.py
richafrank: My instinct is to make a compound pk on those unique columns instead of adding the surrogate key.  Not sure if there's a performance impact...

zipline/assets/asset_writer.py
```diff
@@ -102,6 +104,16 @@
     'exchange': None,
 }
 
+# Default values for the supplementary_mappings DataFrame
+_supplementary_mappings_defaults = {
+    'sid': None,
+    'value': None,
+    'field': None,
+    'start_date': 0,
+    'end_date': 2 ** 62 - 1,
```
llllllllll: did you mean `2 ** 63 - 1`? Maybe `np.iinfo('int64').max` would be more clear.

zipline/assets/assets.py
```diff
@@ -90,7 +94,58 @@
...
+                        pd.Timestamp.max.tz_localize('utc'),
+                        None,
+                        None,
+                        None,
+                    )],
+                ),
+            )
+        ),
+        mappings,
+        factory=lambda: mappings,
```
llllllllll: Why are we passing this as the factory? This makes the function mutate the input.

zipline/assets/assets.py
```diff
@@ -90,7 +94,58 @@
     'auto_close_date',
 })
 
-SymbolOwnership = namedtuple('SymbolOwnership', 'start end sid symbol')
+OwnershipPeriod = namedtuple('OwnershipPeriod', 'start end sid value')
+
+
+def merge_ownership_periods(mappings):
```
llllllllll: Can we add a docstring to this?

zipline/assets/assets.py
llllllllll: Can we add a docstring to this?

tests/test_assets.py
```diff
@@ -966,6 +970,198 @@ def test_sids(self):
...
+
+        asset_1 = af.lookup_by_supplementary_field('ALT_ID', '100000001', dt)
+        self.assertEqual(asset_1.sid, 1)
+
+        asset_0 = af.lookup_by_supplementary_field('ALT_ID', '100000002', dt)
+        self.assertEqual(asset_0.sid, 0)
+
+        # At this point both sids 0 and 2 have held this value, so an
+        # as_of_date is required.
+        with self.assertRaises(MultipleValuesFoundForField):
```
llllllllll: can we add a test for the repr of the error message to make sure we are forwarding the proper info to the message?

zipline/assets/asset_db_schema.py
```diff
@@ -84,6 +85,29 @@
...
+        unique=True,
+        nullable=False,
+        primary_key=True,
+    ),
+    sa.Column(
+        'sid',
+        sa.Integer,
+        sa.ForeignKey(equities.c.sid),
+        nullable=False,
+        index=True,
```
llllllllll: do we need to index this if we are doing `select *`?

yankees714: Probably not, maybe we should change this on the `equity_symbol_mappings` table as well? https://github.com/quantopian/zipline/pull/1612/files#diff-f5e9723ed70ad2f79bc4adb3ba2d15d8R60

zipline/assets/asset_writer.py
yankees714: I'm leaning towards just equities, since we already segregate by asset type in all our existing tables holding this sort of data. Any thoughts?

richafrank: That's fine by me. If we want to add supplementary mappings to other asset classes later, we can always do that.

zipline/assets/assets.py
llllllllll: Because we were doing everything by mutation (and the object being mutated didn't escape until it was done) I passed it as the factory to make the valmap happen inplace.

zipline/assets/asset_writer.py
yankees714: Traced this back to https://github.com/quantopian/zipline/commit/ee9c79a6f849e4d7e9035bc58977bce125ef6f5e. There's no indication what the problem with the actual max int64 was, but this might not be an issue anymore since we no longer support 3.3.

llllllllll:
Other than the iinfo thing this lgtm

yankees714:
@llllllllll Cool thanks! I'm going to leave that as a separate patch - I'd like to see if we can reproduce the issue we saw previously.

PR #1617, opened 2016-12-16
ENH: Allow custom chunks iterators for attach_pipeline

llllllllll:


zipline/algorithm.py
```diff
@@ -2288,13 +2290,13 @@ def attach_pipeline(self, pipeline, name, chunksize=None):
...
+        if chunks is None:
             # Make the first chunk smaller to get more immediate results:
             # (one week, then every half year)
-            chunks = iter(chain([5], repeat(126)))
-        else:
-            chunks = iter(repeat(int(chunksize)))
-        self._pipelines[name] = pipeline, chunks
+            chunks = chain([5], repeat(126))
+        elif isinstance(chunks, int):
+            chunks = repeat(int(chunks))
```
richafrank: I suppose we don't need to call `int` on `chunks` here anymore.

tests/pipeline/test_pipeline_algo.py
```diff
@@ -285,21 +286,33 @@ def before_trading_start(context, data):
...
+            chunks = []
+            st = np.random.RandomState(12345)
+            remaining = (
+                self.dates.get_loc(self.last_asset_end) -
+                self.dates.get_loc(self.first_asset_start)
+            )
+            while remaining > 0:
+                chunk = st.randint(3)
+                chunks.append(chunk)
+                remaining -= chunk
```
richafrank: Subjective, but I think `chunks = [st.randint(3) for _ in range(remaining)]` might be clearer.

llllllllll: Talked IRL, this is slightly different. We are subtracting the random chunksize from remaining in each iteration.

PR #1619, opened 2016-12-16
MAINT: Define default end_date in assets.db as max int64, for clarity

yankees714:
We had previously reduced this value in https://github.com/quantopian/zipline/commit/ee9c79a6f849e4d7e9035bc58977bce125ef6f5e as a workaround, but it should no longer be an issue.

### DIFF
--- a/zipline/assets/asset_writer.py
+++ b/zipline/assets/asset_writer.py
@@ -66,7 +66,7 @@ _equities_defaults = {
     'symbol': None,
     'asset_name': None,
     'start_date': 0,
-    'end_date': 2 ** 62 - 1,
+    'end_date': np.iinfo(np.int64).max,
     'first_traded': None,
     'auto_close_date': None,
     # the canonical exchange name, like "NYSE"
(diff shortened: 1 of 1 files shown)


FreddieV4:
@yankees714 are you blocked on this?

yankees714:
@FreddieV4 This must have just slipped through the cracks, I'll rebase and merge if that's good with you?

zipline/assets/asset_writer.py
```diff
@@ -81,7 +81,7 @@
     'root_symbol': None,
     'asset_name': None,
     'start_date': 0,
-    'end_date': 2 ** 62 - 1,
+    'end_date': np.iinfo(np.int64).max,
```
FreddieV4: My only question here is why we are doubling the size of `end_date`?

```
In [24]: 2 ** 62 - 1
Out[24]: 4611686018427387903

In [25]: (2 ** 62) * 2 - 1
Out[25]: 9223372036854775807

In [26]: np.iinfo(np.int64).max
Out[26]: 9223372036854775807
```

FreddieV4: Ah. Nevermind. I just looked at your original comment in that we _used_ to have `2 ** 63 - 1`, which is just `np.int64`

yankees714: It essentially reverting ee9c79a, to make this more clear. There was apparently some compatibility issue at some point, but I wasn't able to reproduce it.

PR #1620, opened 2016-12-19
Adjustments to Component Dfs

kglowinski:


zipline/data/us_equity_pricing.py
```diff
@@ -1316,3 +1316,45 @@ def get_stock_dividends_with_ex_date(self, assets, date, asset_finder):
...
+                index_col='index',
+                parse_dates={col: 's' for col in date_cols}
+            )
+
+        return {
+            t_name: _get_df_from_table(
+                t_name,
+                KNOWN_DATE_COLS[t_name]
+            )
+            for t_name in KNOWN_DATE_COLS
```
yankees714: Could do:
```
for t_name, date_cols in KNOWN_DATE_COLS.iteritems()
```

zipline/data/us_equity_pricing.py
```diff
@@ -1316,3 +1316,45 @@ def get_stock_dividends_with_ex_date(self, assets, date, asset_finder):
...
+    def unpack_db_to_component_dfs(self):
+        """Returns the set of known tables in the adjustments file in DataFrame
+        form.
+
+        Returns
+        -------
+        dfs : dict{str->DataFrame}
+            Dictionary which maps table name to the corresponding DataFrame
+            version of the table, where all date columns have been coerced back
+             from into to datetime.
```
yankees714: Think there's something missing/extra in this sentence?

zipline/data/us_equity_pricing.py
```diff
@@ -41,8 +41,8 @@
     read_csv,
     Timestamp,
     NaT,
-    DatetimeIndex
-)
+    DatetimeIndex,
+    read_sql)
```
yankees714: Style nitpick, can we leave this paren on the next line?

zipline/data/us_equity_pricing.py
```diff
@@ -1316,3 +1329,33 @@ def get_stock_dividends_with_ex_date(self, assets, date, asset_finder):
...
+
+        def _get_df_from_table(table_name, date_cols):
+
+            # Dates are stored in second resolution as ints in adj.db tables.
+            return read_sql(
+                'select * from "{}"'.format(table_name),
+                self.conn,
+                index_col='index',
+                parse_dates={col: 's' for col in date_cols}
+            ).rename_axis(None)
```
kglowinski: Had to add a rename to `None` here, since we actually have an "index" column in the db itself, which retains it's name when set as the index.

kglowinski:
I object to the fact that adding a test has decreased the amount of coverage. That's not how this should work.

tests/pipeline/test_us_equity_pricing_loader.py
```diff
@@ -395,6 +396,40 @@ def test_load_adjustments_from_sqlite(self):
...
+            ('dividends', DIVIDENDS_EXPECTED)
+        )
+
+        def create_expected_table(df, name):
+            expected_df = df.copy()
+
+            for colname in reader._datetime_int_cols[name]:
+                expected_df[colname] = expected_df[colname].astype(
+                    'datetime64[s]'
+                )
```
yankees714: Not sure of the context where `unpack_db_to_component_dfs` is being used, but it this conversion something we could do within the method before returning?

zipline/data/us_equity_pricing.py
```diff
@@ -60,8 +61,8 @@
 from zipline.utils.functional import apply
 from zipline.utils.preprocess import call
 from zipline.utils.input_validation import (
-    preprocess,
     expect_element,
+    preprocess,
```
yankees714: 👍

zipline/data/us_equity_pricing.py
yankees714: Ah interesting (seems like we might not even need to be writing the index column in the first place, but not worth worrying about here)

kglowinski: I cannot imagine that we do want it. That change is something that I have added to the "Big List of Things that I Want to Change for Corporate Actions" that will be considered after Jan 1's switchover is done.

tests/pipeline/test_us_equity_pricing_loader.py
yankees714: Right, unless we'd prefer that convenience over getting the exact input that's accepted by the writer.

PR #1627, opened 2016-12-28
ENH: Add vectorized lookup_symbol.

ssanderson:


zipline/assets/assets.py
```diff
@@ -827,6 +827,34 @@ def lookup_symbol(self, symbol, as_of_date, fuzzy=False):
...
+            Forwarded to ``lookup_symbol``.
+        fuzzy : bool, optional
+            Forwarded to ``lookup_symbol``.
+
+        Returns
+        -------
+        equities : list[Equity]
+        """
+        memo = {}
+        out = []
```
llllllllll: should this just be a generator?

tests/test_assets.py
```diff
@@ -1475,3 +1476,73 @@ def select_fields(r):
...
+
+    @parameter_space(
+        as_of=pd.to_datetime([
+            '2014-01-02',
+            '2014-01-15',
+            '2014-01-17',
+            '2015-01-02',
+        ], utc=True),
+        symbols=[
+            ['A.'], ['B.'], ['C.'], ['D.'],
```
llllllllll: Why do you have an empty sc ticker? Also, should this just fail to map?

tests/test_assets.py
```diff
@@ -1475,3 +1476,73 @@ def select_fields(r):
...
+
+        with self.assertRaises(SymbolNotFound):
+            af.lookup_symbols(
+                syms,
+                pd.Timestamp('2014-01-15', tz='UTC'),
+                fuzzy=False,
+            )
+
+        results = af.lookup_symbols(syms, dt, fuzzy=True)
+        assert_equal(results, af.retrieve_all([1, 3, 7]))
```
llllllllll: should you compare against a comprehension again?

zipline/assets/assets.py
```diff
@@ -827,6 +827,34 @@ def lookup_symbol(self, symbol, as_of_date, fuzzy=False):
...
+        Parameters
+        ----------
+        symbols : sequence[str]
+            Sequence of ticker symbols to resolve.
+        as_of_date : pd.Timestamp
+            Forwarded to ``lookup_symbol``.
+        fuzzy : bool, optional
+            Forwarded to ``lookup_symbol``.
+
+        Returns
```
llllllllll: Maybe add a note saying that it is the same as the list comprehension but better for performance.

tests/test_assets.py
ssanderson: I'm not sure what you mean by "sc ticker".

llllllllll: shareclass, sorry.

zipline/assets/assets.py
ssanderson: Are you asking if the internal implementation should be a generator, or if the top-level interface should be a generator.  I'm indifferent toward the former.  w/r/t the latter, I can't think of a case where you'd want to call this and not have strict semantics, and returning a generator creates issues for using this with numpy/pandas, so I'd rather have this be eager.

tests/test_assets.py
ssanderson: Ah.  This is the current behavior of lookup_symbol, for better or worse.  I was a bit surprised as well that these resolve, but I don't want to change the semantics here.

llllllllll: Okay, can you not test on that and jus make these `A`, `B`, `C`, and `D` then?

zipline/assets/assets.py
llllllllll: I was talking about the top level impl, but this the numpy case makes sense.

llllllllll:
lgtm

Issue #1566, opened 2016-10-28
No data on or before day error for unknown stock


RamonBeast:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:
# Environment
- Operating System: Mac OS
- Python Version: 2.7.11
- Python Bitness: 64 bit
- How did you install Zipline: pip
- Python packages: `$ pip freeze` or `$ conda list`

Now that you know a little about me, let me tell you about the issue I am
having:

I was ingesting a mixed list of stocks from Russell2000 and SP500, I first incurred in the issue #1241 (hint: _BF_) that was solved after a painful bisection to find the stock that was giving error. Then after all the data was fetched I started to get this:

```
$ zipline ingest -b myindex
Merging daily equity files:  [---#--------------------------------]  2453 100%
Merging daily equity files:  [####################################]
/usr/local/lib/python2.7/site-packages/zipline/assets/asset_writer.py:192: FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)
  row = group.sort('end_date').iloc[-1]
Downloading Yahoo adjustment data:   [####################################]  100%
Traceback (most recent call last):
  File "/usr/local/bin/zipline", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/zipline/__main__.py", line 306, in ingest
    show_progress,
  File "/usr/local/lib/python2.7/site-packages/zipline/data/bundles/core.py", line 443, in ingest
    pth.data_path([name, timestr], environ=environ),
  File "/usr/local/lib/python2.7/site-packages/zipline/data/bundles/yahoo.py", line 177, in ingest
    adjustment_writer.write(splits=splits, dividends=dividends)
[...]

PR #1631, opened 2017-01-02
Handle missing data exceptions

snatchev:
This PR attempts to handle exceptions caused by missing data.

Multiple exceptions are raised in https://github.com/quantopian/zipline/blob/master/zipline/data/us_equity_pricing.py#L678-L686 but only one was being caught.

An example of this problem can be found in https://github.com/quantopian/zipline/issues/1566


FreddieV4:
Hi @snatchev thanks for opening this PR! This issue should be fixed already in the latest zipline release

It looks like this was fixed on master at the time of you opening this PR, but not in the previous release (`1.0.2`)

PR #1632, opened 2017-01-03
ENH: Add a reader writer pair for HDF5 minute bar updates.

ehebert:
This format is intended for storing data for all sids of an asset type,
e.g. equities or futures for a session. bcolz is not used to avoid the overhead
of creating the directories and files for each asset (which numbers around ~8000
for active equities) can be removed since the update is meant to be read at
once, instead of supporting the random access pattern needed by the simulation.

This patch only adds the reader/writer pair, with the management of finding the
paths to delta files and the application of the updates to the bcolz write left
to internal loader code.

Also, the update reader interface is intentionally constrained to the data for
an entire session to allow for an implementation that allows for mid-session updates.


zipline/data/minute_bars.py
```diff
@@ -1242,3 +1246,94 @@ def load_raw_arrays(self, fields, start_dt, end_dt, sids):
...
+        Write the frames to the target HDF5 file, using the format used by
+        ``pd.Panel.to_hdf``
+
+        Parameters
+        ----------
+        frames : iter[(int, DataFrame)] or dict[int -> DataFrame]
+            An iterable or other mapping of sid to the corresponding OHLCV
+            pricing data.
+        """
+        _store = HDFStore(
```
llllllllll: this can be used as a ctx manager

zipline/data/minute_bars.py
```diff
@@ -1242,3 +1246,94 @@ def load_raw_arrays(self, fields, start_dt, end_dt, sids):
...
+            The minutes for which to read the pricing updates.
+        sids : iter[int]
+            The sids for which to read the pricing updates.
+
+        Returns
+        -------
+        data : iter[(int, DataFrame)]
+            Returns an iterable of ``sid`` to the corresponding OHLCV data.
+        """
+        raise NotImplementedError
```
llllllllll: raise an instance instead of the type

zipline/data/minute_bars.py
```diff
@@ -1242,3 +1246,94 @@ def load_raw_arrays(self, fields, start_dt, end_dt, sids):
...
+class H5MinuteBarUpdateWriter(object):
+    """
+    Writer for files containing minute bar updates for consumption by a writer
+    for a ``MinuteBarReader`` format.
+
+    Parameters
+    ----------
+    path : str
+        The destination path.
+    complevel : int
```
llllllllll: `int, optional`

zipline/data/minute_bars.py
```diff
@@ -1242,3 +1246,94 @@ def load_raw_arrays(self, fields, start_dt, end_dt, sids):
...
+            An iterable or other mapping of sid to the corresponding OHLCV
+            pricing data.
+        """
+        _store = HDFStore(
+            self._path, 'w', complevel=self._complevel, complib=self._complib)
+        panel = pd.Panel.from_dict(dict(frames))
+        panel.to_hdf(_store, 'updates')
+        _store.close()
+        h5file = tables.open_file(self._path, mode='r+')
+        metadata = h5file.create_group('/', 'metadata')
```
llllllllll: can you not write metadata with HDFStore?

tests/data/test_minute_bars.py
```diff
@@ -1105,3 +1107,55 @@ def test_early_market_close(self):
...
+            },
+            index=minutes)
+
+        frames = {1: data_1, 2: data_2}
+        update_path = self.instance_tmpdir.getpath('updates.h5')
+        update_writer = H5MinuteBarUpdateWriter(update_path)
+        update_writer.write(frames)
+
+        update_reader = H5MinuteBarUpdateReader(update_path)
+        self.writer.write(update_reader.read(minutes, sids))
```
llllllllll: should we add a helper somewhere like `apply_updates`?

ehebert:
Just added `tables==3.3.0` to requirements to fix build issues.

zipline/data/minute_bars.py
ehebert: Fixing.

zipline/data/minute_bars.py
ehebert: Fixing.

zipline/data/minute_bars.py
ehebert: Could not find a way via HDFStore to set it on the file.
However looked through the docs for `h5file.File` again and found `set_node_attr` so using this instead:
```
with tables.open_file(self._path, mode='r+') as h5file:
       h5file.set_node_attr('/', 'version', 0)
```

tests/data/test_minute_bars.py
ehebert: I am open to it. Ok with my following up with it after dust settles on the review of internal branch (where the internal branch may be changed to use `apply_updates`, with a corresponding PR adding that function.)

PR #1633, opened 2017-01-04
comment out one line in blotter.py

ptoolis:
The cancel_all_orders_for_asset call can induce an assertion failure in the case of a reverse split and delisting.  See example below.
In fact, in running the test from January 2015 through November 26, 2015, the following scenario played out.
1.	 The stock was ordered on 8/25, 2 shares long
2.	The stock experienced a reverse split on 8/25 with a ratio of 7, making the order amount become 2/7 = 0
3.	The stock was halted on 11/25
4.	The past_auto_close_date test for the asset evaluated to true (Quandl set its end date to 11/24)
5.	The cleanup_expired_assets subroutine was invoked, calling cancel_all on the open order for Noranda
6.	Since the order amount was 0, its status was filled, so it was not deleted
7.	The “assert orders” line failed
 If the process finds that the current market data set is missing information about a stock, the order for that stock should be invalidated, but this should be done in a fashion which does not break the execution of assert statements.  In fact, though, the check on status of open in the open_orders list is redundant, and thus can be removed without loss of generality.

### DIFF
--- a/zipline/finance/blotter.py
+++ b/zipline/finance/blotter.py
@@ -75,29 +75,12 @@ class Blotter(object):
         self.current_dt = dt
 
     def order(self, sid, amount, style, order_id=None):
-        """Place an order.
 
-        Parameters
-        ----------
-        asset : zipline.assets.Asset
-            The asset that this order is for.
-        amount : int
-            The amount of shares to order. If ``amount`` is positive, this is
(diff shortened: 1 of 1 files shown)


ptoolis:
resubmitted as separate request with change applied to latest source file.

PR #1634, opened 2017-01-04
amend blotter.py to prevent splits causing 0 order amt, assert failure if stock halted  (was resubmit: comment out one line in blotter.py)

ptoolis:
Update:  This would work, if cancel were called only from the cancel_all_orders_asset method in blotter, but, as API calls can be dispatched to this method on any order id, it might be possible to cancel an order twice, updating the timestamp both times.  Thus, the correct approach to solve the reverse split issue is likely to cancel the order, if the split causes the amout to become zero. 
Initial submission modified an old copy of the file
The cancel_all_orders_for_asset call can induce an assertion failure in the case of a reverse split and delisting. See example below.
In fact, in running the test from January 2015 through November 26, 2015, the following scenario played out.

The stock was ordered on 8/25, 2 shares long
The stock experienced a reverse split on 8/25 with a ratio of 7, making the order amount become 2/7 = 0
The stock was halted on 11/25
The past_auto_close_date test for the asset evaluated to true (Quandl set its end date to 11/24)
The cleanup_expired_assets subroutine was invoked, calling cancel_all on the open order for Noranda
Since the order amount was 0, its status was filled, so it was not deleted
The “assert orders” line failed
If the process finds that the current market data set is missing information about a stock, the order for that stock should be invalidated, but this should be done in a fashion which does not break the execution of assert statements. In fact, though, the check on status of open in the open_orders list is redundant, and thus can be removed without loss of generality.
I did not rerun the test with the latest version of all files, but I see no changes to blotter or tradesimulation which would prevent this error from occurring.

### DIFF
--- a/zipline/finance/blotter.py
+++ b/zipline/finance/blotter.py
@@ -156,11 +156,11 @@ class Blotter(object):
         """
         return [self.order(*order_args) for order_args in order_arg_lists]
 
-    def cancel_on_split(self, order_id, relaya_status=True):
+    def cancel_on_split(self, order_id, relay_status=True):
         if order_id not in self.orders:
                 return
         cur_order = self.orders[order_id]
-        #no open check
+        # no open check
         order_list = self.open_orders[cur_order.sid]
         if cur_order in order_list:
             order_list.remove(cur_order)
@@ -325,7 +325,7 @@ class Blotter(object):
             orders_to_modify = self.open_orders[sid]
             for order in orders_to_modify:
                 order.handle_split(split[1])
-                if(order.amount==0):
+                if(order.amount == 0):
                     self.cancel_on_split(order.id)
 
     def get_transactions(self, bar_data):
--- a/zipline/finance/blotter.py
+++ b/zipline/finance/blotter.py
@@ -162,20 +162,20 @@ class Blotter(object):
 
         cur_order = self.orders[order_id]
 
-        #  if cur_order.open:
-        order_list = self.open_orders[cur_order.sid]
-        if cur_order in order_list:
-            order_list.remove(cur_order)
+        if cur_order.open:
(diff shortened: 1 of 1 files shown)


ptoolis:
Closing, pending a review of alternative solutions.  The problem solved is likely rarer than the multiple-cancel issue created.

ptoolis:
cancel is ineffective, as open will be false.  adding cancel_on_split as special case, whereby the open status will not be checked.

ptoolis:
It is unclear why the integration tests failed, but this commit did remedy the problem of a stock's experiencing a reverse split (causing order amount to be zero) and then being halted (causing the open orders for the stock to be deleted from the blotter) and experiencing an assertion failure.
The question of why an order is added to the blotter before the split is appllied (i.e. why the order amount and limit price are not considered to be post-split) is unclear, as it was my impression that splits are announced pre-open.

ptoolis:
I now understand why market orders were left unexecuted and the reverse split was applied to them (causing a zero order amount and an assertion failure during cancellation).  It seems that, although I scheduled the function to run one hour after the open using the time_rules.market_open(hours=1) function, the order was entered at 16:00 Eastern, which is the closing time of the NYSE. I am unsure about why such a phenomenon occurs.   Nonetheless, even orders entered after-hours should not induce an assertion failure.  The case is an extremely rare one (entered, not traded, reverse split to zero shares, stock delisted all in a single backtest), but I have not detected an issue with my solution which would upset other functionality.  Granted that a separate cancel just for handling the split case is somewhat inelegant, but I discerned no safer, more accurate solution.

ptoolis:
Only the final commit's source need be merged.  However, the comments from all commits are relevant.

ptoolis:
Closing this request.  Will reopen a new one which involves changing only one change in one file (blotter.py)

PR #1636, opened 2017-01-05
BLD: Update numexpr to latest to satisfy tables deps.

ehebert:
tables 3.3.0 requires numexpr >= 2.5.2


richafrank:
If this passes CI, then we should remove our numexpr conda recipe.

ehebert:
Just pushed removal of the `numexpr` recipe

PR #1637, opened 2017-01-06
BUG: reload_symbol_maps should clear the equity_supplementary_maps

llllllllll:


FreddieV4:
Hey @llllllllll could you give me an example of when this occurs/why we want to clear `equity_supplementary_maps`?

llllllllll:
This is useful when you are are writing to the underlying assets db. After each write you need to clear the internal caches.

FreddieV4:
Ok. I understand that. Do you think this warrants any tests to be added?

llllllllll:
I don't think the tests will add much value because it will not automatically ensure that this actually reloads the cache without tightly depending on the implementation of the asset finder. It's probably fine to omit them for now.

PR #1638, opened 2017-01-06
BUG: IchimokuKinkoHyo techinical factor has wrong default inputs

luca-s:
Very small bug fix: creating a IchimokuKinkoHyo factor raises an exception because the default inputs  are not in line with the factor's documentation

### DIFF
--- a/zipline/pipeline/factors/technical.py
+++ b/zipline/pipeline/factors/technical.py
@@ -598,7 +598,7 @@ class IchimokuKinkoHyo(CustomFactor):
         'kijun_sen_length': 26,
         'chikou_span_length': 26,
     }
-    inputs = USEquityPricing.high, USEquityPricing.close
+    inputs = (USEquityPricing.high, USEquityPricing.low, USEquityPricing.close)
     outputs = (
         'tenkan_sen',
         'kijun_sen',


zipline/pipeline/factors/technical.py
```diff
@@ -598,7 +598,7 @@ class IchimokuKinkoHyo(CustomFactor):
         'kijun_sen_length': 26,
         'chikou_span_length': 26,
     }
-    inputs = USEquityPricing.high, USEquityPricing.close
+    inputs = [USEquityPricing.high, USEquityPricing.low, USEquityPricing.close]
```
richafrank: Thanks @luca-s , makes sense to me! Could I ask for two improvements?

1. Would you make this a tuple of inputs instead of a list, so that no one can mistakenly modify this class-level attribute?
1. Would you add a simple test method to `tests.pipeline.test_technical.IchimokuKinkoHyoTestCase` which ensures that this factor doesn't blow up on computing with the default inputs, as you were seeing?

luca-s: I fixed "1" but it's not straightforward to add a test for "2", no other factor has that test.

luca-s: Just to be clear, as I was a little confusing: IchimokuKinkoHyo should have 3 default inputs but they added only 2, it was simply an oversight. I don't believe we need a test for this

richafrank: Ok, sounds fine. Thanks for the update!

PR #1643, opened 2017-01-09
Events loader handle empty raw data

mtydykov:
Fixes a bug where the events loader would break if the raw data it was trying to load happened to be empty.

### DIFF
--- a/tests/pipeline/test_events.py
+++ b/tests/pipeline/test_events.py
@@ -267,6 +267,88 @@ class EventIndexerTestCase(ZiplineTestCase):
                 self.assertEqual(computed_index, -1)
 
 
+class EventsLoaderEmptyTestCase(WithAssetFinder,
+                                WithTradingSessions,
+                                ZiplineTestCase):
+    START_DATE = pd.Timestamp('2014-01-01')
+    END_DATE = pd.Timestamp('2014-01-30')
+
+    @classmethod
+    def init_class_fixtures(cls):
+        cls.ASSET_FINDER_EQUITY_SIDS = [0, 1]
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_events.py
```diff
@@ -267,6 +267,65 @@ def check_next_event_indexer(self,
...
+            EventDataSet.previous_event_date: 'event_date',
+            EventDataSet.previous_float: 'float',
+            EventDataSet.previous_int: 'int',
+            EventDataSet.previous_string: 'string',
+            EventDataSet.previous_string_custom_missing: 'string'
+        }
+        cls.loader = EventsLoader(
+            cls.raw_events, cls.next_value_columns, cls.previous_value_columns
+        )
+        cls.ASSET_FINDER_EQUITY_SIDS = [0, 1]
```
mtydykov: True. Moved the events/loader config into the test method itself.

tests/pipeline/test_events.py
```diff
@@ -267,6 +267,65 @@ def check_next_event_indexer(self,
...
+        cls.ASSET_FINDER_EQUITY_SYMBOLS = ['A', 'B']
+        super(EventsLoaderEmptyTestCase, cls).init_class_fixtures()
```
mtydykov: So, the other config has been moved to the test method, but the call to super's `init_class_fixtures` does have to be after assigning `ASSET_FINDER_EQUITY_SIDS` and `ASSET_FINDER_EQUITY_SYMBOLS` because those need to be set before we call `WithAssetFinder`'s `init_class_fixtures`.

PR #1644, opened 2017-01-10
Don't allow ordering assets after their auto close date

dmichalowicz:
Futures contracts can currently be entered into after their `auto_close_date`. This change prevents that as well as makes `can_trade` return `False` after the `auto_close_date`.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -4294,12 +4294,21 @@ class TestOrderAfterDelist(WithTradingEnvironment, ZiplineTestCase):
     def make_equity_info(cls):
         return pd.DataFrame.from_dict(
             {
+                # Asset whose auto close date is after its end date.
                 1: {
                     'start_date': cls.start,
                     'end_date': cls.day_1,
                     'auto_close_date': cls.day_4,
                     'symbol': "ASSET1",
                     'exchange': "TEST",
+                },
+                # Asset whose auto close date is before its end date.
+                2: {
+                    'start_date': cls.start,
+                    'end_date': cls.day_4,
+                    'auto_close_date': cls.day_1,
+                    'symbol': 'ASSET2',
+                    'exchange': 'TEST',
                 },
             },
             orient='index',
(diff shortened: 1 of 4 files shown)


dmichalowicz:
@ehebert Can I give this to you for a look? Also I think we should discuss the logic in `_can_order_asset`, as the `auto_close_date` could either be before _or_ after the `end_date`.

zipline/_protocol.pyx
```diff
@@ -508,6 +508,9 @@ cdef class BarData:
             # asset isn't alive
             return False
 
+        if asset.auto_close_date and dt >= asset.auto_close_date:
```
richafrank: Should `asset.is_alive_for_session` return `False` instead?

dmichalowicz: Not sure; if `is_alive_for_session` is strictly meant to check if we're between the asset's start and end date then probably not, considering a future's auto close date is a few days before the end date (at which point it could still be called "alive"). If we are being more lenient though, then yes I'd say this logic could be moved to `is_alive_for_session`.

richafrank: Maybe just me, but I'm not sure when a consumer should use is_alive_for_session or check the auto-close or both.

tests/test_algorithm.py
```diff
@@ -4310,7 +4319,10 @@ def init_class_fixtures(cls):
         super(TestOrderAfterDelist, cls).init_class_fixtures()
         cls.data_portal = FakeDataPortal(cls.env)
 
-    def test_order_in_quiet_period(self):
+    @parameterized.expand([[1], [2]])
+    def test_order_in_quiet_period(self, sid):
```
ehebert: When using `expand`, it is useful to include a `name` parameter to show which condition is being tested.
e.g.:
```
@parameterized.expand([('auto_close_before_end_date', [1]), ('auto_close_after_end_date', [2])])
def test_order_in_quiet_period(self, name, sid):
    ...
```

zipline/_protocol.pyx
ehebert: Throughout `zipline.data.resample` `is_alive_for_session` is used to signify whether there should be data for a given a session. (Which is used to place `nan`s for sessions which evaluate to `False`.

Also, I believe this particular line will fail for futures between market open and midnight. Need to either get the market_open of the `auto_close_date`, or use the `session_label`.

tests/test_bar_data.py
```diff
@@ -1153,3 +1162,20 @@ def test_can_trade_restricted(self):
...
             )
             self.assertEqual(bar_data.can_trade(self.ASSET1), info[1])
+
+    def test_can_trade_delisted(self):
+        """
+        Test that can_trade returns False for an asset on or after its auto
+        close date.
+        """
+        dates_to_check = [
+            (pd.Timestamp('2016-01-05', tz='UTC'), True),
```
ehebert: Related to `https://github.com/quantopian/zipline/pull/1644/files#r95641660`, we should add tests that use minute dts.

dmichalowicz:
@ehebert Sorry just got around to updating this. Ready for another look.

PR #1649, opened 2017-01-17
Easier open minute bars for append and truncate

ehebert:
Add a method to open existing minute bar directory.

Remove need for a consumer that is editing an existing minute bars directory to
reread the values which should not change from the metadata.

Add a test to the append on new day and truncate, which would be the common
usage of this method.

Also, add `sorted` to truncate method. (Which was useful when testing the change with real data.)

### DIFF
--- a/tests/data/test_minute_bars.py
+++ b/tests/data/test_minute_bars.py
@@ -445,11 +445,16 @@ class BcolzMinuteBarTestCase(WithTradingCalendars,
             index=[first_minute])
         self.writer.write_sid(sid, data)
 
-        next_day_minute = first_minute + Timedelta(days=1)
+        # Open a new writer to cover `open` method, also a common usage
+        # of appending new days will be writing to an existing directory.
+        cday = self.trading_calendar.schedule.index.freq
+        new_end_session = TEST_CALENDAR_START + cday
+        writer = BcolzMinuteBarWriter.open(self.dest, new_end_session)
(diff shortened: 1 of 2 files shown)


zipline/data/minute_bars.py
```diff
@@ -478,6 +478,28 @@ def __init__(self,
...
             metadata.write(self._rootdir)
 
+    @classmethod
+    def open(cls, rootdir, end_session=None):
+        """
+        Open an existing ``rootdir`` for writing.
+
+        Parameters
+        ----------
+        - end_session : Timestamp (optional)
```
yankees714: Can we remove the `-` for consistency?

tests/data/test_minute_bars.py
```diff
@@ -445,11 +445,14 @@ def test_append_on_new_day(self):
             index=[first_minute])
         self.writer.write_sid(sid, data)
 
-        next_day_minute = first_minute + Timedelta(days=1)
+        cday = self.trading_calendar.schedule.index.freq
+        new_end_session = TEST_CALENDAR_START + cday
+        writer = BcolzMinuteBarWriter.open(self.dest, new_end_session)
```
yankees714: Can we add a comment explaining why we're creating a new writer here and below?

zipline/data/minute_bars.py
ehebert: Yes.

PR #1654, opened 2017-01-19
ENH: add current_session property to BarData

jbredeche:


zipline/_protocol.pyx
```diff
@@ -747,6 +747,12 @@ cdef class BarData:
         def __set__(self, val):
             self._adjust_minutes = val
 
+    property current_session:
+        def __get__(self):
+            return self._trading_calendar.minute_to_session_label(
```
ssanderson: Can we explicitly pass `direction='next'` here to clarify the behavior in `before_trading_start`?

tests/test_bar_data.py
```diff
@@ -204,6 +204,18 @@ def init_class_fixtures(cls):
 
         cls.ASSETS = [cls.ASSET1, cls.ASSET2]
 
+    def test_current_session(self):
+        for minute in self.trading_calendar.minutes_for_sessions_in_range(
```
ssanderson: We should verify the behavior outside of the session as well for BTS.  I think the right behavior there is to roll forward.

zipline/_protocol.pyx
jbredeche: yup, done

tests/test_bar_data.py
jbredeche: agreed, added.

PR #1657, opened 2017-01-24
ENH: Solidify equality comparisons for SlippageModel classes

jbredeche:


jbredeche:
(@richafrank signed off)

PR #1659, opened 2017-01-25
Multipliers were not removed when closing a position

dmichalowicz:
Due to relying on the order of the multipliers dicts, algorithms which mixed futures and equities were hitting the wrong multipliers after closing a position because the closing asset was not being removed from the multipliers dicts.

### DIFF
--- a/tests/test_perf_tracking.py
+++ b/tests/test_perf_tracking.py
@@ -2085,3 +2085,53 @@ class TestPositionTracker(WithTradingEnvironment,
         # Test gross and net exposures
         self.assertEqual(100 + 150000 + 200, pos_stats.gross_exposure)
         self.assertEqual(100 + 150000 - 200, pos_stats.net_exposure)
+
+    def test_close_position(self):
+        future_sid = 1032201401
+        equity_sid = 1
+        pt = perf.PositionTracker(self.env.asset_finder, None)
+        dt = pd.Timestamp('2017/01/04 3:00PM')
+
+        pos1 = perf.Position(
+            sid=future_sid,
+            amount=np.float64(30.0),
+            last_sale_date=dt,
+            last_sale_price=100,
+        )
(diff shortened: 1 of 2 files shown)


tests/test_perf_tracking.py
```diff
@@ -2085,3 +2085,62 @@ def test_update_positions(self):
...
+
+        # Update the positions dictionary with `future_sid` first. The order
+        # matters because it affects the multipliers dictionaries, which are
+        # OrderedDicts. If `future_sid` is not removed from the multipliers
+        # dictionaries, equities will hit the incorrect multiplier when
+        # computing `pt.stats()`.
+        pt.update_positions({future_sid: pos1, equity_sid: pos2})
+
+        for sid in (future_sid, equity_sid):
+            self.assertIn(sid, pt._position_value_multipliers)
```
ehebert: I'm 50/50 on this check.

Generally, it's best to avoid accessing semantically "private" members in tests, since those members are implementation details and could change in the future.

However, checking the value does get at the root cause of the bug.

Checking the `pt.stats()` values as you do below may be enough.

dmichalowicz: Yeah, I was hesitant to add tests for the "private" members as well. I am going to remove them because the tests for `pt.stats()` are still sufficient in that they ensure the expected values are not zero.

Also got some feedback from @ssanderson who similarly suggested I do a bit of refactoring in how these multipliers are set up. I'll merge this fix first though so as not to block on the immediate fix.

ehebert: Curious, what is the refactoring that he suggested?

ssanderson: My argument here was that we should just remove use of `OrderedDict` in `PositionTracker` and instead just align the data ourselves when we're calculating summary statistics.

PR #1661, opened 2017-01-26
Optionally apply deltas adjustments

mtydykov:
This PR makes applying deltas adjustments optional in the blaze core loader. In cases where the caller opts out of deltas adjustments, only novel deltas will be applied. By default, applying deltas adjustments is set to `True` because we only want to avoid applying them in special cases.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -1247,12 +1247,14 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
                       start,
                       end,
                       window_length,
-                      compute_fn):
+                      compute_fn,
+                      apply_deltas_adjustments=True):
         loader = BlazeLoader()
         ds = from_blaze(
             expr,
             deltas,
             checkpoints,
+            apply_deltas_adjustments=apply_deltas_adjustments,
             loader=loader,
             no_deltas_rule='raise',
             no_checkpoints_rule='ignore',
@@ -1480,7 +1482,7 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
             name='delta',
             dshape=self.dshape,
         )
-        expected_views = keymap(pd.Timestamp, {
+        expected_views_all_deltas = keymap(pd.Timestamp, {
             '2014-01-03': np.array([[10.0, 11.0, 12.0],
                                     [10.0, 11.0, 12.0],
                                     [10.0, 11.0, 12.0]]),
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_blaze.py
```diff
@@ -1289,6 +1291,26 @@ def compute(self, today, assets, out, data):
...
+        )
+
+    def _get_expected_output(self,
+                             expected_views,
+                             values,
+                             finder,
+                             asset_info,
+                             nassets):
+        return pd.DataFrame(
+            list(concatv(*[[value] * nassets for value in values])),
```
llllllllll: `concatv` is just:
```python
def concatv(*args):
    return concat(args)
```
so just pass the genexpr directly to concat.

tests/pipeline/test_blaze.py
```diff
@@ -1365,49 +1414,73 @@ def test_deltas_only_one_delta_in_universe(self, asset_info):
...
-                end=dates[-1],
-                window_length=2,
-                compute_fn=np.nanmax,
+                asset_info,
+                nassets
             )
 
+            dates = self.dates
+            dates = dates.insert(len(dates), dates[-1] + timedelta(days=1))
+            for apply_deltas_adjustments, expected_views, expected_output in (
```
llllllllll: Just define the iterator on the line above like:
```python
it = expr_that_spans_multiple_lines
for ... in it:
```

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -206,8 +206,13 @@
...
 
 
-class ExprData(namedtuple('ExprData', 'expr deltas checkpoints odo_kwargs')):
-    """A pair of expressions and data resources. The expresions will be
+class ExprData(
+    namedtuple(
+        'ExprData',
+        'expr deltas checkpoints odo_kwargs apply_deltas_adjustments'
+    )
+):
```
llllllllll: this makes it look like we are closing the class scope. define the super class above as `_expr_data_base`

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -547,20 +562,25 @@ def from_blaze(expr,
...
+        True by default because not applying deltas adjustments is an exception
+        rather than the rule.
     deltas : Expr, 'auto' or None, optional
         The expression to use for the point in time adjustments.
         If the string 'auto' is passed, a deltas expr will be looked up
         by stepping up the expression tree and looking for another field
         with the name of ``expr._name`` + '_deltas'. If None is passed, no
         deltas will be used.
-    deltas : Expr, 'auto' or None, optional
+    checkpoints : Expr, 'auto' or None, optional
```
llllllllll: ty

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -547,20 +562,25 @@ def from_blaze(expr,
...
-               no_checkpoints_rule='warn'):
+               no_checkpoints_rule='warn',
+               apply_deltas_adjustments=True,):
     """Create a Pipeline API object from a blaze expression.
 
     Parameters
     ----------
     expr : Expr
         The blaze expression to use.
+    apply_deltas_adjustments : bool, optional
```
llllllllll: the arguments in the docstring should be in the same order as the function signature

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1094,15 +1117,17 @@ def collect_expr(e, lower):
                                           have_sids=have_sids)
         ffill_across_cols(dense_output, columns, {c.name: c.name
                                                   for c in columns})
+        adjustments_from_deltas = lambda *args: {}
```
llllllllll: don't assign to a lambda, use `def`.

tests/pipeline/test_blaze.py
```diff
@@ -1547,14 +1679,22 @@ def test_novel_deltas_macro(self):
...
                                     [10.0],
                                     [11.0]]),
         })
+        expected_views_novel_deltas = keymap(pd.Timestamp, {
+            '2014-01-03': np.array([[0.0],
+                                    [0.0],
+                                    [0.0]]),
+            '2014-01-06': np.array([[0.0],
+                                    [0.0],
+                                    [11.0]]),
```
llllllllll: why do we get this delta but not the other one?

tests/pipeline/test_blaze.py
```diff
@@ -1480,22 +1575,42 @@ def test_novel_deltas(self, asset_info):
...
                                     [10.0, 11.0, 12.0],
                                     [11.0, 12.0, 13.0]]),
         })
+        expected_views_novel_deltas = keymap(pd.Timestamp, {
+            '2014-01-03': np.array([[0.0, 1.0, 2.0],
+                                    [0.0, 1.0, 2.0],
+                                    [0.0, 1.0, 2.0]]),
+            '2014-01-06': np.array([[0.0, 1.0, 2.0],
+                                    [0.0, 1.0, 2.0],
+                                    [11.0, 12.0, 13.0]]),
```
llllllllll: same question, why are these deltas novel but not the others?

tests/pipeline/test_blaze.py
```diff
@@ -1365,49 +1414,73 @@ def test_deltas_only_one_delta_in_universe(self, asset_info):
...
-        expected_views = keymap(pd.Timestamp, {
+        expected_views_all_deltas = keymap(pd.Timestamp, {
             '2014-01-02': np.array([[0.0, 11.0, 2.0],
                                     [1.0, 2.0, 3.0]]),
             '2014-01-03': np.array([[10.0, 2.0, 3.0],
                                     [2.0, 3.0, 4.0]]),
             '2014-01-04': np.array([[2.0, 3.0, 4.0],
                                     [2.0, 3.0, 4.0]]),
         })
+        expected_views_novel_deltas = keymap(pd.Timestamp, {
```
llllllllll: can you add a comment calling out the `(row_ix, col_ix)` of the difference so people don't miss it.

tests/pipeline/test_blaze.py
```diff
@@ -1315,46 +1337,73 @@ def test_deltas(self, asset_info, add_extra_sid):
...
-                end=dates[-1],
-                window_length=2,
-                compute_fn=np.nanmax,
+                asset_info,
+                nassets
             )
 
+            dates = self.dates
+            dates = dates.insert(len(dates), dates[-1] + timedelta(days=1))
+            for apply_deltas_adjustments, expected_views, expected_output in (
```
llllllllll: instead of running with and without the novel deltas for every test, can you just add one new test case that does this? Right now we are mixing this into all the cases when that is not the core part under test. This will also slow down the test suite a lot.

tests/pipeline/test_blaze.py
mtydykov: By the design of this test, that's the only novel delta here. It's the only case where the base data point came in on a weekend (Saturday, 2014-01-04) and the delta came in the day after, on a Sunday, so that it replaced the most recently known value before we would have had a chance to see it. The other deltas are for data points that we would have seen before seeing the deltas, so they're not novel deltas.

tests/pipeline/test_blaze.py
mtydykov: Same reason as above.

tests/pipeline/test_blaze.py
mtydykov: Sure, I can remove it from `test_deltas`, `test_deltas_only_one_delta_in_universe`,  and `test_deltas_macro`. Wondering if what I've added to `test_novel_deltas` and `test_novel_deltas_macro` already tests what we want, though? Because those tests had both non-novel and novel deltas to begin with, and all I really want to assert is that when we turn off non-novel deltas, that we get only exactly the novel deltas. Don't know what else a new test would assert.

llllllllll:
lgtm, just squash first.

PR #1665, opened 2017-01-27
Determine overwrite type dynamically

mtydykov:
This PR fixes a bug in the blaze core loader where any type of column was acceptable in the baseline table but overwrites for deltas adjustments were assumed to be of type float64. Addresses https://github.com/quantopian/qexec/issues/12024. 

Major changes are:

- Adds `ObjectOverwrite` to the dispatching done in `choose_adjustment_type`, replacing the case where getting a non-float64/datetime64 would raise a `ValueError`. Updates `test_adjustment` tests accordingly.
- Uses `make_adjustment_from_indices` to create overwrites for deltas adjustments in the blaze core loader, allowing overwrites for any column type.

### DIFF
--- a/tests/pipeline/test_adjusted_array.py
+++ b/tests/pipeline/test_adjusted_array.py
@@ -25,6 +25,7 @@ from zipline.lib.adjustment import (
     Float64Multiply,
     Float64Overwrite,
     Float641DArrayOverwrite,
+    Int64Overwrite,
     ObjectOverwrite,
 )
 from zipline.lib.adjusted_array import AdjustedArray, NOMASK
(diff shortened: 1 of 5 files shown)


mtydykov:
This breaks in python3 when we have int values in `test_novel_deltas_macro` with `BufferError: Object is not writable.` 

Looks like this is a bug in cython: http://stackoverflow.com/questions/28203670/how-to-use-cython-typed-memoryviews-to-accept-strings-from-python

So for the time being I'm going to put the adjustment.pyx code back to the way it was. The blaze core loader will still call `make_adjustment_from_indices` to create the overwrite object, but the caller will be restricted to floats/dates.

ssanderson:
LGTM

PR #1667, opened 2017-01-30
Pricing fixture cleanups

ssanderson:
Three semi-related changes here:

1. Add `session_open` and `session_close` methods to `TradingCalendar`.  These produce the same data as `open_and_close_for_session`, but are more ergonomic when you only need one of the two dates.
2. Fixes a bug (introduced by me during the pandas 18 upgrade) where we were failing to raise an error when a history window is misaligned with the dates of the history loader's calendar.
3. Makes EQUITY_DAILY_BAR_LOOKBACK_DAYS default to EQUITY_MINUTE_BAR_LOOKBACK_DAYS when EQUITY_DAILY_BAR_SOURCE_FROM_MINUTE is set.

### DIFF
--- a/zipline/utils/pandas_utils.py
+++ b/zipline/utils/pandas_utils.py
@@ -91,6 +91,39 @@ def mask_between_time(dts, start, end, include_start=True, include_end=True):
     )
 
 
+def find_in_sorted_index(dts, dt):
+    """
+    Find the index of ``dt`` in ``dts``.
+
+    This function should be used instead of `dts.get_loc(dt)` if the index is
+    large enough that we don't want to initialize a hash table in ``dts``. In
+    particular, this should always be used on minutely trading calendars.
+
+    Parameters
+    ----------
+    dts : pd.DatetimeIndex
+        Index in which to look up ``dt``. **Must be sorted**.
+    dt : pd.Timestamp
+        ``dt`` to be looked up.
+
+    Returns
+    -------
+    ix : int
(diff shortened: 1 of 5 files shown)


zipline/utils/pandas_utils.py
```diff
@@ -91,6 +91,39 @@ def mask_between_time(dts, start, end, include_start=True, include_end=True):
...
+
+    Raises
+    ------
+    KeyError
+        If dt is not in ``dts``.
+    """
+    ix = dts.searchsorted(dt)
+    if dts[ix] != dt:
+        raise KeyError(
+            "{0} is not in calendar [{1} ... {2}]".format(dt, dts[0], dts[-1])
```
ehebert: It would be a degenerate case, but the logging here fails if `dts` has no length.
Raising `"{0} is not in calendar {1}".format(dt, dts)` would fix that case, as well as providing extra information about the `DatetimeIndex`, e.g. the timezone, which is useful if this function is used generally, (anectdotallaly, a mismatch between timezones of the sought after `Timestamp` and the `DatetimeIndex` has been a head scratcher at least once.

ssanderson: Agreed that this shouldn't die if the index is of length zero.  Upon looking at this further, I noticed that the errors generated here are pretty unpleasant because `KeyError` `repr`s its input argument, which means multi-line strings get printed with `\n` characters instead of strings.  I'm seeing if I can update use a different error here so that these errors are actually readable.

PR #1671, opened 2017-02-02
Change to a 10.5 hour futures calendar

dmichalowicz:
Change algorithms that use the futures calendar to run from 6:30am to 5:00pm instead of 24 hours. Minutely history calls will still return 24-hour data however.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -3502,6 +3502,115 @@ class TestFutureFlip(WithDataPortal, WithSimParams, ZiplineTestCase):
                 format(i, actual_position, expected_positions[i]))
 
 
+class TestFuturesAlgo(WithDataPortal, WithSimParams, ZiplineTestCase):
+    START_DATE = pd.Timestamp('2016-01-06', tz='utc')
+    END_DATE = pd.Timestamp('2016-01-07', tz='utc')
+    FUTURE_MINUTE_BAR_START_DATE = pd.Timestamp('2016-01-05', tz='UTC')
+
+    SIM_PARAMS_DATA_FREQUENCY = 'minute'
+
+    TRADING_CALENDAR_STRS = ('us_futures',)
+    TRADING_CALENDAR_PRIMARY_CAL = 'us_futures'
+
+    @classmethod
+    def make_futures_info(cls):
+        return pd.DataFrame.from_dict(
+            {
+                1: {
+                    'symbol': 'CLG16',
+                    'root_symbol': 'CL',
+                    'start_date': pd.Timestamp('2015-12-01', tz='UTC'),
+                    'notice_date': pd.Timestamp('2016-01-20', tz='UTC'),
+                    'expiration_date': pd.Timestamp('2016-02-19', tz='UTC'),
+                    'auto_close_date': pd.Timestamp('2016-01-18', tz='UTC'),
+                    'exchange': 'TEST',
+                },
+            },
+            orient='index',
+        )
+
+    def test_futures_history(self):
+        algo_code = dedent(
+            """
+            from datetime import time
+            from zipline.api import (
+                date_rules,
+                get_datetime,
+                schedule_function,
+                sid,
+                time_rules,
+            )
+
+            def initialize(context):
+                context.history_values = []
+
+                schedule_function(
+                    make_history_call,
+                    date_rules.every_day(),
+                    time_rules.market_open(),
+                )
+
+                schedule_function(
+                    check_market_close_time,
+                    date_rules.every_day(),
+                    time_rules.market_close(),
+                )
+
+            def make_history_call(context, data):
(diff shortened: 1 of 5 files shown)


zipline/utils/events.py
```diff
@@ -353,6 +357,13 @@ def calculate_dates(self, dt):
             self.cal.open_and_close_for_session(
                 self.cal.minute_to_session_label(dt)
             )
+
+        # When using the futures calendar, adjust the open and close times to
+        # conform to 6:30am to 5:00pm.
+        if self.cal.name == 'us_futures':
```
jbredeche: How about we add `execution_open` and `execution_close` to the base calendar API. By default it's just the regular open and close, but a calendar can override to provide custom functionality (requiring that execution open is after the regular open, etc).

Then you can just get all the execution opens by asking the calendar.

dmichalowicz:
@richafrank Ready for a look

zipline/utils/calendars/us_futures_calendar.py
```diff
@@ -13,6 +13,12 @@
     Christmas
 )
 
+# Number of hours of offset between the open and close times dictated by this
+# calendar versus the 6:30am to 5:00pm times over which we want to simulate
+# futures algos.
+FUTURES_OPEN_TIME_OFFSET = 12.5
+FUTURES_CLOSE_TIME_OFFSET = -1
```
dmichalowicz: Should I just put these numbers straight in the methods below instead of bothering to define them as constants?

zipline/algorithm.py
```diff
@@ -531,6 +531,19 @@ def _create_clock(self):
             # as the last minute of the session.
             market_opens = market_closes
 
+        # The calendar's execution times are the minutes over which we actually
+        # want to run the clock. Typically the execution times simply adhere to
+        # the market open and close times. In the case of the futures calendar,
+        # for example, we only want to simulate over a subset of the full 24
+        # hour calendar, so the execution times are used to artificially create
```
dmichalowicz: I guess the word artificial doesn't fit here anymore, it made more sense when I was calculating the timedeltas here because we were manually interfering with the simulation times. But since the behavior was moved, it's no longer like an interference, it's just part of the calendar interface

tests/test_algorithm.py
```diff
@@ -3502,6 +3502,96 @@ def check_algo_positions(self, results, expected_positions):
...
+        algo = TradingAlgorithm(
+            script=algo_code,
+            sim_params=self.sim_params,
+            env=self.env,
+            trading_calendar=get_calendar('us_futures'),
+        )
+        algo.run(self.data_portal)
+
+        # Assert that we were able to retrieve history data for minutes outside
+        # of the 6:31am to 5:00pm futures open times.
```
dmichalowicz: > Could you specify the timezone in your comment? 

Yes thanks for pointing that out.

For your second question, I do assert that `get_datetime()` is 6:31 in my scheduled function, which runs at market open. I could add some kind of assertion about market close to add the loop, or just come up with something more comprehensive to fully test that.

richafrank: Ah, I missed that `assert get_datetime()`! Could you add the tz to your comment there too 😄 ?

zipline/utils/calendars/us_futures_calendar.py
```diff
@@ -63,6 +69,12 @@ def close_time(self):
     def open_offset(self):
         return -1
 
+    def execution_time_from_open(self, open_date):
+        return open_date + Timedelta(hours=FUTURES_OPEN_TIME_OFFSET)
+
+    def execution_time_from_close(self, close_date):
+        return close_date + Timedelta(hours=FUTURES_CLOSE_TIME_OFFSET)
```
dmichalowicz: @richafrank Regarding your performance question, so should I just have these methods return a Timedelta and then do the addition in `algorithm.py`?

richafrank: That would work. Do you want to make the interface the offset itself or a method? If you want to implement it as addition here, you could pass the series to this function instead of passing this function to `map`.

tests/test_algorithm.py
```diff
@@ -3502,6 +3502,109 @@ def check_algo_positions(self, results, expected_positions):
...
+        np.testing.assert_array_equal(
+            algo.history_values[0].index,
+            pd.date_range(
+                '2016-01-06 11:27', '2016-01-06 11:31', freq='min', tz='UTC',
+            ),
+        )
+        np.testing.assert_array_equal(
+            algo.history_values[1].index,
+            pd.date_range(
+                '2016-01-07 11:27', '2016-01-07 11:31', freq='min', tz='UTC',
```
richafrank: What do you think about making these literals also in Eastern?

dmichalowicz: Missed these, I will do that

PR #1672, opened 2017-02-03
narrow labelarray

llllllllll:


tests/test_labelarray.py
```diff
@@ -337,3 +339,37 @@ def test_setitem_array(self):
...
+            return [
+                ''.join(cs)
+                for cs in take(
+                    2 ** width + plus_one,
+                    product([chr(c) for c in range(256)], repeat=length),
+                )
+            ]
+
+        #  uint8
+        categories = create_categories(8, plus_one=False)
```
ssanderson: Is this testing that we use a uint8 when we have exactly 8 unique categories?  I think that should only be true if the categories already includes the missing value, otherwise we'll overflow on https://github.com/quantopian/zipline/pull/1672/files#diff-feac3c74134e7ddfc5a86bb82804193cL21.

llllllllll: The argument there is `width`, not `length`.

```
(Pdb) p len(create_categories(8, False))
256
(Pdb) p len(create_categories(8, True))
257
```
The missing value is in the categories by construction.

Also, a uint8 holds 256 values, not 8 ;)

tests/test_labelarray.py
```diff
@@ -337,3 +339,68 @@ def test_setitem_array(self):
...
+        # uint8
+        categories = create_categories(8, plus_one=False)
+        arr = LabelArray(
+            [],
+            missing_value=categories[0],
+            categories=categories,
+        )
+        self.assertEqual(arr.itemsize, 1)
+
+        # uint8 inference
```
ssanderson: For these cases, can we confirm that the label array with the inferred categories round-trips correctly back to strings?  Right now these tests are just asserting that the size is expected, but if we messed up our expected size calculations, then it's possible that we could overflow/underflow in the array, which would cause us to produce incorrect values.

ssanderson:
@llllllllll this looks good.  Just had one request for an extra piece of test coverage.

tests/test_labelarray.py
llllllllll: I am not sure what the failure mode is. The test is asserting that the size is correct so I don't see how it could be (over|under)flowing. Are you saying that the test might be checking this incorrectly?

ssanderson: I'm saying we might be incorrectly calculating the cutoffs in this test and in our code, which would mean we're asserting that the size should be 2, and it is 2, but it should have been 4.

PR #1676, opened 2017-02-07
TEST: Allow parameter_space to work on repeated calls of test

lianga888:
If we have a test that's being called more than once (i.e. two
test cases, both subclasses of the same base test case, with
different setup but calling the same test), allow the subsequent
calls to re-consume the same params


llllllllll:
My only concern is that this will use a lot more ram. Could we move the `product` call into `wrapper` instead?

PR #1677, opened 2017-02-07
MAINT: Removes unnecessary capital_base arg to TradingAlgorithm

yankees714:
Capital base is included in the sim params, so we should define the
value there, or use the default.

This change also unifies the default capital base as 1e5, as was
previously defined in algorithm.py.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -144,8 +144,6 @@ from zipline.gens.sim_engine import MinuteSimulationClock
 from zipline.sources.benchmark_source import BenchmarkSource
 from zipline.zipline_warnings import ZiplineDeprecationWarning
 
-DEFAULT_CAPITAL_BASE = 1e5
-
 
 log = logbook.Logger("ZiplineLog")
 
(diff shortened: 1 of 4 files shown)


zipline/algorithm.py
```diff
@@ -303,16 +299,14 @@ def __init__(self, *args, **kwargs):
...
-        self.capital_base = kwargs.pop('capital_base', DEFAULT_CAPITAL_BASE)
         self.sim_params = kwargs.pop('sim_params', None)
         if self.sim_params is None:
             self.sim_params = create_simulation_parameters(
-                capital_base=self.capital_base,
                 start=kwargs.pop('start', None),
                 end=kwargs.pop('end', None),
                 trading_calendar=self.trading_calendar,
             )
+        self.capital_base = self.sim_params.capital_base
```
richafrank: Looks like this is only used for the `__repr__` now. Could we kill it and have that method reference `self.sim_params.capital_base` instead, so we have one less piece of state to keep track of?

zipline/finance/trading.py
```diff
@@ -128,7 +131,7 @@ def write_data(self, **kwargs):
 class SimulationParameters(object):
     def __init__(self, start_session, end_session,
                  trading_calendar,
-                 capital_base=10e3,
+                 capital_base=DEFAULT_CAPITAL_BASE,
```
richafrank: Not something to change right now, but I think eventually we'll get rid of this default entirely. I think of this class as internal now that users have `run_algorithm`.

zipline/algorithm.py
yankees714: Sounds good!

PR #1679, opened 2017-02-08
TST: Adds TestingSlippage slippage model

yankees714:
Allows specifying a constant number of shares filled per tick.

### DIFF
--- a/zipline/testing/slippage.py
+++ b/zipline/testing/slippage.py
@@ -1,0 +1,32 @@
+from zipline.finance.slippage import SlippageModel
+from zipline.utils.sentinel import sentinel
+
+
+class TestingSlippage(SlippageModel):
+    """
+    Slippage model that fills a constant number of shares per tick, for
+    testing purposes.
+
+    Parameters
+    ----------
+    filled_per_tick : int or TestingSlippage.ALL
+        The number of shares to fill on each call to process_order. If
+        TestingSlippage.ALL is passed, the entire order is filled.
+
+    See also
+    --------
+    zipline.finance.slippage.SlippageModel
+    """
+    ALL = sentinel('ALL')
+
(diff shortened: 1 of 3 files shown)


zipline/testing/slippage.py
```diff
@@ -0,0 +1,32 @@
...
+    ----------
+    filled_per_tick : int or TestingSlippage.ALL
+        The number of shares to fill on each call to process_order. If
+        TestingSlippage.ALL is passed, the entire order is filled.
+
+    See also
+    --------
+    zipline.finance.slippage.SlippageModel
+    """
+    ALL = sentinel('All')
```
ssanderson: In most other places, we use the same name for the sentinel value and the string we pass for its repr.  Any reason not to do that here as well?

zipline/testing/slippage.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(self, filled_per_tick):
+        self.filled_per_tick = filled_per_tick
+
+    def process_order(self, data, order):
+        price = data.current(order.asset, "close")
+        if self.filled_per_tick == self.ALL:
```
ssanderson: For `sentinel` instances `is` is the preferred check.

yankees714: Makes sense 👍

zipline/testing/slippage.py
```diff
@@ -0,0 +1,32 @@
+from zipline.finance.slippage import SlippageModel
+from zipline.utils.sentinel import sentinel
+
+
+class TestingSlippage(SlippageModel):
```
ssanderson: We should probably add a test that exercises this if it's going to live in zipline.

ssanderson:
lgtm

PR #1680, opened 2017-02-09
Eliminate Possible Bias in Volume Rolls

dmichalowicz:
In backtests, volume roll dates are determined by looking back from each auto close date. This introduces potential look ahead bias because if there was ever a double volume flip within the last few days before an auto close, we choose the flip on the later date. Instead we should choose the earlier date because from the perspective of someone seeing the early flip in real time, and being within only a few days of the auto close date, they would likely believe it to be the actual roll date.

### DIFF
--- a/zipline/assets/roll_finder.py
+++ b/zipline/assets/roll_finder.py
@@ -48,7 +48,6 @@ class RollFinder(with_metaclass(ABCMeta, object)):
         back = oc.contract_at_offset(front, 1, dt.value)
         if back is None:
             return front
-        session = self.trading_calendar.minute_to_session_label(dt)
         primary = self._active_contract(oc, front, back, session)
         return oc.contract_at_offset(primary, offset, session.value)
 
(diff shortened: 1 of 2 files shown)


zipline/assets/roll_finder.py
```diff
@@ -146,12 +145,54 @@ def __init__(self, trading_calendar, asset_finder, session_reader):
...
-        if back_vol > front_vol:
+        """
+        Return the active contract based on the previous trading day's volume.
+
+        In the rare case that a double volume switch occurs we treat the first
+        switch as the roll. Take the following case for example:
+
+        | _____             _____
+        |      \   __      /       <--- 'G'
+        |       \_/__\____/__
```
jbredeche: Holy wow that's awesome ASCII art

ehebert: @dmichalowicz could you check in the changes to switch to `+` for the `F` contract?

dmichalowicz: @ehebert all set

zipline/assets/roll_finder.py
```diff
@@ -106,6 +105,9 @@ def get_rolls(self, root_symbol, start, end, offset):
                     if prev < prev_c.contract.auto_close_date:
                         break
                 if back != self._active_contract(oc, front, back, prev):
+                    # TODO: Instead of listing each contract with its roll date
```
ehebert: :+1:

PR #1683, opened 2017-02-13
bugfix: convert index to int so that the pandas indexing doesnt fail

shadiakiki1986:
otherwise, the index could come out as a float because of the division ahead of it

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -549,7 +549,9 @@ class BcolzMinuteBarWriter(object):
         with open(sizes_path, mode='r') as f:
             sizes = f.read()
         data = json.loads(sizes)
-        num_days = data['shape'][0] / self._minutes_per_day
+        # use integer division so that the result is an int
+        # for pandas index later https://github.com/pandas-dev/pandas/blob/master/pandas/tseries/base.py#L247 # noqa
+        num_days = data['shape'][0] // self._minutes_per_day
         if num_days == 0:
             # empty container
             return pd.NaT


FreddieV4:
Hi @shadiakiki1986 thanks for the PR. I was wondering if you could show a case where you caught this so I could try and reproduce it, or how you caught this

shadiakiki1986:
Hello. I'm wrapping zipline's blotter class in a django application [here](https://github.com/shadiakiki1986/django-zipline/). I ran into this issue by pure coincidence. The division yielded a float. I'm not sure where exactly in my code it was triggered, but it might be [here](https://github.com/shadiakiki1986/django-zipline/blob/master/zipline_app/matcher.py#L314)

zipline/data/minute_bars.py
```diff
@@ -550,9 +550,14 @@ def last_date_in_output_for_sid(self, sid):
...
         data = json.loads(sizes)
         num_days = data['shape'][0] / self._minutes_per_day
+
         if num_days == 0:
             # empty container
             return pd.NaT
+
+        # cast to int for pandas
+        # https://github.com/pandas-dev/pandas/blob/master/pandas/tseries/base.py#L247
+        num_days = int(num_days)
```
FreddieV4: I think that if we're going to convert to an int, we should be doing this earlier, before we even check for `num_days == 0`

We can use the `//` operator which does integer division in both Python 2 and 3, and do that on line 552:

`num_days = data['shape'][0] // self._minutes_per_day`

shadiakiki1986: That's a better idea. I just pushed a new commit amending the PR

zipline/data/minute_bars.py
```diff
@@ -549,7 +549,9 @@ def last_date_in_output_for_sid(self, sid):
         with open(sizes_path, mode='r') as f:
             sizes = f.read()
         data = json.loads(sizes)
-        num_days = data['shape'][0] / self._minutes_per_day
+        # use integer division so that the result is an int for pandas index later
+        # https://github.com/pandas-dev/pandas/blob/master/pandas/tseries/base.py#L247
```
FreddieV4: Looks like the flake8 tests failed because line 552 is too long (and probably 553 as well)

shadiakiki1986: shortened the lines

zipline/data/minute_bars.py
```diff
@@ -549,7 +549,9 @@ def last_date_in_output_for_sid(self, sid):
         with open(sizes_path, mode='r') as f:
             sizes = f.read()
         data = json.loads(sizes)
-        num_days = data['shape'][0] / self._minutes_per_day
+        # use integer division so that the result is an int
+        # for pandas index later https://goo.gl/VAW649
```
FreddieV4: Ok great. Just one last thing; we should keep this link expanded (so people know it's a safe link to click). You can make the flake8 tests ignore the length of this line by adding `# noqa` line 553. So it would look like:

`# for pandas index later <link> # noqa`

Should've mentioned that before, sorry! (after that it should be good to merge)

FreddieV4: Also you should squash your commits into 1, and change the commit message to say **BUG: <description>** (to be consistent with our dev guidelines https://github.com/quantopian/zipline/wiki/Zipline-development-guidelines)

shadiakiki1986: No problem :) Done and rebased

shadiakiki1986:
Btw, [here](https://travis-ci.org/shadiakiki1986/django-zipline#L2017) is my travis build which demonstrates this issue

FreddieV4:
Looks good to merge 👍 thanks for the PR

PR #1684, opened 2017-02-15
BUG: fix loader bug for 1 day

mtydykov:
Fixes a bug in the earnings estimates loader where if the only 1 day was requested in the Pipeline run, there would be no overwrites (overwrites for 1 day are impossible), and there are no split adjustments (as is the case for revenue estimates), the overwrites dictionary would be completely empty and would result in a key error. Adds tests for this.

### DIFF
--- a/zipline/pipeline/loaders/earnings_estimates.py
+++ b/zipline/pipeline/loaders/earnings_estimates.py
@@ -318,6 +318,11 @@ class EarningsEstimatesLoader(PipelineLoader):
         sid : int
             The sid for which overwrites should be computed.
         """
+        # If data was requested for only 1 date, there can never be any
+        # overwrites, so skip the extra work.
+        if len(dates) == 1:
+            return
+
         next_qtr_start_indices = dates.searchsorted(
             group[EVENT_DATE_FIELD_NAME].values,
             side=self.searchsorted_side,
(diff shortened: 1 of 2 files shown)


zipline/pipeline/loaders/earnings_estimates.py
```diff
@@ -318,35 +318,38 @@ def collect_overwrites_for_sid(self,
         sid : int
             The sid for which overwrites should be computed.
         """
-        next_qtr_start_indices = dates.searchsorted(
-            group[EVENT_DATE_FIELD_NAME].values,
-            side=self.searchsorted_side,
-        )
+        # If data was requested for only 1 date, there can never be any
+        # overwrites, so skip the extra work.
+        if len(dates) > 1:
```
ehebert: We could also do:

```
# Assuming `dates` of `0` is never passed her.
if len(dates) == 1:
    return
```

or:
```
if len(dates) <= 1:
    return
```

Which would remove the need for the extra level indentation.
I feel the suggested version is more explicit about bailing/short-circuiting on the case of only one day, but could either way.

mtydykov: Ah, I like this much more than what I have. Thanks!

PR #1691, opened 2017-02-28
DOC: Fix typo in -k/--keep-last option description

degtyarevalexey:
-b is shortopt for --bundle, not --before:

```
$ zipline clean --help
Usage: zipline clean [OPTIONS]
...

  -b, --bundle BUNDLE-NAME  The data bundle to clean.  [default: quantopian-
                            quandl]
  -e, --before TIMESTAMP    Clear all data before TIMESTAMP. This may not be
                            passed with -k / --keep-last
...
  -k, --keep-last N         Clear all but the last N downloads. This may not
                            be passed with -b / --before or -a / --after
```


richafrank:
Thanks @degtyarevalexey - Good catch! Merged with https://github.com/quantopian/zipline/commit/02c14583287bb03e08075fb2abec29750a77d65f.

PR #1694, opened 2017-03-01
ENH: Make invalid data behavior optional

ahgnaw:
It is sometimes necessary to test invalid test data, for example in the case that there is missing pricing data. Instead of warning by default, make invalid data behavior configurable by the test if needed.


ssanderson:
LGTM

PR #1696, opened 2017-03-03
ENH: Add winsorize factor

ahgnaw:
This factor is useful when the effects of spurious outliers to be mimized.

### DIFF
--- a/tests/pipeline/test_factor.py
+++ b/tests/pipeline/test_factor.py
@@ -24,7 +24,7 @@ from numpy.random import randn, seed
 import pandas as pd
 from scipy.stats.mstats import winsorize as scipy_winsorize
 
-from zipline.errors import UnknownRankMethod
+from zipline.errors import BadPercentileBounds, UnknownRankMethod
 from zipline.lib.labelarray import LabelArray
 from zipline.lib.rank import masked_rankdata_2d
 from zipline.lib.normalize import naive_grouped_rowwise_apply as grouped_apply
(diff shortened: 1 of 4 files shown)


zipline/pipeline/factors/factor.py
```diff
@@ -833,6 +834,84 @@ def linear_regression(self, target, regression_length, mask=NotSpecified):
...
+            ),
+            'winsor_alt': price.winsorize(limits=0.25, inclusive=(True, False))
+
+        }
+
+        Given a pipeline with columns, defined above, the result for a
+        given day could look like:
+
+                'price'  'winsor_inc' 'winsor_exc'  'winsor_alt'
+        Asset_1    1           2           3             2
```
ssanderson: This would be a good example to turn into a test.  We something similar for `demean` in `test_normalizations_hand_computed` (unfortunately, it looks like we're missing the analogous test for `zscore` :( ).

zipline/pipeline/factors/factor.py
```diff
@@ -833,6 +834,84 @@ def linear_regression(self, target, regression_length, mask=NotSpecified):
             mask=mask,
         )
 
+    def winsorize(self,
```
ahgnaw: Clamp is slightly different than winsorize. In clamping, the upper and lower bounds are provided, in winsorizing they are derived from the data set. For example the lower limit in inclusive winsorizing in generated by arr[int(limit*len(arr))].

zipline/pipeline/factors/factor.py
```diff
@@ -833,6 +834,84 @@ def linear_regression(self, target, regression_length, mask=NotSpecified):
...
+                  limits=0.05,
```
ahgnaw: Ok

zipline/pipeline/factors/factor.py
```diff
@@ -833,6 +834,84 @@ def linear_regression(self, target, regression_length, mask=NotSpecified):
...
+                  inclusive=(False, False),
+                  mask=NotSpecified,
+                  groupby=NotSpecified):
+        """
+        Construct a Factor returns a winsorized row for results. This is useful
```
ahgnaw: The bounds are not clamped by a fixed percentile, rather they are generated from the input array.

zipline/pipeline/factors/factor.py
```diff
@@ -833,6 +834,84 @@ def linear_regression(self, target, regression_length, mask=NotSpecified):
...
+
+        Example
+        -------
+
+        price = USEquityPricing
+        columns={
+            'price': price.latest,
+            'winsor_inc': price.winsorize(limits=0.25, inclusive=(True, True)),
+            'winsor_exc': price.winsorize(
+                limits=0.25, inclusive=(False, False)
```
ahgnaw: Yeah I think changing the behavior of limits makes sense.

tests/pipeline/test_factor.py
```diff
@@ -1062,6 +1168,11 @@ def test_demean_is_window_safe_if_input_is_window_safe(self):
         self.assertFalse(F(window_safe=False).demean().window_safe)
         self.assertTrue(F(window_safe=True).demean().window_safe)
 
+    def test_winsorize_is_window_safe(self):
```
ssanderson: Seems like this is no longer the right name for this test.

zipline/pipeline/factors/factor.py
```diff
@@ -1530,3 +1615,20 @@ def demean(row):
 
 def zscore(row):
     return (row - nanmean(row)) / nanstd(row)
+
+
+def winsorize(row, limits, inclusive):
+    if isinstance(limits, int) or isinstance(limits, float):
```
ssanderson: Rather than doing this logic on every row we process, we should just convert our args into the format expected by `scipy_winsorize` at term construction time in `Factor.winsorize`.

zipline/pipeline/factors/factor.py
```diff
@@ -833,6 +834,90 @@ def linear_regression(self, target, regression_length, mask=NotSpecified):
...
+        Parameters
+        ----------
+        limits : None, tuple of float, optional
+            A tuple of two values between 0 and 100 inclusive. This is the
+            percentage to cut from each tail of the array. A value of None
+            can be used to indicate an open limit.
+        inclusive : a tuple of bool, optional
+            A bool indicating whether the data on each side should be
+            rounded(True) or truncated(False). A value of None can be used if
+            one side is not being winsorized. Default is (False, False).
```
ssanderson: The default listed here doesn't match the one in the function signature.

zipline/pipeline/factors/factor.py
```diff
@@ -833,6 +834,90 @@ def linear_regression(self, target, regression_length, mask=NotSpecified):
...
+        when computing row means and standard deviations, and output NaN
+        anywhere the mask is False.
+
+        If ``groupby`` is supplied, compute by partitioning each row based on
+        the values produced by ``groupby``, winsorizing the partitioned arrays,
+        and stitching the sub-results back together.
+
+        Parameters
+        ----------
+        limits : None, tuple of float, optional
```
ssanderson: Reading through the examples from the tests, I'm still not thrilled with this API for limits.  Reading `f.winsorize(limits=33)` in the tests, for example, feels awkward to me.  It's not clear in that usage that `limits` is a percentile and not an absolute value (an issue that would persist even for more reasonable limits like `5`).

I spent a while thinking about this without coming up with any suggestions that I really liked.  Let's talk about this first thing tomorrow and figure out a plan so that we can get this merged.

ssanderson: One concrete note: I don't think there's any case where it's useful for us to accept `None` for the boundary parameter.

ssanderson:
@ahgnaw changes here generally look good.  I'm still not sold on our current API for describing the boundaries for winsorization.  Let's talk early tomorrow and flesh those out.

zipline/pipeline/factors/factor.py
ahgnaw: GP

tests/pipeline/test_factor.py
```diff
@@ -709,11 +710,156 @@ def test_normalizations_hand_computed(self):
...
+        """
+        Test out of bounds input for factor.winsorize.
+        """
+        f = self.f
+
+        bad_percentiles = [
+            (-.1, 1),
+            (0, 95),
+            (5, 95),
+            (5, 5),
```
ssanderson: One more good case here would be `0.6, 0.5` for the case where the min and max are both valid, but min is greater than max.

PR #1702, opened 2017-03-07
Set data portal last trading session/minute if passed

dmichalowicz:
Only infer the last trading session if the optional argument is not passed.

@yankees714 It looks like you added these arguments, can you take a look? I assume we want to infer the last available minute as well if that is not explicitly passed. cc @ehebert

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -166,15 +166,35 @@ class DataPortal(object):
         self._augmented_sources_map = {}
         self._extra_source_df = None
 
-        self._first_trading_session = first_trading_day
+        self._first_available_session = first_trading_day
 
-        _last_sessions = [r.last_available_dt
-                          for r in [equity_daily_reader, future_daily_reader]
-                          if r is not None]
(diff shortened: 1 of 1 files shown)


zipline/data/data_portal.py
```diff
@@ -168,13 +168,33 @@ def __init__(self,
 
         self._first_trading_session = first_trading_day
 
-        _last_sessions = [r.last_available_dt
-                          for r in [equity_daily_reader, future_daily_reader]
-                          if r is not None]
-        if _last_sessions:
-            self._last_trading_session = min(_last_sessions)
+        if last_available_session:
+            self._last_trading_session = last_available_session
```
yankees714: What would you think of aligning the names here (i.e. `self._last_trading_session` -> `self._last_available_session`)? Seems like that maybe have been a source of confusion here

dmichalowicz: I think that's a good idea, will change.

dmichalowicz:
@yankees714 Updated based on your comment

PR #1706, opened 2017-03-08
Add calendar argument to data portal in test fixtures

dmichalowicz:
@ehebert @yankees714 Look reasonable to you?


ehebert:
Is this intended for creating multiple `DataPortal`s in the same suite?
If not, setting `TRADING_CALENDAR_PRIMARY_CAL = 'us_futures'` on the test suite class would set `self.trading_calendar` appropriately.

PR #1710, opened 2017-03-14
Sort pipeline data on asofdate

mtydykov:
Addresses an issue where values for 2 subsequent dates, d1 and d2, came in for the first time on d3 and had the same timestamp; because d1 happened to be in a later row than d2, `last_in_date_group` selected d1 to be the "latest" value for the next date in the index, which was d5. The fix for this issue is to sort the data on asof_date before we drop the asof_date column.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -776,8 +776,9 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
             check_dtype=False,
         )
 
-    def _test_id_macro(self, df, dshape, expected, finder, add):
-        dates = self.dates
+    def _test_id_macro(self, df, dshape, expected, finder, add, dates=None):
+        if dates is None:
+            dates = self.dates
         expr = bz.data(df, name='expr', dshape=dshape)
         loader = BlazeLoader()
         ds = from_blaze(
@@ -1874,6 +1875,56 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
         })
 
         self._test_checkpoints(checkpoints)
+
+    def test_id_take_last_in_group_sorted(self):
+        """
+        input
+        asof_date     timestamp     other  value
+        2014-01-03    2014-01-04 00     3      3
+        2014-01-02    2014-01-04 00     2      2
+
+        output (expected):
+
+                    other  value
+        2014-01-02    NaN    NaN
+        2014-01-03    NaN    NaN
+        2014-01-06      3      3
+        """
+
+        dates = pd.DatetimeIndex([
+            pd.Timestamp('2014-01-02'),
+            pd.Timestamp('2014-01-03'),
+            pd.Timestamp('2014-01-06'),
+        ])
+
+        T = pd.Timestamp
+        df = pd.DataFrame(
+            columns=['asof_date', 'timestamp', 'other', 'value'],
+            data=[
+                # asof-dates are flipped in terms of order so that if we
+                # don't sort on asof-date before getting the last in group,
+                # we will get the wrong result.
+                [T('2014-01-03'), T('2014-01-04 00'), 3, 3],
+                [T('2014-01-02'), T('2014-01-04 00'), 2, 2],
+            ],
+        )
+        fields = OrderedDict(self.macro_dshape.measure.fields)
+        fields['other'] = fields['value']
+        expected = pd.DataFrame(
+            data=[[np.nan, np.nan],   # 2014-01-02
+                  [np.nan, np.nan],   # 2014-01-03
+                  [3,      3]],       # 2014-01-06
+            columns=['other', 'value'],
+            index=dates,
+        )
+        self._test_id_macro(
+            df,
+            var * Record(fields),
+            expected,
+            self.asset_finder,
+            ('other', 'value'),
+            dates=dates,
+        )
 
 
 class MiscTestCase(ZiplineTestCase):
--- a/zipline/pipeline/loaders/blaze/core.py
+++ b/zipline/pipeline/loaders/blaze/core.py
@@ -1104,6 +1104,11 @@ class BlazeLoader(dict):
             materialized_deltas,
             dates,
         )
+        # If we ever have cases where we find out about multiple asof_dates'
+        # data on the same TS, we want to make sure that last_in_date_group
+        # selects the correct last asof_date's value.
+        sparse_output.sort_values(AD_FIELD_NAME, inplace=True)
+        non_novel_deltas.sort_values(AD_FIELD_NAME, inplace=True)
         if AD_FIELD_NAME not in requested_columns:
             sparse_output.drop(AD_FIELD_NAME, axis=1, inplace=True)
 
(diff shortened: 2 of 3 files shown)


zipline/pipeline/loaders/blaze/core.py
```diff
@@ -1104,6 +1104,11 @@ def collect_expr(e, lower):
             materialized_deltas,
             dates,
         )
+        # If we ever have cases where we find out about multiple asof_dates'
+        # data on the same TS, we want to make sure that last_in_date_group
+        # selects the correct last asof_date's value.
+        sparse_output.sort_values(AD_FIELD_NAME, inplace=True)
```
llllllllll: should `last_in_date_group` just sort the input if it is not already sorted?

tests/pipeline/test_blaze.py
```diff
@@ -1876,6 +1876,54 @@ def test_checkpoints_out_of_bounds(self):
...
+                # we will get the wrong result.
+                [T('2014-01-03'), T('2014-01-04 00'), 3, 3],
+                [T('2014-01-02'), T('2014-01-04 00'), 2, 2],
+            ],
+        )
+        fields = OrderedDict(self.macro_dshape.measure.fields)
+        fields['other'] = fields['value']
+        expected = pd.DataFrame(
+            data=[[np.nan, np.nan],   # 2014-01-02
+                  [np.nan,      np.nan],   # 2014-01-03
```
llllllllll: the indentation here looks like it got messed up

tests/pipeline/test_blaze.py
```diff
@@ -1876,6 +1876,54 @@ def test_checkpoints_out_of_bounds(self):
...
+        output (expected):
+
+                    other  value
+        2014-01-02    NaN    NaN
+        2014-01-03    NaN    NaN
+        2014-01-06      3      3
+         """
+        T = pd.Timestamp
+        df = pd.DataFrame(
+            columns=['asof_date',        'timestamp', 'other', 'value'],
```
llllllllll: what's with the whitespace here?

tests/pipeline/test_blaze.py
```diff
@@ -1876,6 +1876,54 @@ def test_checkpoints_out_of_bounds(self):
...
+        super(LastInGroupSortingTestCase, cls).init_class_fixtures()
+        cls.dates = pd.DatetimeIndex([
+            pd.Timestamp('2014-01-02'),
+            pd.Timestamp('2014-01-03'),
+            pd.Timestamp('2014-01-06'),
+        ])
+
+    def test_id_take_last_in_group_sorted(self):
+        """
+        output (expected):
```
llllllllll: can you also show the input?

tests/pipeline/test_blaze.py
```diff
@@ -1876,6 +1876,54 @@ def test_checkpoints_out_of_bounds(self):
         self._test_checkpoints(checkpoints)
 
 
+class LastInGroupSortingTestCase(BlazeToPipelineTestCase):
```
llllllllll: why is this in its own test suite?

llllllllll:
looks like the blaze pipeline tests are now failing

mtydykov:
Yeah, looks like they only fail with the addition of the new test - the fix itself doesn't seem to make them fail.

tests/pipeline/test_blaze.py
mtydykov: Because I wanted a different set of dates to work with

zipline/pipeline/loaders/blaze/core.py
mtydykov: We could add some conditional logic to sort the input on some specified fields. 

My thought here was that `last_in_date_group` should just expect data that's already sorted so that all it needs to do is to group and select that last item in each group, as the name of the method suggests.

tests/pipeline/test_blaze.py
llllllllll: then you can pass your own dates around in the function instead of using self.dates.

zipline/pipeline/loaders/blaze/core.py
llllllllll: Idk, it seems like we could have this bug at all of the other call sites for this function which makes me want to have the function check itself.

mtydykov:
Other call sites may need to use different fields to sort the data, and so they would have to tell the function how to check itself anyway.

tests/pipeline/test_blaze.py
mtydykov: Yeah, but then other helper functions like`_test_id_macro` use `self.dates`. I was trying not to change the testing structure around too much.

mtydykov: Though I forgot that subclassing won't work here because it's trying to run all the other tests under this class.

zipline/pipeline/loaders/blaze/core.py
mtydykov: Other call sites may need to use different fields to sort the data, and so they would have to tell the function how to check itself anyway.

mtydykov: Then again, maybe having that extra, optional argument on that method would make people think about the sorting aspect when calling it. Either that, or we just add clarify in the docstring that the input df must be sorted so that the correct 'last' items are chosen. I think at this point I'm 50-50 on either option.

tests/pipeline/test_blaze.py
```diff
@@ -1875,6 +1876,56 @@ def test_checkpoints_out_of_bounds(self):
 
         self._test_checkpoints(checkpoints)
 
+    def test_id_take_last_in_group_sorted(self):
+        """
+        input
+        asof_date     timestamp     other  value
+        2014-01-02    2014-01-04 00     3      3
```
llllllllll: this isn't the input, the point is that these are flipped.

zipline/pipeline/loaders/utils.py
```diff
@@ -316,6 +318,8 @@ def last_in_date_group(df,
     )]]
     if have_sids:
         idx += [SID_FIELD_NAME]
+    if not extra_groupers:
```
llllllllll: we probably want `is None` to be more clear and allow for empty sequences

zipline/pipeline/loaders/utils.py
```diff
@@ -281,15 +281,17 @@ def last_in_date_group(df,
...
     Determine the last piece of information known on each date in the date
-    index for each group.
+    index for each group. Input df MUST be sorted such that the correct last
+    item is chosen from each group.
 
     Parameters
     ----------
     df : pd.DataFrame
-        The DataFrame containing the data to be grouped.
+        The DataFrame containing the data to be grouped. Must be sorted so that
```
llllllllll: +1

zipline/pipeline/loaders/utils.py
```diff
@@ -281,15 +281,17 @@ def last_in_date_group(df,
                        assets,
                        reindex=True,
                        have_sids=True,
-                       extra_groupers=[]):
+                       extra_groupers=None):
```
llllllllll: alternativly, we could make this a tuple

tests/pipeline/test_blaze.py
mtydykov: Yeah, good catch, flipped the order by accident.

llllllllll:
lgtm, can you just squash this into 1 commit?

mtydykov:
👍

PR #1715, opened 2017-03-16
Better error message for non-existent root symbol

dmichalowicz:


zipline/assets/assets.py
```diff
@@ -996,8 +996,13 @@ def _get_root_symbol_exchange(self, root_symbol):
 
         fields = (fc_cols.exchange,)
 
-        return sa.select(fields).where(
-            fc_cols.root_symbol == root_symbol).execute().fetchone()[0]
+        exchange_tuple = sa.select(fields).where(
+            fc_cols.root_symbol == root_symbol).execute().fetchone()
```
richafrank: Would there ever be more than one match? Should we use `scalar()` instead of `fetchone()` to ensure the cursor is closed? Then we also don't need to index the first column ourselves.

dmichalowicz: Good point, afaik there would never be more than one match, so `scalar()` is better here.

dmichalowicz:
@richafrank Updated and added a small test.

richafrank:
:+1: from me!

PR #1718, opened 2017-03-22
Add ContinuousFuture to lookup_generic

dmichalowicz:


zipline/assets/assets.py
```diff
@@ -1161,6 +1161,10 @@ def lookup_generic(self,
         matches = []
         missing = []
 
+        # If the input is a ContinuousFuture just return it as-is.
+        if isinstance(asset_convertible_or_iterable, ContinuousFuture):
+            return asset_convertible_or_iterable
+
```
ehebert: This could also be part of the check that follows with an elif.
(Being an `AssetConvertible` is probably the most common case, so would want to check for that first.)

dmichalowicz:
@ehebert Updated with a test.

PR #1719, opened 2017-03-22
Allow DataPortal.get_spot_value to accept multiple assets

dmichalowicz:


zipline/data/data_portal.py
```diff
@@ -463,41 +463,76 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
-            elif field == "contract":
-                return self._get_current_contract(asset, dt)
+        for asset in assets:
+            if self._is_extra_source(
+                    asset, field, self._augmented_sources_map):
+                spot_values.append(self._get_fetcher_value(asset, field, dt))
+                continue
+
+            if field not in BASE_FIELDS:
+                raise KeyError("Invalid column: " + str(field))
```
dmichalowicz: I'll pull this check out of the `for` loop.

zipline/data/data_portal.py
```diff
@@ -463,41 +463,76 @@ def get_spot_value(self, asset, field, dt, data_frequency):
             ``field`` is 'volume' the value will be a int. If the ``field`` is
             'last_traded' the value will be a Timestamp.
         """
-        if self._is_extra_source(asset, field, self._augmented_sources_map):
-            return self._get_fetcher_value(asset, field, dt)
+        assets_is_scalar = False
+        if isinstance(assets, (Asset, ContinuousFuture, str)):
```
ehebert: Could we use `AssetConvertible` or `PricingDataAssociable` here instead, which each have the Python string types registered (and handles Python 2/3).
Checking for just `str` is not sufficient.

zipline/data/data_portal.py
ehebert: :+1:

zipline/data/data_portal.py
```diff
@@ -463,41 +463,76 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
+                        )
+                    )
+                elif field == "contract":
+                    spot_values.append(self._get_current_contract(asset, dt))
+                else:
+                    spot_values.append(
+                        self._get_minute_spot_value(asset, field, dt)
+                    )
+
+        return spot_values[0] if assets_is_scalar else spot_values
```
ehebert: I have not profiled it, but always creating the out array may be a place to watch for a performance regression if/when `get_spot_value` is called in a tight loop.

A good way to check is to profile a 'buy and hold' type algorithm over a 500+ sized universe to make sure that that the the position tracker checks for the end of day values does not cause too much of a regression.
(Also the position tracker could be changed to pass a vector of assets, to take advantage of this change.)

dmichalowicz: Ok, I'm going to make it so that we only create the `out` list if `assets_is_scalar` is False.

zipline/data/data_portal.py
```diff
@@ -463,41 +469,65 @@ def get_spot_value(self, asset, field, dt, data_frequency):
...
-                return 0
-            elif field != "last_traded":
-                return np.NaN
-            elif field == "contract":
-                return None
-
-        if data_frequency == "daily":
-            if field == "contract":
-                return self._get_current_contract(asset, session_label)
+        def get_single_asset_value(asset):
```
dmichalowicz: I made this helper function so that we can apply it to a list of assets only if we have to.

dmichalowicz:
@ehebert Updated based on your comments.

PR #1721, opened 2017-03-24
Negative indexing in cython without wraparound

dmichalowicz:
When calling `history` for `price` or `close`, it was possible that we would index into an array with a `-1`, and with `@wraparound(False)` this caused undefined behavior.

### DIFF
--- a/tests/data/test_resample.py
+++ b/tests/data/test_resample.py
@@ -143,6 +143,8 @@ _FUTURE_CASES = (
             ('none_missing', 'day_0_back'))),
     (1003, (('missing_last', 'day_0_back'),
             ('missing_first', 'day_1_front'))),
+    (1004, (('all_missing', 'day_0_back'),
+            ('none_missing', 'day_1_front'))),
 )
 
 FUTURE_CASES = OrderedDict()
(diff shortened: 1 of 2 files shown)


dmichalowicz:
@ehebert I'd be interested to hear your thoughts on this one; I'm wondering if this is the correct fix or if we actually want the `-1` index, meaning we should change `@wraparound` to `True`.

ehebert:
@dmichalowicz and I talked about this in person.

This version of `_minute_to_session_close` handles the boundary condition by decrementing after the access into the data array, instead of before.

```
@boundscheck(False)
@wraparound(False)
cpdef void _minute_to_session_close(intp_t[:] close_locs,
                                    float64_t[:] data,
                                    float64_t[:] out):
    cdef intp_t out_i, data_i = 0
    cdef float64_t val
    num_out = len(out)
    for out_i in range(num_out - 1, -1, -1):
        data_i = close_locs[out_i]
        if out_i > 0:
            prev_close_i = close_locs[out_i - 1]
        else:
            prev_close_i = -1
        val = nan
        while isnan(val) and data_i > prev_close_i:
            val = data[data_i]
            data_i -= 1
        out[out_i] = val
```
I'll push that version to this branch. (And may do a separate cleanup PR as well.)

dmichalowicz:
The added test case is hit in `TestResampleSessionBars.test_resample` and `TestResampleSessionBars.test_get_value`.

PR #1722, opened 2017-03-24
ENH: Use regular rounding to calculate order amounts.

jbredeche:
We previously tried to prevent accidental over-ordering by truncating
orders down unless they were within 1e-4 of the next higher integer.
Unfortunately, this makes it easy for a sell order to be one share short
of the desired position.

Using regular rounding treats both buys and sells in the same way.


zipline/algorithm.py
```diff
@@ -1441,6 +1438,13 @@ def _calculate_order(self, asset, amount,
                                                         style)
         return amount, style
 
+    @staticmethod
+    def _round_order(amount):
```
ssanderson: Since the intent of this change is to make this overridable by subclasses, can we make this a non-underscore method and put a docstring on it?

ssanderson:
LGTM (round two)

PR #1723, opened 2017-03-24
ENH: Remove path.py as its not a dependency

### LINKED ISSUE #1237: Docs build script requires `path.py` without specifying it as a dependency.
We should probably just remove it.

FreddieV4:
Should fix #1237.

I traced the call to `path()` down to this section of [path.py](https://github.com/jaraco/path.py/blob/master/path.py)

```
def __enter__(self):
    self._old_dir = self.getcwd()
    os.chdir(self)
    return self
```

And changed our `deploy.py` to just do the same thing

### DIFF
--- a/docs/deploy.py
+++ b/docs/deploy.py
@@ -2,7 +2,6 @@
 from __future__ import print_function
 from contextlib import contextmanager
 from glob import glob
-from path import path
 import os
 from os.path import abspath, basename, dirname, exists, isfile
 from shutil import move, rmtree
@@ -32,8 +31,11 @@ def ensure_not_exists(path):
 
 
 def main():
+    old_dir = os.getcwd()
     print("Moving to %s." % HERE)
-    with path(HERE):
+    os.chdir(HERE)
+
+    try:
         print("Building docs with 'make html'")
         check_call(['make', 'html'])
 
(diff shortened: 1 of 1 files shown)


docs/deploy.py
```diff
@@ -32,38 +31,42 @@ def ensure_not_exists(path):
...
+
+        print("Copying built files:")
+        for file_ in glob(TEMP_LOCATION_GLOB):
+            base = basename(file_)
+
+            print("%s -> %s" % (file_, base))
+            ensure_not_exists(base)
+            move(file_, '.')
+
+    os.chdir(old_dir)
```
richafrank: Should this be in a `finally` block, like the [`with`](https://docs.python.org/2/reference/compound_stmts.html#with)'s `__exit__` is?

FreddieV4: Hmm I'm not sure what you mean by that. Could you elaborate?

I know that using `with` is equivalent to

```
VAR = EXPR
VAR.__enter__()
try:
    BLOCK
finally:
    VAR.__exit__()
```

but in this context, I'm not sure what you mean about using a `finally` block

richafrank: Yup, so if we're inlining the `Path` context manager here, then what was a call to its `__exit__` method should end up in a `finally` block as you pasted, e.g.

```python
old_dir = self.getcwd()
os.chdir(HERE)  # was __enter__()
try:
    print("Building docs with 'make html'")
    check_call(['make', 'html'])
    ...
finally:
    os.chdir(old_dir)  # was __exit__()
```

As it is in the diff currently, an exception raised in `BLOCK`, e.g. in one of the `check_call`s, will not change the directory back to the original, whereas `with Path(HERE):` does.

FreddieV4:
Made the suggested changes and squashed commits

richafrank:
LGTM!

PR #1698, opened 2017-03-05
ENH: Adding CFE Adhoc Holidays

trical09:
@FreddieV4 Fixed, redone #1682 

CBOE and CFE observed the USNationalDaysofMourning and HurricaneSandyClosings:

USNationalDaysofMourning
http://www.mondovisione.com/news/cboe-and-cfe-will-be-closed-on-tuesday-january-2-to-honor-former-president-geral/
https://www.cboe.com/aboutcboe/legal/bulletins/archives/documents/eb062504.pdf

HurricaneSandyClosings
http://ir.cboe.com/press-releases/2012/29-oct-2012.aspx
http://www.cboe.com/publish/RegCir/RG12-146.pdf

exchange_calendar_cfe.CFEExchangeCalendar

should have the following:

```
@property
 def adhoc_holidays(self):
     return list(chain(
         HurricaneSandyClosings,
         USNationalDaysofMourning,
     ))
```

PR #1725, opened 2017-03-27
TST: Adhoc holidays in CFEExchangeCalendar

FreddieV4:
Thought of this after pulling latest master & looking at https://github.com/quantopian/zipline/pull/1698


tests/calendars/test_cfe_calendar.py
```diff
@@ -40,3 +40,15 @@ def test_2016_early_closes(self):
...
         self.assertEqual(15, market_close.minute)
+
+    def test_adhoc_holidays(self):
+        # hurricane sandy: oct 29 2012, oct 30 2012
+        # national days of mourning:
+        # - apr 27 1994
+        # - june 11 2004
+        # - jan 2 2007
+        for day in ["1994-04-27", "2004-06-11", "2007-01-02",
+                    "2012-10-29", "2012-10-30"]:
```
richafrank: Do you think it's worth testing these separately, or should we fold these dates into `test_2016_holidays` above?

FreddieV4: We could fold them into `test_2016_holidays`, and then I'd just need to refactor the test name (because they're not all in 2016)

richafrank: Ah, indeed! Then I'd probably just leave them as is.

FreddieV4:
Talked off of GitHub. Gonna merge this 😃

PR #1726, opened 2017-03-27
Make continuous future adjustment style an argument

dmichalowicz:
Allow specifying an adjustment style as a parameter, e.g. `continuous_future('CL', offset=0, roll='volume', adjustment='add')` is equivalent to `continuous_future('CL', offset=0, roll='volume').adj('add')`

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -161,7 +161,10 @@ def build_lookup_generic_cases(asset_finder_type):
         ]
         fof14 = finder.retrieve_asset(fof14_sid)
         cf = finder.create_continuous_future(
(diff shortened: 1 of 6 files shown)


ehebert:
To confirm, the intent is not just to make an equivalent API to `.adj`, but to replace it in favor of passing the adjustment arg.

If so, I agree with the API change and the implementation.
(The setting up of `adjustment_children` was awkward.)

abhijeetkalyan:
Sorry about the review request, misclick

dmichalowicz:
> To confirm, the intent is not just to make an equivalent API to .adj, but to replace it in favor of passing the adjustment arg.

Yes, that's correct. Thanks for the review Eddie.

PR #1729, opened 2017-03-29
Use 'us_futures' calendar in test fixtures

dmichalowicz:
Changing certain fixtures to use the `us_futures` calendar was causing the `benchmark_returns` file written to `/.zipline/data/^GSPC_benchmark.csv` to initially get written with dates from the futures calendar (2000-present), causing tests that involved equities which relied on data in the 1990s to retrieve the cached file and fail. To fix that, have the `TradingEnvironment`, which consumes the benchmark returns file, use a static csv in source which has all the dates necessary to cover both calendars.


zipline/testing/fixtures.py
```diff
@@ -481,10 +482,26 @@ class WithTradingEnvironment(WithAssetFinder,
...
     def make_load_function(cls):
-        return None
+        """
+        Load benchmark returns and treasury curves data from a static csv file.
+
+        This ensures that we have data over a consistent date range for the
+        entirety of the zipline tests, and guarantees that the data covers the
+        dates required for tests on both the equity and futures calendars.
+        """
+        benchmark_csv_path = join(cls.RESOURCE_DIR, '^GSPC_benchmark.csv')
```
richafrank: What do you think of moving this behavior to `ensure_benchmark_data`, so it would first check for the repo resource, then the current `~/.zipline` data, then download, at each step checking for the date range?

Then this isn't specific to tests, but covers the tests, and means that in many cases, one can run zipline without needing to download benchmarks.

richafrank: If you dislike that idea, then I'd at least like this to be robust to needing a wider range of data in the future. Right now tests would just fail again, like you were seeing.

dmichalowicz:
@richafrank Thanks for the suggestion, I went with your idea of moving the code. I ended up doing a slight refactoring to move the shared code between `ensure_benchmark_data` and `ensure_treasury_data` into a helper function called `_load_cached_data`. I'll comment on the key addition.

zipline/data/loader.py
```diff
@@ -273,26 +247,68 @@ def ensure_treasury_data(bm_symbol, first_date, last_date, now):
...
+            return data
+    except (OSError, IOError, ValueError) as e:
+        # These can all be raised by various versions of pandas on various
+        # classes of malformed input.
+        logger.info(
+            'Loading data from source path {path!r} failed with error '
+            '[{error}].',
+            path=source_path,
+            error=e,
+        )
```
dmichalowicz: This logic above is the main addition that checks for a csv in source (and checks its dates).

zipline/data/loader.py
```diff
@@ -43,6 +43,8 @@
     (treasuries, 'treasury_curves.csv', 'www.federalreserve.gov'),
 }
 
+MARKET_DATA_DIR = 'zipline/resources/market_data'
```
richafrank: What do you think about making this relative to the zipline module, as we did in in [`security_list`](https://github.com/quantopian/zipline/blob/3b5031a82959371b6c303f2381a7c15acae34e5b/zipline/utils/security_list.py#L17-L17). That method also ensures that we use the correct path separator between subdirectories - not sure if that's an issue or not, given that the tests passed on appveyor.

richafrank: If you make these new directories packages, then `zipline.resources.market_data.__file__` would even work. Not sure if there were reasons we didn't do that already...

zipline/data/loader.py
```diff
@@ -273,26 +247,68 @@ def ensure_treasury_data(bm_symbol, first_date, last_date, now):
...
+    # necessary data.
+    logger.info('Downloading treasury data for {symbol!r}.', symbol=symbol)
+
+    try:
+        data = loader_module.get_treasury_data(first_date, last_date)
+        data.to_csv(get_data_filepath(filename))
+    except (OSError, IOError, HTTPError):
+        logger.exception('failed to cache treasury data')
+    if not has_data_for_dates(data, first_date, last_date):
+        logger.warn("Still don't have expected data after redownload!")
```
richafrank: If we're consolidating these helper methods, I wonder if we could remove even this remaining duplication. Not a blocker.

dmichalowicz: I considered consolidating these as well, but figured that there are enough minute differences in method calls and logging language that sharing the code wasn't worth it for only ~8 lines of code.

richafrank: Ok, sounds fine.

zipline/data/loader.py
dmichalowicz: updated

dmichalowicz:
@richafrank I think this is finally all set to merge, unless you had any final comments.

zipline/data/loader.py
```diff
@@ -273,26 +249,68 @@ def ensure_treasury_data(bm_symbol, first_date, last_date, now):
...
+
+
+def _load_cached_data(filename, first_date, last_date, now, resource_name):
+    if resource_name == 'benchmark':
+        from_csv = pd.Series.from_csv
+    else:
+        from_csv = pd.DataFrame.from_csv
+
+    # First try to retrieve the data from a static csv file in source.
+    source_path = os.path.join(MARKET_DATA_DIR, filename)
```
richafrank: Do we still want this new behavior? Seems like this file will not be useful outside of tests.

dmichalowicz: No you're right, this file will no longer be used as soon as `pd.Timestamp.now` is greater than 2 days after the static end date of the file. I'll remove.

dmichalowicz:
@richafrank Good now I think? 😄

richafrank:
:+1:

PR #1730, opened 2017-03-29
Add safeguard if current contract of continuous future is None

dmichalowicz:
Make sure the current contract of a continuous future is not `None` before calling `retrieve_asset` on it.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -500,10 +500,10 @@ class DataPortal(object):
                      session_label > asset.end_date):
                 if field == "volume":
                     return 0
-                elif field != "last_traded":
-                    return np.NaN
                 elif field == "contract":
                     return None
+                elif field != "last_traded":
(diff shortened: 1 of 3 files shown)


zipline/data/data_portal.py
```diff
@@ -500,10 +500,10 @@ def get_single_asset_value(asset):
                      session_label > asset.end_date):
                 if field == "volume":
                     return 0
-                elif field != "last_traded":
-                    return np.NaN
                 elif field == "contract":
                     return None
+                elif field != "last_traded":
```
ehebert: Any reason for moving this?

dmichalowicz: Th `elif field == "contract"` wasn't getting hit because `elif field != "last_traded"` was `True`, so if `field` was "contract" this would return NaN instead of None.

dmichalowicz:
Will squash before merging, thanks @ehebert

PR #1735, opened 2017-04-03
PERF: Avoid repeated recursive calls when getting forward-filled close

yankees714:
Instead of recursively calling `DailyHistoryAggregator.closes` until we find a non-nan close, we can instead call `load_raw_arrays` once, and find the value from the returned array.

### DIFF
--- a/tests/data/test_resample.py
+++ b/tests/data/test_resample.py
@@ -30,6 +30,7 @@ from zipline.data.resample import (
     ReindexSessionBarReader,
 )
 
+from zipline.testing import parameter_space
 from zipline.testing.fixtures import (
     WithEquityMinuteBarData,
     WithBcolzEquityMinuteBarReader,
@@ -253,6 +254,7 @@ EXPECTED_SESSIONS = {
 
 
 class MinuteToDailyAggregationTestCase(WithBcolzEquityMinuteBarReader,
+                                       WithBcolzFutureMinuteBarReader,
                                        ZiplineTestCase):
 
     #    March 2016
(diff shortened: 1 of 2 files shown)


ehebert:
@yankees714 do you have profiles/timings of this change using a contract which fits in the happy case (i.e. the last minute of the day is not `nan`.) as well as the worst case that this is fixing?
(I am slightly worried about incurring cost of creating arrays in cases that they are not needed.)

Depending on those results, we may want to have the logic assume best case and get the last value, and then assume the worst if it is nan.

yankees714:
Think I may have mis-interpreted @ehebert's question here at first - the case where the minute passed to `closes()` has a non-nan close should not be effected. We do see a bit of a slowdown where the value is nan and the value for the previous minute is non-nan (`~180 µs` -> `~220 µs`), but we don't think this is significant enough to add extra logic.

tests/data/test_resample.py
```diff
@@ -479,6 +480,101 @@ def test_skip_minutes_multiple(self, field):
...
+    # 20 21 22 23 24 25 26
+    # 27 28 29 30 31
+
+    TRADING_CALENDAR_STRS = ('us_futures',)
+    TRADING_CALENDAR_PRIMARY_CAL = 'us_futures'
+
+    TRADING_ENV_MIN_DATE = START_DATE = pd.Timestamp(
+        '2016-03-01', tz='UTC',
+    )
+    TRADING_ENV_MAX_DATE = END_DATE = pd.Timestamp(
```
dmichalowicz: I don't think `TRADING_ENV_MIN_DATE` and `TRADING_ENV_MAX_DATE` are necessary here (the test works without them)

dmichalowicz:
@yankees714 The new test looks good to me, though it would be nice to be able to share a lot of that code. What do you think of moving the test into the existing `MinuteToDailyAggregationTestCase` suite and creating a shared helper function? Also, is there any way to use this test to compare the old behavior to the new? (i.e. does it fail or is it just slower, in which case I'm not immediately sure how to compare)

yankees714:
@dmichalowicz Good call on sharing a helper function - I was initially tentative to include both the futures and equities setup in the same test case, but I like the way it came out.

For the behavior comparison, the new test fails on master (with the max recursion depth exceeded issue).

dmichalowicz:
The shared test looks great, nice work. This PR looks good by me, unless @ehebert had any other comments

PR #1738, opened 2017-04-04
Add preliminary support for Futures slippage models

dmichalowicz:
Add futures slippage and commission models to the Blotter, which dispatches to the appropriate models according to asset type.


zipline/algorithm.py
```diff
@@ -324,10 +321,8 @@ def __init__(self, *args, **kwargs):
             self.blotter = Blotter(
                 data_frequency=self.data_frequency,
                 asset_finder=self.asset_finder,
-                slippage_func=VolumeShareSlippage(),
-                commission=PerShare(),
```
dmichalowicz: Instead of adding in the two new future parameters here I just removed these two arguments because they were passing the same values as the defaults anyway.

dmichalowicz:
cc @ssanderson in case you were interested.

jbredeche:
nice @dmichalowicz, this looks good to me.

dmichalowicz:
Thanks for the review @jbredeche, I just squashed and rebased, then will merge.

PR #1739, opened 2017-04-04
Fix zipline and pandas bug

mtydykov:
Fixes:
1) A pandas bug whereby if `tz_localize(None)` is called on a non-sorted DatetimeIndex, some dates are pushed back an hour. The implications of this are that all loaders that use `normalize_timestamp_to_query_time` (e.g. the blaze core loader, the events loader, etc) could result in incorrect timestamps and thus data being surfaced early.

```
In [10]: d = pd.DatetimeIndex(['2013-12-30', '2013-06-01'])

In [11]: d.tz_localize('US/Eastern').tz_localize(None)
Out[11]: DatetimeIndex(['2013-12-30 00:00:00', '2013-05-31 23:00:00'], dtype='datetime64[ns]', freq=None)
```

2) A zipline bug where datetime arrays were not being properly compared in a testing utility.

### DIFF
--- a/zipline/testing/core.py
+++ b/zipline/testing/core.py
@@ -406,7 +406,7 @@ def check_arrays(x, y, err_msg='', verbose=True, check_dtypes=True):
         )
         # Fill NaTs with zero for comparison.
         x = np.where(x_isnat, np.zeros_like(x), x)
-        y = np.where(x_isnat, np.zeros_like(x), x)
+        y = np.where(y_isnat, np.zeros_like(y), y)
 
     return assert_array_equal(x, y, err_msg=err_msg, verbose=verbose)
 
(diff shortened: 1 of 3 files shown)


zipline/testing/core.py
```diff
@@ -406,7 +406,7 @@ def check_arrays(x, y, err_msg='', verbose=True, check_dtypes=True):
         )
         # Fill NaTs with zero for comparison.
         x = np.where(x_isnat, np.zeros_like(x), x)
-        y = np.where(x_isnat, np.zeros_like(x), x)
+        y = np.where(y_isnat, np.zeros_like(y), y)
```
llllllllll: oh wow, good catch. Do all of the tests still pass?

mtydykov: Yup, looks like the travis build passed.

PR #1740, opened 2017-04-04
BUG: Return from can_trade in same order as input.

ssanderson:
This matches the behavior of history and data.current.


ssanderson:
Spoke about this with @richafrank, who okay'ed it.

PR #1741, opened 2017-04-05
Remove ContinuousFuture adjustment method

dmichalowicz:
This method is no longer needed as the `adjustment` field is now [passed as a parameter](https://github.com/quantopian/zipline/pull/1726).


dmichalowicz:
@yankees714 Can you take a glance to make sure I didn't miss anything?

PR #1742, opened 2017-04-06
MAINT: Refactor DataPortal._get_minute_spot_value to avoid two lookups

yankees714:


zipline/data/data_portal.py
```diff
@@ -650,47 +650,37 @@ def get_adjusted_value(self, asset, field, dt,
...
-            asset.sid,
-            last_traded_dt,
-            column
-        )
-
-        if np.isnan(result):
-            return np.nan
-
-        if dt == last_traded_dt or dt.date() == last_traded_dt.date():
+        if not ffill or (dt == query_dt) or (dt.date() == query_dt.date()):
```
ehebert: Not part of this change. But I just noticed that this particular check on `date` does not play well with adjustments of Futures.

PR #1743, opened 2017-04-06
OrderedContracts chain could sometimes terminate on first contract

dmichalowicz:
It is possible for the first contract of a `ContinuousFuture` to have a start date which is after its auto close date. In that case the continuous future terminates immediately, giving it a premature end date. To handle this, throw out the first contract if it hits this case because it is not tradable anyway.


dmichalowicz:
@ehebert This fixes the issue we discussed with yesterday regarding broken continuous futures. Mind taking a quick look?

PR #1745, opened 2017-04-07
Make certain continuous future arguments optional

dmichalowicz:
Use a default offset of `0` and default roll style of `volume`.


dmichalowicz:
@ehebert Could you take a peek?

PR #1746, opened 2017-04-10
MAINT: Add better repr for transactions

FreddieV4:
The `repr` for a Transaction object now looks like:

`Transaction(order_id=20, sid=Asset(0), dt=2015-01-01 00:00:00, amount=10, price=50)`

### DIFF
--- a/tests/finance/test_transaction.py
+++ b/tests/finance/test_transaction.py
@@ -1,0 +1,35 @@
+#
+# Copyright 2017 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pandas as pd
+from unittest import TestCase
+
+from zipline.assets import Equity
+from zipline.finance.transaction import Transaction
+
+
+class TransactionTestCase(TestCase):
+
+    def test_transaction_repr(self):
+        dt = pd.Timestamp('2017-01-01')
+
+        asset = Equity(1, exchange='test')
+        txn = Transaction(asset, amount=100, dt=dt, price=10, order_id=0)
+
+        expected = (
+            "Transaction(asset=Equity(1), dt=2017-01-01 00:00:00,"
+            " amount=100, price=10)"
+        )
+
+        self.assertEqual(repr(txn), expected)
--- a/zipline/finance/transaction.py
+++ b/zipline/finance/transaction.py
@@ -35,6 +35,20 @@ class Transaction(object):
     def __getitem__(self, name):
         return self.__dict__[name]
 
+    def __repr__(self):
+        template = (
+            "{cls}(asset={asset}, dt={dt},"
+            " amount={amount}, price={price})"
+        )
+
+        return template.format(
+            cls=type(self).__name__,
+            asset=self.asset,
+            dt=self.dt,
+            amount=self.amount,
+            price=self.price
+        )
+
     def to_dict(self):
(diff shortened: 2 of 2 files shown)


jbredeche:
Nice!  Can you add `price` too?

zipline/finance/transaction.py
```diff
@@ -36,6 +36,17 @@ def __init__(self, sid, amount, dt, price, order_id, commission=None):
     def __getitem__(self, name):
         return self.__dict__[name]
 
+    def __repr__(self):
+        return """{typename}(order_id={order_id}, sid={sid}, \
+dt={dt}, amount={amount}, price={price})""".format(
+            typename=type(self).__name__,
+            order_id=self.order_id,
```
ssanderson: My only immediate reaction to this is that I'm not sure that it's useful to include `order_id` in the repr.  I can't think of a case where I've wanted to see the order id when looking at a transaction.

Also, we probably want a test for this.  People often forget to update a repr when removing or changing a field name, so test coverage is important for them.

FreddieV4: Where would be a good place to add test coverage on this?

ssanderson:
I'd probably just add a new file in `tests/finance` like
`tests/finance/test_transaction.py'.

FreddieV4:
Updated example `repr` without `order_id`:

`Transaction(sid=Asset(1), dt=2015-01-01 00:00:00, amount=10, price=50)`

tests/finance/test_transaction.py
```diff
@@ -0,0 +1,33 @@
...
+from zipline.assets import Asset
+from zipline.finance.transaction import Transaction
+
+
+class TransactionTestCase(TestCase):
+
+    def test_transaction_repr(self):
+        dt = pd.Timestamp('2017-01-01')
+
+        asset = Asset(1, exchange='test')
```
ssanderson: I would probably make this an `Equity` or a `Future`.  The base `Asset` class isn't really meant to be instantiated directly.

tests/finance/test_transaction.py
```diff
@@ -0,0 +1,33 @@
...
+        txn = Transaction(asset, amount=100, dt=dt, price=10, order_id=0)
+
+        expected = "Transaction(sid=Asset(1), dt=2017-01-01 00:00:00," + \
```
ssanderson: You can also write a long string literal like this by putting it in parentheses, which removes the need for concatenating and escaping:

```
expected = (
    "Transaction(sid=Asset(1), dt=2017-01-01 00:00:00,"
    "amount=100, price=10)"
)
```

zipline/finance/transaction.py
```diff
@@ -36,6 +36,16 @@ def __init__(self, sid, amount, dt, price, order_id, commission=None):
     def __getitem__(self, name):
         return self.__dict__[name]
 
+    def __repr__(self):
```
ssanderson: Very minor style point: this might be slightly nicer to read broken out over multiple lines:

```python
    def __repr__(self):
        template = "{cls}(sid={sid}, dt={dt}, amount={amount}, price={price})"
        return template.format(
            cls=type(self).__name__,
            sid=self.sid,
            dt=self.dt,
            amount=self.amount,
            price=self.price
        )
```

ssanderson: Also, as a heads-up, I think `sid` here is going to change to `asset` in @jbredeche's latest (`sid` originally stood for "security identifier", but in modern zipline it just refers to the unique integer value assigned to every asset; in either case, the use of `sid` here is actually misleading, since the expected value is an `Equity` or `Future`) in https://github.com/quantopian/zipline/pull/1757.  Might be easiest to wait for that to merge (which should be later today) and then update here.

ssanderson: xref: https://github.com/quantopian/zipline/pull/1757/files#diff-82f717facf7980359a913c2407bed230R24

tests/finance/test_transaction.py
FreddieV4: TIL. That's pretty cool 😄 I'll make that change

zipline/finance/transaction.py
FreddieV4: @ssanderson Updated the `repr` after rebasing against Jean's changes

ssanderson:
LGTM

PR #1747, opened 2017-04-11
Improve TradingCalendar perf with scalar-optimized accesses

yankees714:
When retrieving the open and close for a given session, we only care about the scalar values, so using `DataFrame.at` instead of `DataFrame.loc` is significantly faster.

Also improves `MinuteResampleSessionBarReader` by only retrieving session close when we need just the close, instead of the open and close.

### DIFF
--- a/zipline/data/resample.py
+++ b/zipline/data/resample.py
@@ -521,7 +521,7 @@ class MinuteResampleSessionBarReader(SessionBarReader):
         sessions = self._calendar.sessions_in_range(start_dt, end_dt)
         m_closes = np.zeros(len(sessions), dtype=np.dtype('datetime64[ns]'))
         for i, s in enumerate(sessions):
-            close = self._calendar.open_and_close_for_session(s)[1]
+            close = self._calendar.session_close(s)
             m_closes[i] = close.value
         m_locs = np.searchsorted(dts, m_closes)
         results = []
--- a/zipline/utils/calendars/trading_calendar.py
+++ b/zipline/utils/calendars/trading_calendar.py
@@ -460,7 +460,10 @@ class TradingCalendar(with_metaclass(ABCMeta)):
         pd.DateTimeIndex
             All the minutes for the given session.
         """
-        return self.minutes_in_range(*self.schedule.loc[session_label])
+        return self.minutes_in_range(
+            start_minute=self.schedule.at[session_label, 'market_open'],
+            end_minute=self.schedule.at[session_label, 'market_close'],
+        )
 
     def minutes_window(self, start_dt, count):
         start_dt_nanos = start_dt.value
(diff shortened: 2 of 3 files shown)


ssanderson:
@yankees714 do you have a profiling run that shows the difference here?

zipline/data/resample.py
```diff
@@ -521,7 +521,7 @@ def _get_resampled(self, columns, start_dt, end_dt, assets):
         sessions = self._calendar.sessions_in_range(start_dt, end_dt)
         m_closes = np.zeros(len(sessions), dtype=np.dtype('datetime64[ns]'))
         for i, s in enumerate(sessions):
-            close = self._calendar.open_and_close_for_session(s)[1]
+            close = self._calendar.session_close(s)
```
ssanderson: :+1:

zipline/utils/calendars/trading_calendar.py
```diff
@@ -460,7 +460,10 @@ def minutes_for_session(self, session_label):
         pd.DateTimeIndex
             All the minutes for the given session.
         """
-        return self.minutes_in_range(*self.schedule.loc[session_label])
+        return self.minutes_in_range(
```
ssanderson: If we're willing to assume that trading days are contiguous (i.e., that there are no holes in the calendar), then we can skip `minutes_in_range` entirely and just return `pd.date_range(start_minute, end_minute)` here.

zipline/utils/calendars/trading_calendar.py
```diff
@@ -635,22 +638,24 @@ def open_and_close_for_session(self, session_label):
...
-        o_and_c = self.schedule.loc[session_label]
+        sched = self.schedule
 
         # `market_open` and `market_close` should be timezone aware, but pandas
         # 0.16.1 does not appear to support this:
         # http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#datetime-with-tz  # noqa
-        return (o_and_c['market_open'].tz_localize('UTC'),
-                o_and_c['market_close'].tz_localize('UTC'))
+        return (
+            sched.at[session_label, 'market_open'].tz_localize('UTC'),
```
ssanderson: As long as we're micro-optimizing these functions, it might be worth looking at how this compares to something like:

```
row = self.schedule.index.get_loc(session_label)
return sched.iat[row, 0], sched.iat[row, 1]
```

yankees714:
@ssanderson With this change it looks like:
```
In [3]: %%timeit
   ...: nyse_cal.open_and_close_for_session(today)
   ...:
10000 loops, best of 3: 41.6 µs per loop
```

vs. master:
```
In [2]: %%timeit
   ...: nyse_cal.open_and_close_for_session(today)
   ...:
1000 loops, best of 3: 274 µs per loop
```

ssanderson:
@yankees714 changes look reasonable.  Had a few suggestions for possible further improvements.  I'm a little curious what prompted this change.  Is `open_and_close_for_session` is a significant bottleneck somewhere?  I've tended not to focus too much on scalar operations because my assumption has been that anyone calling a scalar function enough times to matter should probably be calling a vectorized version instead.

zipline/utils/calendars/trading_calendar.py
yankees714: Looks like a similar result:
```
In [3]: %%timeit
   ...: nyse_cal.open_and_close_for_session(today)
   ...:
10000 loops, best of 3: 48.9 µs per loop
```
with:
```
        row_i = sched.index.get_loc(session_label)
        return (
            sched.iat[row_i, 0].tz_localize('UTC'),
            sched.iat[row_i, 1].tz_localize('UTC'),
        )
```

yankees714:
The bottleneck this is aimed at is: https://github.com/quantopian/zipline/pull/1747/files#diff-64eeb1b15a7a2ce1ec06af6c9a2f4b47R524. Seems like we can probably unfold that into a vectorized lookup?

ssanderson:
> Seems like we can probably unfold that into a vectorized lookup?

Yeah, that function looks like it's doing a ton of manual looping over arrays.  If the goal is to improve perf in `_get_resampled`, the right fix is to vectorize the implementation, not optimize scalar operations.

jbredeche:
I'll still take a 6x improvement on the scalar operation, though.

ssanderson:
> I'll still take a 6x improvement on the scalar operation, though.

Yeah, no reason to drop these improvements, but I suspect we can get more like 100x on the function that's the actual bottleneck here.

PR #1748, opened 2017-04-11
Add slippage and commission models for futures

dmichalowicz:
This PR adds the actual slippage and commissions models for futures, taking advantage of the infrastructure built [here](https://github.com/quantopian/zipline/pull/1738).

In terms of the API, the `set_slippage` and `set_commission` functions now accept an `equities` argument and a `futures` argument for setting two different models according to the type of asset being traded. Furthermore, the `SlippageModel` and `CommissionModel` classes have been broken out into Equity and Future subclasses. This is to help enforce that the models passed to these new parameters are of the correct type. For example, the `equities` parameter for `set_slippage` must inherit from `EquitySlippageModel`, while the `futures` parameter must inherit from `FutureSlippageModel`.

For futures commissions, we use `PerContract` by default, which is essentially the equivalent of `PerShare` for the number of contracts traded. Unlike `PerShare` however, the commission for each different root symbol can be toggled by passing a dictionary mapping root symbol to its respective commission cost.

### DIFF
--- a/tests/finance/test_commissions.py
+++ b/tests/finance/test_commissions.py
@@ -1,8 +1,18 @@
 from datetime import timedelta
 from textwrap import dedent
 
+from nose_parameterized import parameterized
+from pandas import DataFrame
+
 from zipline import TradingAlgorithm
-from zipline.finance.commission import PerTrade, PerShare, PerDollar
+from zipline.errors import IncompatibleCommissionModel
+from zipline.finance.commission import (
+    PerContract,
+    PerDollar,
+    PerFutureTrade,
+    PerShare,
+    PerTrade,
+)
 from zipline.finance.order import Order
 from zipline.finance.transaction import Transaction
 from zipline.testing import ZiplineTestCase, trades_by_sid_to_dfs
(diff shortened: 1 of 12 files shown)


zipline/finance/commission.py
```diff
@@ -79,56 +204,114 @@ def __init__(self,
...
+        if isinstance(self.exchange_fee, float):
+            exchange_fee = self.exchange_fee
+        else:
+            # Exchange fee is a dictionary. If the user's dictionary does not
+            # provide an exchange fee for a certain contract, fall back on the
+            # default.
+            root_symbol = order.asset.root_symbol
+            exchange_fee = self.exchange_fee.get(
+                root_symbol, FUTURE_EXCHANGE_FEES_BY_SYMBOL[root_symbol],
+            )
```
dmichalowicz: The logic here that differs according to the type of `exchange_fee` and `cost_per_contract` seems well suited for something like multiple dispatch, but I'm not sure how that would work if these values aren't parameters to the method.

dmichalowicz:
@ssanderson This is ready for a first look. cc @jbredeche in case you were interested.

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+    annualized daily volatility.
+    """
+
+    def __init__(self):
+        super(WithWindowData, self).__init__()
+        self._window_data_cache = ExpiringCache()
+
+    def _get_window_data(self, data, asset, window_length):
+        """
+        Internal utility method to return the trailing 20-day mean volume and
```
prsutherland: We can probably remove the "20-day" in the doc now that it is parameterized and add it to the doc's parameter list.

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+        asset : The Asset whose data we are fetching.
+
+        Returns
+        -------
+        (mean volume, volatility)
+        """
+        try:
+            values = self._window_data_cache.get(asset, data.current_session)
+        except KeyError:
+            # Add a day because we wan 'window_length' complete days.
```
prsutherland: Small typo in "we want".

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+        if not volume:
+            return None, None
+
+        # Price to use is the average of the minute bar's open and close.
+        price = np.mean([minute_data['high'], minute_data['low']])
+
+        eta = ROOT_SYMBOL_TO_ETA[order.asset.root_symbol]
+        mean_volume, volatility = self._get_window_data(data, order.asset, 20)
+
+        if mean_volume == 0:
```
prsutherland: You probably want to check volatility too. If you have less than 20 days but more than 0, the volatility will be NaN but the mean volume may not be 0

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+            # If this is the first day the contract exists and there is no
+            # volume history, default to an impact of 10 bps.
+            simulated_impact = price * (10.0 / 10000)
+
+            impacted_price = \
+                price + math.copysign(simulated_impact, order.direction)
+        else:
+            txn_volume = int(
```
prsutherland: I'm wondering if we should refactor this method so that it can be shared between VolatilityVolumeShare and MarketImpact and this else block can be a call to get simulated_impact because that appears to be the only place this logic differs.

tests/finance/test_commissions.py
```diff
@@ -35,65 +59,147 @@ def generate_order_and_txns(self):
...
         self.assertEqual(0, model.calculate(order, txns[1]))
         self.assertEqual(0, model.calculate(order, txns[2]))
 
+    def test_per_trade(self):
+        # Test per trade model for equities.
+        model = PerTrade(cost=10)
+        self.verify_per_trade_commissions(model, expected_commission=10, sid=1)
+
+        # Test per trade model for futures.
+        model = PerFutureTrade(cost_per_trade=10)
```
dmichalowicz: Will change to `cost`

tests/finance/test_commissions.py
```diff
@@ -35,65 +59,147 @@ def generate_order_and_txns(self):
...
         for i, commission_total in enumerate(commission_totals):
             order.commission += model.calculate(order, txns[i])
             self.assertAlmostEqual(commission_total, order.commission)
             order.filled += txns[i].amount
 
+    def test_per_contract_no_minimum(self):
+        # Note that the exchange fee is a one-time cost that is only applied to
+        # the first fill of an order.
+        #
+        # The commission on the first fill is (230 * 0.01) + 0.3 = 2.6
```
dmichalowicz: Agreed, will change.

tests/finance/test_commissions.py
```diff
@@ -249,6 +440,29 @@ def test_per_dollar(self):
 
         self.verify_capital_used(results, [-1010, -1010, -1010])
 
+    def test_incorrectly_set_futures_model(self):
+        with self.assertRaises(UnsupportedCommissionModel):
```
dmichalowicz: Good idea

tests/finance/test_slippage.py
```diff
@@ -631,6 +660,177 @@ def test_volume_share_slippage(self):
...
+
+        orders_txns = list(
+            slippage_model.simulate(bar_data, self.ASSET1000, open_orders)
+        )
+
+        self.assertEquals(len(orders_txns), 1)
+        _, txn = orders_txns[0]
+
+        expected_txn = {
+            'price': float(5.005),
```
dmichalowicz: > Why are we calling float here?

Good question, will remove.

> Why is this the expected transaction?

I'll add some comments here

tests/finance/test_slippage.py
```diff
@@ -631,6 +660,177 @@ def test_volume_share_slippage(self):
...
+            'sid': [1000],
+            'root_symbol': ['CL'],
+            'symbol': ['CLF07'],
+            'start_date': [cls.START_DATE],
+            'end_date': [cls.END_DATE],
+            'multiplier': [500],
+            'exchange': ['CME'],
+        })
+
+    def create_order(self, asset=None, open_amount=0, limit=None, dt=None):
```
dmichalowicz: Makes sense, I'm just going to remove this helper entirely and call `Order`

tests/finance/test_slippage.py
```diff
@@ -631,6 +660,177 @@ def test_volume_share_slippage(self):
...
+        minute = self.trading_calendar.minutes_for_session(session)[1]
+        data = self.create_bardata(simulation_dt_func=lambda: minute)
+        asset = self.asset_finder.retrieve_asset(1)
+
+        mean_volume, volatility = WithWindowData()._get_window_data(
+            data, asset, window_length=20,
+        )
+
+        #                            close  volume
+        # 2006-01-31 00:00:00+00:00   29.0   119.0
```
dmichalowicz: Ok, I'll add that set up

zipline/finance/commission.py
```diff
@@ -17,9 +17,84 @@
 from abc import abstractmethod
 from six import with_metaclass
 
-DEFAULT_PER_SHARE_COST = 0.0075         # 0.75 cents per share
-DEFAULT_MINIMUM_COST_PER_TRADE = 1.0    # $1 per trade
-DEFAULT_FUTURE_COST_PER_TRADE = 2.35
+DEFAULT_PER_SHARE_COST = 0.0075       # 0.75 cents per share
+DEFAULT_PER_CONTRACT_COST = 0.85      # $0.85 per future contract
+DEFAULT_PER_DOLLAR_COST = 0.0015      # 0.15 cents per dollar
+DEFAULT_MINIMUM_COST_PER_TRADE = 1.0  # $1 per trade
```
dmichalowicz: Ya probably worth splitting this.

zipline/finance/commission.py
```diff
@@ -17,9 +17,84 @@
...
+FUTURE_EXCHANGE_FEES_BY_SYMBOL = {    # Default fees by root symbol
```
dmichalowicz: @jmccorriston provided this data to me, I don't know if there is an online reference somewhere.

zipline/finance/commission.py
```diff
@@ -59,7 +134,57 @@ def calculate(self, order, transaction):
         raise NotImplementedError('calculate')
 
 
-class PerShare(CommissionModel):
+class EquityCommissionModel(CommissionModel):
```
dmichalowicz: I updated the class hierarchies to be simpler, we now use an abstractproperty called `allowed_asset_types` to define what is supported.

> A short docstring on these classes would be nice (e.g. for sphinx docs).

Will add

zipline/finance/commission.py
```diff
@@ -59,7 +134,57 @@ def calculate(self, order, transaction):
...
+    pass
+
+
+class FutureCommissionModel(CommissionModel):
+    pass
+
+
+class PerUnit(object):
```
dmichalowicz: Yup good point, will change

zipline/finance/commission.py
```diff
@@ -59,7 +134,57 @@ def calculate(self, order, transaction):
...
+            commission.
+
+            If the order has paid a commission, start paying additional
+            commission once the minimum commission has been reached.
+
+        If there is no minimum commission:
+            Pay commission based on number of units in the transaction.
+        """
+        additional_commission = abs(transaction.amount * cost_per_unit)
+        min_trade_cost = self.min_trade_cost or 0
```
dmichalowicz: yup

zipline/finance/commission.py
```diff
@@ -79,56 +204,114 @@ def __init__(self,
...
+            )
+        )
 
     def calculate(self, order, transaction):
-        """
-        If there is a minimum commission:
-            If the order hasn't had a commission paid yet, pay the minimum
-            commission.
+        return self._calculate(
+            order, transaction, self.cost_per_share, exchange_fee=0,
```
dmichalowicz: yup will change

zipline/finance/commission.py
dmichalowicz: Thanks for the suggestion, this is definitely an improved way of doing it.

zipline/finance/commission.py
```diff
@@ -155,17 +343,41 @@ def calculate(self, order, transaction):
...
+        if isinstance(self.cost_per_trade, float):
+            cost_per_trade = self.cost_per_trade
+        else:
+            cost_per_trade = '[varies]'
+        return '{class_name}(cost_per_trade={cost_per_trade})'.format(
+            class_name=self.__class__.__name__, cost_per_trade=cost_per_trade,
+        )
+
+
+class PerDollarBase(object):
```
dmichalowicz: good observation, I'll remove.

zipline/finance/slippage.py
```diff
@@ -18,21 +18,98 @@
...
 BUY = 1 << 1
 STOP = 1 << 2
 LIMIT = 1 << 3
 
+SQRT_252 = math.sqrt(252)
 
-class LiquidityExceeded(Exception):
-    pass
+DEFAULT_VOLUME_SLIPPAGE_BAR_LIMIT = 0.025
+ROOT_SYMBOL_TO_ETA = {
```
dmichalowicz: I moved these constants to their own file and added some comments and docstrings to explain them

zipline/finance/slippage.py
```diff
@@ -162,8 +239,17 @@ def asdict(self):
...
-    """Model slippage as a function of the volume of shares traded.
+class EquitySlippageModel(SlippageModel):
+    pass
+
+
+class FutureSlippageModel(SlippageModel):
+    pass
+
+
+class VolumeSlippage(object):
```
dmichalowicz: d'oh, I'll change back. I think `VolumeShareSlippage` can just be used as is for both Equities and Futures

zipline/finance/slippage.py
dmichalowicz: good catch

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+            values = {
+                'volume': volume_history[:-1].mean(),
+                'close': close_history[:-1].pct_change().std() * SQRT_252,
+            }
+            self._window_data_cache.set(asset, values, data.current_session)
+
+        return values['volume'], values['close']
+
+
+class VolatilityVolumeShare(WithWindowData, FutureSlippageModel):
```
dmichalowicz: will add

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+
+        minute_data = data.current(order.asset, ['volume', 'high', 'low'])
+        volume = minute_data['volume']
+        if not volume:
+            return None, None
+
+        # Price to use is the average of the minute bar's open and close.
+        price = np.mean([minute_data['high'], minute_data['low']])
+
+        eta = ROOT_SYMBOL_TO_ETA[order.asset.root_symbol]
```
dmichalowicz: Yup, I'll treat this the same as passing custom commission costs

zipline/finance/slippage.py
dmichalowicz: I added a test for this case (`test_calculate_impact_without_history`)

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+        mean_volume, volatility = self._get_window_data(data, order.asset, 20)
+
+        if mean_volume == 0:
+            # If this is the first day the contract exists and there is no
+            # volume history, default to an impact of 10 bps.
+            simulated_impact = price * (10.0 / 10000)
```
dmichalowicz: yes

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+
+            impacted_price = \
```
dmichalowicz: yes

zipline/finance/slippage.py
dmichalowicz: Good idea, will do

zipline/finance/slippage.py
```diff
@@ -252,3 +349,124 @@ def process_order(self, data, order):
...
+                price + math.copysign(simulated_impact, order.direction)
+        else:
+            txn_volume = int(
+                min(volume * self.volume_limit, abs(order.open_amount))
+            )
+            psi = txn_volume / mean_volume
+
+            market_impact = eta * volatility * math.sqrt(psi)
```
dmichalowicz: This is just according to the model that @marketneutral prescribed.

zipline/finance/slippage.py
```diff
@@ -234,21 +266,224 @@ def process_order(self, data, order):
...
+        MI = eta * sigma * sqrt(psi)
+
+    Eta is a constant which varies by root symbol.
+    Sigma is 20-day annualized volatility.
+    Psi is the volume traded in the given bar divided by 20-day ADV.
+    """
+    allowed_asset_types = (Future,)
+
+    def __init__(self, volume_limit, eta=ROOT_SYMBOL_TO_ETA):
+        super(VolatilityVolumeShare, self).__init__()
```
jbredeche: might be useful to have documentation on this method explaining the types of the parameters, especially `eta` - took me a while to figure out what the allowed types were for it.

zipline/finance/slippage.py
```diff
@@ -234,21 +266,224 @@ def process_order(self, data, order):
...
+            close_volatility = close_history[:-1].pct_change()[1:].std(
+                skipna=False,
+            )
+            values = {
+                'volume': volume_history[:-1].mean(),
+                'close': close_volatility * SQRT_252,
+            }
+            self._window_data_cache.set(asset, values, data.current_session)
+
+        # from nose.tools import set_trace; set_trace()
```
dmichalowicz: oops will remove

dmichalowicz:
Got the OK from @jbredeche to merge.

PR #1754, opened 2017-04-18
Allow gaps between auto close and start date

dmichalowicz:
If ever in the middle of a futures chain a contract auto closes before the start date of the next contract, the chain would end prematurely. Instead, set the current contract during that gap to be the next upcoming contract, essentially as a placeholder.


dmichalowicz:
Thanks @ehebert

PR #1755, opened 2017-04-18
Add calendar factory for the schedule_function API

dmichalowicz:


zipline/utils/events.py
```diff
@@ -603,6 +605,24 @@ class time_rules(object):
...
+    @property
+    def us_equities(self):
+        return get_calendar('NYSE')
+
+    @property
+    def us_futures(self):
+        return get_calendar('us_futures')
+
+
+class calendars(six.with_metaclass(CalendarsMeta)):
```
ssanderson: FYI, we also have `zipline.utils.classproperty`.

dmichalowicz: Oh that's perfect! Thanks for pointing that out

zipline/algorithm.py
```diff
@@ -1106,7 +1109,12 @@ def schedule_function(self,
         # Check the type of the algorithm's schedule before pulling calendar
         # Note that the ExchangeTradingSchedule is currently the only
         # TradingSchedule class, so this is unlikely to be hit
-        cal = calendar or self.trading_calendar
+        if calendar is calendars.US_EQUITIES:
+            cal = get_calendar('NYSE')
+        elif calendar is calendars.US_FUTURES:
+            cal = get_calendar('us_futures')
+        else:
```
richafrank: Should this explicitly check for `calendar is None`? I would expect that passing some other value should raise an exception or warn me that it's being ignored.

zipline/api.pyi
```diff
@@ -545,6 +545,8 @@ def schedule_function(func, date_rule=None, time_rule=None, half_days=True, cale
         The rule for the times to execute this function.
     half_days : bool, optional
         Should this rule fire on half days?
+    calendar : Sentinel, optional
```
richafrank: FYI We regenerate this file when we [release](http://www.zipline.io/release-process.html#updating-the-python-stub-files) a new version of zipline.

zipline/utils/events.py
```diff
@@ -603,6 +605,11 @@ class time_rules(object):
     every_minute = Always
 
 
+class calendars(object):
+    US_EQUITIES = sentinel('us_equities_calendar')
```
richafrank: Nice use of `sentinel`.

zipline/api.py
```diff
@@ -53,5 +54,6 @@
     'execution',
     'math_utils',
     'slippage',
-    'time_rules'
+    'time_rules',
+    'calendars',
```
richafrank: I didn't recall discussing this addition. @ssanderson This align with your understanding?

dmichalowicz: I guess we never did discuss this piece of it. I had added it here initially for consistency with `date_rules` and `time_rules`, but I'm totally open to changing this to a more appropriate place.

zipline/api.pyi
dmichalowicz: Ok, so this addition is unnecessary then?

zipline/algorithm.py
dmichalowicz: Ah good catch. I'll add a check for `None`, and the `else` case should probably raise an exception.

zipline/api.pyi
richafrank: Yea. It certainly doesn't hurt anything, but want you to know it's not required.

zipline/api.py
richafrank: The consistency with `date_rules` and `time_rules` makes sense to me.

zipline/utils/events.py
ssanderson: Most of our uses of `sentinel` right now use the same string for both the variable name and the repr.  Any reason not to do that here?

zipline/api.py
ssanderson: I'm :+1: on namespacing this under `zipline.api.calendars`.

zipline/utils/events.py
dmichalowicz: Nope, only that I didn't know that was the convention 😄 . I do think "us_equities_calendar" is a little clearer, but am also fine with changing it to be the same as "US_EQUITIES" if that's more consistent.

dmichalowicz:
@richafrank Updated from your feedback

richafrank:
👍 from me!

PR #1757, opened 2017-04-20
Assets everywhere ... and a futures fix.

jbredeche:
The original point of this branch was to fix a bug in how we calculate a position's cost basis for futures (it wasn't taking into account contract size).

This led me into a refactoring that I had been meaning to do for a while, in which all mentions of `sid` were removed in favor of `asset` (the `sid` properties were almost always `Asset` objects already).

### DIFF
--- a/tests/test_data_portal.py
+++ b/tests/test_data_portal.py
@@ -401,6 +401,10 @@ class DataPortalTestBase(WithDataPortal,
                          "Asset 10000 had a trade on fourth minute, so should "
                          "return that as the last trade on the fifth.")
 
+    def test_get_empty_splits(self):
+        splits = self.data_portal.get_splits([], self.trading_days[2])
+        self.assertEqual([], splits)
+
 
 class TestDataPortal(DataPortalTestBase):
     DATA_PORTAL_LAST_AVAILABLE_SESSION = None
--- a/tests/test_finance.py
+++ b/tests/test_finance.py
@@ -187,8 +187,8 @@ class FinanceTestCase(WithLogger,
         # if present, expect transaction amounts to match orders exactly.
         complete_fill = params.get('complete_fill')
 
-        sid = 1
-        metadata = make_simple_equity_info([sid], self.start, self.end)
+        asset1 = self.asset_finder.retrieve_asset(1)
+        metadata = make_simple_equity_info([asset1.sid], self.start, self.end)
         with TempDirectory() as tempdir, \
                 tmp_trading_env(equities=metadata) as env:
 
@@ -206,7 +206,7 @@ class FinanceTestCase(WithLogger,
 
                 price_data = np.array([10.1] * len(minutes))
                 assets = {
-                    sid: pd.DataFrame({
+                    asset1.sid: pd.DataFrame({
                         "open": price_data,
                         "high": price_data,
                         "low": price_data,
@@ -305,7 +305,7 @@ class FinanceTestCase(WithLogger,
                     # place an order
                     direction = alternator ** len(order_list)
                     order_id = blotter.order(
-                        blotter.asset_finder.retrieve_asset(sid),
+                        asset1,
                         order_amount * direction,
                         MarketOrder())
                     order_list.append(blotter.orders[order_id])
@@ -333,7 +333,7 @@ class FinanceTestCase(WithLogger,
 
             for i in range(order_count):
                 order = order_list[i]
-                self.assertEqual(order.sid, sid)
+                self.assertEqual(order.asset, asset1)
                 self.assertEqual(order.amount, order_amount * alternator ** i)
 
             if complete_fill:
@@ -351,15 +351,19 @@ class FinanceTestCase(WithLogger,
 
             self.assertEqual(len(transactions), expected_txn_count)
 
-            cumulative_pos = tracker.position_tracker.positions[sid]
+            cumulative_pos = tracker.position_tracker.positions[asset1]
             if total_volume == 0:
                 self.assertIsNone(cumulative_pos)
             else:
                 self.assertEqual(total_volume, cumulative_pos.amount)
 
-            # the open orders should not contain sid.
+            # the open orders should not contain the asset.
             oo = blotter.open_orders
-            self.assertNotIn(sid, oo, "Entry is removed when no open orders")
+            self.assertNotIn(
+                asset1,
+                oo,
+                "Entry is removed when no open orders"
+            )
 
     def test_blotter_processes_splits(self):
         blotter = Blotter('daily', self.env.asset_finder,
@@ -376,23 +380,24 @@ class FinanceTestCase(WithLogger,
         blotter.process_splits([(2, 0.3333)])
 
         for sid in [1, 2]:
-            order_lists = blotter.open_orders[sid]
+            order_lists = \
+                blotter.open_orders[blotter.asset_finder.retrieve_asset(sid)]
             self.assertIsNotNone(order_lists)
             self.assertEqual(1, len(order_lists))
 
-        aapl_order = blotter.open_orders[1][0].to_dict()
-        fls_order = blotter.open_orders[2][0].to_dict()
+        aapl_order = blotter.open_orders[1][0]
+        fls_order = blotter.open_orders[2][0]
 
         # make sure the aapl order didn't change
-        self.assertEqual(100, aapl_order['amount'])
-        self.assertEqual(10, aapl_order['limit'])
-        self.assertEqual(1, aapl_order['sid'])
+        self.assertEqual(100, aapl_order.amount)
+        self.assertEqual(10, aapl_order.limit)
+        self.assertEqual(1, aapl_order.asset)
 
         # make sure the fls order did change
         # to 300 shares at 3.33
-        self.assertEqual(300, fls_order['amount'])
-        self.assertEqual(3.33, fls_order['limit'])
-        self.assertEqual(2, fls_order['sid'])
+        self.assertEqual(300, fls_order.amount)
+        self.assertEqual(3.33, fls_order.limit)
+        self.assertEqual(2, fls_order.asset)
 
 
 class TradingEnvironmentTestCase(WithLogger,
--- a/tests/test_finance.py
+++ b/tests/test_finance.py
@@ -275,8 +275,7 @@ class FinanceTestCase(WithLogger,
             else:
                 slippage_func = None
 
-            blotter = Blotter(sim_params.data_frequency, self.env.asset_finder,
-                              slippage_func)
+            blotter = Blotter(sim_params.data_frequency, slippage_func)
 
             start_date = sim_params.first_open
 
@@ -366,22 +365,19 @@ class FinanceTestCase(WithLogger,
             )
 
     def test_blotter_processes_splits(self):
-        blotter = Blotter('daily', self.env.asset_finder,
-                          equity_slippage=FixedSlippage())
+        blotter = Blotter('daily',  equity_slippage=FixedSlippage())
 
         # set up two open limit orders with very low limit prices,
         # one for sid 1 and one for sid 2
-        blotter.order(
-            blotter.asset_finder.retrieve_asset(1), 100, LimitOrder(10))
-        blotter.order(
-            blotter.asset_finder.retrieve_asset(2), 100, LimitOrder(10))
+        blotter.order(self.asset_finder.retrieve_asset(1), 100, LimitOrder(10))
+        blotter.order(self.asset_finder.retrieve_asset(2), 100, LimitOrder(10))
 
         # send in a split for sid 2
         blotter.process_splits([(2, 0.3333)])
 
         for sid in [1, 2]:
             order_lists = \
-                blotter.open_orders[blotter.asset_finder.retrieve_asset(sid)]
+                blotter.open_orders[self.asset_finder.retrieve_asset(sid)]
             self.assertIsNotNone(order_lists)
             self.assertEqual(1, len(order_lists))
 
--- a/tests/test_finance.py
+++ b/tests/test_finance.py
@@ -369,11 +369,16 @@ class FinanceTestCase(WithLogger,
 
         # set up two open limit orders with very low limit prices,
         # one for sid 1 and one for sid 2
-        blotter.order(self.asset_finder.retrieve_asset(1), 100, LimitOrder(10))
-        blotter.order(self.asset_finder.retrieve_asset(2), 100, LimitOrder(10))
+        asset1 = self.asset_finder.retrieve_asset(1)
+        asset2 = self.asset_finder.retrieve_asset(2)
+        asset133 = self.asset_finder.retrieve_asset(133)
 
-        # send in a split for sid 2
-        blotter.process_splits([(2, 0.3333)])
+        blotter.order(asset1, 100, LimitOrder(10))
+        blotter.order(asset2, 100, LimitOrder(10))
+
+        # send in splits for assets 133 and 2.  We have no open orders for
+        # asset 133 so it should be ignored.
+        blotter.process_splits([(asset133, 0.5), (asset2, 0.3333)])
 
         for sid in [1, 2]:
             order_lists = \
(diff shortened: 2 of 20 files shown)


zipline/protocol.py
```diff
@@ -242,6 +247,7 @@ def __repr__(self):
     __getitem__ = _deprecated_getitem_method(
         'position', {
             'sid',
+            'asset',
```
jbredeche: now the user-facing `Position` object will have both `sid` and `asset` properties.  is that OK?

jbredeche:
cc @ssanderson , the refactoring we have been discussing.

zipline/finance/performance/period.py
```diff
@@ -392,15 +392,15 @@ def _calculate_execution_cash_flow(self, txn):
         # Check if the multiplier is cached. If it is not, look up the asset
         # and cache the multiplier.
         try:
-            multiplier = self._execution_cash_flow_multipliers[txn.sid]
+            multiplier = self._execution_cash_flow_multipliers[txn.asset]
         except KeyError:
-            asset = self.asset_finder.retrieve_asset(txn.sid)
+            asset = self.asset_finder.retrieve_asset(txn.asset)
```
ssanderson: Do we still need to retrieve the asset here?

ssanderson: (If not do we still need the multiplier caches here?)

zipline/finance/performance/position.py
```diff
@@ -167,14 +173,18 @@ def adjust_commission_cost_basis(self, sid, cost):
             return
 
         prev_cost = self.cost_basis * self.amount
-        new_cost = prev_cost + cost
+        if isinstance(asset, Future):
```
jbredeche: Correct

zipline/finance/performance/position.py
```diff
@@ -186,7 +196,7 @@ def to_dict(self):
         Returns a dict object of the form:
         """
         return {
-            'sid': self.sid,
+            'sid': self.asset,
```
jbredeche: I think so. This is how the perf packets are constructed.

zipline/finance/performance/position.py
dmichalowicz: +1 on adding a `multiplier` to `Equity`.

jbredeche: I'm +1 on that too, but not doing it in this PR.

zipline/data/data_portal.py
```diff
@@ -1121,24 +1121,24 @@ def _ensure_asset_dates(self, asset):
...
 
         Returns
         -------
-        splits : list[(int, float)]
-            List of splits, where each split is a (sid, ratio) tuple.
+        splits : list[(asset, float)]
+            List of splits, where each split is a (asset, ratio) tuple.
         """
-        if self._adjustment_reader is None or not sids:
+        if self._adjustment_reader is None or not assets:
```
jbredeche: good find, thanks

zipline/data/data_portal.py
```diff
@@ -1149,7 +1149,9 @@ def get_splits(self, sids, dt):
             "SELECT sid, ratio FROM SPLITS WHERE effective_date = ?",
             (seconds,)).fetchall()
 
-        splits = [split for split in splits if split[0] in sids]
+        splits = [split for split in splits if split[0] in assets]
+        splits = [(self.asset_finder.retrieve_asset(split[0]), split[1])
```
jbredeche: totally agreed

zipline/finance/blotter.py
```diff
@@ -306,18 +303,19 @@ def process_splits(self, splits):
...
         splits: list
-            A list of splits.  Each split is a tuple of (sid, ratio).
+            A list of splits.  Each split is a tuple of (asset, ratio), where
+            sid is an integer.
 
         Returns
         -------
         None
         """
         for split in splits:
```
jbredeche: agreed, fixed

zipline/data/data_portal.py
ehebert: :+1: on using adjustment reader.

zipline/finance/performance/position.py
ehebert: I am closer to :-1: than :1: on making the change to Equity to include a multiplier value.

- There are places where we do dispatch because `Equities`and `Futures` behave differently. 
e.g. cases like `period.PerformancePeriod.handle_execution` where that value should be a no-op for `Equity`s. I believe in most cases we do want to dispatch on Asset type because we want to consider whether the contract contributes to exposure, should calculate a payout, etc.
- `multiplier` is something that is inherit to the definition of a future contract; adding it to an `Equity` object would be clouding the modelling of an equity which may lead to regressions in treating Assets alike in cases where they should not be.

Needless to say, 'not for this PR', but have we considered modelling the portfolio by having an `EquityPortfolio` and a `FuturePortfolio`, where the overall `Portfolio` is compromised of the two?
This may allow us to model the unique behaviors of each more cleanly.

jbredeche:
@ssanderson thanks for the review.  You were completely correct about the places that were now doing redundant lookups, and I've since removed assetfinder and multiplier caching from `PositionTracker` and the assetfinder from  `PerformancePeriod`.

zipline/finance/performance/position_tracker.py
```diff
@@ -133,41 +136,42 @@ def __init__(self, asset_finder, data_frequency):
...
-                self._position_exposure_multipliers[sid] = asset.multiplier
+                self._position_value_multipliers[asset] = 0
+                self._position_exposure_multipliers[asset] = asset.multiplier
 
     def update_positions(self, positions):
         # update positions in batch
         self.positions.update(positions)
-        for sid, pos in iteritems(positions):
-            self._update_asset(sid)
+        for asset, pos in iteritems(positions):
```
jbredeche: Entire method is gone now.

zipline/finance/performance/position_tracker.py
```diff
@@ -133,41 +136,42 @@ def __init__(self, asset_finder, data_frequency):
...
+            self._position_exposure_multipliers[asset]
         except KeyError:
             # Check if there is an AssetFinder
             if self.asset_finder is None:
                 raise PositionTrackerMissingAssetFinder()
 
-            # Collect the value multipliers from applicable sids
-            asset = self.asset_finder.retrieve_asset(sid)
+            # Collect the value multipliers from applicable assets
+            asset = self.asset_finder.retrieve_asset(asset)
```
jbredeche: nope. all gone.

zipline/finance/performance/position_tracker.py
```diff
@@ -216,7 +221,8 @@ def handle_splits(self, splits):
         Parameters
         ----------
         splits: list
-            A list of splits.  Each split is a tuple of (sid, ratio).
+            A list of splits.  Each split is a tuple of (sid, ratio), where
```
jbredeche: I just typo-ed the comment, we were already receiving assets.

zipline/finance/performance/position_tracker.py
```diff
@@ -133,41 +136,42 @@ def __init__(self, asset_finder, data_frequency):
 
         self.data_frequency = data_frequency
 
-    def _update_asset(self, sid):
+    def _update_asset(self, asset):
```
jbredeche: yup!

zipline/protocol.py
```diff
@@ -121,6 +123,7 @@ class Order(Event):
         'order', {
             'dt',
             'sid',
+            'asset',
```
jbredeche: fixed

zipline/finance/performance/period.py
```diff
@@ -362,7 +362,7 @@ def record_order(self, order):
     def handle_execution(self, txn):
         self.cash_flow += self._calculate_execution_cash_flow(txn)
 
-        asset = self.asset_finder.retrieve_asset(txn.sid)
+        asset = self.asset_finder.retrieve_asset(txn.asset)
```
jbredeche: no more assetfinder

zipline/finance/performance/period.py
jbredeche: no more caches

tests/test_data_portal.py
```diff
@@ -401,6 +401,10 @@ def test_get_last_traded_dt_minute(self):
                          "Asset 10000 had a trade on fourth minute, so should "
                          "return that as the last trade on the fifth.")
 
+    def test_get_empty_splits(self):
```
ssanderson: :+1: on having an explicit test for this case.

tests/test_finance.py
```diff
@@ -351,48 +350,55 @@ def transaction_sim(self, **params):
...
-        # send in a split for sid 2
-        blotter.process_splits([(2, 0.3333)])
+        blotter.order(asset1, 100, LimitOrder(10))
+        blotter.order(asset2, 100, LimitOrder(10))
+
+        # send in splits for assets 133 and 2.  We have no open orders for
+        # asset 133 so it should be ignored.
+        blotter.process_splits([(asset133, 0.5), (asset2, 0.3333)])
 
         for sid in [1, 2]:
```
ssanderson: This could just be `for asset in asset1, asset2:`.

zipline/finance/performance/position_tracker.py
```diff
@@ -226,13 +187,12 @@ def handle_splits(self, splits):
         total_leftover_cash = 0
 
         for split in splits:
-            sid = split[0]
-            if sid in self.positions:
+            asset = split[0]
```
jbredeche: I should probably do `for asset, ratio in splits` here

zipline/finance/performance/period.py
```diff
@@ -392,15 +390,15 @@ def _calculate_execution_cash_flow(self, txn):
         # Check if the multiplier is cached. If it is not, look up the asset
         # and cache the multiplier.
         try:
-            multiplier = self._execution_cash_flow_multipliers[txn.sid]
+            multiplier = self._execution_cash_flow_multipliers[txn.asset]
```
jbredeche: this `_execution_cash_flow_multipliers` cache can probably be removed

tests/test_perf_tracking.py
```diff
@@ -1974,6 +1954,16 @@ class TestPositionTracker(WithTradingEnvironment,
...
 
     @classmethod
+    def init_class_fixtures(cls):
+        super(TestPositionTracker, cls).init_class_fixtures()
+
+        cls.EQUITY1 = cls.asset_finder.retrieve_asset(1)
+        cls.EQUITY2 = cls.asset_finder.retrieve_asset(2)
+        cls.FUTURE3 = cls.asset_finder.retrieve_asset(3)
+        cls.FUTURE4 = cls.asset_finder.retrieve_asset(4)
+        cls.FUTURE5 = cls.asset_finder.retrieve_asset(1032201401)
```
ssanderson: Is there a reason this future is way different from the others?

tests/test_perf_tracking.py
```diff
@@ -2015,17 +2004,27 @@ def test_empty_positions(self):
             self.assertNotIsInstance(val, (bool, np.bool_))
 
     def test_position_values_and_exposures(self):
-        pt = perf.PositionTracker(self.env.asset_finder, None)
+        pt = perf.PositionTracker(None)
```
ssanderson: `data_frequency=None` seems odd here...

tests/test_perf_tracking.py
jbredeche: Was just porting the existing test over.  Don't know why it's such a high number.

tests/test_perf_tracking.py
```diff
@@ -2015,17 +2004,27 @@ def test_empty_positions(self):
...
-        pos1 = perf.Position(1, amount=np.float64(10.0),
-                             last_sale_date=dt, last_sale_price=10)
-        pos2 = perf.Position(2, amount=np.float64(-20.0),
-                             last_sale_date=dt, last_sale_price=10)
-        pos3 = perf.Position(3, amount=np.float64(30.0),
-                             last_sale_date=dt, last_sale_price=10)
-        pos4 = perf.Position(4, amount=np.float64(-40.0),
-                             last_sale_date=dt, last_sale_price=10)
-        pt.update_positions({1: pos1, 2: pos2, 3: pos3, 4: pos4})
+        pt.update_position(
```
ssanderson: Is there a reason this changed from a single call to `update_positions` to multiple calls to `update_position`?

jbredeche: I removed the `update_positions` method, as its primary usecase was to update all those internal caches we no longer need.

I can put it back, and just have it call `update_position` on all the incoming positions.

zipline/errors.py
```diff
@@ -665,18 +665,6 @@ class UnsupportedDatetimeFormat(ZiplineError):
            "coercible to a pandas.Timestamp object.")
 
 
-class PositionTrackerMissingAssetFinder(ZiplineError):
```
ssanderson: :tada:

zipline/finance/order.py
```diff
@@ -86,9 +87,16 @@ def to_dict(self):
         if self.broker_order_id is None:
             del dct['broker_order_id']
 
+        dct['sid'] = self.asset
```
ssanderson: Can we add a comment here explaining that we use `sid` here rather than `'asset'` for backwards compat with downstream consumers of the perf packet stream?

zipline/finance/performance/position_tracker.py
```diff
@@ -52,37 +48,36 @@
...
-        price * amount * multiplier for
-        price, amount, multiplier in iter_amount_price_multiplier
-    ]
+def calc_position_values(positions):
+    values = []
+
+    for position in positions:
+        if isinstance(position.asset, Future):
+            # Futures don't have an inherent position value.
+            values.append(0)
```
ssanderson: You probably want `0.0` here to avoid unnecessary coercions later.

zipline/finance/performance/position_tracker.py
```diff
@@ -119,55 +114,26 @@ def calc_gross_value(long_value, short_value):
...
-        # Arrays for quick calculations of positions value
-        self._position_value_multipliers = OrderedDict()
-        self._position_exposure_multipliers = OrderedDict()
         self._unpaid_dividends = {}
         self._unpaid_stock_dividends = {}
         self._positions_store = zp.Positions()
 
         self.data_frequency = data_frequency
 
-    def _update_asset(self, sid):
```
ssanderson: :tada:

zipline/finance/transaction.py
```diff
@@ -39,6 +38,9 @@ def __getitem__(self, name):
     def to_dict(self):
         py = copy(self.__dict__)
         del py['type']
+        del py['asset']
+        py['sid'] = self.asset
```
ssanderson: same note here re: having a comment about this.

zipline/protocol.py
```diff
@@ -225,14 +227,17 @@ def __repr__(self):
...
+    @expect_types(asset=Asset)
+    def __init__(self, asset):
+        self.asset = asset
         self.amount = 0
         self.cost_basis = 0.0  # per share
         self.last_sale_price = 0.0
         self.last_sale_date = None
 
+        # for backwards compatibility
+        self.sid = asset
```
ssanderson: Is there a reason we're aliasing this here but making it a property in https://github.com/quantopian/zipline/pull/1757/files#diff-506712f0fea50ba9a660883d95d114d4R94 ?

ssanderson: I'd probably prefer to use a property here unless you're concerned about performance overhead.  As this stands, it's possible for `sid` and `asset` to diverge if someone mutates one and not the other after construction.  (Admittedly, doing that is probably a bad idea.)

zipline/utils/factory.py
```diff
@@ -150,27 +152,27 @@ def create_split(sid, ratio, date):
...
-        'sid': sid,
-        'amount': amount,
-        'dt': datetime,
-        'price': price,
-        'type': DATASOURCE_TYPE.TRANSACTION,
-        'source_id': 'MockTransactionSource'
-    })
-    return txn
+@expect_types(asset=Asset)
+def create_txn(asset, price, amount, datetime, order_id):
```
ssanderson: FWIW, if you felt like being fancy, you could just write this as `create_txn = expect_types(asset=Asset)(Transaction)`.

ssanderson: (Probably not worth it though).

zipline/utils/factory.py
```diff
@@ -150,27 +152,27 @@ def create_split(sid, ratio, date):
...
                        trading_calendar):
     txns = []
     current = sim_params.first_open
 
     for price, amount in zip(priceList, amtList):
-        current = get_next_trading_dt(current, interval, trading_calendar)
+        dt = get_next_trading_dt(current, interval, trading_calendar)
 
-        txns.append(create_txn(sid, price, amount, current))
+        txns.append(create_txn(asset, price, amount, dt, None))
```
ssanderson: Should we generate order_ids here?

ssanderson:
@jbredeche finished a second pass.  Had a few small comments, but mostly looks good to me.

zipline/utils/factory.py
jbredeche: nah, mine is easier to read.

zipline/utils/factory.py
jbredeche: maybe, but it wasn't needed for my work so I didn't see the need to do it now.

zipline/protocol.py
jbredeche: sounds good, switched to property.

zipline/finance/transaction.py
jbredeche: yup

zipline/finance/performance/position_tracker.py
jbredeche: good call

zipline/finance/order.py
jbredeche: yup

tests/test_perf_tracking.py
jbredeche: agreed. was already there, can fix later.

zipline/finance/performance/period.py
jbredeche: removed

PR #1762, opened 2017-04-25
MAINT: Use a quarterly cycle for JY, CD, AD, and BP continuous futures

yankees714:
The March, June, September, and December contracts for these futures
contain most of the trading activity, so we exclude the other more
sparsely traded contracts from the chain.


yankees714:
@ehebert Good point, I'll just take care of that while I'm here, thanks!

PR #1767, opened 2017-04-26
More futures slippage cleanup

dmichalowicz:
If a backtest starts on a date which does not have enough lookback days for the `VolatilityVolumeShare` slippage model, fall back on a predefined constant.

Confirmed this behavior with @gusgordon 

This PR also adds a feature that always for subclassing `EquitySlippageModel` and `FutureSlippageModel` together and getting the expected behavior that it supports both equities and futures.

### DIFF
--- a/tests/finance/test_slippage.py
+++ b/tests/finance/test_slippage.py
@@ -26,13 +26,15 @@ import numpy as np
 import pandas as pd
 import pytz
 
-from zipline.assets import Equity
+from zipline.assets import Equity, Future
 from zipline.data.data_portal import DataPortal
 from zipline.finance.asset_restrictions import NoRestrictions
 from zipline.finance.order import Order
 from zipline.finance.slippage import (
+    EquitySlippageModel,
     fill_price_worse_than_limit_price,
-    NO_DATA_VOLATILITY_SLIPPAGE_IMPACT,
+    FutureSlippageModel,
+    SlippageModel,
     VolatilityVolumeShare,
     VolumeShareSlippage,
 )
@@ -104,6 +106,47 @@ class SlippageTestCase(WithCreateBarData,
 
         self.assertEqual(vol1.__dict__, vol1.asdict())
         self.assertEqual(vol2.__dict__, vol2.asdict())
+
+    def test_allowed_asset_types(self):
+        # Custom equities model.
+        class MyEquitiesModel(EquitySlippageModel):
+            def process_order(self, data, order):
+                return 0, 0
+
+        self.assertEqual(MyEquitiesModel.allowed_asset_types, (Equity,))
+
+        # Custom futures model.
+        class MyFuturesModel(FutureSlippageModel):
+            def process_order(self, data, order):
+                return 0, 0
+
+        self.assertEqual(MyFuturesModel.allowed_asset_types, (Future,))
+
+        # Custom model for both equities and futures.
+        class MyMixedModel(EquitySlippageModel, FutureSlippageModel):
+            def process_order(self, data, order):
+                return 0, 0
+
+        self.assertEqual(MyMixedModel.allowed_asset_types, (Equity, Future))
+
+        # Equivalent custom model for both equities and futures.
+        class MyMixedModel(SlippageModel):
+            def process_order(self, data, order):
+                return 0, 0
+
+        self.assertEqual(MyMixedModel.allowed_asset_types, (Equity, Future))
+
+        SomeType = type('SomeType', (object,), {})
+
+        # A custom model that defines its own allowed types should take
+        # precedence over the parent class definitions.
+        class MyCustomModel(EquitySlippageModel, FutureSlippageModel):
+            allowed_asset_types = (SomeType,)
+
+            def process_order(self, data, order):
+                return 0, 0
+
+        self.assertEqual(MyCustomModel.allowed_asset_types, (SomeType,))
 
     def test_fill_price_worse_than_limit_price(self):
         non_limit_order = TestOrder(limit=None, direction=1)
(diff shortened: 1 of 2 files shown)


zipline/finance/slippage.py
```diff
@@ -175,14 +175,45 @@ def asdict(self):
...
+    example) would override the 'allowed_asset_types' class attribute to only
+    include equities. However, this is probably not the expected behavior for a
+    reasonable user, so the metaclass will handle this specific case by
+    manually allowing both equities and futures for the class being created.
+    """
+    def __new__(mcls, name, bases, dict_):
+        try:
+            if EquitySlippageModel in bases and FutureSlippageModel in bases:
+                dict_['allowed_asset_types'] = (Equity, Future)
+        except NameError:
```
richafrank: The tested behavior looks good to me! 👍 

Implementation-wise I'd probably make this a bit more generic and potentially more robust by de-coupling the metaclass from its current marker instances (`EquitySlippageModel`, `FutureSlippageModel`). The `NameError->pass` makes me nervous that we might mask some unexpected failure.

We can pull out a marker metaclass from this base metaclass and make the base metaclass handle any marker instance, instead of just the two. Then we can add more markers without renaming the metaclass and changing its implementation.

```python
class AssetSlippageMeta(ABCMeta):
    def __new__(mcls, name, bases, dict_):
        if 'allowed_asset_types' not in dict_:
            allowed_asset_types = tuple(
                chain.from_iterable(marker.allowed_asset_types 
                                    for marker in bases
                                    if isinstance(marker, AssetSlippageMarker))
            )
            if allowed_asset_types:
                dict_['allowed_asset_types'] = allowed_asset_types

        return super(AssetSlippageMeta, mcls).__new__(
            mcls, name, bases, dict_,
        )


class AssetSlippageMarker(AssetSlippageMeta):
    pass


class SlippageModel(with_metaclass(AssetSlippageMeta)):
...
class EquitySlippageModel(with_metaclass(AssetSlippageMarker, SlippageModel)):
...
class FutureSlippageModel(with_metaclass(AssetSlippageMarker, SlippageModel)):
...
```

tests/finance/test_slippage.py
```diff
@@ -105,6 +106,28 @@ def test_equality_and_comparison(self):
...
+                return 0, 0
+
+        self.assertEqual(MyFuturesModel.allowed_asset_types, (Future,))
+
+        # Custom model for both equities and futures.
+        class MyMixedModel(EquitySlippageModel, FutureSlippageModel):
+            def process_order(self, data, order):
+                return 0, 0
+
+        self.assertEqual(MyMixedModel.allowed_asset_types, (Equity, Future))
```
richafrank: The are great. Should we document (i.e. test) what happens if your custom model subclasses just `SilppageModel`, or how a custom model that defines `allowed_asset_types` itself interacts with its base classes?

zipline/finance/slippage.py
dmichalowicz: I like this! Only comment I have is that in the `for` loop the word "marker" is slightly confusing in that it's only maybe a marker; might just say "base_class".

tests/finance/test_slippage.py
dmichalowicz: Yes good point, I'll add these

zipline/finance/slippage.py
```diff
@@ -77,7 +78,45 @@ def fill_price_worse_than_limit_price(fill_price, order):
     return False
 
 
-class SlippageModel(with_metaclass(ABCMeta)):
+class SlippageModelMeta(ABCMeta):
```
dmichalowicz: @richafrank Thanks for the feedback, I updated based on your comments. I also decided to call this `SlilppageModelMeta` instead of `AssetSlippageMeta` since we are now using it as the metaclass for the core `SlippageModel` class, but let me know if you have feelings against this.

richafrank: Sounds fine to me.

zipline/finance/slippage.py
```diff
@@ -77,7 +78,45 @@ def fill_price_worse_than_limit_price(fill_price, order):
...
+    reasonable user, so the metaclass will handle this specific case by
+    manually allowing both equities and futures for the class being created.
+    """
+
+    def __new__(mcls, name, bases, dict_):
+        if 'allowed_asset_types' not in dict_:
+            allowed_asset_types = tuple(
+                chain.from_iterable(
+                    base_class.allowed_asset_types
+                    for base_class in bases
```
richafrank: > I like this! Only comment I have is that in the for loop the word "marker" is slightly confusing in that it's only maybe a marker; might just say "base_class".

Yup, I see your point. The challenge is that there are two uses here: the first is the for-loop var which is an element of `bases`; the second is an instance of `AssetSlippageMarker` which has the `allowed_asset_types` attribute (line above this can make that assumption). I find it surprising to see `base_class.allowed_asset_types`, but you find it confusing to name each element of `bases` as `marker`. Both are valid, so feel free to use what you have here.

tests/finance/test_slippage.py
```diff
@@ -105,6 +107,47 @@ def test_equality_and_comparison(self):
...
+
+        # A custom model that defines its own allowed types should take
+        # precedence over the parent class definitions.
+        class MyCustomModel(EquitySlippageModel, FutureSlippageModel):
+            allowed_asset_types = (SomeType,)
+
+            def process_order(self, data, order):
+                return 0, 0
+
+        self.assertEqual(MyCustomModel.allowed_asset_types, (SomeType,))
```
dmichalowicz: @richafrank updated

richafrank: 👍

PR #1770, opened 2017-04-26
Futures slippage model could have zero transaction volume

dmichalowicz:
If the computed transaction volume is less than 1 just bail.


jbredeche:
LGTM pending tests.

PR #1775, opened 2017-04-30
ENH: upgrade pandas to 0.19.2

kennell:
Hi folks. This upgrades Pandas to version 0.19.2 and fixes some issues dealing with the pickle'd output when using Python 3.6, see https://github.com/pandas-dev/pandas/issues/14679 for details on the issue.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -17,7 +17,7 @@ requests-file==1.4.1
 # scipy and pandas are required for statsmodels,
 # statsmodels in turn is required for some pandas packages
 scipy==0.17.1
-pandas==0.18.1
+pandas==0.19.2
 pandas-datareader==0.2.1
 # Needed for parts of pandas.stats
 patsy==0.4.0
--- a/setup.py
+++ b/setup.py
@@ -162,7 +162,7 @@ def _filter_requirements(lines_iter, filter_names=None,
 
 REQ_UPPER_BOUNDS = {
     'bcolz': '<1',
-    'pandas': '<0.19',
+    'pandas': '<0.19.3',
 }
 
 


Peque:
Maybe project admins can re-run the Travis build, which seems it never reported back. Otherwise maybe you could try to push some changes to force update?.

I would use `'pandas': '<0.20'` in `REQ_UPPER_BOUNDS` instead (just like there was `<0.19` before). That way future minor `0.19.x` versions can be used as well (minor are usually just bug fixes and do not break things).

kennell:
Folks, any chance we can get this merged? Regarding the travis test, perhaps someone can re-trigger.

richafrank:
Hey @kennell , could you tell us more about what this fixes?

I see two changes here: 

1. The first increments the lower bound of required pandas versions, which I don't think we want to do. The quantopian platform on the website is running zipline with the lower pandas version, so we very much want to support that version right now.
1. The second increments the upper bound of support, which I don't have a strong opinion on, assuming zipline works. It looks like travis and appveyor are running tests only with the lower bound version, so could you update the test matrix to assess both the lower and upper? I believe @ssanderson last bumped that upper bound, so probably he could share how he chose the version listed on master.

Thanks!

richafrank:
> Folks, any chance we can get this merged? Regarding the travis test, perhaps someone can re-trigger.

Travis doesn't seem to know about that build, so there's nothing I can re-run. Want to update your branch to trigger a new run?

Peque:
Note that you may need to update some other files (see #1810).

Also, I would rather set `'pandas': '<0.20'` for the upper bounds, to allow future minor versions, which could fix bugs but should not break compatibility.

FreddieV4:
Going to close this due to inactivity. Feel free to re-open or open a new PR 🙂

Peque:
Created #1975 for those who wish to subscribe/comment.

PR #1777, opened 2017-05-02
MAINT: Create a __eq__ for commission objects

lianga888:
Just like for the slippage objects, allows commission object
instances to equate if they're called with the same params

### DIFF
--- a/tests/finance/test_commissions.py
+++ b/tests/finance/test_commissions.py
@@ -79,6 +79,16 @@ class CommissionUnitTests(WithAssetFinder, ZiplineTestCase):
         self.assertEqual(0, model.calculate(order, txns[1]))
         self.assertEqual(0, model.calculate(order, txns[2]))
 
+    def test_equality_and_comparison(self):
+        perdollar1 = PerDollar(cost=0.2)
+        perdollar2 = PerDollar(cost=0.2)
+
+        self.assertEqual(perdollar1, perdollar2)
+        self.assertEqual(hash(perdollar1), hash(perdollar2))
+
+        self.assertEqual(perdollar1.__dict__, perdollar1.asdict())
+        self.assertEqual(perdollar2.__dict__, perdollar2.asdict())
+
     def test_per_trade(self):
         # Test per trade model for equities.
         model = PerTrade(cost=10)
--- a/zipline/finance/commission.py
+++ b/zipline/finance/commission.py
@@ -16,7 +16,7 @@ import abc
 from abc import abstractmethod
 from collections import defaultdict
 
-from six import with_metaclass
+from six import with_metaclass, iteritems
 from toolz import merge
 
 from zipline.assets import Equity, Future
@@ -68,6 +68,18 @@ class CommissionModel(with_metaclass(abc.ABCMeta)):
             this order.
         """
         raise NotImplementedError('calculate')
+
+    def __eq__(self, other):
+        return self.asdict() == other.asdict()
+
+    def __hash__(self):
+        return hash((
+            type(self),
+            tuple(sorted(iteritems(self.asdict())))
+        ))
+
+    def asdict(self):
+        return self.__dict__
 
 
 class EquityCommissionModel(CommissionModel):


zipline/finance/commission.py
```diff
@@ -69,6 +69,18 @@ def calculate(self, order, transaction):
         """
         raise NotImplementedError('calculate')
 
+    def __eq__(self, other):
+        return self.asdict() == other.asdict()
+
+    def __hash__(self):
+        return hash((
+            type(self),
+            tuple(sorted(iteritems(self.asdict())))
```
richafrank: Is it important that these be hashable? I ask only because this implementation implies hash-ability of all the instance's attributes, which could be restrictive for subclass writers.

tests/finance/test_commissions.py
```diff
@@ -79,6 +79,16 @@ def verify_per_trade_commissions(self,
...
 
+    def test_equality_and_comparison(self):
+        perdollar1 = PerDollar(cost=0.2)
+        perdollar2 = PerDollar(cost=0.2)
+
+        self.assertEqual(perdollar1, perdollar2)
+        self.assertEqual(hash(perdollar1), hash(perdollar2))
+
+        self.assertEqual(perdollar1.__dict__, perdollar1.asdict())
+        self.assertEqual(perdollar2.__dict__, perdollar2.asdict())
```
richafrank: My instinct is to remove these last two assertions, since they imply implementation requirements - as opposed to interface requirements, which you're checking above. I could see asserting that the return value is `{'cost_per_dollar': ...}`, but I'd probably remove it instead, since `asdict` is really more of an implementation detail of `__eq__` and `__hash__`.

tests/finance/test_commissions.py
```diff
@@ -79,6 +79,16 @@ def verify_per_trade_commissions(self,
         self.assertEqual(0, model.calculate(order, txns[1]))
         self.assertEqual(0, model.calculate(order, txns[2]))
 
+    def test_equality_and_comparison(self):
+        perdollar1 = PerDollar(cost=0.2)
+        perdollar2 = PerDollar(cost=0.2)
+
+        self.assertEqual(perdollar1, perdollar2)
```
richafrank: 👍 Very clear test.

zipline/finance/commission.py
lianga888: Hmm yeah that's a good point. My main rationale for implementing it was simply that is was implemented for slippage as well https://github.com/quantopian/zipline/blob/master/zipline/finance/slippage.py#L168

I don't feel strongly either way, but I think commissions and slippage should both implement the `__hash__` or neither. Unless there's something special about slippage that reasonably restricts its attributes to all being hashable

richafrank: Barring some known reason to keep them, my tendency would be to remove 'em. Probably run all integration tests to check though. @ssanderson Any opinion?

zipline/finance/commission.py
```diff
@@ -69,6 +69,18 @@ def calculate(self, order, transaction):
         """
         raise NotImplementedError('calculate')
 
+    def __eq__(self, other):
+        return self.asdict() == other.asdict()
```
ssanderson: A few notes here:

1. I think this `__eq__` implementation will currently incorrectly compare equal for objects of different types but with the same instance dictionary.  That's actually a realistic case too, b/c we have futures/equities versions of some models that are otherwise very similar.  The easy fix for this is to check if `type(other) == type(self)` along with whatever other comparisons are being made.  However...
2. In general, I don't think it's safe to just use comparison against `__dict__` for equality comparisons.  In particular, this implementation will only work if commission models only ever hold objects with nicely-behaved `__eq__` implementations/  Off the top of my head I can think of at least three cases of objects that can't be stored on a commission model without breaking this implementation: NaN, numpy/pandas objects, and custom objects that don't define `__eq__`.

The NaN case means that equality will almost always return false if one of the dictionary values is `NaN`:
```
In [23]: data = np.full(5, np.nan)

In [24]: d1 = {'a': 1, 'b': data[0]}

In [25]: d2 = {'a': 1, 'b': data[1]}

In [26]: d1 == d2
Out[26]: False
```

I say **almost**, because the dictionary constructor will short-circuit on object identity, so you can actually sometimes get the expected result depending on where your NaNs came from:

```
In [27]: d1 = {'a': 1, 'b': np.nan}

In [28]: d2 = {'a': 1, 'b': np.nan}

In [29]: d1 == d2
Out[29]: True
```

w/r/t numpy/pandas objects, `__eq__` returns a new numpy/pandas object, which will then raise an error when the dict comparator tries to call `__bool__`.

w/r/t other objects, there's no guarantee in general that every instance attribute of every commission model implements comparison the way you'd expect.  Notably, for example, the [`DummyMapping` class used by `PerContract`](https://github.com/quantopian/zipline/pull/1777/files#diff-0c146aa0be7e8ccc3c741f35e4928fb4R178) doesn't implement a `__eq__`, so this implementation will probably say that all `PerContract` instances are unequal.  (I think we also have a similar bug with the `VolatilityVolumeShare` slippage model)

ssanderson: The conclusion from all of the above, I think, is that there's no safe and general way to implement `__eq__` in terms of the instance members of an arbitrary subclass.  If we want to be able to equality-compare any two instances of CommissionModel, then that imposes constraints on what a subclass of CommissionModel can do, which means we probably need tests for each subclass of CommissionModel to ensure that they actually abide by those constraints.  In particular, as noted above, I think `__eq__` is broken right now for `PerContract`.

PR #1778, opened 2017-05-03
MAINT: Handle gaps in input to daily bars writer

yankees714:
Previously, a dataframe pass into BcolzDailyBarWriter.write that was
missing an expected session between its first and last sessions would be
written incorrectly. Upon converting the dataframe to a ctable, the
values for all days following the gap would be shifted backwards, and
nans would be shifted in at the end.

This commit handles the issue by reindexing the input dataframe against
the expected sessions, which will fill any gaps with nans at the
expected index. Also adds BcolzDailyBarWriterMissingDataTestCase to
catch any regressions.

### DIFF
--- a/zipline/data/us_equity_pricing.py
+++ b/zipline/data/us_equity_pricing.py
@@ -50,6 +50,7 @@ from six import (
     string_types,
     viewkeys,
 )
+from toolz import compose
 
 from zipline.data.session_bars import SessionBarReader
 from zipline.data.bar_reader import (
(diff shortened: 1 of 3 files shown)


zipline/data/us_equity_pricing.py
```diff
@@ -355,15 +344,30 @@ def iterator(iterator=iterator, assets=set(assets)):
...
+
+            asset_sessions = sessions[
+                sessions.slice_indexer(asset_first_day, asset_last_day)
+            ]
+            assert len(table) == len(asset_sessions), (
+                'Got {} rows for daily bars table with first day={}, last '
+                'day={}, expected {} rows.'.format(
+                    len(table),
+                    asset_first_day.date(),
+                    asset_last_day.date(),
```
ehebert: Wouldn't block on it. But we could put the different values between `table['days']` and `sessions` in to the error message.

yankees714: Agree the diff would be useful - going to merge as-is to get the change in, then come back to this. Thanks!

PR #1779, opened 2017-05-03
Gold/Silver futures contracts only trade certain months

dmichalowicz:
Sources: These are the month codes for silver, http://www.mrci.com/web/online-explanation-pages/symbols-a-codes.html and https://www.theice.com/products/31500923/Mini-Silver-Future

Fix silver continuous futures (calendar roll) from looking like this:

screen shot 2017-05-03 at 3 12 19 pm

to this:

screen shot 2017-05-03 at 3 12 42 pm


dmichalowicz:
Got the ok from @yankees714 and @jmccorriston

PR #1780, opened 2017-05-04
MAINT: Display diff if input to daily bar writer has gaps/extra bars

yankees714:


yankees714:
Follow-up to https://github.com/quantopian/zipline/pull/1778#discussion_r114679440.

PR #1783, opened 2017-05-05
Slippage cleanup

richafrank:
- `process_order` is a method. Call super's `__init__` to set up base state
- Fixed abstractness of `MarketImpactBase`

### DIFF
--- a/zipline/finance/slippage.py
+++ b/zipline/finance/slippage.py
@@ -193,8 +193,6 @@ class VolumeShareSlippage(SlippageModel):
     """
     Model slippage as a function of the volume of contracts traded.
     """
-    allowed_asset_types = (Equity, Future)
-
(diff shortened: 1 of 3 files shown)


zipline/finance/slippage.py
```diff
@@ -290,7 +291,7 @@ def process_order(self, data, order):
         )
 
 
-class MarketImpactBase(object):
+class MarketImpactBase(with_metaclass(ABCMeta)):
```
dmichalowicz: What do you think of having this subclass `SlippageModel` instead of setting the metaclass, and then in `VolatilityVolumeShare` remove `FutureSlippageModel` as a parent class, since it already defines `allowed_asset_types = (Future,)`? That way this is more of a base slippage class than a mixin.

richafrank: Sounds reasonable to me. I'll make that change.

dmichalowicz:
LGTM!

PR #1784, opened 2017-05-05
Some futures prices need more precision when rounding

dmichalowicz:
Previously, _all_ asset prices were being rounded to 3 decimal places because this is the most precision we ever needed for equities. Some futures however have tick sizes with precision less than 0.001, so we need to dynamically account for that when rounding.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -58,7 +58,8 @@ from zipline.data.us_equity_pricing import NoDataOnDate
 from zipline.utils.math_utils import (
     nansum,
     nanmean,
-    nanstd
+    nanstd,
+    number_of_decimal_places,
 )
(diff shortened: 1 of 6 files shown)


zipline/data/data_portal.py
```diff
@@ -906,6 +910,32 @@ def get_history_window(self, assets, end_dt, bar_count, frequency, field,
         else:
             raise ValueError("Invalid frequency: {0}".format(frequency))
 
+        # Round prices according to asset type. The number of decimal places
+        # used for rounding is constant for all equities, but varies by future
+        # contract according to tick size.
+        decimals = {}
+        for asset in df:
```
dmichalowicz: Looping here is probably too costly, I'll have to figure out a better way to do this.

dmichalowicz:
Closing in favor of https://github.com/quantopian/zipline/pull/1788

PR #1785, opened 2017-05-08
Fixes equity history calls on the futures calendar

yankees714:
- Makes `ReindexBarReader.get_value` handle missing data, so errors won't be raised on valid futures minutes where the equities calendar is closed.
- `get_history_window` with '1d' now handles midnight on the futures calendar correctly, instead of interpreting it as a request for daily frequency data, by now  taking a `data_frequency` arg.
- Small fix for `_handle_minute_history_out_of_bounds`, which should be using `minute_to_session_label` instead of normalizing the date.
- Adds test coverage by creating subclasses of `DailyEquityHistoryTestCase` and `MinuteEquityHistoryTestCase` which use the 'us_futures' calendar.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -735,8 +735,12 @@ class DataPortal(object):
             )
         return tds[start_loc:end_loc + 1]
 
-    def _get_history_daily_window(self, assets, end_dt, bar_count,
-                                  field_to_use):
+    def _get_history_daily_window(self,
+                                  assets,
+                                  end_dt,
+                                  bar_count,
+                                  field_to_use,
+                                  data_frequency):
         """
         Internal method that returns a dataframe containing history bars
         of daily frequency for the given sids.
@@ -750,7 +754,7 @@ class DataPortal(object):
                                 columns=None)
 
         data = self._get_history_daily_window_data(
-            assets, days_for_window, end_dt, field_to_use
+            assets, days_for_window, end_dt, field_to_use, data_frequency
         )
         return pd.DataFrame(
             data,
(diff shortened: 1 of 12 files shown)


zipline/data/data_portal.py
```diff
@@ -735,8 +735,12 @@ def _get_days_for_window(self, end_date, bar_count):
...
         return tds[start_loc:end_loc + 1]
 
-    def _get_history_daily_window(self, assets, end_dt, bar_count,
-                                  field_to_use):
+    def _get_history_daily_window(self,
+                                  assets,
+                                  end_dt,
+                                  bar_count,
+                                  field_to_use,
+                                  data_frequency):
```
ehebert: :+1:

zipline/data/data_portal.py
```diff
@@ -762,10 +766,9 @@ def _get_history_daily_window_data(self,
                                        assets,
                                        days_for_window,
                                        end_dt,
-                                       field_to_use):
-        ends_at_midnight = (end_dt.hour == end_dt.minute == 0)
```
ehebert: Thank you for removing this inference of the mode based on the input.

zipline/data/resample.py
```diff
@@ -652,7 +649,14 @@ def first_trading_day(self):
         return self._reader.first_trading_day
 
     def get_value(self, sid, dt, field):
-        return self._reader.get_value(sid, dt, field)
+        # Give an empty result if no data is present.
+        try:
+            return self._reader.get_value(sid, dt, field)
+        except NoDataOnDate:
+            if field == 'volume':
+                return 0.0
```
ehebert: Could you return `0` (i.e. an integer) instead, as we do in most other cases where we return a 'no volume' value?

Issue #1675, opened 2017-02-07
Consider shortening Asset.__repr__?


ssanderson:
The repr for asset currently displays all the information that can be associated with an asset, and the verbosity makes it unpleasant to use, especially in collections where the default behavior is to repr the elements of the collection.  For example, this is what a list of 5 assets from the Q database currently looks like:

```
In [7]: assets[:5]
Out[7]:
[Equity(2, symbol=u'ARNC', asset_name=u'ARCONIC INC', exchange=u'NYSE', start_date=Timestamp('2002-01-01 00:00:00+0000', tz='UTC'), end_date=Timestamp('2017-02-02 00:00:00+0000', tz='UTC'), first_traded=None, auto_close_date=Timestamp('2017-02-07 00:00:00+0000', tz='UTC'), exchange_full=u'NEW YORK STOCK EXCHANGE'),
 Equity(6, symbol=u'AACE', asset_name=u'ACE CASH EXPRESS INC', exchange=u'NASDAQ', start_date=Timestamp('2002-01-01 00:00:00+0000', tz='UTC'), end_date=Timestamp('2006-10-05 00:00:00+0000', tz='UTC'), first_traded=None, auto_close_date=Timestamp('2006-10-10 00:00:00+0000', tz='UTC'), exchange_full=u'NASDAQ GLOBAL SELECT MARKET'),
 Equity(14, symbol=u'AALA', asset_name=u'AMERALIA INC', exchange=u'NASDAQ', start_date=Timestamp('2002-01-01 00:00:00+0000', tz='UTC'), end_date=Timestamp('2006-03-30 00:00:00+0000', tz='UTC'), first_traded=None, auto_close_date=Timestamp('2006-04-04 00:00:00+0000', tz='UTC'), exchange_full=u'NASDAQ GLOBAL MARKET'),
 Equity(21, symbol=u'AAME', asset_name=u'ATLANTIC AMERICAN CORP', exchange=u'NASDAQ', start_date=Timestamp('2002-01-01 00:00:00+0000', tz='UTC'), end_date=Timestamp('2017-02-02 00:00:00+0000', tz='UTC'), first_traded=None, auto_close_date=Timestamp('2017-02-07 00:00:00+0000', tz='UTC'), exchange_full=u'NASDAQ GLOBAL MARKET'),
[...]

PR #1786, opened 2017-05-08
MAINT: Shorten __repr__ for different assets

FreddieV4:
x-ref https://github.com/quantopian/zipline/issues/1675

This change shortens the `__repr__` method for `Asset`, and uses the same repr for different Asset classes (e.g. Futures and Equities). 

Only concern I have here is the comment from Eddie and others on the Futures repr:

image

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -30,7 +30,7 @@ from nose_parameterized import parameterized
 from numpy import full, int32, int64
 import pandas as pd
 from pandas.util.testing import assert_frame_equal
-from six import PY2, viewkeys
+from six import viewkeys
 import sqlalchemy as sa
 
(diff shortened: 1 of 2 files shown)


ssanderson:
LGTM.  We probably should wait to merge this on built check.  We've seen instances in the past of Quantopian users getting object attributes by doing regex matching on the repr/str of API objects, so we probably need to run this through a regression test suite to make sure it doesn't break anything.

ssanderson:
```
1.95s$ flake8 zipline tests
tests/test_assets.py:33:1: F401 'PY2' imported but unused
The command "flake8 zipline tests" exited with 1.
```

FreddieV4:
@ssanderson Fixed the flake8 failure

PR #1788, opened 2017-05-09
Do not explicitly round asset prices

dmichalowicz:
Previously, all asset prices were being rounded to 3 decimal places because this is the most precision we ever needed for equities. Some futures however have tick sizes with precision less than 0.001, so we don't want to cut off those price values.

Since we already cutoff the decimal places for equities to three places by default (https://github.com/quantopian/zipline/blob/master/zipline/data/minute_bars.py#L55) the explicit rounding when reading is unnecessary anywhere. And futures prices should be already cutoff to the appropriate precision according to their `ohlc_ratios_per_sid` value.

### DIFF
--- a/tests/test_data_portal.py
+++ b/tests/test_data_portal.py
@@ -19,41 +19,64 @@ from numpy.testing import assert_almost_equal
 import pandas as pd
 from pandas.tslib import Timedelta
 
-from zipline.assets import Equity
+from zipline.assets import Equity, Future
+from zipline.data.data_portal import HISTORY_FREQUENCIES, OHLCV_FIELDS
+from zipline.data.minute_bars import (
+    FUTURES_MINUTES_PER_DAY,
+    US_EQUITIES_MINUTES_PER_DAY,
+)
+from zipline.testing import parameter_space
 from zipline.testing.fixtures import (
     ZiplineTestCase,
     WithTradingSessions,
     WithDataPortal,
     alias,
 )
+from zipline.testing.predicates import assert_equal
+from zipline.utils.numpy_utils import float64_dtype
 
 
 class DataPortalTestBase(WithDataPortal,
                          WithTradingSessions,
                          ZiplineTestCase):
 
-    ASSET_FINDER_EQUITY_SIDS = (1,)
+    ASSET_FINDER_EQUITY_SIDS = (1, 2)
     START_DATE = pd.Timestamp('2016-08-01')
     END_DATE = pd.Timestamp('2016-08-08')
 
-    TRADING_CALENDAR_STRS = ('NYSE', 'CME')
+    TRADING_CALENDAR_STRS = ('NYSE', 'us_futures')
 
     EQUITY_DAILY_BAR_SOURCE_FROM_MINUTE = True
 
+    # Since the future with sid 10001 has a tick size of 0.0001, its prices
+    # should be rounded out to 4 decimal places. To test that this rounding
+    # occurs correctly, store its prices out to 5 decimal places by using a
+    # multiplier of 100,000 when writing its values.
+    OHLC_RATIOS_PER_SID = {10001: 100000}
+
+    @classmethod
+    def make_root_symbols_info(self):
+        return pd.DataFrame({
+            'root_symbol': ['BAR', 'BUZ'],
+            'root_symbol_id': [1, 2],
+            'exchange': ['CME', 'CME'],
+        })
+
     @classmethod
     def make_futures_info(cls):
-        trading_sessions = cls.trading_sessions['CME']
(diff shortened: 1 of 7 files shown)


tests/test_data_portal.py
```diff
@@ -102,13 +110,25 @@ def make_equity_minute_bar_data(cls):
...
+            index=asset1_df.index,
+        )
+        yield 2, asset2_df
 
     @classmethod
     def make_future_minute_bar_data(cls):
-        asset = cls.asset_finder.retrieve_asset(10000)
-        trading_calendar = cls.trading_calendars[asset.exchange]
-        trading_sessions = cls.trading_sessions[asset.exchange]
+        trading_calendar = cls.trading_calendars['us_futures']
```
yankees714: We should be pass the `Future` type here to make it clear that we're getting the primary calendar for futures, and in case we ever want change it.

yankees714: I think that would allow us to remove the override of `TRADING_CALENDAR_STRS` as well.

dmichalowicz: Will do

zipline/lib/adjusted_array.py
```diff
@@ -58,6 +58,9 @@
     OBJECT_DTYPES,
 )
 
+# Default number of decimal places used for rounding asset prices.
+DEFAULT_ASSET_PRICE_DECIMALS = 3
```
dmichalowicz: I'm not sure if this file is the best place for this constant to live.

dmichalowicz:
@yankees714 Ok this should be ready for another look.

zipline/data/history_loader.py
```diff
@@ -343,6 +358,20 @@ def _calendar(self):
...
         pass
 
+    def _decimal_places_for_asset(self, asset, reference_date):
+        if isinstance(asset, Future) and asset.tick_size:
+            return number_of_decimal_places(asset.tick_size)
+        elif isinstance(asset, ContinuousFuture):
+            rf = self._roll_finders[asset.roll_style]
+            contract_sid = rf.get_contract_center(
+                asset.root_symbol, reference_date, asset.offset,
+            )
```
yankees714: Can you just use the asset finder for this?
```python
contract_sid = self._asset_finder.get_ordered_contracts(
    asset.root_symbol).contract_before_auto_close(reference_date.value)
```

Seems like the offset probably doesn't matter?

dmichalowicz: Yeah the offset doesn't matter. Then this way I don't need to set a `_roll_finders` attribute either.

zipline/lib/adjusted_array.py
```diff
@@ -227,6 +227,7 @@ def traverse(self,
             offset,
             window_length,
             perspective_offset,
+            rounding_places=None,
```
yankees714: So this means we don't want to round here?

zipline/lib/_windowtemplate.pxi
```diff
@@ -40,6 +40,7 @@ cdef class AdjustedArrayWindow:
         readonly Py_ssize_t window_length
         Py_ssize_t anchor, max_anchor, next_adj
         Py_ssize_t perspective_offset
+        object rounding_places
```
yankees714: Can we add some explanation of this to the docstring?

zipline/lib/adjusted_array.py
dmichalowicz: Correct, for instance in pipeline, if there's a price adjustment that causes prices to go out to more than 3 decimal places, I don't believe any rounding currently occurs.

dmichalowicz:
@yankees714 Updated based on your latest comments

zipline/lib/_windowtemplate.pxi
```diff
@@ -138,6 +145,8 @@ cdef class AdjustedArrayWindow:
         new_out = asanyarray(self.data[anchor - self.window_length:anchor])
         if view_kwargs:
             new_out = new_out.view(**view_kwargs)
+        if self.rounding_places is not None and new_out.dtype is dtype('float64'):
```
yankees714: Seems like `np.issubdtype` might be the more robust way to do this? Also any reason not to just import `float64` directly?

dmichalowicz: Good call on the `np.issubdtype`, will change.

> Also any reason not to just import float64 directly?

I've seen throughout zipline we tend to favor using `dtype('float64')` as seen [here](https://github.com/quantopian/zipline/blob/master/zipline/utils/numpy_utils.py#L35), so I was just following that convention.

yankees714: Got it, makes sense.

PR #1798, opened 2017-05-19
With equity pipeline engine

ahgnaw:
I want to be able to test the pipeline result of a factor so I created a pipeline engine fixture with equity pricing data.


jbredeche:
LGTM

PR #1811, opened 2017-05-23
Run chunked pipeline

ahgnaw:
Running a pipeline in chunks can be more efficient when running over a long time period. This pr ultimately introduces a function to do just that, but other required functions are included.

Pipelines can sometimes have a column that is of type 'categorical', e.g. a country code. In order to combine the separate pipeline chunks, the categories of each column must be consolidated because pd.concat cannot concatenate two DataFrame columns with different categories. For this the function `categorical_df_concat` is added.

To roll back input to the last trading day (inclusive), the function `roll_dates_to_previous_session` is added. 

Tests for each new function are included.


twiecki:
This PR will be very helpful for the ML workflow.

zipline/pipeline/engine.py
```diff
@@ -62,6 +65,37 @@ def run_pipeline(self, pipeline, start_date, end_date):
         """
         raise NotImplementedError("run_pipeline")
 
+    @abstractmethod
+    def run_chunked_pipeline(self, pipeline, start_date, end_date, chunksize):
+        """
+        Compute values for `pipeline` for date range chunks the size
+        `chunksize`. Computing in chunks could reduce the amount of memory
```
ssanderson: Is there something more specific we can say here besides 'could reduce the amount of memory'?

zipline/utils/date_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    sessions : pandas.tseries.index.DatetimeIndex
+        The list of valid session dates.
+    *dates : pd.Timestamp
+        The dates for which the last trading date is needed.
+
+    Returns
+    -------
+    rolled_dates: (np.datetime64, np.datetime64)
+        The last trading date of the input dates, inclusive.
+
```
ssanderson: Is the extra whitespace here intentional?

zipline/utils/date_utils.py
```diff
@@ -0,0 +1,65 @@
+from toolz import partition_all
+
+
+def roll_dates_to_previous_session(sessions, *dates):
```
ssanderson: I don't think this function is actually used anywhere in zipline anymore?

tests/utils/test_date_utils.py
```diff
@@ -0,0 +1,89 @@
...
+class TestDateUtils(ZiplineTestCase):
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(TestDateUtils, cls).init_class_fixtures()
+        cls.calendar = get_calendar('NYSE')
+
+    @parameterized.expand([
+        (
+            Timestamp('05-19-2017', tz='UTC'),  # actual trading date
```
ahgnaw: TIL

tests/utils/test_date_utils.py
```diff
@@ -0,0 +1,86 @@
...
+        # End date not found in calendar
+        with self.assertRaises(KeyError) as cm:
+            compute_date_range_chunks(
+                self.calendar.all_sessions,
+                T('2017-05-01'),
+                T('2017-05-27'),  # Saturday
+                None
+            )
+        self.assertEqual(
+            cm.exception.__str__(),
```
ssanderson: Usually the more idiomatic way to write this would just be `str(cm.exception)`.

tests/utils/test_pandas_utils.py
```diff
@@ -80,3 +84,105 @@ def test_nearest_unequal_bad_input(self):
...
+                    'A': pd.Series(['c', 'b', 'd'], dtype='category'),
+                    'X': pd.Series([103, 102, 104], dtype='int64'),
+                }
+            ),
+        ]
+
+        with self.assertRaises(ValueError) as cm:
+            categorical_df_concat(mismatched_dtypes)
+        self.assertEqual(
+            cm.exception.message,
```
ssanderson: `Exception.message` is deprecated in python2 and doesn't exist in python 3.  You just want `str(cm.exception)`.

zipline/pipeline/engine.py
```diff
@@ -12,6 +12,7 @@
     with_metaclass,
 )
 from numpy import array
+from odo.utils import copydoc
```
ssanderson: `odo` is an optional requirement for Zipline.  We should just copy this function and put it somewhere like `zipline.utils.sharedoc`.

zipline/utils/pandas_utils.py
```diff
@@ -222,3 +223,45 @@ def clear_dataframe_indexer_caches(df):
...
+    if not all([(df.dtypes.equals(df_i.dtypes)) for df_i in df_list[1:]]):
+        raise ValueError("Input DataFrames must have the same columns/dtypes.")
+
+    categorical_columns = df.columns[df.dtypes == 'category']
+
+    for col in categorical_columns:
+        new_categories = sorted(
+            set().union(
+                *(frame[col].cat.categories for frame in df_list)
+            ) - {None}
```
ssanderson: Why are we removing `None` from the categories here?  This will change results for missing values.

ahgnaw: `None` is a deprecated category and this could produce a `FutureWarning`. It doesn't matter regardless because any array entry lacking a categorical code is assumed to be `None`.

`Setting NaNs in `categories` is deprecated and will be removed in a future version of pandas.`

ssanderson: Right.  This is a known issue that we've punted on updating because it's a breaking change for consumers of the API.  In `LabelArray.as_categorical`, for example,  we [explicitly silence this warning](https://github.com/quantopian/zipline/blob/master/zipline/lib/labelarray.py#L324).

The issue here is that when we first added support for `Classifiers` in Pipeline, the api allowed for consumers to specify how missing data should be represented.  For strings data, we default to `None` as the missing value, which means that pipeline results for string-dtype classifiers contain `None` for locations where we don't have any data.  When we upgraded to pandas 18, pandas started emitting warnings about `None` values in categories, but we ignored the warning because what pandas wants us to do instead is let it use its preferred value of `NaN`, which is a significant behavioral change for users.  I think we want to revisit that issue at some point, but I don't think right now is the time for that change (my guess is that we'll deal with this the next time we upgrade pandas).

For the time being, my recommendation would be to wrap this for-loop in the same `ignore_pandas_categorical_nan_warning` block that we use in LabelArray.

ssanderson: >  It doesn't matter regardless because an array entry missing a categorical code is assumed to be `None`.

To be clear, I don't think this is true; you'd get `NaN` back, rather than `None`, which has very different behavior.  In particular, `NaN` compares unequal with any other python object, including itself, which means you need to test for its presence explicitly.

ssanderson:
LGTM!

Issue #1805, opened 2017-05-22
Remove deprecated Yahoo code


freddiev4:
As a result of https://github.com/quantopian/zipline/issues/1776, we'll no longer be able to use the Yahoo API to load data for bundles or benchmarks.

Some of the deprecated code can be founder in [loader.py](https://github.com/quantopian/zipline/blob/be8ea7c54c940b3cb3d804db3ec3af7c0fb374f0/zipline/data/loader.py), and also references to `load_bars_from_yahoo` can be removed.

And also the entire data bundle in `zipline.data.bundles` in `yahoo.py`

souravsingh:
I am interested in working on the issue.

FreddieV4:
Hi @souravsingh thanks for offering to work on this issue. I think @mmargenot is working on this currently, because he's been assigned to this issue; if he says he's no longer working on this then feel free to work on it.

Also need to finish up some work on this [branch](https://github.com/quantopian/zipline/tree/google-finance-for-benchmarking) beforehand.

FreddieV4:
x-ref https://github.com/quantopian/zipline/pull/1867

Vaibhavdixit02:
Hey, I would like to work on this issue, can someone help me out?

souravsingh:
The issue can be closed since it is fixed.

FreddieV4:
Hi @Vaibhavdixit02 I actually removed the remaining code in this PR https://github.com/quantopian/zipline/pull/1972 so just waiting to merge that. @souravsingh I'll close this once I merge that PR 🙂

freddiev4:
#2032 should close this

ferhatelmas:
@freddiev4 ping :smile:

freddiev4:
Hi @ferhatelmas this code was removed in the PR linked above so I'll close this issue on the next release 🙂

HyperionSounds:
I'm pretty new to zipline - I've tried writing custom bundles with google to limited success. Is there something comparable that I can use to load_bars_from_yahoo?

freddiev4:
Hi @HyperionSounds the `load_bars_from_yahoo` work was removed a while ago, so the closest thing we have to that now is the ability to [ingest .csv files](http://www.zipline.io/bundles.html#ingesting-data-from-csv-files)

PR #1812, opened 2017-05-24
BUG/MAINT: Switch over to Google for benchmarking

FreddieV4:
Due to recent changes in the Yahoo API, we're switching to Google Finance for benchmarking data to `SPY`. The endpoint we're hitting is also undocumented so this will serve as just a fix until we can come up with a more robust solution. 

We're using [pandas-datareader](https://github.com/pydata/pandas-datareader) to pull the data from Google Finance.

The one potential issue here is that the history that GF provides only goes back to around June 2001, and also data from `2009-08-11` and `2012-02-02` is missing. This breaks zipline backtests that need older benchmark data. One thought is to ffill for those dates but I'm not quite sure if that's an appropriate change.
 
x-ref https://github.com/quantopian/zipline/issues/1805 and https://github.com/quantopian/zipline/issues/1776


FreddieV4:
Tests currently failing and looking like this:

image

tests/risk/test_risk_period.py
```diff
@@ -385,7 +385,7 @@ def test_benchmarkrange(self):
     def test_partial_month(self):
 
         start_session = self.trading_calendar.minute_to_session_label(
-            pd.Timestamp("1991-01-01", tz='UTC')
+            pd.Timestamp("1993-02-01", tz='UTC')
```
richafrank: Why this change?

tests/test_examples.py
```diff
@@ -54,10 +53,7 @@ def init_class_fixtures(cls):
             serialization='pickle',
         )
 
-        copy_market_data(WithTradingEnvironment.MARKET_DATA_DIR,
-                         cls.tmpdir.getpath('example_data/root'))
```
richafrank: Does this mean the tests will query google for the data instead of using the data saved in the repo?

zipline/data/loader.py
```diff
@@ -215,7 +218,13 @@ def ensure_benchmark_data(symbol, first_date, last_date, now, trading_day,
 
     # If no cached data was found or it was missing any dates then download the
     # necessary data.
-    logger.info('Downloading benchmark data for {symbol!r}.', symbol=symbol)
+    logger.info(
+        ('Downloading benchmark data for {symbol!r} '
+            'from {first_date} to {last_date}'),
```
richafrank: 👍 for the improved message

zipline/data/loader.py
```diff
@@ -260,7 +269,7 @@ def ensure_treasury_data(symbol, first_date, last_date, now, environ=None):
     path.
     """
     loader_module, filename, source = INDEX_MAPPING.get(
-        symbol, INDEX_MAPPING['^GSPC'],
+        symbol, INDEX_MAPPING['SPY'],
```
richafrank: Not a blocker, but I wonder if we should move this default benchmark symbol out into a global somewhere.

zipline/data/treasuries.py
```diff
@@ -67,9 +67,10 @@ def get_treasury_data(start_date, end_date):
         "&from="  # An unbounded query is ~2x faster than specifying dates.
         "&to="
         "&filetype=csv"
+        "&label=include"
         "&layout=seriescolumn"
         "&type=package",
-        skiprows=1,  # First row is a useless header.
+        skiprows=5,  # First 5 rows are useless headers.
```
richafrank: Why are we changing the treasuries download? Is this related to the yahoo/google change?

zipline/finance/trading.py
```diff
@@ -78,7 +78,7 @@ class TradingEnvironment(object):
     def __init__(
         self,
         load=None,
-        bm_symbol='^GSPC',
+        bm_symbol='SPY',
```
richafrank: Should we delete the old benchmark csv from the repo?

tests/test_examples.py
```diff
@@ -54,10 +53,7 @@ def init_class_fixtures(cls):
...
-
-    @parameterized.expand(examples.EXAMPLE_MODULES)
+    @parameterized.expand(sorted(examples.EXAMPLE_MODULES))
```
richafrank: 👍 for consistent test order/names

tests/resources/rebuild_example_data
```diff
@@ -14,6 +15,8 @@ from zipline.testing import test_resource_path, tmp_dir
 from zipline.utils.cache import dataframe_cache
 
 
+matplotlib.use('Agg')
```
FreddieV4: It's required to rebuild the example data only because the example algos call an `analyze` function, which plots graphs of performance results; for some reason (most likely because of our recent update to the matplotlib version) a matplotlib window will open up and the `rebuild_example_data` script waits on the user to manually close that window (thus killing that process) before it continues.

By explicitly stating the `Agg` backend, this issue doesn't occur and no matplotlib windows show up when running the script.

tests/risk/test_risk_period.py
FreddieV4: I made this change because the furthest date for which we have data from Google Finance is `1993-02-01`, so this test fails because it's looking for data at a start date before the oldest date.

tests/resources/rebuild_example_data
richafrank: 👍

zipline/data/treasuries.py
FreddieV4: This is just a problem I ran into and noticed when I wanted to be sure that the benchmarking works (by running a small algo I made locally). When I deleted the cached `treasury_curves.csv` from `~/.zipline`, I ran into the `'Time Period' not in list` error when trying to download new treasury curves data. 

I downloaded a csv of the treasury data manually to check it and it has 5 useless headers at the beginning of the file. Removing those headers fixed the download issue.

tests/risk/test_risk_period.py
richafrank: Oh, I thought it went back only to 2001?

> The one potential issue here is that the history that GF provides only goes back to around June 2001, and also data from 2009-08-11 and 2012-02-02 is missing.

zipline/finance/trading.py
FreddieV4: Yes. We should do that. We should also remove data in the `example_data` tarball for older versions of pandas.

zipline/data/treasuries.py
richafrank: Aha, gotcha. And what's the deal with the label?

tests/risk/test_risk_period.py
FreddieV4: It does not. I have to change that comment as I had misunderstood how google provides data. The one important restriction is that we only can get the last 4000 days of data at a time.

I stitched multiple csvs of data all the way back to 1993, the furthest possible date google provides data for.

FreddieV4:
@richafrank Added the `ensure_file()` call and removed the extra data we had saved

zipline/data/loader.py
```diff
@@ -50,6 +51,7 @@ def last_modified_time(path):
     """
     Get the last modified time of path as a Timestamp.
     """
+    ensure_file(path)
```
richafrank: Oh, I think we don't want to do this, since it basically subverts the check below in production code. This function no longer just returns the last modified time, it first makes it "now". I think where we were copying over the market data previously (for tests only), we want to replace that with touching the market data.

tests/test_examples.py
```diff
@@ -54,10 +55,11 @@ def init_class_fixtures(cls):
             serialization='pickle',
         )
 
-        copy_market_data(WithTradingEnvironment.MARKET_DATA_DIR,
-                         cls.tmpdir.getpath('example_data/root'))
+        market_data = ('SPY_benchmark.csv', 'treasury_curves.csv')
+        for file in market_data:
+            ensure_file(cls.tmpdir.getpath('example_data/root/data/') + file)
```
richafrank: Just some nits now: Let's not name this variable `file`, so it doesn't mask the builtin. And let's use `os.path.join` instead of `+` to ensure the path separator is correct on windows and elsewhere - in fact, if we move `+ file` into the `getpath` call, it looks like it uses `os.path.join` for us under the hood. 

Otherwise LGTM!

FreddieV4:
@richafrank fixed up stuff on your most recent comment!

zipline/finance/trading.py
yiorgosn: If i may make one final recommendation on all the benchmark data as you program this.  As a 20+ year hedge fund professional I can attest that the benchmark is good to have and good business practice to include it in your studies and presentations, so do NOT remove it like some have suggested. However to save valuable programming time and provide a top notch solution when it comes to missing data chunks (like we currently have with the time series from Google) filling in the missing periods with the backfill function would be totally acceptable, so long as you alert the user.  This can be a third choice: if a file is only partially available online.  So  maybe 
 the decision tree here regarding benchmarks could be: a. See if available online, b. see if the user can upload a csv, 3, if available online with NaN backfill and d.  All else fails grab an old benchmark file and backfill the missing data for the time range and alert the user.  Finally, if you can, don't call it the S&P benchmark for file names and inside the code, call it the equity benchmark as you may choose to change it later on.  You guys are doing a stellar job here.  Thank you!

yiorgosn:
Another benchmark error that has to do again with missing index data for the user specified date range:  Apparently, very recent benchmark data is also missing with the current google solution such as yesterday's prices.  Trying to simulate a portfolio on June 6, 2017 with the Quantopian-Quandl bundle that i ingested and verified that it includes closing prices for June 5th, 2017.  The simulation can NOT be run because the benchmark loader throws an error due to the benchmark data file not including June 5th, 2017.  The benchmark SPY or  specifically the ^GSPC csv file actually gets written or modified today, June 6 2017, but it only includes benchmark data up until June 2nd 2017 (<-3999 entry).   I checked manually on the google finance website and the data for June 5, 2017 is available.  If i do simulations not including yesterday's date for periods when i know the benchmark data is available from 2013 - 6-2-2017 all the simulation runs fine.  As I mentioned in my earlier post if the benchmark modules can just fill in missing data this should not be an issue.

Here is the error:
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\pandas\core\indexing.py", line 1296, in _has_valid_type
    error()
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\pandas\core\indexing.py", line 1291, in error
    (key, self.obj._get_axis_name(axis)))
KeyError: 'the label [2017-06-05 00:00:00+00:00] is not in the [index]'

During handling of the above exception, another exception occurred:

  File "C:\Users\George\Anaconda3\envs\py34\Scripts\zipline-script.py", line 11, in <module>
    load_entry_point('zipline==1.0.2', 'console_scripts', 'zipline')()
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\click\core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\click\core.py", line 697, in main
    rv = self.invoke(ctx)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\click\core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\click\core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\click\core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\zipline\__main__.py", line 97, in _
    return f(*args, **kwargs)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\click\decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\zipline\__main__.py", line 240, in run
    environ=os.environ,
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\zipline\utils\run_algo.py", line 180, in _run
    overwrite_sim_params=False,
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\zipline\algorithm.py", line 688, in run
    for perf in self.get_generator():
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\zipline\gens\tradesimulation.py", line 228, in transform
    handle_benchmark(normalize_date(dt))
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\zipline\gens\tradesimulation.py", line 188, in handle_benchmark
    benchmark_source.get_value(date)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\zipline\sources\benchmark_source.py", line 73, in get_value
    return self._precalculated_series.loc[dt]
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\pandas\core\indexing.py", line 1198, in __getitem__
    return self._getitem_axis(key, axis=0)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\pandas\core\indexing.py", line 1342, in _getitem_axis
    self._has_valid_type(key, axis)
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\pandas\core\indexing.py", line 1304, in _has_valid_type
    error()
  File "C:\Users\George\Anaconda3\envs\py34\lib\site-packages\pandas\core\indexing.py", line 1291, in error
    (key, self.obj._get_axis_name(axis)))
KeyError: 'the label [2017-06-05 00:00:00+00:00] is not in the [index]'

FreddieV4:
Hey @yiorgosn thanks for posting about this. I also noticed this issue earlier today and am working on a fix 😄

yiorgosn:
@FreddieV4  Thank you!  Changes made to fix the benchmark missing data and code error issues work GREAT!  
Files I updated locally:
\data\benchmarks.py
\data\loader.py
\finance\trading.py
\sources\benchmark_source.py
\utils\date_utils.py

saitam1:
I am having this error when running the example buyapple.py
```
request.py line 1320, in do_open raise URLError(err)
urllib.error.URLError: <urlopen error [Errno 11001] getaddrinfo failed>
```
Does it have to do with the Yahoo Finance API change?

FreddieV4:
@saitam1 what is the command you ran for the `buyapple.py` example algo?

PR #1813, opened 2017-05-24
BLD: Swap conda build args to check CI label/channel first

FreddieV4:
Making this change because the build failed when we started building Python 3.5 packages for the first time; we were checking the Quantopian channel `main` label first to see if there is an existing built package for packages like `sortedcontainers` or `setuptools_scm`.

Order matters when passing `-c` flags to `conda build`. 

We've also been unnecessarily building packages in our CI builds because the URL we specified in `make_conda_packages.py` needed to be updated to just be the name of the channel and label.

### DIFF
--- a/ci/make_conda_packages.py
+++ b/ci/make_conda_packages.py
@@ -31,8 +31,8 @@ def main(env, do_upload):
                "--python", env['CONDA_PY'],
                "--numpy", env['CONDA_NPY'],
                "--skip-existing",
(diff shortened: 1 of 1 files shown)


richafrank:
@FreddieV4 Assuming this fixed it, will you make the same change to our Travis config?

FreddieV4:
@richafrank Sure! x-ref https://github.com/quantopian/zipline/pull/1814

PR #1814, opened 2017-05-25
BLD: Modify conda label for Travis CI

FreddieV4:
We've been unnecessarily building packages in our CI builds because the URL we specified in `make_conda_packages.py` needed to be updated to just be the name of the channel and label.

When we pass in the specified channel via the `-c` flag, we can just say `channel_name/label/label_name`, in this case it's `quantopian/label/ci`, so we skip building existing packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ install:
     for recipe in $(ls -d conda/*/ | xargs -I {} basename {}); do
       if [[ "$recipe" = "zipline" ]]; then continue; fi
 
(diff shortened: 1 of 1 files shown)


richafrank:
This looks good to me @FreddieV4 . 

You mentioned that the order matters - do you think we need to swap the order here, or is that not the issue? As we're currently using anaconda, I think we probably won't see packages for the same dependency in both channels, so I think the order is fine. I'm just going to remove that comment and merge this.

PR #1816, opened 2017-05-25
MAINT: Remove environment as an argument to benchmark source.

ehebert:
To allow the BenchmarkSource class to be more easily used in contexts other than
a TradingAlgorithm, remove the TradingEnvironment as an argument to the
benchmark source.

Instead:
- Pass a benchmark Asset, instead of a bencmark sid; so that the asset_finder
does not need to be passed to the benchmark source.
- Pass the pre-calculated benchmark_returns instead of an env,
which contains the benchmark_returns; a consumer can let the benchmark_returns
stay as the default of `None` when using an asset.

We may want to further refactor and make two different classes, instead of
relying on a combination of existence/non-existence of benchmark_asset and
benchmark_returns. That refactoring should be easier to do with this change.

### DIFF
--- a/zipline/sources/benchmark_source.py
+++ b/zipline/sources/benchmark_source.py
@@ -49,8 +49,6 @@ class BenchmarkSource(object):
                     self.data_portal
                 )
         elif benchmark_returns is not None:
-            # get benchmark info from trading environment, which defaults to
-            # downloading data from Yahoo.
             daily_series = benchmark_returns[sessions[0]:sessions[-1]]
 
             if self.emission_rate == "minute":
--- a/zipline/sources/benchmark_source.py
+++ b/zipline/sources/benchmark_source.py
@@ -23,19 +23,21 @@ from zipline.errors import (
 
 
 class BenchmarkSource(object):
-    def __init__(self, benchmark_sid, env, trading_calendar, sessions,
-                 data_portal, emission_rate="daily"):
-        self.benchmark_sid = benchmark_sid
-        self.env = env
+    def __init__(self,
+                 benchmark_asset,
+                 trading_calendar,
+                 sessions,
+                 data_portal,
+                 emission_rate="daily",
+                 benchmark_returns=None):
+        self.benchmark_asset = benchmark_asset
         self.sessions = sessions
         self.emission_rate = emission_rate
         self.data_portal = data_portal
 
         if len(sessions) == 0:
             self._precalculated_series = pd.Series()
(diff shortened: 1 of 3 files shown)


zipline/sources/benchmark_source.py
```diff
@@ -46,11 +48,10 @@ def __init__(self, benchmark_sid, env, trading_calendar, sessions,
                     self.sessions,
                     self.data_portal
                 )
-        else:
+        elif benchmark_returns is not None:
             # get benchmark info from trading environment, which defaults to
             # downloading data from Yahoo.
-            daily_series = \
-                env.benchmark_returns[sessions[0]:sessions[-1]]
+            daily_series = benchmark_returns[sessions[0]:sessions[-1]]
```
jbredeche: the comment just above this line needs updating now

ehebert: Good catch. I'll move that to where the returns are retrieved in `algorithm.py`.

zipline/sources/benchmark_source.py
```diff
@@ -77,27 +79,27 @@ def _validate_benchmark(self, benchmark_asset):
         # error suggesting that the user pick a different asset to use
         # as benchmark.
         stock_dividends = \
-            self.data_portal.get_stock_dividends(self.benchmark_sid,
+            self.data_portal.get_stock_dividends(self.benchmark_asset,
```
srklonaris: Does this method already require a sid or an asset?

ehebert: `get_stock_dividends` accepts either `Asset` or an `int`. (It calls `int` on the `sid` value before creating the sql query.)
The docstring should probably be updated on that method to reflect that either is accepted; though in general we have aimed to pass `Asset`s to `DataPortal` methods.

srklonaris: okay, makes sense

srklonaris:
lgtm!  Thanks for doing this, its really helpful for me when needing to build a `BenchmarkSource` in qids

zipline/sources/benchmark_source.py
```diff
@@ -68,6 +67,9 @@ def __init__(self, benchmark_sid, env, trading_calendar, sessions,
                 self._precalculated_series = minute_series
             else:
                 self._precalculated_series = daily_series
+        else:
+            raise Exception("Must provide either benchmark_asset or "
```
richafrank: What do you think about `ValueError` or something else more specific?

Issue #1817, opened 2017-05-26
Error retrieving H15 interest rates - ValueError: 'Time Period' is not in list


rudolf-bauer:
Dear Zipline Maintainers,

# Environment

* Operating System: Linux 3d62a3c1924c 4.4.0-66-generic #87-Ubuntu SMP Fri Mar 3 15:29:05 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
* Python Version: Python 3.5.2 :: Anaconda 4.3.0 (64-bit)
* Python Bitness: 64
* How did you install Zipline: pip
* Python packages: 
```
alabaster==0.7.9
alembic==0.9.2
anaconda-client==1.6.0
anaconda-navigator==1.4.3
appdirs==1.4.3
APScheduler==3.3.1
arctic==1.38.0
argcomplete==1.0.0
astroid==1.4.9
astropy==1.3
Babel==2.3.4
backports.shutil-get-terminal-size==1.0.0
bayesian-optimization==0.4.0
bcolz==0.12.1
beautifulsoup4==4.5.3
bitarray==0.8.1
blaze==0.10.1
bokeh==0.12.4
boto==2.45.0
Bottleneck==1.2.1
butils==0.2.13
cachetools==2.0.0
cffi==1.9.1
chardet==2.3.0
chest==0.2.3
click==6.7
cloudpickle==0.2.2
clyent==1.2.2
colorama==0.3.7
conda==4.3.9
conda-build==2.1.3
conda-verify==2.0.0
configobj==5.0.6
contextlib2==0.5.5
coverage==4.3.4
coveralls==1.1
cryptography==1.7.1
cycler==0.10.0
cyordereddict==1.0.0
Cython==0.25.2
cytoolz==0.8.2
dask==0.13.0
data==0.3.7
datashape==0.5.4
deap==1.1.0
decorator==4.0.11
dill==0.2.5
docopt==0.6.2
docutils==0.13.1
dynd==0.7.3.dev1
empyrical==0.2.2
enum34==1.1.6
et-xmlfile==1.0.1
fastcache==1.0.2
filelock==2.0.7
Flask==0.12
Flask-Cors==3.0.2
future==0.16.0
gevent==1.2.1
greenlet==0.4.11
h5py==2.6.0
HeapDict==1.0.0
hmmlearn==0.2.0
idna==2.2
imagesize==0.7.1
inspyred==1.0.1
intervaltree==2.1.0
ipykernel==4.5.2
[...]

PR #1818, opened 2017-05-26
MAINT: Modify broken treasury data link

FreddieV4:
After moving around on the Federal Reserve site, the link that we're using for `get_treasury_data()` isn't valid (x-ref https://github.com/quantopian/zipline/issues/1817), and so I removed `label=omit` from the URL params to make a working link.

You can see that the same URL is generated when going to `https://www.federalreserve.gov/datadownload/`, clicking on "Selected Interest Rates (H.15)", clicking "Treasury Constant Maturities" and then clicking "Go to Download", lastly click "Download file".

Full URL `https://www.federalreserve.gov/datadownload/Download.aspx?rel=H15&series=bf17364827e38702b42a58cf8eaa3f78&lastobs=&from=&to=&filetype=csv&label=include&layout=seriescolumn&type=package`

### DIFF
--- a/zipline/data/treasuries.py
+++ b/zipline/data/treasuries.py
@@ -60,14 +60,13 @@ def earliest_possible_date():
 
 def get_treasury_data(start_date, end_date):
     return pd.read_csv(
-        "http://www.federalreserve.gov/datadownload/Output.aspx"
+        "https://www.federalreserve.gov/datadownload/Output.aspx"
         "?rel=H15"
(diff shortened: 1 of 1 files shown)


richafrank:
If this gives us the correct data again, looks good to me.

> You can see that the same URL is generated when going to https://www.federalreserve.gov/datadownload/

Shall we use https as well?

FreddieV4:
@richafrank it looks like it's giving us the same data as before. Also changed the URL to use https

PR #1822, opened 2017-05-30
Add portfolio weights method to algorithms

dmichalowicz:
1. Add a property to `context.portfolio` that computes the current portfolio's weights of each of its holdings. That is, for every equity/future currently held, calculate its value held divided by the portfolio's total exposure.

~~2. Add a rolling realized CVaR metric to algorithms which looks at the past 200 days of day-to-day returns and averages the bottom 5% of those values.~~ To be added in a following PR.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -123,6 +123,7 @@ from zipline.test_algorithms import (
     TestOrderPercentAlgorithm,
     TestOrderStyleForwardingAlgorithm,
     TestOrderValueAlgorithm,
+    TestPositionWeightsAlgorithm,
     TestRegisterTransformAlgorithm,
     TestTargetAlgorithm,
     TestTargetPercentAlgorithm,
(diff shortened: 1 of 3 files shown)


zipline/finance/risk/cumulative.py
```diff
@@ -281,6 +286,12 @@ def update(self, dt, algorithm_returns, benchmark_returns, leverage):
         self.max_leverage = self.calculate_max_leverage()
         self.max_leverages[dt_loc] = self.max_leverage
 
+        self.cvar[dt_loc] = conditional_value_at_risk(
+            returns=pd.DataFrame(self.algorithm_returns[-CVAR_LOOKBACK_DAYS:]),
```
richafrank: How often is this called? Any concern about the speed of making a `DataFrame` here?

When we originally moved the other metrics to empyrical, they slowed down the zipline test suite dramatically because of all the `Series` constructions, whereas zipline had been using raw numpy arrays previously. Eventually empyrical was updated to support either an `ndarray` or a `Series/DataFrame` as input to each function, without having to wrap the arrays in pandas objects. This dropped the test times back down.

dmichalowicz: That's a great point and something I realized right after putting this up; what I'm unsure of is how to handle the `weights` parameter in the case that we pass an array (or anything of ndim 1). In the case of an array, do we just ignore the weights parameter? Seems odd if it's required but does nothing. (either way you're right that I should change this not to construct a DataFrame).

dmichalowicz:
Those broken tests are failing for me even on master.

tests/test_algorithm.py
```diff
@@ -1095,10 +1096,63 @@ class TestPositions(WithLogger,
     START_DATE = pd.Timestamp('2006-01-03', tz='utc')
     END_DATE = pd.Timestamp('2006-01-06', tz='utc')
 
-    sids = ASSET_FINDER_EQUITY_SIDS = [1, 133]
+    equity_sids = ASSET_FINDER_EQUITY_SIDS = (1, 133)
```
yankees714: Is there a reason we're defining equity sids here but not futures sids?

zipline/protocol.py
```diff
@@ -169,6 +169,21 @@ def __repr__(self):
         },
     )
 
+    @property
+    def current_portfolio_weights(self):
+        def asset_multiplier(asset):
+            return asset.multiplier if isinstance(asset, Future) else 1
```
yankees714: Would it be worth pulling this out of the property, so we don't have to redefine the function on each access?

tests/test_algorithm.py
dmichalowicz: Nope, I had just left this here since it was defined before. I think I'll just delete it though as it doesn't really have a significant use other than as a convenient reference.

zipline/protocol.py
dmichalowicz: Yup good point.

dmichalowicz:
@yankees714 Updated from your comments

zipline/protocol.py
```diff
@@ -191,7 +191,7 @@ def current_portfolio_weights(self):
             )
             for asset, position in self.positions.items()
         })
-        return position_values / position_values.abs().sum()
+        return position_values / self.portfolio_value
```
yankees714: 👍

PR #1343, opened 2016-07-22
Improve dockerfile and add a turn-key bash script

JasonGiedymin:
Improve dockerfile and add a turn-key bash script to help lower the bar for new community users and developers.

PR #1823, opened 2017-06-01
DEV: Improve dockerfile and add a turn-key bash script

freddiev4:
Builds on https://github.com/quantopian/zipline/pull/1343 and https://github.com/quantopian/zipline/pull/1825 by fixing merge conflicts and making some other small changes


zipline/examples/.gitignore
```diff
@@ -0,0 +1 @@
+.ipynb_checkpoints
```
FreddieV4: @ssanderson Should this be added to the top level `.gitignore` file rather than adding another `.gitignore`?

containers/Dockerfile
```diff
@@ -42,10 +42,12 @@ ENV PROJECT_DIR=/projects \
 # install TA-Lib and other prerequisites
 #
 
-RUN mkdir ${PROJECT_DIR} \
-    && apt-get -y update \
-    && apt-get -y install libfreetype6-dev libpng-dev libopenblas-dev liblapack-dev gfortran \
-    && curl -L https://downloads.sourceforge.net/project/ta-lib/ta-lib/0.4.0/ta-lib-0.4.0-src.tar.gz | tar xvz
+RUN mkdir ${PROJECT_DIR}
```
Gaffney: nit: `mkdir -p ${PROJECT_DIR}`

Without knowledge of what these variables point to we do not know if these paths are more than one level deep.

JasonGiedymin: Up a couple lines they are set:

 ENV PROJECT_DIR=/projects

Gaffney: @JasonGiedymin It's a two character change that allows things to not break when PROJECT_DIR changes... any reason not to use `-p`?

JasonGiedymin: I can update sure thing.

JasonGiedymin: Actually, this is now @FreddieV4's pull request. :-)
👍 on @Gaffney's recommendation, better safe than sorry.
@FreddieV4, do the honors for an update?

PR #1824, opened 2017-06-02
MAINT: Delete unused files.

ssanderson:


ernestoeperez88:
Looks like this has gone stale.

PR #1829, opened 2017-06-06
TST: Update modified time of market data in test_examples

FreddieV4:
The `ensure_file` call doesn't actually update the current modified time of the file that we touch, so had to add a util function that calls `os.utime()`; this will fix broken tests as well.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -25,7 +25,7 @@ from zipline.testing import test_resource_path
 from zipline.testing.fixtures import WithTmpDir, ZiplineTestCase
 from zipline.testing.predicates import assert_equal
 from zipline.utils.cache import dataframe_cache
-from zipline.utils.paths import ensure_file
+from zipline.utils.paths import update_modified_time
 
 
 # Otherwise the next line sometimes complains about being run too late.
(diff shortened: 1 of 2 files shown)


zipline/utils/paths.py
```diff
@@ -59,6 +59,21 @@ def ensure_file(path):
     open(path, 'a+').close()  # touch the file
 
 
+def update_modified_time(path, times):
+    """
+    Updates the modified time of an existing file.
```
richafrank: Like `ensure_file`, do you think it would be worth mentioning here that this creates any necessary parent directories?

tests/test_examples.py
```diff
@@ -57,7 +57,12 @@ def init_class_fixtures(cls):
 
         market_data = ('SPY_benchmark.csv', 'treasury_curves.csv')
         for data in market_data:
-            ensure_file(cls.tmpdir.getpath('example_data/root/data/' + data))
+            update_modified_time(
+                cls.tmpdir.getpath(
+                    'example_data/root/data/' + data
+                ),
+                None
```
richafrank: You were asking yesterday about the default that was added in python 3. You could abstract that away in our code by making `None` the default value of the `times` argument to `update_modified_time`.

PR #1833, opened 2017-06-07
Add support for relabeling classifiers.

ssanderson:


zipline/lib/labelarray.py
```diff
@@ -565,6 +567,64 @@ def f_to_use(x):
...
+                     otypes=allowed_outtypes):
+
+            if x == missing_value:
+                return x
+
+            ret = f(x)
+
+            if not isinstance(ret, otypes):
+                raise TypeError(
+                    "Expected f() to return a string. Got %s." % (
```
llllllllll: should we also include the value in the error message?

zipline/pipeline/classifiers/classifier.py
```diff
@@ -382,6 +402,48 @@ def short_repr(self):
...
+
+    def _compute(self, arrays, dates, assets, mask):
+        relabeler = self.params['relabeler']
+        data = arrays[0]
+
+        if isinstance(data, LabelArray):
+            result = data.map(relabeler)
+            result[~mask] = data.missing_value
+        else:
+            raise TypeError(
```
llllllllll: should this be a `NotImplementedError`?

tests/pipeline/test_classifier.py
```diff
@@ -468,6 +468,70 @@ class C(Classifier):
...
+
+        class C(Classifier):
+            inputs = ()
+            dtype = categorical_dtype
+            missing_value = None
+            window_length = 0
+
+        c = C()
+
+        raw = np.asarray(
```
llllllllll: this doesn't test the f(missing) -> missing

tests/pipeline/test_classifier.py
```diff
@@ -468,6 +468,70 @@ class C(Classifier):
...
+        class C(Classifier):
+            inputs = ()
+            dtype = int64_dtype
+            missing_value = -1
+            window_length = 0
+
+        c = C()
+
+        with self.assertRaises(TypeError) as e:
+            c.relabel(lambda x: 0 / 0)  # Function should never be called.
```
llllllllll: nice

tests/test_labelarray.py
```diff
@@ -111,6 +111,65 @@ def test_compare_to_str(self,
...
+            dtype=object,
+        )
+        la = LabelArray(data, missing_value=None)
+
+        numpy_transformed = np.vectorize(f)(data)
+        la_transformed = la.map(f).as_string_array()
+
+        assert_equal(numpy_transformed, la_transformed)
+
+    def test_map_ignores_missing_value(self):
```
llllllllll: can we test the None case as well?

tests/test_labelarray.py
```diff
@@ -111,6 +111,65 @@ def test_compare_to_str(self,
...
+        la = LabelArray(self.strs, missing_value=None)
+
+        with self.assertRaises(TypeError):
+            la.map(f)
+
+    def test_map_can_only_return_none_if_missing_value_is_none(self):
+
+        # Should work.
+        la = LabelArray(self.strs, missing_value=None)
+        la.map(lambda x: None)
```
llllllllll: should we test that this returns `[None] * len(self.strs)` or was this tested somewhere else?

tests/test_labelarray.py
```diff
@@ -111,6 +111,65 @@ def test_compare_to_str(self,
...
+
+        def increment_char(c):
+            return chr(ord(c) + 1)
+
+        result = la.map(increment_char)
+        expected = LabelArray(['A', 'C', 'D'], missing_value='A')
+        assert_equal(result.as_string_array(), expected.as_string_array())
+
+    @parameter_space(
+        __fail_fast=True,
```
llllllllll: Why are we using fail fast on all of these? are the tests that slow?

Also, why is it `__` instead of just `_`?

zipline/lib/labelarray.py
```diff
@@ -565,6 +567,64 @@ def f_to_use(x):
...
+
+        new_categories_with_duplicates = (
+            np.vectorize(f_to_use, otypes=[object])(self.categories)
+        )
+
+        # If f() maps multiple inputs to the same output, then we can end up
+        # with the same code duplicated multiple times. Compress the categories
+        # by running them through np.unique, and then use the reverse lookup
+        # table to compress codes as well.
+        new_categories, bloated_reverse_index = np.unique(
```
llllllllll: could we use `factorize_strings` for this to get the `None` aware unique?

tests/test_labelarray.py
ssanderson: I generally use `__fail_fast` with `parameter_space` because it makes the error message easier to read and lets me use `nosetests --pdb`.  Without it, `--pdb` drops me in at the point where `subtest` is re-raising a multi-exception.

llllllllll: should that be the default then?

tests/test_labelarray.py
ssanderson: 6bb31b2544bd091bf1c9c2007db725ebefa11298

tests/test_labelarray.py
ssanderson: I don't think we have an explicit test for this case.  I'll add one.

tests/test_labelarray.py
ssanderson: Sure.

ssanderson: I'll probably do that in a different PR though to avoid noise here.

llllllllll: kk

zipline/lib/labelarray.py
ssanderson: 09cc54e08a

zipline/lib/labelarray.py
ssanderson: The nice thing about `unique` is that it gives me back the inverse index, which I'm then using to map the original codes onto the new codes.  I think I have an idea for a fix for this.

tests/pipeline/test_classifier.py
ssanderson: Added a test here: https://github.com/quantopian/zipline/pull/1833/files#diff-7a35549bd3377d8f50dd1c675db9dac7R520.

zipline/lib/labelarray.py
```diff
@@ -650,3 +729,17 @@ def element_of(self, container):
...
+
+
+@total_ordering
+class _SortableSentinel(object):
+    """Dummy object that sorts before any other python object.
+    """
+    def __eq__(self, other):
+        return isinstance(other, _SortableSentinel)
+
+    def __lt__(self, other):
```
llllllllll: can you just make this `return True`?

zipline/lib/labelarray.py
```diff
@@ -565,6 +567,83 @@ def f_to_use(x):
...
+        # If f() maps multiple inputs to the same output, then we can end up
+        # with the same code duplicated multiple times. Compress the categories
+        # by running them through np.unique, and then use the reverse lookup
+        # table to compress codes as well.
+        new_categories, bloated_inverse_index = np.unique(
+            new_categories_with_duplicates,
+            return_inverse=True
+        )
+
+        if new_categories[0] == _sortable_sentinel:
```
llllllllll: since this is a sentinel could we use `is`?

PR #1834, opened 2017-06-08
ENH: Match window_safe status of multi output factor with factor

ahgnaw:
Marking a multi-output factor as window safe does not currently result in window_safe output terms.


ssanderson:
LGTM.  Feel free to merge when travis is green.

ahgnaw:
LGTM

PR #1839, opened 2017-06-09
Expected Shortfall

dmichalowicz:
- Create a method called `calculate_expected_shortfall` on `context.portfolio` which computes the expected shortfall of the portfolio according the 2-year returns history of the assets currently being held, taking into account the weights of each holding in the portfolio.
- Add expected shortfall (aka CVaR) as a metric to algorithm daily results.


zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
...
+        self.current_date = None
+
+    @property
+    def expected_shortfall(self):
+        """
+        Function for computing expected shortfall (also known as CVaR, or
+        Conditional Value at Risk) for the portfolio according to the assets
+        currently held and their respective weight in the portfolio.
+
+        This function requires a data portal in order to retrieve price
```
ssanderson: Is it important to a caller of this method where the portfolio object reads its data from?

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
...
+        """
+        data_portal = self.data_portal
+        current_date = self.current_date
+
+        # If we do not even have a year's worth of data to look back on
+        # then the expected shortfall calculation will not be reliable, so
+        # just return NaN. If we only have between one and two years of
+        # data just use what is available.
+        num_days_of_data = data_portal.trading_calendar.session_distance(
+            self.start_date, current_date,
```
ssanderson: Where does `self.start_date` come from?

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
...
+            cutoff=CVAR_CUTOFF,
+        )
+
+    def _asset_for_history_call(self, asset):
+        calendar = self.data_portal.trading_calendar
+        asset_finder = self.data_portal.asset_finder
+        current_date = self.current_date
+
+        if isinstance(asset, Equity):
+            num_days_of_data = calendar.session_distance(
```
ssanderson: Elsewhere we use `asset.exchange` to decide what calendar to use when asking questions about what data is available for an asset, but here it seems like we're always using the data portal's calendar?

zipline/protocol.py
```diff
@@ -59,6 +61,10 @@
     'share_count',
 ]
 
+TRADING_DAYS_PER_YEAR = 252
```
ssanderson: These constants are specific to the NYSE calendar.  I don't think we should be enforcing them on all Zipline algorithms.

zipline/protocol.py
```diff
@@ -59,6 +61,10 @@
...
+CVAR_LOOKBACK_DAYS = TRADING_DAYS_PER_YEAR * 2
+CVAR_CUTOFF = 0.05
```
ssanderson: I would call this `DEFAULT_CVAR_CUTOFF` to emphasize that other values could be used.

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
         return position_values / self.portfolio_value
 
 
+class AlgorithmPortfolio(Portfolio):
+
+    def __init__(self, data_portal, benchmark=None):
+        super(AlgorithmPortfolio, self).__init__()
+        self.data_portal = data_portal
+        self.benchmark = benchmark
+        self.current_date = None
```
ssanderson: What external system is responsible for updating this value?

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
...
+        )
+        if num_days_of_data < TRADING_DAYS_PER_YEAR:
+            return np.NaN
+        elif num_days_of_data < CVAR_LOOKBACK_DAYS:
+            num_lookback_days = num_days_of_data
+        else:
+            num_lookback_days = CVAR_LOOKBACK_DAYS
+
+        # Series mapping each asset to its portfolio weight.
+        weights = self.current_portfolio_weights
```
ssanderson: Where does current_portfolio_weights come from?

zipline/finance/risk/cumulative.py
```diff
@@ -281,6 +287,9 @@ def update(self, dt, algorithm_returns, benchmark_returns, leverage):
         self.max_leverage = self.calculate_max_leverage()
         self.max_leverages[dt_loc] = self.max_leverage
 
+        if expected_shortfall is not None:
```
ssanderson: When would we expect expected_shortfall to be None?

zipline/finance/performance/tracker.py
```diff
@@ -385,11 +400,15 @@ def handle_market_close(self, dt, data_portal):
 
             benchmark_value = self.all_benchmark_returns[completed_session]
 
+            portfolio = self.get_portfolio(False)
```
ssanderson: Style note: it's almost always clearer to pass boolean function arguments by keyword.  `self.get_portfolio(False)` provides very little information to a reader, but `self.get_portfolio(performance_needs_update=False)` is much more useful.

zipline/finance/performance/tracker.py
```diff
@@ -385,11 +400,15 @@ def handle_market_close(self, dt, data_portal):
...
+            expected_shortfall = portfolio.expected_shortfall
```
ssanderson: Do we need to calculate this every day during the backtest (as opposed to just once at the end when we can fetch all the data we need in one call)?  I'm very worried about the performance impact of this change.

zipline/finance/performance/tracker.py
```diff
@@ -77,7 +78,7 @@ class PerformanceTracker(object):
     """
     Tracks the performance of the algorithm.
     """
-    def __init__(self, sim_params, trading_calendar, env):
```
ssanderson: Do we still need separate trading calendar and env parameters if this class is accepting a data portal now?

zipline/finance/performance/tracker.py
```diff
@@ -125,12 +126,20 @@ def __init__(self, sim_params, trading_calendar, env):
                     create_first_day_stats=True
                 )
 
+        try:
+            benchmark = self.asset_finder.lookup_symbol(
```
ssanderson: Doing our own ad-hoc symbol resolution here seems like an easy way for the benchmarks to get out of sync across different zipline subsystems. For example, I think the benchmark resolution is supposed to respect `set_symbol_lookup_date`.  Is there a reason we can't just pass the benchmark asset itself at construction time for this class?

zipline/finance/performance/tracker.py
```diff
@@ -350,10 +362,13 @@ def handle_minute_close(self, dt, data_portal):
         # cumulative returns
         bench_since_open = (1. + bench_returns).prod() - 1
 
+        # Do not calculate expected shortfall every minute as it requires a
```
ssanderson: If feels odd to me that all the other risk metrics are managed by the RiskMetricsCumulative class, but then `expected_shortfall` is passed in as a parameter.

zipline/finance/performance/tracker.py
ssanderson: Another reason to make this a method rather than a property is that this is expensive to call; `property` generally should only be used for attribute accesses that are computationally cheap.

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
...
+    def _asset_for_history_call(self, asset):
+        calendar = self.data_portal.trading_calendar
+        asset_finder = self.data_portal.asset_finder
+        current_date = self.current_date
+
+        if isinstance(asset, Equity):
+            num_days_of_data = calendar.session_distance(
+                asset.start_date, current_date,
+            )
+            if num_days_of_data < TRADING_DAYS_PER_YEAR and \
```
ssanderson: This feels like it should be a parameter to the class and/or this function.  252 isn't a reasonable default if we're not a NYSE algo.

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
...
+           end_dt=current_date,
+           bar_count=num_lookback_days,
+           frequency='1d',
+           field='price',
+           data_frequency='daily',
+        )
+        asset_returns = prices.pct_change()
+
+        return conditional_value_at_risk(
+            returns=asset_returns.fillna(0).dot(weights.values),
```
ssanderson: Filling NaNs with zeros seems like a pretty significant decision here.  If I have mostly missing data for  an asset, filling those values with zeros will cause us to think that the CVaR for the asset is low (since it never loses money if most of the inputs have been filled with zeros)

When do we expect to have NaNs still in the data at this point?

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
...
+        Conditional Value at Risk) for the portfolio according to the assets
+        currently held and their respective weight in the portfolio.
+
+        This function requires a data portal in order to retrieve price
+        histories of the assets in the portfolio.
+        """
+        data_portal = self.data_portal
+        current_date = self.current_date
+
+        # If we do not even have a year's worth of data to look back on
```
ssanderson: I don't think it's correct for us to be basing this decision on a relative offset from our start date.  A cardinal rule in Zipline is that simulation behavior shouldn't change depending on when my backtest starts and ends, but in this case, asking for the expected shortfall of a portfolio at time T depends on the relative offset of T from the start of the backtest.

I think the right behavior here is to fetch prices for all relevant assets and then filter those assets based on how much missing data they have.  One of the interesting questions is whether/how we want to redistribute the weights associated with the assets that get filtered out.  For example, if I have a portfolio of three assets, and one of them just IPOed, it's probably not right to just ignore 33% of the portfolio.  One option might be to redistribute the removed weight across other assets (though you'd probably want to treat longs and shorts separately).

zipline/finance/performance/tracker.py
dmichalowicz: We don't need `trading_calendar`, but I think we still need `env`?

zipline/finance/performance/tracker.py
dmichalowicz: > Is there a reason we can't just pass the benchmark asset itself at construction time for this class?

Nope I will do that

zipline/assets/continuous_futures.pyx
```diff
@@ -403,6 +403,24 @@ cdef class OrderedContracts(object):
...
+        """
+        Get the offset of the given contract relative to the contract with an
+        offset of zero on the given date.
+        """
+        current_contract_sid = self.contract_before_auto_close(dt_value)
+        if sid < current_contract_sid:
+            raise ValueError(
+                'The given sid has already expired by the given date!',
+            )
+        offset = 0
```
jbredeche: Should set a type for `offset` for performance reasons.  Same with `current_contract_sid`

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
...
+class AlgorithmPortfolio(Portfolio):
+
+    def __init__(self, data_portal, benchmark=None):
+        super(AlgorithmPortfolio, self).__init__()
+        self.data_portal = data_portal
+        self.benchmark = benchmark
+        self.current_date = None
+
+    @property
+    def expected_shortfall(self):
```
dmichalowicz: Will change. On that note, do you think I should make the `current_portfolio_weights` property a method as well?

zipline/protocol.py
```diff
@@ -194,6 +200,87 @@ def current_portfolio_weights(self):
         return position_values / self.portfolio_value
 
 
+class AlgorithmPortfolio(Portfolio):
```
dmichalowicz: No reason other than that I felt I these were pretty radical additions to add to the (previously simple) `Portfolio` class, so I thought it warranted their own sublclass. I wouldn't be opposed to keeping it all in `Portfolio` though if you think that makes more sense.

zipline/protocol.py
dmichalowicz: It gets updated by the `PerformanceTracker` when calling `get_portfolio`

zipline/protocol.py
dmichalowicz: Not really, I'll remove.

zipline/protocol.py
dmichalowicz: These come from a property of `Portfolio` that I just recently added: https://github.com/quantopian/zipline/pull/1822

zipline/protocol.py
dmichalowicz: Like we discussed this is incorrect, but I think what I want here is the start date of _all_ data. Either that, or later on when I do a history call I'll have to catch `HistoryWindowStartsBeforeData` errors.

zipline/protocol.py
dmichalowicz: This relates to your point above regarding redistributing weights. Lets discuss later today.

zipline/protocol.py
dmichalowicz: Yeah it should probably be parameter, although I think we should decide on a default to represent both equities and futures calendars (250 here? 500 max?); the point being that 1) I don't think these are meant to be strict numbers, they are just approximations to use to represent about 2 years of history lookbacks when doing the calculations. 2) If this varies by calendar, that means that equity values could differ slightly on the futures calendar, which seems to go against the idea of keeping backtest runs consistent.

zipline/protocol.py
dmichalowicz: Correct, as long as you consider a forward filled value as a price point. Otherwise in futures algos we'd have to segregate the history call for equities to use its own calendar.

zipline/finance/performance/tracker.py
ssanderson: I'm pretty sure the only thing we're using `env` for is to get the asset finder, which is also on the data portal.

zipline/finance/performance/tracker.py
```diff
@@ -189,7 +199,9 @@ def get_portfolio(self, performance_needs_update):
         if performance_needs_update:
             self.update_performance()
             self.account_needs_update = True
-        return self.cumulative_performance.as_portfolio()
+        portfolio = self.cumulative_performance.as_portfolio()
+        portfolio.current_date = self._current_session
```
ssanderson: I don't think it's guaranteed that `get_portfolio` will called every day.  Even if that's true now, I wouldn't necessarily expect that to be the case in the future.  If the portfolio needs to know the current date, then I think we want an explicit documented API that includes a contract for when the date should be updated and whose responsible for performing that update.

ssanderson: Alternatively, we might want to pass a callback to the portfolio to get the current datetime.  We do this in a bunch of other places for objects that are tied to the context of the simulation.

zipline/finance/performance/tracker.py
ssanderson: We're using the `treasury_curves` attribute, but I believe that's now effectively unused.  @ehebert can you confirm that nothing should be using the treasury curves anymore?

zipline/protocol.py
ssanderson: Noted in person: I would move the lookback length to be a parameter here with a default.

zipline/finance/performance/tracker.py
ehebert: It is no longer used for as the input of risk calculations, since the change over to using empyrical.

However, the `treasury_period_return` value is set on the `cumulative` and `period` risk metric objects.

Last I checked that value is not actually read or shown on the Quantopian platform.

However, processes which consume backtest results would have to be updated to the 'schema' change of removing that value from the risk metrics object.

We could deprecate that value by hardcoding `treasury_period_return` to 0.0. (We effectively are already doing this hardcoding on the Quantopian platform at the data level.)

dmichalowicz: Is it also ok to zero out the `excess_return` field as well? (The excess return of the algorithm over the treasuries)

tests/test_algorithm.py
```diff
@@ -1223,6 +1224,337 @@ def test_position_weights(self):
...
+            sids_and_amounts=[(sid, 1)],
+            start=pd.Timestamp('2015-01-06', tz='UTC'),
+            end=self.sim_params.end_session,
+            env=self.env,
+            benchmark_sid=8554,
+        )
+        daily_stats = algo.run(self.data_portal)
+
+        #
+        self.assertTrue((daily_stats.expected_shortfall == 0).all())
```
dmichalowicz: This needs fixing and explanation

zipline/algorithm.py
```diff
@@ -845,6 +858,109 @@ def _create_daily_stats(self, perfs):
...
+        # futures contracts from the weights dataframe as we only want to look
+        # at the continuous futures when doing a history call later on.
+        futures_held = filter(
+            lambda asset: isinstance(asset, Future),
+            weights.columns,
+        )
+        for day in weights.index:
+            for asset in futures_held:
+                if weights.loc[day, asset] != 0:
+                    cf = zp.assets_for_history_call(asset, asset_finder, day)
```
dmichalowicz: This could be costly; consider recording the continuous futres when recording the weights

zipline/algorithm.py
```diff
@@ -845,6 +858,109 @@ def _create_daily_stats(self, perfs):
...
+                    if cf not in weights:
+                        weights[cf] = 0
+                    weights.loc[day, cf] = weights.loc[day, asset]
+        weights.drop(futures_held, axis=1, inplace=True)
+        assets = weights.columns.tolist()
+
+        benchmark = self._create_benchmark_source().benchmark_asset
```
dmichalowicz: `benchmark_sid`

zipline/algorithm.py
```diff
@@ -845,6 +858,109 @@ def _create_daily_stats(self, perfs):
...
+        if benchmark is not None:
+            assets.append(benchmark)
+
+        # If we are near the start date of our data, just use the data
+        # available. Otherwise, use the full default number of lookback days.
+        days_before_start = min(
+            self.trading_calendar.session_distance(
+                data_portal._first_available_session,
```
dmichalowicz: could make a public method for this

zipline/algorithm.py
```diff
@@ -845,6 +858,109 @@ def _create_daily_stats(self, perfs):
...
+
+        # Any assets that came into existence after the start date of the
+        # simulation have their missing returns values proxied with the
+        # benchmark's returns values.
+        if benchmark is not None:
+            for column in asset_returns:
+                asset_returns[column].fillna(
+                    asset_returns[benchmark], inplace=True,
+                )
+            asset_returns.drop(benchmark, axis=1, inplace=True)
```
dmichalowicz: 1) whats the right benchmark for futures.
2) Is there a better pandas operation for filling the nans here

zipline/algorithm.py
```diff
@@ -845,6 +858,109 @@ def _create_daily_stats(self, perfs):
...
+                weights_to_use = weights.loc[date_to_use]
+            except KeyError:
+                return np.NaN
+
+            return conditional_value_at_risk(
+                returns=df.dot(weights_to_use.values),
+                cutoff=zp.DEFAULT_CVAR_CUTOFF,
+            )
+
+        # Compute rolling CVaR over the returns data frame.
```
dmichalowicz: Is there a way to do a 3D vectorized version of this?

zipline/utils/pandas_utils.py
```diff
@@ -266,3 +266,58 @@ def categorical_df_concat(df_list, inplace=False):
                 df[col].cat.set_categories(new_categories, inplace=True)
 
     return pd.concat(df_list)
+
+
```
dmichalowicz: Add an optional starting index to this function

dmichalowicz:
@ssanderson I think this is ready for another round of review. Some changes I made since our last discussion:
- Moved the expected shortfall calculation out of `TradingAlgorithm` and into the perf tracker (in `handle_simulation_end`). I also refactored that logic to share some its code with the `Portfolio`'s expected shortfall method.
- Changed the way I do continuous future lookups. Previously, at the end of the backtest, I was looping through all days and futures assets and turning them into continuous futures in order to do the `get_pricing` call. Now, I'm doing the conversion day-by-day when recording the portfolio weights. On top of that, I added caching so that the asset finder does not have to create new continuous futures every time. This improved performance quite a bit.
- Changed the way I'm filling in benchmark missing values. Previously, I was filling in NaNs column-by-column, since (as we discussed) pandas doesn't provide a direct way to fill all the columns of the dataframe at once. Now, I create an entire dataframe that copies the single benchmark column and do a `fillna` with that.
- After more profiling, I realized that the rolling apply function over the dataframe was actually not that costly, so I did not attempt to make any improvements there. The bottleneck at this point is doing the history call, which takes up the majority of time spent. I think this is fine/expected, because it will be improved once we have futures daily bars.

tests/test_algorithm.py
```diff
@@ -1223,6 +1225,416 @@ def test_position_weights(self):
...
+        day_before_auto_close = calendar.previous_session_label(
+            future_1002.auto_close_date,
+        )
+        self.assertIn(cf_offset_2, pt_weights[day_before_auto_close])
+        self.assertIn(cf_offset_1, pt_weights[future_1002.auto_close_date])
+        day_before_auto_close = calendar.previous_session_label(
+            future_1003.auto_close_date,
+        )
+        self.assertIn(cf_offset_1, pt_weights[day_before_auto_close])
+        self.assertIn(cf_offset_0, pt_weights[future_1003.auto_close_date])
```
ssanderson: I had a hard time following the structure in this section.  I'm generally :+1: on adding a little bit of duplication to tests in the interest of being explicit about what's being checked, but in this case I felt like the clarity of the explicitness didn't help because I couldn't see the pattern of what was being checked.

ssanderson: Alternative suggestion for testing the same behavior: https://github.com/quantopian/zipline/pull/1874/commits/856fdef331d84d87d3a25ef07b0ad2bda07206f9.

zipline/utils/pandas_utils.py
```diff
@@ -266,3 +266,58 @@ def categorical_df_concat(df_list, inplace=False):
...
+    """
+    if min_periods is None:
+        min_periods = window_length
+    elif min_periods > window_length:
+        raise ValueError(
+            "'min_periods' argument ({0}) can't be more than the given window "
+            "length ({1}).".format(min_periods, window_length)
+        )
+
+    slider = pd.lib.BlockSlider(df)
```
ssanderson: I prototyped this suggestion in https://github.com/quantopian/zipline/pull/1874/commits/856fdef331d84d87d3a25ef07b0ad2bda07206f9.  It looks to me like it produces the same results, but runs about 20x faster.

zipline/finance/performance/tracker.py
```diff
@@ -459,14 +486,82 @@ def handle_simulation_end(self):
...
+
+        asset_returns = zp.asset_returns_for_cvar(
+            assets=assets,
+            benchmark=self.benchmark_asset,
+            data_portal=data_portal,
+            end_date=sim_params.end_session,
+            lookback_days=len(sim_params.sessions) + days_before_start,
+        )
+
+        def cvar_of_df(df):
```
ssanderson: I prototyped an alternative to this calculation using just numpy here: https://github.com/quantopian/zipline/pull/1874/commits/e8fe43db58c09f1b25377d4c2ecf27992ebb3108.

zipline/finance/performance/tracker.py
```diff
@@ -149,6 +157,7 @@ def __init__(self, sim_params, trading_calendar, env):
             # initial cash is your capital base.
             starting_cash=self.capital_base,
             data_frequency=self.sim_params.data_frequency,
+            portfolio=portfolio,
```
ssanderson: Having looked through this code some more, I don't think we ever inspect to call `as_portfolio` on the daily performance period; in fact, it would be a serious bug to do so, because that method mutates the shared portfolio object in place.  Given that that's the case, I think what we should do here is pass a dummy portfolio object that raises an error if we try to do anything with it.

Longer term, I think we should either merge the cumulative and daily performance periods together into one object or break them out into separate classes.  As things currently stand, they're both instances of the same `PerformancePeriod` class, but they behave and are used very differently.

dmichalowicz:
@ssanderson I'm just getting back to this now, but it's going to take me a while to go through all your comments. I'm going to start with your branch-off and go from there.

zipline/algorithm.py
```diff
@@ -563,13 +563,21 @@ def _create_generator(self, sim_params):
...
+
         if self.perf_tracker is None:
             # HACK: When running with the `run` method, we set perf_tracker to
             # None so that it will be overwritten here.
+            portfolio = zp.Portfolio(
+                data_portal=self.data_portal,
+                current_dt_callback=self.get_datetime,
+                benchmark_asset=benchmark_source.benchmark_asset,
+            )
             self.perf_tracker = PerformanceTracker(
```
dmichalowicz: As far as I can tell, if we construct our own `PerformanceTracker`, the check for `if self.perf_tracker is None:` will prevent this code path from ever being hit then, meaning I think this is actually ok?

zipline/finance/performance/tracker.py
ehebert: I believe so.

zipline/protocol.py
```diff
@@ -193,6 +265,74 @@ def current_portfolio_weights(self):
...
+            self.asset_for_history_call, date=current_date,
+        )
+        assets = list(map(convert_futures, weights.index))
+        asset_returns = asset_returns_for_cvar(
+            assets, benchmark, data_portal, current_date, lookback_days,
+        )
+
+        cvar = conditional_value_at_risk(
+            returns=asset_returns.dot(weights.values),
+            cutoff=DEFAULT_CVAR_CUTOFF,
```
dmichalowicz: Nope will add

zipline/test_algorithms.py
```diff
@@ -709,7 +710,13 @@ def handle_data(self, data):
                 self.order(self.sid(s), amount)
             self.ordered = True
 
-        self.record(position_weights=self.portfolio.current_portfolio_weights)
+        self.record(
+            position_weights=self.portfolio.current_portfolio_weights()
```
dmichalowicz: This hadn't occurred to me, I'll try to change.

zipline/utils/run_algo.py
```diff
@@ -130,8 +130,10 @@ def _run(handle_data,
                 str(bundle_data.asset_finder.engine.url),
             )
         env = TradingEnvironment(asset_db_path=connstr, environ=environ)
-        first_trading_day =\
-            bundle_data.equity_minute_bar_reader.first_trading_day
+        first_trading_day = max(
+            bundle_data.equity_daily_bar_reader.first_trading_day,
```
dmichalowicz: Probably a good idea, yes.

tests/test_algorithm.py
```diff
@@ -1223,6 +1225,416 @@ def test_position_weights(self):
...
+        amounts = (1, 1)
+        calendar = self.trading_calendar
+        af = self.asset_finder
+        equity_1, future_1004 = af.retrieve_all(sids)
+
+        algo = TestPositionWeightsAlgorithm(
+            sids_and_amounts=zip(sids, amounts),
+            sim_params=self.sim_params,
+            env=self.env,
+            benchmark_sid=8554,
```
dmichalowicz: > Is that intentional?

Yes, running in daily makes things easier to reason about and all the expected shortfall calculations are done on daily data so I don't think running in minute mode is necessary, but would be happy to talk more in person.

tests/test_algorithm.py
```diff
@@ -1223,6 +1225,416 @@ def test_position_weights(self):
             assert_equal(daily_stats.iloc[i]['position_weights'], expected)
 
 
+class TestPortfolio(WithDataPortal, WithSimParams, ZiplineTestCase):
+    START_DATE = pd.Timestamp('2014-01-06', tz='UTC')
+    END_DATE = pd.Timestamp('2016-11-09', tz='UTC')
+    SIM_PARAMS_START = pd.Timestamp('2014-01-10', tz='UTC')
```
dmichalowicz: > Do we really need to run a two year backtest to exercise all the functionality here?

Yes, in order to make sure we cover the full 504 day window length that will be standard for expected shortfall.

zipline/testing/core.py
```diff
@@ -491,35 +491,33 @@ def write_daily_data(tempdir, sim_params, sids, trading_calendar):
 
 def create_data_portal(asset_finder, tempdir, sim_params, sids,
                        trading_calendar, adjustment_reader=None):
-    if sim_params.data_frequency == "daily":
-        daily_path = write_daily_data(tempdir, sim_params, sids,
-                                      trading_calendar)
-
-        equity_daily_reader = BcolzDailyBarReader(daily_path)
+    daily_path = write_daily_data(tempdir, sim_params, sids, trading_calendar)
```
dmichalowicz: We always need daily data to do the `expected_shortfall` calculation. However, I added in a flag to turn off that calculation in most cases, so this change is actually no longer necessary for the current state of things. That being said, if anyone wanted to run an algo in a test minute mode that also wants to calculate `expected_shortfall`, this change would be necessary. I don't think it hurts to just always have daily data, but what do you think?

tests/test_algorithm.py
```diff
@@ -1202,25 +1213,577 @@ def test_position_weights(self):
...
+        # Create a price series according to how we want returns to look.
+        # Include five returns values of -0.05, five of -0.02, and the rest are
+        # zero or greater. This allows for testing expected shortfall using an
+        # easily known average of low returns.
+        returns = [0, 0, 0, 0, 0, 0, -0.02, 1.0 / 49.0, -0.05, 1.0 / 19.0]
+        returns *= 5
+        while len(returns) < len(sessions) - 1:
+            returns.append(0)
+        prices = prices_with_returns(initial_price=50, returns=returns)
+        yield 3, frame(prices)
```
dmichalowicz: Added this for testing multiple different low returns values.

tests/test_algorithm.py
```diff
@@ -1202,25 +1213,577 @@ def test_position_weights(self):
...
+        expected[248::2] = second_expected_shortfall_value
+        expected[249::2] = first_expected_shortfall_value
+
+        actual = pd.DataFrame(
+            algo.risk_report['daily'], index=sim_params.sessions,
+        )['expected_shortfall']
+
+        assert_equal(actual, expected)
+
+    def test_expected_shortfall_averaging(self):
```
dmichalowicz: This test is new

tests/test_algorithm.py
```diff
@@ -1202,25 +1213,577 @@ def test_position_weights(self):
...
+        expected = pd.DataFrame(
+            algo.risk_report['daily'],
+            index=calendar.sessions_in_range(start_date, end_date),
+        )['expected_shortfall']
+
+        assert_equal(daily_stats.recorded_expected_shortfall, expected)
+
+    @parameterized.expand([
+        (2, equity_2_start_date, equity_2_start_price),
+        (2000, future_fv_start_date, future_fv_start_price),
```
dmichalowicz: Added a futures case to this test.

zipline/algorithm.py
```diff
@@ -403,6 +403,13 @@ def noop(*args, **kwargs):
         if 'data_frequency' in kwargs:
             self.data_frequency = kwargs.pop('data_frequency')
 
+        # Flag for whether or not to compute rolling expected shortfall of the
+        # algorithm. By default this is False because it can be a costly
+        # computation, so only do it if explicitly specified.
+        self.calculate_expected_shortfall = kwargs.pop(
+            'calculate_expected_shortfall', False,
+        )
```
dmichalowicz: I added this in order to avoid doing the `expected_shortfall` calculation unless we specifically pass `True`. Of course on the Quantopian platform this will be `True`.

zipline/assets/assets.py
```diff
@@ -287,6 +287,10 @@ def __init__(self, engine, future_chain_predicates=CHAIN_PREDICATES):
         # retrieve_asset will populate the cache on first retrieval.
         self._caches = (self._asset_cache, self._asset_type_cache) = {}, {}
 
+        # A cache specifically for looking up continuous futures by their
+        # specifications rather than by sid.
+        self._continuous_future_cache = {}
```
dmichalowicz: Moved this cache from the `Portfolio` object.

zipline/assets/assets.py
```diff
@@ -1026,6 +1030,28 @@ def get_ordered_contracts(self, root_symbol):
             self._ordered_contracts[root_symbol] = oc
             return oc
 
+    def offset_of_contract(self, asset, date):
```
dmichalowicz: Moved this method from the cython.

zipline/data/bundles/core.py
```diff
@@ -472,7 +472,7 @@ def most_recent_data(bundle_name, timestamp, environ=None):
         environ : dict, optional
             An environment dict to forward to zipline_root.
         """
-        if bundle_name not in bundles:
+        if bundle_name not in bundles and bundle_name != 'test':
```
dmichalowicz: This change is because there is no longer a 'test' bundle, so I used the 'quantopian-quandl' bundle in order to run the script that updates the `test_examples` data (`python tests/resources/rebuild_example_data`). This changed because of zeroing out the treasury data.

zipline/finance/performance/tracker.py
```diff
@@ -459,14 +479,57 @@ def handle_simulation_end(self):
...
+        return pd.Series(
+            rolling_expected_shortfall(
+                asset_returns=asset_returns.values,
+                weights=weights.values,
+                cutoff=zp.DEFAULT_EXPECTED_SHORTFALL_CUTOFF,
+                window_length=zp.DEFAULT_EXPECTED_SHORTFALL_LOOKBACK_DAYS,
+                min_periods=zp.DEFAULT_EXPECTED_SHORTFALL_MINIMUM_DAYS,
+            ),
+            index=sim_params.sessions,
+        )
```
dmichalowicz: This is now using your numpy rolling function.

zipline/testing/core.py
```diff
@@ -1198,6 +1196,30 @@ def create_empty_splits_mergers_frame():
...
+                              first_value=0,
+                              second_value=1,
+                              dtype=None):
+    if dtype is None:
+        dtype = type(first_value)
+    alternating = np.empty(length, dtype=dtype)
+    alternating[::2] = first_value
+    alternating[1::2] = second_value
+    return alternating
+
```
dmichalowicz: These are new --^

zipline/utils/dummy.py
```diff
@@ -8,3 +12,14 @@ def __init__(self, value):
 
     def __getitem__(self, key):
         return self._value
+
+
+class DummyPortfolio(object):
```
dmichalowicz: Added this.

zipline/utils/pandas_utils.py
```diff
@@ -311,3 +311,58 @@ def days_at_time(days, t, tz, day_offset=0):
         seconds=t.second,
     )
     return (days + delta).tz_localize(tz).tz_convert('UTC')
+
+
+def sliding_apply(df, window_length, f, min_periods=None):
```
dmichalowicz: I left this function here even though it isn't being used anymore because it could still be a useful function, even if  only for testing purposes and such.

dmichalowicz:
@ssanderson Ok, I think this is ready for another look. I went through and tried to answer your comments and also note where I made changes. If you want to discuss anything in person just let me know.

zipline/protocol.py
```diff
@@ -153,6 +216,12 @@ def __init__(self):
         self.start_date = None
         self.positions_value = 0.0
 
+        self.benchmark_asset = benchmark_asset
+
+        self._data_portal = data_portal
+        self._current_dt_callback = current_dt_callback
+        self._expiring_cache = ExpiringCache()
```
dmichalowicz: I still need to solve the problem of hiding these private attributes (or at least the data portal)

dmichalowicz: Here's one possible solution: https://github.com/quantopian/qexec/pull/13780

zipline/finance/performance/tracker.py
```diff
@@ -355,7 +367,7 @@ def handle_minute_close(self, dt, data_portal):
         minute_packet = self.to_dict(emission_type='minute')
         return minute_packet
 
-    def handle_market_close(self, dt, data_portal):
+    def handle_market_close(self, dt, data_portal, calculate_position_weights):
```
dmichalowicz: I'm wondering if there's a better way to flag whether or not to do this computation other than just passing it as an argument. Or, if it should just always be calculated.

dmichalowicz:
@ehebert I merged the treasury curves change and rebased this PR off of master, so it should be ready for you to take a look, and let me know if you have any questions in the process.

zipline/gens/tradesimulation.py
```diff
@@ -296,7 +298,9 @@ def _get_daily_message(self, dt, algo, perf_tracker):
         Get a perf message for the given datetime.
         """
         perf_message = perf_tracker.handle_market_close(
-            dt, self.data_portal,
+            dt=dt,
+            data_portal=self.data_portal,
+            calculate_position_weights=algo.calculate_expected_shortfall,
```
dmichalowicz: This feels like a poor design to me but I'm not sure a better way to go about it.

zipline/finance/performance/tracker.py
```diff
@@ -388,6 +400,20 @@ def handle_market_close(self, dt, data_portal):
...
+
+            # For each day of the simulation, check to see if any futures were
+            # held. If so, convert them to continuous futures.
+            convert_futures = partial(
+                portfolio.asset_for_history_call, date=portfolio.current_date,
+            )
+            position_weights.index = list(
+                map(convert_futures, position_weights.index),
+            )
+            self.position_weights.append(dict(position_weights))
```
ehebert: Had you investigated other structures for storing the position weights?
e.g. a dict of numpy arrays or a sparse `DataFrame`, which would add a new column for each asset as a position is held?

We convert this data to a `DataFrame` anyway in `_calculate_rolling_expected_shortfall`.

I haven't sketched it out, but I think the overhead of creating and storing a dict with the keys for each position for every backtest day may be more costly than storing the data in a sparse `DataFrame`.

The caveat with storing as a `DataFrame` is that as each position is held a new column will be added throughout the lifetime of the backtest, with many possible rows where there is no value; however the sparse representation should mitigate that growth.

conda/empyrical/meta.yaml
```diff
@@ -1,11 +1,12 @@
 package:
   name: empyrical
-  version: "0.2.2"
+  version: '0.3.2'
```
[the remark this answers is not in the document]
dmichalowicz: It could be done ahead of time/any time, there just wasn't a need for it besides this PR so I just lumped it in here.

zipline/assets/assets.py
dmichalowicz: The latter

zipline/assets/assets.py
```diff
@@ -1026,6 +1030,28 @@ def get_ordered_contracts(self, root_symbol):
...
+        """
+        Get the offset of the given contract relative to the contract with an
+        offset of zero on the given date.
+        """
+        oc = self.get_ordered_contracts(asset.root_symbol)
+        current_contract_sid = oc.contract_before_auto_close(date.value)
```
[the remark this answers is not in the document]
dmichalowicz: The `.__sub__` is an interesting thought! I can definitely see that being useful in making this code more concise, my only worry is that it might be just a bit _too_ clever. That is, having not seen it before it's slightly more difficult to reason about than the verbose, explicit code. Regardless, in heeding your final comment, I think this would be getting a tad too out of scope for this PR.

zipline/data/data_portal.py
```diff
@@ -420,6 +420,10 @@ def get_last_traded_dt(self, asset, dt, data_frequency):
         return self._get_pricing_reader(data_frequency).get_last_traded_dt(
             asset, dt)
 
+    @property
+    def first_day_of_data(self):
```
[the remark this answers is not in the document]
dmichalowicz: good point, will change

zipline/finance/performance/tracker.py
dmichalowicz: I think @ssanderson meant to make this comment for `self.todays_performance` below, where I've changed the portfolio to be `portfolio=DummyPortfolio()`. For `self.cumulative_performance` here I think we still need the actual portfolio.

zipline/finance/performance/tracker.py
dmichalowicz: > Had you investigated other structures for storing the position weights?

No I hadn't really considered other structures, mainly due to your point about a DataFrame being mostly empty for a lot of values. But I had not thought of using a sparse `DataFrame`, so I can investigate to see if that would help.

dmichalowicz: @ehebert Sketched out a possibility here: https://github.com/quantopian/zipline/pull/1952

dmichalowicz: Addressed this in #1952, but TLDR is I think what we have is a better option overall.

dmichalowicz:
@ehebert I think this is nearly ready to merge, wondering if you might have time to discuss the build failure as I never managed to quite figure it out.

ehebert:
Here are the errors that I see in the AppVeyor build:

```
  File "c:\python27_64\envs\testenv\lib\site-packages\testfixtures\rmtree.py", line 8, in _waitfor
    func(pathname)
  File "c:\python27_64\envs\testenv\lib\site-packages\testfixtures\rmtree.py", line 44, in _rmtree_inner
    _waitfor(_rmtree_inner, fullname, waitall=True)
  File "c:\python27_64\envs\testenv\lib\site-packages\testfixtures\rmtree.py", line 8, in _waitfor
    func(pathname)
  File "c:\python27_64\envs\testenv\lib\site-packages\testfixtures\rmtree.py", line 47, in _rmtree_inner
    os.unlink(fullname)
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'c:\\users\\appveyor\\appdata\\local\\temp\\1\\tmp5of1ll\\tmpwwuett\\example_data\\root\\data\\test\\2016-12-13T19;59;34.153449\\adjustments.sqlite'
```

The only thing that I see related to `test_examples` in this PR is https://github.com/quantopian/zipline/pull/1839#discussion_r131475825, but I don't think that would affect the ability to read the generated fixture data from any particular test.

dmichalowicz:
Long live CVAR.

PR #1842, opened 2017-06-14
MAINT: Update asset warning

freddiev4:
Saw this warning when running a zipline backtest earlier:

x-ref #1850 

```
[2017-06-14 18:01:12.479077] WARNING: ZiplineLog: Cannot place order for AAPL, as it has de-listed. Any existing positions for this asset will be liquidated on 2017-06-03 00:00:00+00:00.
```

And thought something broke or something happened to our `quantopian-quandl` bundle. Turns out I just needed to run `zipline ingest` again to have data that's more up to date. Changed the log warning message to reflect that.

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -1379,10 +1379,15 @@ class TradingAlgorithm(object):
                 # If we are after the asset's end date or auto close date, warn
                 # the user that they can't place an order for this asset, and
                 # return None.
-                log.warn("Cannot place order for {0}, as it has de-listed. "
-                         "Any existing positions for this asset will be "
-                         "liquidated on "
-                         "{1}.".format(asset.symbol, asset.auto_close_date))
+                log.warn(
+                    ("Cannot place order for {asset}, either because "
+                     "your data bundle is not up to date, "
+                     "or the asset has de-listed. "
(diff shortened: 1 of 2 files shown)


FreddieV4:
@richafrank this should be good for review now 🙂

zipline/algorithm.py
```diff
@@ -1379,10 +1379,15 @@ def _can_order_asset(self, asset):
                 # If we are after the asset's end date or auto close date, warn
                 # the user that they can't place an order for this asset, and
                 # return None.
-                log.warn("Cannot place order for {0}, as it has de-listed. "
-                         "Any existing positions for this asset will be "
-                         "liquidated on "
-                         "{1}.".format(asset.symbol, asset.auto_close_date))
+                log.warn(
+                    ("Cannot place order for {asset}, either because "
+                     "your data bundle is not up to date, "
```
ssanderson: @FreddieV4 we surface this warning on Quantopian to users who aren't going to have any idea what a bundle is (nor can they do anything to update the bundle).  If a user tries to run a backtest whose end date is after the end date of their bundle, IMO the right thing to is to error at that point and tell them to run over a different period or update the bundle.

PR #1846, opened 2017-06-15
PERF: Avoid extra copy if we only need one chunk.

ssanderson:


zipline/pipeline/engine.py
```diff
@@ -82,9 +82,8 @@ def run_chunked_pipeline(self, pipeline, start_date, end_date, chunksize):
             The start date to run the pipeline for.
         end_date : pd.Timestamp
             The end date to run the pipeline for.
-        chunksize : int or None
-            The number of days to execute at a time. If None, then
-            results will be calculated for entire date range at once.
+        chunksize : int
+            The number of days to execute at a time.
```
ssanderson: If a user is going to explicitly pass None, they should probably just use `run_pipeline`.

zipline/pipeline/engine.py
```diff
@@ -330,6 +329,11 @@ def run_chunked_pipeline(self, pipeline, start_date, end_date, chunksize):
         )
         chunks = [self.run_pipeline(pipeline, s, e) for s, e in ranges]
 
+        if len(chunks) == 1:
```
ssanderson: We'd care about this if we add a default (large-ish) chunksize on the Q research environment.

ahgnaw: 👍

ahgnaw:
LGTM!

zipline/pipeline/engine.py
```diff
@@ -330,6 +329,11 @@ def run_chunked_pipeline(self, pipeline, start_date, end_date, chunksize):
...
+            # OPTIMIZATION: Don't make an extra copy in `categorical_df_concat`
+            # if we don't have to.
+            return chunks[0]
+
         return categorical_df_concat(chunks, inplace=True)
```
llllllllll: why not move this optimization to categotical_df_concat`?

ssanderson: The `inplace` parameter implies that a copy is going to be made, and it seems confusing to me to have that not always be true if `chunks` is of length 1.

PR #1849, opened 2017-06-19
BUG per unit commissions should not be negative when shorting or selling

prsutherland:
Any order that 1) fills over multiple transactions/bars and 2) Reduces the number of shares owned (shorts, exits / shrinks a long position) will have a zero or negative commission applied. The first bar applies a positive commission, so unfortunately it isn't as straight forward as just taking the absolute value the order's commission.

For example if you were selling 30 shares of a stock with one share fill per bar with a PerShare(cost=0.0007, min_trade_cost=None), the cumulative commission on the order would look like:

```
Bar | Total Commission for Order
-----------------
 1  |  0.0007
 2  |  0.0
 3  | -0.0007
 4  | -0.0014
 5  | -0.0021
 6  | -0.0028
 7  | -0.0035
...
30  | -0.0203
```

This PR fixes the bug by making the commission values consistently positive numbers.


zipline/finance/performance/position.py
```diff
@@ -172,13 +172,13 @@ def adjust_commission_cost_basis(self, asset, cost):
         if self.amount == 0:
             return
 
-        prev_cost = self.cost_basis * self.amount
+        prev_cost = self.cost_basis * abs(self.amount)
```
ehebert: @ssanderson and I had an in person chat were we discussed whether this was the correct change or not.

That conversation hinges on what we determine on what we intend `cost_basis` to mean. If it is meant to be a value to represent a price at which the the position would break even, the current implementation was correct. 

However, if cost basis is meant as the cost basis for tax reporting purposes, then I believe your change is correct. Since, during a short sale, the stocks are briefly held and then sold at current market price. I can definitely be wrong here, and we should verify, but in that case for tax reasons the cost basis would be from that sale of the shares.

(I checked `Position.update`, and there I do believe we are correctly handling the `cost_basis` for the second definition, when covering a short, by holding on to the `cost_basis` of the initial short sale until the short is fully covered.)

If that clarification blocks, we could separate this commit out from the per units fix.

prsutherland: After talking with @Jstauth and @abhijeetkalyan, treating cost basis as the price at which a position would break even is the correct call. Which means, the code is correct prior to this PR.

I'm going to take another pass to document that behavior and add test coverage for it.

tests/finance/test_commissions.py
```diff
@@ -147,14 +147,38 @@ def test_per_trade(self):
...
-            sid=1, order_amount=500, fill_amounts=[230, 170, 100],
+            sid=1, order_amount=500, fill_amounts=fill_amounts
         )
+        expected_commissions = [1.725, 1.275, 0.75]
 
         # make sure each commission is pro-rated
-        self.assertAlmostEqual(1.725, model.calculate(order, txns[0]))
-        self.assertAlmostEqual(1.275, model.calculate(order, txns[1]))
-        self.assertAlmostEqual(0.75, model.calculate(order, txns[2]))
+        for fill_amount, expected_commission, txn in \
```
ehebert: Small style recommendation.

The following would avoid a line break:

```
        for fill_amount, expected_commission, txn in zip(
            fill_amounts,
            expected_commissions,
            txns
        ):
```

ehebert:
LGTM, could you squash/fixup the revert before merging?

PR #1854, opened 2017-06-20
MAINT: Removed use of empyrical's information_ratio

richafrank:
which was removed from empyrical in https://github.com/quantopian/empyrical/pull/47


boos1993:
Thanks @richafrank!

PR #1861, opened 2017-06-26
TST: Test RSI edge behavior

ahgnaw:
Technical factors are tested in `test_technical.py`. This pr moves the RSI tests into that file.

Test that RSI is equal if the returns series is the same, though prices may differ.

Test that RSI is 100 when close prices increases for 14 days.

Test that RSI is 0 when close prices decrease for 14 days.

### DIFF
--- a/tests/pipeline/test_technical.py
+++ b/tests/pipeline/test_technical.py
@@ -561,35 +561,57 @@ class RSITestCase(ZiplineTestCase):
 
         check_allclose(expected, out)
 
-    def test_rsi_edge_cases(self):
+    def test_rsi_all_positive_returns(self):
         """
-        Test against example RSI calculation found online. Test cases where
-        close prices are always increasing and where close prices are always
-        decreasing. Scaling price series by a factor should not affect
-        the result.
+        RSI indicator should be 100 in the case of 14 days of positive returns.
         """
 
         rsi = RSI()
 
         today = np.datetime64(1, 'ns')
-        assets = np.arange(4)
-        out = np.empty((4,), dtype=float)
+        assets = np.arange(1)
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_technical.py
```diff
@@ -532,6 +533,65 @@ def test_moving_average_convergence_divergence(self,
...
+        test_data = np.abs(np.random.randn(15, 3))
+
+        out = np.empty((3,), dtype=float)
+        rsi.compute(today, assets, out, test_data)
+
+        check_allclose(expected, out)
+
+    def test_rsi_edge_cases(self):
+        """
+        Test against example RSI calculation found online. Test cases where
```
ehebert: Might be useful to have a href here to that example.

tests/pipeline/test_technical.py
```diff
@@ -532,6 +533,65 @@ def test_moving_average_convergence_divergence(self,
...
+
+        example_case = np.array([46.125, 47.125, 46.4375, 46.9375, 44.9375,
+                                 44.25, 44.625, 45.75, 47.8125, 47.5625, 47.,
+                                 44.5625, 46.3125, 47.6875, 46.6875])
+        starting_val = 46.0
+        increasing = np.linspace(starting_val, starting_val + 1.4, num=15)
+        decreasing = np.linspace(starting_val, starting_val - 1.4, num=15)
+        double = example_case * 2
+
+        closes = np.vstack((increasing, decreasing, example_case, double)).T
```
ehebert: What do you think of breaking this into 3 separate tests? One for `always_increasing`, one for `always_decreasing`, and one for `same_returns`?

Each of those cases may independently fail, and it would be useful to isolate which one fails if they do.

PR #1862, opened 2017-06-27
TST: add asset_names to mock equities

jbredeche:


zipline/assets/synthetic.py
```diff
@@ -60,7 +60,8 @@ def make_rotating_equity_info(num_assets,
 def make_simple_equity_info(sids,
                             start_date,
                             end_date,
-                            symbols=None):
+                            symbols=None,
+                            names=None):
```
richafrank: Would you add this to the docstring?

zipline/testing/fixtures.py
```diff
@@ -346,6 +346,7 @@ class WithAssetFinder(WithDefaultDateBounds):
     """
     ASSET_FINDER_EQUITY_SIDS = ord('A'), ord('B'), ord('C')
     ASSET_FINDER_EQUITY_SYMBOLS = None
+    ASSET_FINDER_EQUITY_NAMES = None
```
richafrank: We should add this to the docstring as well.

zipline/assets/synthetic.py
```diff
@@ -82,11 +83,16 @@ def make_simple_equity_info(sids,
...
+
+    if names is None:
+        names = map(lambda x: str(x) + " INC.", symbols)
+
     return pd.DataFrame(
         {
             'symbol': list(symbols),
             'start_date': pd.to_datetime([start_date] * num_assets),
             'end_date': pd.to_datetime([end_date] * num_assets),
+            'asset_name': list(names),
```
richafrank: This branch has failing python3 tests in the case where `symbols` is passed as a generator (actually the result of a call to `map`, which in python2 returns a `list`). The map is already consumed by the call to `list(symbols)`, so `names` is an empty list, which is a different length from the length of the list of symbols.

zipline/assets/synthetic.py
```diff
@@ -82,11 +83,16 @@ def make_simple_equity_info(sids,
     num_assets = len(sids)
     if symbols is None:
         symbols = list(ascii_uppercase[:num_assets])
+
+    if names is None:
+        names = map(lambda x: str(x) + " INC.", symbols)
```
richafrank: My instinct is to use a list comprehension here, but definitely just a style choice.

zipline/testing/fixtures.py
jbredeche: done

zipline/assets/synthetic.py
jbredeche: done

zipline/assets/synthetic.py
jbredeche: I'll switch it.  My Ruby instincts kicked in there.

zipline/assets/synthetic.py
jbredeche: thanks.  fixed

zipline/assets/synthetic.py
```diff
@@ -82,11 +87,18 @@ def make_simple_equity_info(sids,
...
         symbols = list(ascii_uppercase[:num_assets])
+    else:
+        symbols = list(symbols)
+
+    if names is None:
+        names = [str(s) + " INC." for s in symbols]
+
     return pd.DataFrame(
         {
             'symbol': list(symbols),
```
richafrank: nitpick - You don't need this extra call to `list` now. Not a big deal.

PR #1863, opened 2017-06-28
Change auto close to happen at session end

dmichalowicz:
Now that the futures calendar does not run for 24 hours, it makes more sense for futures to auto-close at the end of the day (5pm) rather than midnight UTC. This PR moves the auto-close execution to the end of day _on_ the auto-close date, rather than the night before it. This means that you _can_ trade futures on their auto-close dates now.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -484,13 +484,13 @@ def log_nyse_close(context, data):
         for minute in algo.nyse_opens:
             # each minute should be a nyse session open
             session_label = nyse.minute_to_session_label(minute)
-            session_open = nyse.open_and_close_for_session(session_label)[0]
+            session_open = nyse.session_open(session_label)
             self.assertEqual(session_open, minute)
 
         for minute in algo.nyse_closes:
             # each minute should be a minute before a nyse session close
             session_label = nyse.minute_to_session_label(minute)
-            session_close = nyse.open_and_close_for_session(session_label)[1]
+            session_close = nyse.session_close(session_label)
             self.assertEqual(session_close - timedelta(minutes=1), minute)
 
         # Test that passing an invalid calendar parameter raises an error.
@@ -4212,9 +4212,7 @@ class TestEquityAutoClose(WithTradingEnvironment, WithTmpDir, ZiplineTestCase):
         else:
             final_prices = {
                 asset.sid: trade_data_by_sid[asset.sid].loc[
-                    self.trading_calendar.open_and_close_for_session(
-                        asset.end_date
-                    )[1]
+                    self.trading_calendar.session_close(asset.end_date)
                 ].close
                 for asset in assets
             }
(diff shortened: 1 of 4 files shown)


zipline/gens/tradesimulation.py
```diff
@@ -272,14 +272,20 @@ def past_auto_close_date(asset):
...
+        # date. These orders get processed immediately because otherwise they
+        # would not be processed until the first bar of the next day.
         blotter = algo.blotter
-        assets_to_cancel = \
-            set([asset for asset in blotter.open_orders
-                 if past_auto_close_date(asset)])
+        assets_to_cancel = [
+            asset for asset in blotter.open_orders
+            if past_auto_close_date(asset)
+        ]
```
yankees714: Any reason this changed from a list to a set? (If not, could also use a set comprehension here)

zipline/gens/tradesimulation.py
```diff
@@ -272,14 +272,20 @@ def past_auto_close_date(asset):
...
         for asset in assets_to_cancel:
             blotter.cancel_all_orders_for_asset(asset)
+        for order in copy(blotter.new_orders):
```
yankees714: Why do we need the `copy()` here?

tests/test_algorithm.py
```diff
@@ -4423,7 +4421,9 @@ def transactions_for_date(date):
             {
                 'amount': -order_size,
                 'commission': 0.0,
-                'dt': assets[0].auto_close_date,
+                'dt': self.trading_calendar.open_and_close_for_session(
+                    assets[0].auto_close_date,
+                )[1],
```
yankees714: There's now a `session_close()` method on the trading calendar

tests/test_algorithm.py
```diff
@@ -4359,13 +4359,11 @@ def test_daily_delisted_equities(self,
             )
 
         # Check expected cash.
-        self.assertEqual(algo.cash, expected_cash)
```
yankees714: Why remove this?

tests/test_algorithm.py
```diff
@@ -4479,6 +4481,10 @@ def handle_data(context, data):
                 context.order(context.sid(0), 10)
                 assert len(context.get_open_orders()) == 1
             elif today_session == first_asset_auto_close_date:
+                # We do not cancel open orders until the end of the auto close
+                # date, so our open order should still exist at this point.
+                assert len(context.get_open_orders()) == 1
+            elif today_session == first_asset_auto_close_date + trading_day:
```
yankees714: This might be more clear as `self.trading_calendar.next_session(first_asset_auto_close_date)`.

zipline/gens/tradesimulation.py
dmichalowicz: No reason pertinent to this PR. But `blotter.open_orders` is just a dictionary keyed by assets, which means they're already unique, so I didn't see a need for this to be a set. And because all we do with it later is just iterate over it.

zipline/gens/tradesimulation.py
dmichalowicz: Inside this loop I remove elements from the `blotter.new_orders` list, so I had to make a copy such that I wasn't modifying the same list I was iterating over.

dmichalowicz: I'll add a comment.

tests/test_algorithm.py
dmichalowicz: Oh ok, perfect.

tests/test_algorithm.py
dmichalowicz: Didn't know about that method, will change.

tests/test_algorithm.py
dmichalowicz: The way things worked previously is that we were doing the auto-closing before market open, so when the algo recorded its `cash` on the auto-close date, the cash had already been adjusted. Since the auto closing now happens at market close, when the algo records its `cash` on the auto close date, the cash adjustment hasn't happened yet. That means the `algo.cash` is essentially off-by-one day from the `'ending_cash'`. If you think it is worth it/not too confusing, I could keep this assertion here but shift the values by a day?

zipline/gens/tradesimulation.py
yankees714: Makes sense, thanks!

tests/test_algorithm.py
yankees714: Ah that makes sense, I'll leave it up to you as to whether you think the second check is worth it.

PR #1865, opened 2017-06-29
MAINT: Improve error when missing bcolz file for a sid is missing

yankees714:
Replaces the IOError raised by bcolz, which isn't very informative, with
NoDataOnDate.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -38,7 +38,7 @@ from zipline.data._minute_bar_internal import (
 
 from zipline.gens.sim_engine import NANOS_IN_MINUTE
 
-from zipline.data.bar_reader import BarReader, NoDataOnDate
+from zipline.data.bar_reader import BarReader, NoDataForSid, NoDataOnDate
 from zipline.data.us_equity_pricing import check_uint32_safe
 from zipline.utils.calendars import get_calendar
 from zipline.utils.cli import maybe_show_progress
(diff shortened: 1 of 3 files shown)


zipline/data/minute_bars.py
```diff
@@ -1057,9 +1057,13 @@ def _open_minute_file(self, field, sid):
...
-            carray = self._carrays[field][sid] = \
-                bcolz.carray(rootdir=self._get_carray_path(sid, field),
-                             mode='r')
+            try:
+                carray = self._carrays[field][sid] = bcolz.carray(
+                    rootdir=self._get_carray_path(sid, field),
+                    mode='r',
+                )
+            except IOError:
+                raise NoDataOnDate('No minute data for sid {}.'.format(sid))
```
ehebert: Though technically there is no data for that date, this error really occurs when there is no data at all for the sid.
Should we have a different exception indicating that there is no data for any date, instead of `NoDataOnDate` which is just for the date requested?

ehebert:
I agree it is a good to catch the IOError and raise a Zipline specific error. This lends itself well to writing another storage backend, where the underlying storage library may not raise an IOError when missing data.
Just had one question about the name of the error that we are raising.

Also, @ssanderson fixed the issue with the error message returned from bcolz with https://github.com/Blosc/bcolz/pull/237; however we are on a pretty old version of bcolz. (0.12.1 vs. 1.1.2).

yankees714:
@ehebert FWIW the error that was being raised here is from the `carray` constructor, and looked like

> IOError: [Errno 2] No such file or directory: '[...]/meta/sizes'

So it's not really the lack of the path in the error message that I'm addressing here, it's that the bcolz error isn't very clear as to what went wrong in the zipline context.

PR #1868, opened 2017-06-30
DOC: Add whatsnew for zipline 1.1.1

FreddieV4:


docs/source/whatsnew/1.1.1.txt
```diff
@@ -1,23 +1,49 @@
...
-   This release is still under active development.  All changes listed are
-   subject to change at any time.
+:Date: June 30, 2017
 
 
 Highlights
 ~~~~~~~~~~
 
-None
+Since the the release of Zipline ``1.1.0``, Quantopian has officially released support for Futures in Zipline, as well as Python 3.5 support.
```
richafrank: As a reader, I find "since 1.1.0" ambiguous: does it mean these are features that first appeared in 1.1.0, or features that first appeared after 1.1.0?

FreddieV4: I think we could just say "After the release of Zipline 1.1.0"

richafrank: What do you think about "Zipline now has broad support for futures, in addition to equities. It's also being tested and packaged for Python 3.5.", ignoring official-ness or the version?

FreddieV4: Sounds good to me 😄

FreddieV4:
@richafrank also generated type stubs

richafrank:
Thanks for that!

Issue #1838, opened 2017-06-08
docker container fails to start the jupyter notebook without --allow-root option


jay-johnson:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

This issue is about running the docker container built from the latest Dockerfile on master: https://github.com/quantopian/zipline/blob/master/Dockerfile

* Operating System: (Windows Version or `$ uname --all`) - the non-developer docker container
* Python Version: `$ python --version` - the non-developer docker container 
* Python Bitness: `$ python -c 'import math, sys;print(int(math.log(sys.maxsize + 1, 2) + 1))'` - the non-developer docker container
* How did you install Zipline: (`pip`, `conda`, or `other (please explain)`) - the non-developer docker container
* Python packages: `$ pip freeze` or `$ conda list` - the non-developer docker container

Now that you know a little about me, let me tell you about the issue I am
having:

# Description of Issue

* What did you expect to happen?
* What happened instead?

Here is how you can reproduce this issue on your machine:

I built the non-developer docker container and it fails to start the notebook without the `--allow-root` option:

```
root@zipline:/projects# jupyter notebook -y --no-browser --notebook-dir=${PROJECT_DIR} --certfile=${SSL_CERT_PEM} --keyfile=${SSL_CERT_KEY} --ip='*' --config=${CONFIG_PATH}
[C 19:42:34.067 NotebookApp] Running as root is not recommended. Use --allow-root to bypass.
```

Confirm it's not running somewhere in the background:
```
root@zipline:/projects# ps auwwx | grep notebook
root       205  0.0  0.0  12820  1028 ?        S+   19:42   0:00 grep notebook
```

Add the `--allow-root` option and retry...it works:
```
root@zipline:/projects# jupyter notebook -y --no-browser --notebook-dir=${PROJECT_DIR} --certfile=${SSL_CERT_PEM} --keyfile=${SSL_CERT_KEY} --ip='*' --config=${CONFIG_PATH} --allow-root
[I 19:52:01.869 NotebookApp] Serving notebooks from local directory: /projects
[I 19:52:01.869 NotebookApp] 0 active kernels 
[I 19:52:01.869 NotebookApp] The Jupyter Notebook is running at: https://[all ip addresses on your system]:8888/
[I 19:52:01.869 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
```

can this be added to the `docker_cmd.sh`?

```
[...]

PR #1871, opened 2017-07-02
Add '--allow-root' to jupyter command to avoid error starting container.

njgraham:
Alleviates the symptoms noted in #1838.


ssanderson:
@njgraham thanks for the PR!  I think my long term preference would probably be to add a non-root user to the container so that we don't need to worry about running as root.

Given that we were already running root, however, this seems like a good incremental improvement.

PR #1874, opened 2017-07-03
Var and cvar feedback

ssanderson:
References from PR review point here.

### DIFF
--- a/zipline/finance/performance/tracker.py
+++ b/zipline/finance/performance/tracker.py
@@ -178,10 +178,7 @@ class PerformanceTracker(object):
 
         # The weights for each day should be a dictionary mapping assets to
         # their respective position weights in the current portfolio.
-        self.position_weights = pd.Series(
-            np.full(len(self.sim_params.sessions), {}, dtype=object),
-            index=self.sim_params.sessions,
-        )
+        self.position_weights = []
 
     def __repr__(self):
         return "%s(%r)" % (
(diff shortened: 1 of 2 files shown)


zipline/finance/performance/tracker.py
```diff
@@ -523,12 +517,31 @@ def _calculate_rolling_expected_shortfall(self, data_portal):
...
-            assets=assets,
+            assets=list(weights.columns),
             benchmark=self.benchmark_asset,
             data_portal=data_portal,
             end_date=sim_params.end_session,
             lookback_days=len(sim_params.sessions) + days_before_start,
         )
+        asset_returns_values = asset_returns.values
+
+        def rolling_shortfall():
```
ssanderson: If we decide to use this, I'd probably break this out into a function that takes `asset_returns`, `weights`, and `cvar_cutoff` as parameters.

zipline/finance/performance/tracker.py
```diff
@@ -523,12 +517,31 @@ def _calculate_rolling_expected_shortfall(self, data_portal):
...
+            cvar_cutoff = zp.DEFAULT_CVAR_CUTOFF
+            out = np.full(len(weights), np.nan)
+
+            # Compute from back to front, since that simplifies the task of
+            # aligning the correct row of `weights with the correct slice of
+            # `asset_returns`.
+            last_end = -len(weights)
+            end = -1
+            while end >= last_end:
+                # TODO: Bail if we don't have enough input data.
```
dmichalowicz: I don't think we want to bail, we want to just compress the window and use what we have, down to a certain point.

dmichalowicz:
@ssanderson I'm finally getting back to this. I think the stuff you have here is good, so to make it easier to work with I'm just going to merge this into my PR now if that's ok.

ssanderson:
@dmichalowicz sounds good to me.

PR #1877, opened 2017-07-07
BUG: Fix reference to floating type.

ehebert:
Use float64_t instead, since `isnan` is only used when the array's
dtype kind is `'f'`.

Fixes this compilation error:
```
zipline/pipeline/loaders/blaze/_core.pyx:30:16: 'floating' is not a type identifier
```

### DIFF
--- a/zipline/pipeline/loaders/blaze/_core.pyx
+++ b/zipline/pipeline/loaders/blaze/_core.pyx
@@ -27,7 +27,7 @@ from zipline.pipeline.common import (
 )
 
 
-cdef bint isnan(floating value):
+cdef bint isnan(np.float64_t value):
     # this isn't defined in libc on windows...
     return value != value
 


llllllllll:
I was going for `cython.floating` which is a fused type of `float32` and `float64`

ehebert:
Ah, I can this over to that. Will update commit message.

ehebert:
```
(qexec) [ehebert@ehebert-ubuntu-desktop zipline_repo (use-float64-for-isnan)]
$ git diff
diff --git a/zipline/pipeline/loaders/blaze/_core.pyx b/zipline/pipeline/loaders/blaze/_core.pyx
index b004398c..fb275c45 100644
--- a/zipline/pipeline/loaders/blaze/_core.pyx
+++ b/zipline/pipeline/loaders/blaze/_core.pyx
@@ -27,7 +27,7 @@ from zipline.pipeline.common import (
 )


-cdef bint isnan(np.float64_t value):
+cdef bint isnan(cython.floating value):
     # this isn't defined in libc on windows...
     return value != value
```

Results in the following during compilation:
```
zipline/pipeline/loaders/blaze/_core.pyx:310:16: Invalid use of fused types, type cannot be specialized

Error compiling Cython file:
------------------------------------------------------------
...
            missing_value.view('int64'),
            is_missing_value[np.int64_t],
            array_kind,
        )
    elif kind == 'f':
        if isnan(missing_value):
               ^
------------------------------------------------------------
```

llllllllll:
fine, just make this a np.float64

PR #1884, opened 2017-07-11
Perform account validation at the end of the day

dmichalowicz:


dmichalowicz:
@richafrank Updated. I removed the test I added because this behavior is already being tested here: https://github.com/quantopian/zipline/blob/master/tests/test_algorithm.py#L3511

richafrank:
> I removed the test I added because this behavior is already being tested

Nice - does that mean the existing test didn't care about minutely versus daily?

dmichalowicz:
> does that mean the existing test didn't care about minutely versus daily?

Correct, it's simply checking that the appropriate exception is raised when running an algo under the leverage constraint. Do you think this is sufficient?

richafrank:
I suppose we could test to see how far the backtest got. I don't have a strong opinion in this case.

dmichalowicz:
@richafrank I added a time check to the test.

dmichalowicz:
@richafrank Good to go?

richafrank:
:shipit:

PR #1886, opened 2017-07-14
BUG: fix downsampled rowwise reductions

llllllllll:
When an asset stops existing partway through the downsample period we can get different results for rank (and demean and friends) on the same day if we start the pipeline inside the same downsample period but after the asset stopped existing.

### DIFF
--- a/tests/pipeline/test_downsampling.py
+++ b/tests/pipeline/test_downsampling.py
@@ -1,6 +1,9 @@
 """
 Tests for Downsampled Filters/Factors/Classifiers
 """
+from functools import partial
+
+import numpy as np
 import pandas as pd
 from pandas.util.testing import assert_frame_equal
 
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_downsampling.py
```diff
@@ -684,3 +689,80 @@ def test_errors_on_bad_downsample_frequency(self):
...
+class TestDownsampledRowwiseOperation(WithAssetFinder, ZiplineTestCase):
+    T = partial(pd.Timestamp, tz='utc')
+    START_DATE = T('2014-01-01')
+    END_DATE = T('2014-02-01')
+    HALF_WAY_POINT = T('2014-01-15')
+
+    dates = pd.date_range(START_DATE, END_DATE)
+    factor = TestingDataSet.float_col.latest
+
+    @classmethod
```
ssanderson: Is there a particular reason that we want this test to use `populate_initial_workspace` to produce `factor`?  Another way we've produced this same factor in tests elsewhere has been to define a factor like:

```python
class Sid(CustomFactor):
    inputs = []
    window_length = 1
    def compute(self, today, assets, out):
        out[:] = assets
```

llllllllll: This was just the first way I thought to do it. I wanted to execute as little pipeline machinery as possible so the most natural way in my mind was to prepopulate the workspace with values

ssanderson: If I put back in the old implementation, we don't actually hit the assertion error in the test; we crash in pipeline because of a shape mismatch, which I think is happening because the start and end dates specified in the test aren't trading days. I put up a change to the test that actually completes the pipeline but computes an incorrect result with the old implementation here: https://github.com/quantopian/zipline/pull/1887

llllllllll: yeah, I had to manually change it to get the test to fail correctly on master. I made some comments on your pr.

PR #1887, opened 2017-07-14
MAINT: Update test to not crash with the old impl.

ssanderson:


tests/pipeline/test_downsampling.py
```diff
@@ -691,40 +691,23 @@ def test_errors_on_bad_downsample_frequency(self):
         self.assertEqual(str(e.exception), expected)
 
 
-class TestDownsampledRowwiseOperation(WithAssetFinder, ZiplineTestCase):
+class TestDownsampledRowwiseOperation(WithSeededRandomPipelineEngine,
+                                      ZiplineTestCase):
     T = partial(pd.Timestamp, tz='utc')
-    START_DATE = T('2014-01-01')
-    END_DATE = T('2014-02-01')
+    START_DATE = T('2014-01-02')
```
llllllllll: we aren't using a trading calendar, this doesn't matter

tests/pipeline/test_downsampling.py
```diff
@@ -750,14 +733,14 @@ def test_downsampled_rank(self):
         downsampled_rank = self.factor.rank().downsample('month_start')
         pipeline = Pipeline({'rank': downsampled_rank})
 
-        results_month_start = self.pipeline_engine.run_pipeline(
+        results_month_start = self.seeded_random_engine.run_pipeline(
```
llllllllll: why use a seeded random loader, I want the rank value to be their sid, it makes it easier to explain. Also, given a different seed this could still pass

ssanderson: The seeded random loader is never actually used in this example (that's probably an argument against using that fixture, since that's not clear)

tests/pipeline/test_downsampling.py
ssanderson: It's being used by the seeded random loader fixture, but I'd be fine with ditching that in favor of a manual engine setup.

llllllllll: do you want to do that or should I?

tests/pipeline/test_downsampling.py
```diff
@@ -692,38 +692,30 @@ def test_errors_on_bad_downsample_frequency(self):
...
 
-            return workspace
+    factor = SidFactor()
 
+    @classmethod
+    def init_class_fixtures(cls):
+        super(TestDownsampledRowwiseOperation, cls).init_class_fixtures()
         cls.pipeline_engine = SimplePipelineEngine(
-            get_loader=lambda column: ExplodingObject(),
+            get_loader=lambda c: ExplodingObject(),
```
llllllllll: ok

PR #1890, opened 2017-07-18
MAINT: Updates for better docs generation.

ssanderson:
- Add compat function for `functools.wraps`.
- Add `is_sentinel` check.
- Add docstrings to some pipeline methods.

### DIFF
--- a/zipline/utils/sentinel.py
+++ b/zipline/utils/sentinel.py
@@ -7,6 +7,12 @@ import sys
 from textwrap import dedent
 
 
+class _Sentinel(object):
+    """Base class for Sentinel objects.
+    """
+    __slots__ = ('__weakref__',)
+
+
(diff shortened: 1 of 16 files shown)


zipline/utils/sentinel.py
```diff
@@ -27,9 +33,8 @@ def sentinel(name, doc=None):
         ) % (name, value.__doc__, doc))
 
     @object.__new__   # bind a single instance to the name 'Sentinel'
-    class Sentinel(object):
+    class Sentinel(_Sentinel):
```
FreddieV4: Should we add a comment here explaining why we did this?

ssanderson: I moved `is_sentinel` up to the top of this file to make it clearer that the base class exists so that we can use `isinstance` for `is_sentinel`.

PR #1892, opened 2017-07-18
Check contract exists when using futures daily bar reader

dmichalowicz:
The `VolumeRollFinder` failed get the active contract when using a daily bar reader.

### DIFF
--- a/tests/test_continuous_futures.py
+++ b/tests/test_continuous_futures.py
@@ -33,13 +33,15 @@ from zipline.assets.continuous_futures import (
     OrderedContracts,
     delivery_predicate
 )
+from zipline.assets.roll_finder import VolumeRollFinder
 from zipline.data.minute_bars import FUTURES_MINUTES_PER_DAY
 from zipline.errors import SymbolNotFound
 from zipline.testing.fixtures import (
     WithAssetFinder,
+    WithBcolzFutureDailyBarReader,
+    WithBcolzFutureMinuteBarReader,
     WithCreateBarData,
     WithDataPortal,
-    WithBcolzFutureMinuteBarReader,
     WithSimParams,
     ZiplineTestCase,
 )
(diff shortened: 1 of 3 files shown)


zipline/assets/roll_finder.py
```diff
@@ -90,6 +90,7 @@ def get_rolls(self, root_symbol, start, end, offset):
                                         tc.minute_to_session_label(end))
         freq = sessions.freq
         if first == front:
+            # Compute at least one roll date no matter what.
```
yankees714: Thanks for this comment, mind adding a little more detail on how this follows from the code? I still have to think about it for a while even though I've had a lot of exposure to this

zipline/testing/fixtures.py
```diff
@@ -1058,6 +1058,8 @@ class level fixtures.
     # options are: 'warn', 'raise', 'ignore'
     BCOLZ_FUTURE_DAILY_BAR_INVALID_DATA_BEHAVIOR = 'warn'
 
+    _write_method_name = 'write'
```
yankees714: Ah thanks for fixing, this was my fault because I was only using this fixture along with `WithBcolzEquityDailyBarReader`. Although now we won't be able to set this field separately between the two fixtures, maybe we should make this `BCOLZ_FUTURE_DAILY_BAR_WRITE_METHOD_NAME`?

tests/test_continuous_futures.py
```diff
@@ -1284,6 +1286,148 @@ def test_history_close_minute_adjusted_volume_roll(self):
...
+                2017-02-10        0       1000__     5000
+                2017-02-13        0       1000  `--> 5000
+                2017-02-14        0       1000       5000
+        ACD --> 2017-02-15        0       1000       5000
+                2017-02-16        0       1000       5000
+                2017-02-17        0       1000       5000
+                2017-02-20        0          0       5000
+                    ...
+                2017-03-16        0          0       5000
+                2017-03-17        0          0       5000
```
yankees714: This is great 🌈

tests/test_continuous_futures.py
```diff
@@ -1284,6 +1286,148 @@ def test_history_close_minute_adjusted_volume_roll(self):
...
+            [
+                (1000, pd.Timestamp('2017-01-18', tz='UTC')),
+                (1001, pd.Timestamp('2017-02-13', tz='UTC')),
+                (1002, None),
+            ],
+        )
+
+        # If we call 'get_rolls' with start and end dates that do not have any
+        # rolls between them, we should still expect the last roll date to be
+        # computed successfully.
```
yankees714: Thoughts on making this its own test?

dmichalowicz: Yeah probably worth making it its own

zipline/assets/roll_finder.py
dmichalowicz: I'll try my best 😬

PR #1894, opened 2017-07-28
Still hitting NoDataBeforeDate edge cases

dmichalowicz:
@yankees714 From what we discussed yesterday. The failing contract was 'EL', which is on a quarterly cycle.

### DIFF
--- a/zipline/assets/continuous_futures.pyx
+++ b/zipline/assets/continuous_futures.pyx
@@ -51,6 +51,7 @@ march_cycle_delivery_predicate = partial(delivery_predicate,
                                          set(['H', 'M', 'U', 'Z']))
 
 CHAIN_PREDICATES = {
+    'EL': march_cycle_delivery_predicate,
     'ME': march_cycle_delivery_predicate,
     'PL': partial(delivery_predicate, set(['F', 'J', 'N', 'V'])),
     'PA': march_cycle_delivery_predicate,
--- a/zipline/assets/roll_finder.py
+++ b/zipline/assets/roll_finder.py
@@ -179,19 +179,19 @@ class VolumeRollFinder(RollFinder):
         front_contract = oc.sid_to_contract[front].contract
         back_contract = oc.sid_to_contract[back].contract
 
+        tc = self.trading_calendar
+        trading_day = tc.day
+        prev = dt - trading_day
+        get_value = self.session_reader.get_value
+
         # If the front contract has reached its auto close date, the back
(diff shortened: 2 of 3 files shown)


zipline/assets/continuous_futures.pyx
```diff
@@ -51,6 +51,7 @@ march_cycle_delivery_predicate = partial(delivery_predicate,
                                          set(['H', 'M', 'U', 'Z']))
 
 CHAIN_PREDICATES = {
+    'EL': march_cycle_delivery_predicate,
```
dmichalowicz: 'EL' is on March quarterly cycle: http://www.cmegroup.com/trading/interest-rates/stir/eurodollar_contract_specifications.html

zipline/assets/roll_finder.py
```diff
@@ -179,20 +179,20 @@ def _active_contract(self, oc, front, back, dt):
         front_contract = oc.sid_to_contract[front].contract
         back_contract = oc.sid_to_contract[back].contract
 
+        tc = self.trading_calendar
+        trading_day = tc.day
+        prev = dt - trading_day
+        get_value = self.session_reader.get_value
```
yankees714: The changes in this file are just cosmetic, correct?

dmichalowicz: These four lines that moved are, but below I had to change `dt` to `prev` because `prev` is the actual day that we are calling `get_value` on.

zipline/assets/roll_finder.py
```diff
@@ -179,20 +179,23 @@ def _active_contract(self, oc, front, back, dt):
...
         get_value = self.session_reader.get_value
 
+        # If the front contract is past its auto close date it cannot be the
+        # active contract, so return the back contract. Similarly, if the back
+        # contract has not even started yet, just return the front contract.
+        # The reason for comparing the back contract's start date to 'prev'
+        # instead of to 'dt' is because we need to get each contract's volume
+        # on the previous day, so we need to make sure the back contract exists
+        # on 'prev' in order to call 'get_value' below.
+        if dt > front_contract.auto_close_date:
```
dmichalowicz: @yankees714 Just FYI, I made this change as you pointed out to be `>` instead of `>=` (and Jamie confirmed), and updated the tests accordingly (basically just moving rolls up by a day where necessary).

yankees714: Sounds good, thanks for the comment!

PR #1895, opened 2017-07-28
ENH: performance optimizations around adjustments in blaze core loader

llllllllll:


zipline/lib/labelarray.py
```diff
@@ -383,6 +382,30 @@ def __setitem__(self, indexer, value):
                 ),
             )
 
+    def set_scalar_value(self, indexer, value):
```
ssanderson: I'd probably just call this `set_scalar`

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -329,7 +329,7 @@ cdef _array_for_column_impl(object dtype,
                 # upper bound doesn't include the timestamp because we've
                 # already included the timestamp-date in the baseline.
                 end = max(ts_ix - 1, 0)
-                if end >= asof_ix:
+                if end >= asof_ix and out_array[asof_ix, column_ix] != value:
```
ssanderson: Can we add a comment here explaining why we don't need to emit an adjustment if the new value equals the already-written value?

ssanderson: Also, this is technically unsafe for the special case of "baseline = N, event arrives that amends baseline to N +1, then event arrives that amends the baseline **back to N**, right?  We're just deciding that the performance benefit here is worth losing that second amendment?

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -364,7 +371,14 @@ cdef _array_for_column_impl(object dtype,
             # of v1. However, if we look back from t6, we should see v0 for the
             # period from t0 to t1.
             end = max(non_null_ad_ixs[ix] - 1, 0)
-            if end >= asof_ix:
+            if end >= asof_ix and out_array[asof_ix, column_ix] != value:
+                # The second condition of this ``and`` is technically
```
ssanderson: I'm not sure it helps to have this duplicated.  I'd probably just put a "see note above" on this.

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -329,7 +329,14 @@ cdef _array_for_column_impl(object dtype,
                 # upper bound doesn't include the timestamp because we've
                 # already included the timestamp-date in the baseline.
                 end = max(ts_ix - 1, 0)
-                if end >= asof_ix:
+                if end >= asof_ix and out_array[asof_ix, column_ix] != value:
+                    # The second condition of this ``and`` is technically
```
ssanderson: This doesn't explain to a reader why we're skipping these events in the first place.

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -329,7 +329,24 @@ cdef _array_for_column_impl(object dtype,
                 # upper bound doesn't include the timestamp because we've
                 # already included the timestamp-date in the baseline.
                 end = max(ts_ix - 1, 0)
-                if end >= asof_ix:
+                if end >= asof_ix and out_array[asof_ix, column_ix] != value:
+                    # The first condition ensures that the adjustment spans at
```
ssanderson: :+1:

ssanderson:
LGTM

PR #1898, opened 2017-07-31
Need to handle case of auto closing after end date

dmichalowicz:
It looks like we are already handling the case of auto close dates being after end dates, both in [data.can_trade](https://github.com/quantopian/zipline/blob/master/zipline/_protocol.pyx#L500-L502) and [when ordering](https://github.com/quantopian/zipline/blob/master/zipline/algorithm.py#L1374). So this (the roll finder) is the only other spot I know of that needs this change.

### DIFF
--- a/zipline/assets/roll_finder.py
+++ b/zipline/assets/roll_finder.py
@@ -187,13 +187,17 @@ class VolumeRollFinder(RollFinder):
         # If the front contract is past its auto close date it cannot be the
         # active contract, so return the back contract. Similarly, if the back
         # contract has not even started yet, just return the front contract.
-        # The reason for comparing the back contract's start date to 'prev'
-        # instead of to 'dt' is because we need to get each contract's volume
-        # on the previous day, so we need to make sure the back contract exists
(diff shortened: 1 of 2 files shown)


zipline/assets/roll_finder.py
```diff
@@ -191,10 +191,15 @@ def _active_contract(self, oc, front, back, dt):
...
-        if dt > front_contract.auto_close_date:
+        if dt > min(front_contract.auto_close_date, front_contract.end_date):
             return back
         elif prev < back_contract.start_date:
             return front
+        elif prev > back_contract.end_date:
+            # This is an extreme edge case, but it is possible for the back
+            # contract to end before the front contract. At this point we know
+            # the front contract still exists, so just return that.
+            return front
```
yankees714: Could we collapse the last two conditions into one with `back_contract.asset.is_alive_for_session(prev)`, like in `data.can_trade`?

dmichalowicz: Yes, not sure how I didn't think of that after I literally just looked at the `can_trade` code 😆

PR #1902, opened 2017-08-02
Set RollingLinearRegressionOfReturns factor to be window_safe.

dstaple:
As far as I can tell, RollingLinearRegressionOfReturns is window safe, but is not marked as such, preventing it from being used in CustomFactor classes.  I have not tested building zipline with this one-line change, but I successfully subclassed RollingLinearRegressionOfReturns on Quantopian and verified that it functions without error.


ssanderson:
@dstaple thanks for the PR!

PR #1905, opened 2017-08-03
Allow column names to be different from the underlying expression

llllllllll:
The primary use case for this is aliasing the `asof_date`, `sid`, or `timestamp` columns without copying the data at read time.

### DIFF
--- a/zipline/pipeline/loaders/blaze/_core.pyx
+++ b/zipline/pipeline/loaders/blaze/_core.pyx
@@ -524,6 +524,13 @@ cdef array_for_column(object dtype,
         raise TypeError('unknown column dtype: %r' % input_array.dtype)
 
 
+cpdef str getname(object column):
+    try:
+        return column.metadata['blaze_column_name']
+    except KeyError:
+        return column.name
+
+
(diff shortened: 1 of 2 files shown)


zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -524,6 +524,13 @@ cdef array_for_column(object dtype,
         raise TypeError('unknown column dtype: %r' % input_array.dtype)
 
 
+cpdef str getname(object column):
+    try:
+        return column.metadata['blaze_column_name']
```
ehebert: We should have a test that writes the `blaze_column_name` so that this code path is exercised. 

Also will `metatadata` always exist? If not, we may want to add catching an `AttributeError`

llllllllll: column.metadata always exists, it was added in: https://github.com/quantopian/zipline/commit/cd09c5bc5c2a2f626855fec307ebd80d8fe532d8#diff-ca55a97f3934c34c38c2be0840a85340R178`

PR #1910, opened 2017-08-11
Remove and zero out unused treasury curves

dmichalowicz:
Remove unused treasury curves attribute from the performance tracker, and zero out the `treasury_period_return` value rather than removing it so as not to require a schema change for backtest consumers.


zipline/finance/risk/period.py
```diff
@@ -104,12 +88,7 @@ def calculate_metrics(self):
         self.benchmark_volatility = annual_volatility(self.benchmark_returns)
         self.algorithm_volatility = annual_volatility(self.algorithm_returns)
 
-        self.treasury_period_return = choose_treasury(
-            self.treasury_curves,
-            self._start_session,
-            self._end_session,
-            self.trading_calendar,
-        )
+        self.treasury_period_return = 0
```
ehebert: Should we add a comment about how this is needed to retain API/protocol compatibility, but that it could removed?

PR #1915, opened 2017-08-15
Volume roll current contracts could flip back and forth

dmichalowicz:
In `get_contract_center` of the `VolumeRollFinder`, we only look at a single date when trying to determine what the current contract is. This means that when deciding between the given front and back contracts we just choose the one with the higher volume. Since this method has no knowledge of any other dates, it can potentially flip flop on the current contract from day to day. This means that getting the active contract on a single date needs knowledge of the rolls of surrounding dates.

This symptom of seeing flip-flopping rolls could also be seen in history calls due to appending the current contract to the last day here: https://github.com/quantopian/zipline/blob/master/zipline/data/data_portal.py#L811

The one-off changes to the dates in the test are a result of https://github.com/quantopian/zipline/pull/1863, which allows contracts to be on active on their auto close date.

### DIFF
--- a/zipline/assets/roll_finder.py
+++ b/zipline/assets/roll_finder.py
@@ -15,6 +15,11 @@
 from abc import ABCMeta, abstractmethod
 from six import with_metaclass
 
+# Number of days over which to compute rolls when finding the current contract
+# for a volume-rolling contract chain. For more details on why this is needed,
+# see `VolumeRollFinder.get_contract_center`.
+ROLL_DAYS_FOR_CURRENT_CONTRACT = 90
+
 
 class RollFinder(with_metaclass(ABCMeta, object)):
     """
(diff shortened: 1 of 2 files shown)


zipline/assets/roll_finder.py
```diff
@@ -226,3 +229,8 @@ def _active_contract(self, oc, front, back, dt):
             if back_vol > front_vol:
                 return back
         return front
+
+    def get_contract_center(self, root_symbol, dt, offset):
+        day = self.trading_calendar.day
+        rolls = self.get_rolls(root_symbol, dt, dt + (180 * day), offset)
+        return self.asset_finder.retrieve_asset(rolls[0][0])
```
dmichalowicz: The number `180` is up for discussion as what it should be, or what the bounds on the `get_rolls` call should be. Either way this should probably be defined as a constant somewhere.

yankees714: Just a style thing, what would you think of writing the final line as:
```python
sid, _ = rolls[0][0]
return self.asset_finder.retrieve_asset(sid)
```
I like explicitly unpacking the tuple, since it makes it clear what the value represents.

zipline/assets/roll_finder.py
```diff
@@ -25,7 +25,7 @@ class RollFinder(with_metaclass(ABCMeta, object)):
     def _active_contract(self, oc, front, back, dt):
         raise NotImplementedError
 
-    def get_contract_center(self, root_symbol, dt, offset):
+    def _get_contract_center(self, root_symbol, dt, offset):
```
yankees714: So it seems like `_get_contract_center` and `get_contract_center` now mean different things. `_get_contract_center` returns what we would choose as the active contract with just the info we would have on `dt`, while `get_contract_center` returns the center given the calculated rolls (where in the base class, these are equivalent). Maybe we could rename `_get_contract_center` to make this clearer? To me the inheritance also feels a little convoluted otherwise.

dmichalowicz: great point, will revise

zipline/assets/roll_finder.py
dmichalowicz: yup clarity is good

PR #1917, opened 2017-08-17
BUG: Handle rolling to contract with auto-close date outside window

yankees714:
This fixes a bug with the edge case where:

- A contract (C_0) in the chain has an auto-close date equal to the `end` date passed to get_rolls
- The contract following C_0 (C_1), is *not* the active contract relative to the contract after it (C2), e.g. C_1 has lower volume than C_2, etc.

In this case, we need to roll from C_2 back to C_0, in which we would previously jump to the day following C_0's auto-close date. However, in this edge case, that would be outside the desired window (past `end`). Instead, in this case, we'll jump to `end` instead.

Also includes some cosmetic cleanup.

### DIFF
--- a/tests/test_continuous_futures.py
+++ b/tests/test_continuous_futures.py
@@ -1289,7 +1289,7 @@ def record_current_contract(algo, data):
 class RollFinderTestCase(WithBcolzFutureDailyBarReader, ZiplineTestCase):
 
     START_DATE = pd.Timestamp('2017-01-03', tz='UTC')
-    END_DATE = pd.Timestamp('2017-04-21', tz='UTC')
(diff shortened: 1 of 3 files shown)


tests/test_continuous_futures.py
```diff
@@ -1523,6 +1558,27 @@ def test_end_before_auto_close(self):
...
+            start=self.last_start_date,
+            end=self.fifth_auto_close_date,
+            offset=0,
+        )
+        self.assertEqual(
+            rolls,
+            [
+                (1003, pd.Timestamp('2017-04-18', tz='UTC')),
+                (1004, pd.Timestamp('2017-05-19', tz='UTC')),
+                (1006, None),
```
dmichalowicz: This is failing for me.

Issue #1901, opened 2017-08-01
Add info about using a virtualenv for installs


FreddieV4:
It seems that there have been many attempts to install zipline just by running `pip install` or `conda install`, which results in a bunch of common errors like:

- Downgrading of packages
- Mismatch of package versions in the global `package-list` or `conda packages` with those that Zipline uses
- Conflicting versions with other libraries; i.e. now you can't use another library because Zipline changed the package version or vice versa
- Installing on an unsupported Python version

I was thinking that we should suggest that people use virtualenvs or conda envs for their installs. Any thoughts on this @richafrank?

richafrank:
Environment isolation is a great thing for us to mention, especially for new folks to python. It won't solve the cases where people actually want to use libraries that conflict with zipline, but this education could help to prevent other cases of conflict.

z-Runmin:
Just submitted a PR on this issue.

PR #1922, opened 2017-08-20
DOC: Add info about using a conda envs for installs

z-Runmin:
I added information about how to use conda envs to install Zipline. (x-ref [#1901](https://github.com/quantopian/zipline/issues/1901))

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -124,34 +124,46 @@ channel:
 .. _`Hitchhiker's Guide to Python` : http://docs.python-guide.org/en/latest/
 .. _`Homebrew` : http://brew.sh
 
-Installing in ``conda`` environments
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Managing ``conda`` environments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 It is recommended to install Zipline in an isolated ``conda`` environment. 
-Installing Zipline in ``conda`` environment will not interfere your default 
+Installing Zipline in ``conda`` environments will not interfere your default 
 Python deployment or site-packages, which will prevent any possible conflict 
-with your global libraries. For more information on ``conda environment``, see 
+with your global libraries. For more information on ``conda`` environment, see 
 the `Conda User Guide <https://conda.io/docs/user-guide/tasks/manage-environments.html>`_
 
-Assuming ``conda`` has been set up, you can create a virtual environment:
+Assuming ``conda`` has been set up, you can create a ``conda`` environment:
+
+- Python 2.7:
 
 .. code-block:: bash
 
-    $ conda create -n zipline
+    $ conda create -n env_zipline python=2.7
 
-Then you should activate the conda environment by using the command
-
-.. code-block:: bash
-
-    (zipline) $ source activate zipline
-
-You can install Zipline by simply run
+- Python 3.5:
 
 .. code-block:: bash
 
-    (zipline) $ conda install -c Quantopian zipline
+    $ conda create -n env_zipline python=3.5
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -123,3 +123,35 @@ channel:
...
+
+    $ conda create -n zipline
+
+Then you should activate the conda environment by using the command
+
+.. code-block:: bash
+
+    (zipline) $ source activate zipline
+
+You can install Zipline by simply run
```
richafrank: typo - should be "You can install Zipline by running"

docs/source/install.rst
```diff
@@ -123,3 +123,35 @@ channel:
...
+Installing Zipline in ``conda`` environment will not interfere your default 
+Python deployment or site-packages, which will prevent any possible conflict 
+with your global libraries. For more information on ``conda environment``, see 
+the `Conda User Guide <https://conda.io/docs/user-guide/tasks/manage-environments.html>`_
+
+Assuming ``conda`` has been set up, you can create a virtual environment:
+
+.. code-block:: bash
+
+    $ conda create -n zipline
```
richafrank: It's possible I'm remembering incorrectly, but something that might be confusing at this step is that users will end up with a conda env with a python version that matches their root env. If they installed latest conda, then I think they'll have python3.6, for which we don't currently have zipline packages. So probably we should specify the python version for the newly created env or explain this situation.

docs/source/install.rst
```diff
@@ -123,3 +123,35 @@ channel:
...
+
+Installing in ``conda`` environments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+It is recommended to install Zipline in an isolated ``conda`` environment. 
+Installing Zipline in ``conda`` environment will not interfere your default 
+Python deployment or site-packages, which will prevent any possible conflict 
+with your global libraries. For more information on ``conda environment``, see 
+the `Conda User Guide <https://conda.io/docs/user-guide/tasks/manage-environments.html>`_
+
+Assuming ``conda`` has been set up, you can create a virtual environment:
```
richafrank: Let's call this a "conda env", just to disambiguate it from a "virtualenv" virtual environment.

docs/source/install.rst
richafrank: Let's mention somewhere that we're creating an env named "zipline" here, as opposed to installing zipline, which could be confusing for new folks.

docs/source/install.rst
```diff
@@ -123,3 +123,35 @@ channel:
 .. _`Arch Linux` : https://www.archlinux.org/
 .. _`Hitchhiker's Guide to Python` : http://docs.python-guide.org/en/latest/
 .. _`Homebrew` : http://brew.sh
+
+Installing in ``conda`` environments
```
richafrank: Since the previous block gives the command required to install in a conda environment (possibly the root environment), what do you think about titling this section "Managing conda environments"?

z-Runmin:
Thanks for the edits @richafrank ! I have updated the doc accordingly.

docs/source/install.rst
```diff
@@ -123,3 +123,47 @@ channel:
...
+
+- Python 3.5:
+
+.. code-block:: bash
+
+    $ conda create -n env_zipline python=3.5
+
+
+Now you have set up an isolated environment called ``env_zipline``, a sandbox-like
+structre to install Zipline. Then you should activate the conda environment
```
richafrank: typo "structre" -> "structure"

z-Runmin: Sorry for the typo. I just fixed it.

richafrank:
Thanks @z-Runmin !

PR #1927, opened 2017-08-29
BUG: take a copy of the categories list before mutating it in place

llllllllll:


llllllllll:
added the regression test

PR #1930, opened 2017-08-30
fix blaze core loader adjustment application for stale data

llllllllll:
We need to keep track of the asof date stored at index 0 because once we convert date-space to index-space, we cannot disambiguate data that falls before the start of the window. This change fixes a problem where we have an adjustment for a data point from before the first data in a pipeline window which was overwriting a more recent asof date that was written at index 0.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -41,6 +41,7 @@ from zipline.testing import (
 from zipline.testing.fixtures import WithAssetFinder
 from zipline.testing.predicates import assert_equal, assert_isidentical
 from zipline.utils.numpy_utils import float64_dtype, int64_dtype
+from zipline.utils.pandas_utils import empty_dataframe
 
 
 nameof = op.attrgetter('name')
@@ -458,19 +459,16 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
                    Equity(66 [B])      NaT  False
                    Equity(67 [C])      NaT  False
         """
-        df = pd.DataFrame(np.array(
-            [],
-            dtype=[
-                ('sid', 'int64'),
-                ('float_value', 'float64'),
-                ('str_value', 'object'),
-                ('int_value', 'int64'),
-                ('bool_value', 'bool'),
-                ('dt_value', 'datetime64[ns]'),
-                ('asof_date', 'datetime64[ns]'),
-                ('timestamp', 'datetime64[ns]'),
-            ],
-        ))
+        df = empty_dataframe(
+            ('sid', 'int64'),
+            ('float_value', 'float64'),
+            ('str_value', 'object'),
+            ('int_value', 'int64'),
+            ('bool_value', 'bool'),
+            ('dt_value', 'datetime64[ns]'),
+            ('asof_date', 'datetime64[ns]'),
+            ('timestamp', 'datetime64[ns]'),
+        )
 
         expr = bz.data(
             df,
@@ -1284,7 +1282,7 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
                       start,
                       end,
                       window_length,
-                      compute_fn):
+                      compute_fn=None):
         loader = BlazeLoader()
         ds = from_blaze(
             expr,
@@ -1299,6 +1297,15 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
 
         # prevent unbound locals issue in the inner class
         window_length_ = window_length
+
+        if compute_fn is None:
+            self.assertIsNone(
+                expected_output,
+                'expected_output must be None if compute_fn is None',
+            )
+
+            def compute_fn(data):
+                return data[0]
 
         class TestFactor(CustomFactor):
             inputs = ds.value,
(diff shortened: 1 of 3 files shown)


tests/pipeline/test_blaze.py
```diff
@@ -452,9 +458,19 @@ def test_cols_with_all_missing_vals(self):
                    Equity(66 [B])      NaT  False
                    Equity(67 [C])      NaT  False
         """
-        df = pd.DataFrame(columns=['sid', 'float_value', 'str_value',
-                                   'int_value', 'bool_value', 'dt_value',
-                                   'asof_date', 'timestamp'])
+        df = pd.DataFrame(np.array(
```
ssanderson: We should just put this in pandas_utils as like, `empty_dataframe`.

tests/pipeline/test_blaze.py
```diff
@@ -1377,7 +1393,87 @@ def test_deltas(self, asset_info, add_extra_sid):
                 compute_fn=np.nanmax,
             )
 
-    @with_extra_sid
+    @with_ignore_sid()
+    def test_deltas_before_index_0(self, asset_info, add_extra_sid):
```
ssanderson: Should this do something with `add_extra_sid`?

tests/pipeline/test_blaze.py
```diff
@@ -1377,7 +1393,87 @@ def test_deltas(self, asset_info, add_extra_sid):
...
+        })
+
+        column_constant = np.arange(len(asset_info)) * 100
+        expected_views = {
+            k: v + column_constant
+            for k, v in expected_views_single_sid.items()
+        }
+        with tmp_asset_finder(equities=asset_info) as finder:
+            nassets = len(asset_info)
+            expected_max = np.array([0, 1, 1, 3]) + (nassets - 1) * 100
```
ssanderson: The `max` calculations here aren't really relevant to the behavior under test here; can we just pass a sentinel or something for the case where we don't care about the pipeline calculation itself?

tests/pipeline/test_blaze.py
```diff
@@ -1377,7 +1393,87 @@ def test_deltas(self, asset_info, add_extra_sid):
...
+                ('sid', 'int64'),
+                ('value', 'float64'),
+                ('asof_date', 'datetime64[ns]'),
+                ('timestamp', 'datetime64[ns]'),
+            ],
+        ))
+        expr = bz.data(df, name='expr', dshape=self.dshape)
+
+        T = pd.Timestamp
+        deltas_df_single_sid = pd.DataFrame({
```
ssanderson: This made sense to me after you explained it in person, but I wouldn't have figured it out easily by myself.  A couple sentences explaining what's interesting about the data under test here would be helpful.

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -258,6 +259,15 @@ cdef _array_for_column_impl(object dtype,
     cdef list non_null_ts_ixs_by_column_ix = [
         set() for _ in range(out_array.shape[1])
     ]
+
+    cdef np.ndarray[np.int64_t] index_0_asof_date_by_column_ix
+    if AsArrayKind is AsAdjustedArray:
+        index_0_asof_date_by_column_ix = np.full(
+            out_array.shape[1],
+            pd.Timestamp.min.asm8.view('int64'),
```
ssanderson: Slightly clearer here might be `pd.Timestamp.min.value`?

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -573,6 +600,11 @@ cdef arrays_from_rows(DatetimeIndex_t dates,
             out_shape,
             size,
             ts_ixs,
+            (
+                all_rows[AD_FIELD_NAME].values.view('int64')
+                if len(all_rows) else
```
ssanderson: Might be worth noting that we're working around pandas losing dtype metadata for empty objects here.

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -314,6 +324,17 @@ cdef _array_for_column_impl(object dtype,
             else:
                 adjustment_list = <list> adjustment_list_ptr
 
+            if asof_ix == 0:
+                # If the asof_ix == 0, this value may fall before the start of
+                # our window. In this case, we only want to apply data that is
+                # more recent than the current asof date stored at index 0.
+                with cython.boundscheck(False), cython.wraparound(False):
+                    asof_date = asof_dates[n]
+                    if asof_date >= index_0_asof_date_by_column_ix[column_ix]:
```
ssanderson: I might comment here that this is `>=` rather than `>` because if we have the same asof but a newer timestamp this value should take precedence.

tests/pipeline/test_blaze.py
llllllllll: done, I add an extra asset to the input which is just ignored so the expected output is the same.

zipline/pipeline/loaders/blaze/_core.pyx
llllllllll: done

zipline/pipeline/loaders/blaze/_core.pyx
llllllllll: done

PR #1939, opened 2017-09-12
BUG: fix issue with sparse output date queries in the blaze core loader

llllllllll:
when multiple rows occur between the requested dates we don't correctly use asof date to tie-break the value written to the index. This is a generalization of the issue fixed in #1930 

cc @ssanderson @ehebert @jmccorriston


tests/pipeline/test_blaze.py
```diff
@@ -1483,6 +1483,112 @@ def test_deltas_before_index_0(self, asset_info, add_extra_sid):
...
+            ('sid', 'int64'),
+            ('value', 'float64'),
+            ('asof_date', 'datetime64[ns]'),
+            ('timestamp', 'datetime64[ns]'),
+        )
+        expr = bz.data(df, name='expr', dshape=self.dshape)
+
+        T = pd.Timestamp
+
+        # These data are interesting because we have pairs of rows that come on
```
ssanderson: This is super clear 👍.

tests/pipeline/test_blaze.py
```diff
@@ -1483,6 +1483,112 @@ def test_deltas_before_index_0(self, asset_info, add_extra_sid):
...
+            for k, v in expected_views_single_sid.items()
+        }
+        with tmp_asset_finder(equities=asset_info) as finder:
+            # The dates queried are non-contiguous. We have two day groups to
+            # capture the two day pairs in the input data.
+            dates = pd.to_datetime(['2014-01-03', '2014-01-05', '2014-01-07'])
+            self._run_pipeline(
+                expr,
+                deltas,
+                None,
```
ssanderson: What are the `None`s here?  Might help to either pass these by keyword or comment inline.

llllllllll: it's expected output, which can now be None if `compute_fn` is None to just check the views but not the full output. This is what you suggested in the other PR. I will pass these by kwarg.

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -314,6 +312,21 @@ cdef _array_for_column_impl(object dtype,
...
+                # last recorded asof_date at the given index and we should
+                # treat this value as the best known row. We use >=
+                # because a more recent row with the same asof_date
+                # should be treated as an adjustment and the new value
+                # becomes the best-known.
+                most_recent_asof_date_for_ix[asof_ix, column_ix] = asof_date
+            else:
+                # The asof_date is earlier than the asof_date written
+                # at the given index. Ignore this row.
+                continue
```
ssanderson: Checking my own comprehension: `continue`-ing here is saying that "if we get records with asof_dates of saturday and sunday, but we only care about the value as of monday, then there's no case where we ever care about the record whose asof is saturday, so we should throw that record in the trash."?

llllllllll: Correct, we didn't know about the Saturday record on Friday, and by the time Monday comes we have more up to date information. There is no row in the output where that record would be the most recently known data for the time.

tests/pipeline/test_blaze.py
```diff
@@ -1645,6 +1751,164 @@ def test_deltas_before_index_0_macro(self):
...
+        )
+        expr = bz.data(df, name='expr', dshape=self.macro_dshape)
+
+        T = pd.Timestamp
+
+        # These data are interesting because they exercise the tie breaking of
+        # adjustments. Here we have 4 rows which we learn about within a single
+        # calendar index. The first row provides the most recently known value
+        # for some day in the window. All of the following rows are adjustments
+        # to the same historical value. We expect that the first row's value is
```
ssanderson: "an earlier historical value" might make it clearer that the adjustments are for a different date than the first value.

llllllllll: "earlier historical" seems a bit redundant

ssanderson:
LGTM

PR #1942, opened 2017-09-12
Travis update

richafrank:
Travis updated our build image including installed packages, etc, which broke our conda packaging. This is to fix the build by updating the conda tools we're using.


richafrank:
FYI At some point we should bootstrap conda packages with new build strings, which will include recipe hashes, but for now we're still using old build strings.

Issue #1941, opened 2017-09-12
Remove dead code links from _modules on zipline.io


FreddieV4:
It appears we still have links to dead code in zipline.io via the `_modules` directory.

e.g. http://www.zipline.io/_modules/zipline/data/bundles/yahoo.html

Noticed this because it was mentioned in #1937

richafrank:
With #1943 merged, I believe we just need to deploy updated docs. If we want to deploy the version from the latest release instead, can just run `make clean` first.

richafrank:
Updated zipline.io after cleaning my output directory

PR #1943, opened 2017-09-12
Clean docs

richafrank:
For https://github.com/quantopian/zipline/issues/1941

@FreddieV4 Will you confirm that running deploy.py from this branch removes the yahoo bundle html for you? Since this is about stale files, I want to make sure I didn't pollute my test env.

### DIFF
--- a/docs/deploy.py
+++ b/docs/deploy.py
@@ -36,6 +36,8 @@ def main():
     os.chdir(HERE)
 
     try:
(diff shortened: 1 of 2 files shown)


ernestoeperez88:
I tested this as well. I generated old docs using 1.1.0, and then generated new ones using this branch. The yahoo html file gets removed from the new docs.

richafrank:
Thanks @ernestoeperez88 !

PR #1946, opened 2017-09-14
ENH: Change default commission to .001

srklonaris:
Change default commission to .001 to be consistent with the contest.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -1488,11 +1488,12 @@ class TestBeforeTradingStart(WithDataPortal,
 
         # Starting portfolio value is 10000. Order for the asset fills on the
         # second bar of 1/06, where the price is 391, and costs the default
-        # commission of 1. On 1/07, the price is 780, and the increase in
-        # portfolio value is 780-392-1
+        # commission of 0. On 1/07, the price is 780, and the increase in
+        # portfolio value is 780-392-0
         self.assertEqual(results.port_value.iloc[0], 10000)
         self.assertAlmostEqual(results.port_value.iloc[1],
-                               10000 + 780 - 392 - 1)
+                               10000 + 780 - 392 - 0,
+                               places=2)
 
     def test_portfolio_bts_with_overnight_split(self):
         algo_code = dedent("""
@@ -1572,7 +1573,7 @@ class TestBeforeTradingStart(WithDataPortal,
         # On 1/07, portfolio value is the same as without split
         self.assertEqual(results.port_value.iloc[0], 10000)
         self.assertAlmostEqual(results.port_value.iloc[1],
-                               10000 + 780 - 392 - 1)
+                               10000 + 780 - 392 - 0, places=2)
 
 
 class TestAlgoScript(WithLogger,
@@ -1762,7 +1763,7 @@ def handle_data(context, data):
     @parameterized.expand(
         [
             ('no_minimum_commission', 0,),
-            ('default_minimum_commission', 1,),
+            ('default_minimum_commission', 0,),
             ('alternate_minimum_commission', 2,),
         ]
     )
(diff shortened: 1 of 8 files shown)


jbredeche:
what about `DEFAULT_MINIMUM_COST_PER_EQUITY_TRADE`?  do we still want a minimum of $1/trade?

srklonaris:
After discussion with Josh, I also updated `DEFAULT_MINIMUM_COST_PER_EQUITY_TRADE` and `DEFAULT_MINIMUM_COST_PER_FUTURE_TRADE` to both be `0.0`.  

I had to update some of the tests too.  Because the issue here: https://github.com/quantopian/zipline/blob/master/tests/resources/rebuild_example_data#L105 I can't rebuild the example data, so I had to manually set the commisions in each example to the old commision values so that the results would be the same.  This should be a temporary fix until we can rebuld the example data again.

@dmichalowicz do you mind taking a look again?

tests/test_algorithm.py
```diff
@@ -1390,11 +1390,12 @@ def handle_data(context, data):
...
         # second bar of 1/06, where the price is 391, and costs the default
-        # commission of 1. On 1/07, the price is 780, and the increase in
-        # portfolio value is 780-392-1
+        # commission of 0. On 1/07, the price is 780, and the increase in
+        # portfolio value is 780-392-0
         self.assertEqual(results.port_value.iloc[0], 10000)
         self.assertAlmostEqual(results.port_value.iloc[1],
-                               10000 + 780 - 392 - 1)
+                               10000 + 780 - 392 - 0,
+                               places=2)
```
srklonaris: I was getting this failure `'10387.99899997449 != 10388 within 7 places'` without adding `places=2`  I'm not sure if there is a better way to do this.

dmichalowicz: I think this is fine.

zipline/examples/buy_and_hold.py
```diff
@@ -22,6 +23,12 @@ def initialize(context):
     context.has_ordered = False
     context.stocks = stocks
 
+    # Explicitly set the commission to the "old" value until we can
+    # rebuild example data.
+    # github.com/quantopian/zipline/blob/master/tests/resources/
+    # rebuild_example_data#L105
```
dmichalowicz: TLDR: What you have here is good.

It's actually still possible to rebuild the example data (with a couple minor tweaks), I did so [here](https://github.com/quantopian/zipline/pull/1910) only about a month ago. We now get benchmark data from Google instead of Yahoo, as seen [here](https://github.com/quantopian/zipline/blob/master/zipline/data/benchmarks.py#L46-L51).

_However_, it appears that as of only a week or two ago, Google changed the URL from which they are serving their financial data, causing pandas datareader to break. As a result, after checking out your branch I was unable to rebuild the data like I was previously (without having to either (1) make our own pandas-datareader fork, or (2) make a bunch of extra changes to the tests and test dates; neither of which I want or think we need to do).

For more info see the following:
- https://github.com/pydata/pandas-datareader/issues/391
- https://stackoverflow.com/questions/46259528/datareader-google-finance-date-not-working/46356247#46356247
- https://stackoverflow.com/questions/46243360/google-api-changed-for-data-from-google-finance

All that being said, I think your solution of keeping the old parameters for now is best. When the pandas datareader issue is fixed though I think this may be worth revisiting.

@richafrank @FreddieV4 Is this breakage something we should be worried about anywhere else?

dmichalowicz:
@srklonaris Looks good to me

dmichalowicz:
@srklonaris Quick note before you merge, I would rebase this off of master first as it looked like it was branched off a really old commit

zipline/examples/buy_and_hold.py
richafrank: Yea, I think we need to revisit our benchmark downloading, either fixing google, adding back yahoo (via pandas-datareader if it supports it), or both.

dmichalowicz: Filed https://github.com/quantopian/zipline/issues/1965

PR #1952, opened 2017-09-18
PERF: Use a sparse data frame for recording position weights

dmichalowicz:
Potential change for https://github.com/quantopian/zipline/pull/1839/files#r138711152

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -1553,7 +1553,7 @@ class TestPortfolio(WithDataPortal, WithSimParams, ZiplineTestCase):
         # Since we always hold exactly one contract in sid 1004, we should see
         # our weight transition from being at offset 4 to offset 3, to offset
         # 2, etc.
-        for day, positions in zip(sim_params.sessions[1:], pt_weights[1:]):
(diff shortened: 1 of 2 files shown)


ehebert:
Do you have a comparison of memory usage between the two implementations?

Also would be interested in seeing a comparison of time profiling.

dmichalowicz:
@ehebert Did some profiling, TLDR I don't think this is going to help enough to justify.

The problem with creating the DataFrame first is that because we don't know the number of columns beforehand, it means we have to append every new row (day) of data as we go, creating new columns as we see new assets. I didn't do a super in-depth memory look, but from what I can glean this will likely hurt more time-wise than it could possibly save memory-wise. In just running the expected shortfall test, which trades a single asset over a 2-year backtest, the time without this change was ~2.7 seconds, and with this change went up to ~5.7 seconds.

I'm going to close this PR for now, but if you want to revisit or talk in person feel free to let know.

ehebert:
I agree, that slow down does not seem worth it.

I'm not sure if there is way to implement the same logic without calling `to_sparse` on every market_close.

Let's keep this closed, but may need to revisit if memory issues do become an issue.

Issue #1771, opened 2017-04-26
Yahoo Bundle Factories example not working


DMTSource:
I am attempting to use the example for the Yahoo Bundle Factories:
http://www.zipline.io/bundles.html#yahoo-bundle-factories

But I get the following errors. Ubuntu 1404, just upgraded to zipline 1.1.0. Using the exact example from the link above. My hope is to, when its working, add the SPY symbol so I can use that price data in zipline.

zipline ingest -b my-yahoo-equities-bundle
Merging daily equity files:  [------------------------------------]  2#]  100%
Merging daily equity files:  [####################################]   
Downloading Yahoo adjustment data:   [####################################]  100%
Traceback (most recent call last):
  File "/usr/local/bin/zipline", line 11, in <module>
    load_entry_point('zipline==1.1.0', 'console_scripts', 'zipline')()
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/zipline/__main__.py", line 312, in ingest
    show_progress,
  File "/usr/local/lib/python2.7/dist-packages/zipline/data/bundles/core.py", line 451, in ingest
    pth.data_path([name, timestr], environ=environ),
  File "/usr/local/lib/python2.7/dist-packages/zipline/data/bundles/yahoo.py", line 178, in ingest
    adjustment_writer.write(splits=splits, dividends=dividends)
  File "/usr/local/lib/python2.7/dist-packages/zipline/data/us_equity_pricing.py", line 1169, in write
    self.write_frame('splits', splits)
  File "/usr/local/lib/python2.7/dist-packages/zipline/data/us_equity_pricing.py", line 931, in write_frame
    frame,
  File "/usr/local/lib/python2.7/dist-packages/zipline/data/us_equity_pricing.py", line 891, in _write
[...]

Issue #1641, opened 2017-01-08
Test test_perf_tracking.py 


fysdbd:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

* Operating System: Linux ubuntu 4.8.6-x86_64-linode78 #1 SMP Tue Nov 1 14:51:21 EDT 2016 x86_64 x86_64 x86_64 GNU/Linux
* Python Version: Python 2.7.6
* Python Bitness: 64
* How did you install Zipline: cloned from github
* Python packages:
alabaster==0.7.6
alembic==0.7.7
AlgoTrading==0.2.2
apt-xapian-index==0.45
Babel==1.3
backports-abc==0.5
backports.shutil-get-terminal-size==1.0.0
backports.ssl-match-hostname==3.4.0.2
bcolz==0.12.1
beautifulsoup4==4.2.1
blaze==0.10.1
boto==2.20.1
Bottleneck==1.2.0
certifi==2015.4.28
chardet==2.0.1
click==4.0
cloudpickle==0.2.2
colorama==0.2.5
conda==4.2.7
configparser==3.5.0
contextlib2==0.4.0
cookies==2.2.1
coverage==4.0.3
cvxopt==1.1.4
cycler==0.10.0
cyordereddict==0.2.2
Cython==0.25.2
dask==0.13.0
DataAPI==0.8.2
datashape==0.5.2
decorator==4.0.0
Django==1.6.1
docopt==0.6.2
docutils==0.12
empyrical==0.2.1
entrypoints==0.2.2
enum34==1.1.6
Finance-Python==0.4.3
flake8==2.4.1
Flask==0.12
Flask-Cors==3.0.2
funcsigs==0.4
functools32==3.2.3.post2
futures==3.0.5
gevent==1.2.0
greenlet==0.4.11
guppy==0.1.9
h5py==2.6.0
HelloWorld==0.1
html5lib==0.999
http==0.2
httplib2==0.8
ImageHash==3.1
intervaltree==2.1.0
iotop==0.6
ipykernel==4.5.2
ipython==5.1.0
ipython-genutils==0.1.0
ipywidgets==5.2.2
itsdangerous==0.24
Jinja2==2.7.3
joblib==0.7.1
jsonschema==2.5.1
jupyter==1.0.0
jupyter-client==4.4.0
jupyter-console==5.0.0
jupyter-core==4.2.1
Keras==1.1.2
latexcodec==1.0.4
locket==0.2.0
[...]

PR #1954, opened 2017-09-19
BUG: Fix frame column validation in Python 2.7.5 (#1771)

tibkiss:


zipline/data/us_equity_pricing.py
```diff
@@ -915,7 +915,8 @@ def _write(self, tablename, expected_dtypes, frame):
                 np.array([], dtype=list(expected_dtypes.items())),
             )
         else:
-            if frozenset(frame.columns) != viewkeys(expected_dtypes):
+            if frozenset(frame.columns) != \
+               frozenset(viewkeys(expected_dtypes)):
```
ssanderson: Is there still a point in calling `viewkeys` here if we're just going to pass the input to `frozenset` anyway?

ssanderson:
@tibkiss this seems reasonable to me.  I assume this is working around https://bugs.python.org/issue8404 ?  Just had one small perf-related question.

richafrank:
Looks like a fix for https://github.com/quantopian/zipline/issues/1641

zipline/data/us_equity_pricing.py
tibkiss: good point, updated the PR accordingly

tibkiss:
@richafrank / @ssanderson : Thanks for the quick review & sorry for the late response. 

@ssanderson : Even though your linked issue looks really similar I don't think it is exactly the same. In issue8404 the affectedVersion is a Python RC and the problem I'm workarounding here is present on python 2.7.5. 

Nonetheless this fix solves #1771 & #1641 (which are essentially the same).

richafrank:
CI tests pass, and this looks safe to me. Merging.

richafrank:
Thanks @tibkiss !

PR #1956, opened 2017-09-19
TEST: Add test coverage for datetime factors.

ssanderson:
We previously didn't have a testcase for a factor that works on datetime
inputs from a real loader.


ssanderson:
@llllllllll is adding a test that exercises this behavior via the blaze core loader, which is more indicative of real use.

PR #1958, opened 2017-09-20
BUG: don't mask out values in AdjustedArray

llllllllll:


zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -352,7 +351,7 @@ cdef _array_for_column_impl(object dtype,
                 # upper bound doesn't include the timestamp because we've
                 # already included the timestamp-date in the baseline.
                 end = max(ts_ix - 1, 0)
-                if end >= asof_ix and out_array[asof_ix, column_ix] != value:
+                if end >= asof_ix: # and out_array[asof_ix, column_ix] != value:
```
ssanderson: Do we still want this branch commented here?  What did the perf impact of this end up being?

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -404,7 +403,7 @@ cdef _array_for_column_impl(object dtype,
             # of v1. However, if we look back from t6, we should see v0 for the
             # period from t0 to t1.
             end = max(non_null_ad_ixs[ix] - 1, 0)
-            if end >= asof_ix and out_array[asof_ix, column_ix] != value:
+            if end >= asof_ix: #  and out_array[asof_ix, column_ix] != value:
```
ssanderson: Same questions as above.

zipline/pipeline/loaders/blaze/_core.pyx
```diff
@@ -436,12 +435,11 @@ cdef _array_for_column_impl(object dtype,
         )
     else:
         # this cast prevents a compiler crash
-        baseline_array = <object> out_array
+        baseline_array = out_array.view(dtype)
```
ssanderson: Why this change?

llllllllll: Oh, there were actually two other issues here:

1. we need to view the array as the correct type or datetime64[ns] and bool arrays will appear as ints
2. the unsafe optimization was actually shown to be unsafe in real data and the perf impact of removing it is fine

ssanderson:
This looks good to me assuming we don't need to do more work on adjustment perf.

llllllllll:
For 1 year of the Q1500US, which has good coverage of different fields and adjustments, we get:
```
master: 32.664
branch: 35.976
```

so we get a slight slowdown in exchange for ensured correctness. I think that is okay.

PR #1963, opened 2017-09-26
BUG: Fix ffill parameter handling in DataPortal

tibkiss:


richafrank:
@ehebert Could you take a look at this?

tibkiss:
@ehebert : In zipline-live there is a DataPortalLive class, which extends DataPortal with IB Data feed support. I'm working on a patch where we override get_history_window to merge ingested historical data with realtime collected data. In that method I rely on ffill=False.

The current version looks as follows:
`class DataPortalLive(DataPortal):
    def __init__(self, broker, *args, **kwargs):
        self.broker = broker
        super(DataPortalLive, self).__init__(*args, **kwargs)

    def get_last_traded_dt(self, asset, dt, data_frequency):
        return self.broker.get_last_traded_dt(asset)

    def get_spot_value(self, assets, field, dt, data_frequency):
        return self.broker.get_spot_value(assets, field, dt, data_frequency)

    def get_history_window(self,
                           assets,
                           end_dt,
                           bar_count,
                           frequency,
                           field,
                           data_frequency,
                           ffill=True):
        # This method is responsible for merging the ingested historical data
        # with the real-time collected data through the Broker.
        # DataPortal.get_history_window() is called with ffill=False to mark
        # the missing fields with NaNs. After merge on the historical and
        # real-time data the missing values (NaNs) are filled based on their
        # next available values in the requested time window.
        #
        # Warning: setting ffill=True in DataPortal.get_history_window() call
        # results a wrong behavior: The last available value reported by
        # get_spot_value() will be used to fill the missing data - which is
        # always representing the current spot price presented by Broker.

        ingested_historical = super(DataPortalLive, self).get_history_window(
            assets, end_dt, bar_count, frequency, field, data_frequency,
            ffill=False)

        realtime_historical = self.broker.get_realtime_history(
            assets, frequency)

        # Broker.get_realtime_history() returns the asset as level 0 column,
        # open, high, low, close, volume returned as level 1 columns.
        # To filter for field the levels needs to be swapped
        realtime_historical = realtime_historical.swaplevel(0, 1, axis=1)

        field = 'close' if field == 'price' else field

        combined_historical = ingested_historical.copy()
        combined_historical.update(realtime_historical[field])

        if ffill:
            # Simple forward fill is not enough here as the last ingested
            # value might be outside of the requested time window. That case
            # the time series starts with NaN and forward filling won't help.
            # To provide values for such cases we backward fill.
            # Backward fill as a second operation will have no effect if the
            # forward-fill was successful.
            combined_historical.fillna(method='ffill', inplace=True)
            combined_historical.fillna(method='bfill', inplace=True)

        return combined_historical
`

ehebert:
@tibkiss, thanks for the greater context.

On my first reading, your implementation loses some of the existing semantics of only ever applying forward-filling to the `price` column. e.g. a `volume` field would be forward and back filled, since `_protocol.pyx` would be calling `get_history_window` with a value of `ffill=True` which would be an error.

My recommendation would be to put the logic which combines previous ingested and live data in to a bar reader which dispatches reads, based on the dates of the index, to two readers which implement the `MinuteBarReader` interface, one for previously ingested reads and one for current day from the broker.

This reader would be used by init-ing a non-subclassed `DataPortal` with the live dispatch reader as the equity minute bar reader parameter. `DataPortal.get_history_window` would then use this combined reader since the `__init__` would pass it as a parameter to the  `MinuteHistoryLoader` instance. The `MinuteHistoryLoader` would then call the `load_adjusted_arrays` provided by the new reader.

Moving the complexity for straddling live and historical data into the reader provides these benefits:

-   Reduces overall complexity by avoiding having to wire in a separate `LiveDataPortal`.
-   Reduces chances for subtle drifting in behavior and semantics between live and backtesting (e.g. the aforementioned forward and back filling of `volume`)
-   Is less brittle to refactorings or API changes to `DataPortal`, such as my proposal earlier in the thread to remove `ffill` completely.

The following is loosely based on a class we have internally, which has started on the path of putting the logic in a minute bar reader. (Ours still relies on some logic in `LiveDataPortal`, but this version attempts to removes those dependencies, and not been run anywhere, so caveat emptor.)

```
class LiveDispatchBarReader(object):

    def __init__(self,
                 trading_calendar,
                 live_day,
                 historical_minute_reader,
                 live_minute_reader):
        self._trading_calendar = trading_calendar
        self._live_day = live_day
        self._live_market_open = \
            self.trading_calendar.open_and_close_for_session(self._live_day)[0]
        self._historical_minute_reader = historical_minute_reader
        self._live_minute_reader = live_minute_reader

    def get_value(self, sid, dt, field):
        if dt < self._live_day:
            return self._minute_reader.get_value(sid, dt, field)
        else:
            return self._live_data_portal._get_live_value(sid, field, dt)

    def load_raw_arrays(self, fields, start_dt, end_dt, sids):
        mm = self._trading_calendar.all_minutes
        minutes_for_window = mm[mm.slice_indexer(start_dt, end_dt)]

        out = []
        for field in fields:
            if minutes_for_window[-1] < self._live_day:
                # entire window is in the past
                data = self._historical_minute_reader.load_raw_arrays(
                     field, start_dt, end_dt, sids)
            elif minutes_for_window[0] > self._live_day:
                # entire window is live
                data = self._live_minute_reader.load_raw_arrays(
                     field, start_dt, end_dt, sids)
            else:
                # blended historical and live
                open_idx = minutes_for_window.searchsorted(
                    self._live_market_open)
                old_minutes = minutes_for_window[0:open_idx]
                today_minutes = minutes_for_window[open_idx:]

                # Read old data.
                old_data = self._historical_minute_reader.load_raw_arrays
                        sids, field, old_minutes
                    )

                # Read current data.

                today_data = self._live_minute_reader.load_raw_arrays(
                    fields, start_dt, end_dt, sids)

                # Combine data.
                data = np.concatenate([old_data, today_data])

        return data
```

tibkiss:
@ehebert: Thanks for the detailed answer and the possible solution.

I pushed the original implementation into zipline-live and will make note of your approach as improvement. The reason behind is that changing the implementation would take me at least an additional week, while the original version was already tested.

ehebert:
@tibkiss shall we close this issue without merging? If you get to it in the future, we can work on upstreaming the dispatch approach.

tibkiss:
@ehebert : Sure, we can close this. Thanks for your support!

Issue #1965, opened 2017-09-27
Benchmark downloading is broken


dmichalowicz:
Fix benchmark downloading from Google with pandas-datareader. This issue was originally brought up [here](https://github.com/quantopian/zipline/pull/1946/files#r141126481).

We now get benchmark data from Google instead of Yahoo, as seen [here](https://github.com/quantopian/zipline/blob/master/zipline/data/benchmarks.py#L46-L51).

However, it appears that as of only a week or two ago, Google changed the URL from which they are serving their financial data, causing pandas datareader to break. This is also preventing us from rebuilding the `test_examples` data. (For more info see the original post above).

cyniphile:
Had this issue too. Just changed back to yahoo in the meantime in `get_benchmark_returns` which seems to work.

MBattagl:
Same issue here https://github.com/quantopian/zipline/issues/1953

FreddieV4:
x-ref https://github.com/pydata/pandas-datareader/issues/395

yiorgosn:
Quick solution: Use a manually downloaded local copy of SPY (yahoo lets you download manually the entire history).  I modified the benchmarks.py to try google first and if it fails to look for a local csv copy instead.  I attach the modified benchmarks.py file it should replace the existing one (so make a a copy of the original first before you overwrite it).  The benchmarks.py file is usually found in:  %USERPROFILE%\Anaconda3\envs\py34\Lib\site-packages\zipline\data.  If you didn't create a unique environment for it then don't specify py34 after envs.

 Also make sure that your local directory is reflected in this line in the code: 
new_dir = 'c:/Downloaded_csv' 
[benchmarks.txt](https://github.com/quantopian/zipline/files/1354884/benchmarks.txt)

tanaytrivedi:
Hi,
Is there an official solution out there for running backtests and not having the system break every time because of this bechmark issue? @yiorgosn solution doesn't work for me, I think you have to do more than just replace the file. I have the exact same failure even with his file. Of course, I have replaced the file directory to make sure it looks in the right place for the csv.

Is there a way I can run the backtest without doing a benchmark until it is fixed? Without, that is, ripping up the code and removing any mention of benchmarks.
Thanks

brian-from-quantrocket:
You can try setting the benchmark to an asset that's already in your bundle. For example if running the example algos with AAPL, tell Zipline to use AAPL as your benchmark.

```python
from zipline.api import symbol, set_benchmark

[...]

PR #1967, opened 2017-09-28
BUG/MAINT: Adjust get_benchmark_returns to match Google API change

FreddieV4:
This zeros out any dates that come before the 251 day limit that google sets, if we request > 251 days worth of benchmark data.


FreddieV4:
So this won't actually work b/c of https://github.com/quantopian/zipline/issues/1965 and https://github.com/pydata/pandas-datareader/issues/395. It's not that we need to chain data (which is probably better in the general case)

PR #1969, opened 2017-10-02
EHH: New filter NoMissingValues with tests

ahgnaw:


zipline/pipeline/filters/filter.py
```diff
@@ -533,3 +535,34 @@ class StaticAssets(StaticSids):
...
+
+        return super(NoMissingValues, cls).__new__(
+            cls,
+            inputs=inputs,
+            window_length=window_length,
+            mask=mask,
+        )
+
+    def compute(self, today, assets, out, value):
+        if isinstance(value, LabelArray):
```
ssanderson: My inner micro-optimizer is slightly twitchy that we're doing this dispatch every day even though we know the dtypes will never change.  In practice though, I think the cost is probably negligible

```
In [28]: %%timeit
    ...: data.any(axis=0)
    ...:
    ...:
100000 loops, best of 3: 3.61 µs per loop

In [30]: %%timeit
    ...: if isinstance(x, int):
    ...:     data.any(axis=0)
    ...: else:
    ...:     data.any(axis=0)
    ...:
100000 loops, best of 3: 3.77 µs per loop
```

0.16 microseconds per factor-dayends up being basically nothing.  For example, for a 10-year pipeline, that adds up to 0.0004 seconds.

zipline/pipeline/filters/filter.py
```diff
@@ -533,3 +535,34 @@ class StaticAssets(StaticSids):
...
+            out[:] = ~np_any(value.is_missing(), axis=0)
```
ssanderson: One thing that probably **is** worth doing here, though, is passing `out=out` to np.any instead of assigning the result to `out[:]`.

zipline/pipeline/filters/filter.py
```diff
@@ -533,3 +535,34 @@ class StaticAssets(StaticSids):
     def __new__(cls, assets):
         sids = frozenset(asset.sid for asset in assets)
         return super(StaticAssets, cls).__new__(cls, sids)
+
+
+class NoMissingValues(CustomFilter, SingleInputMixin):
```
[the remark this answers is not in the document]
ssanderson: I tend to try to name filters after a positive property rather than the negation of a property, e.g., `AllPresent` instead of `NoMissingValues`.  The main reason for that preference that I think the positive assertion is easier to parse if it gets negated: `~AllPresent()` ("not all present") is easier for me to think about than `~NoMissingValues` ("not no missing values").  Thoughts?

tests/pipeline/test_filter.py
```diff
@@ -379,6 +380,154 @@ def test_isfinite(self):
...
+            dtype = object_dtype
+            inputs = ()
+            window_length = 0
+            missing_value = ''
+            window_safe = True
+
+        input_factor = SomeWindowSafeStringClassifier()
+
+        shape = (10, 6)
+        data = choice(
```
ssanderson: I think this should be a LabelArray if we're trying to simulate what the term would actually get passed?

ahgnaw: I think classifier inputs gets converted to a LabelArray when it is an input to `compute()`

zipline/pipeline/filters/filter.py
ahgnaw: That would make a difference, however the output should be the negation of the `np.any` results.

zipline/pipeline/filters/filter.py
```diff
@@ -533,3 +535,34 @@ class StaticAssets(StaticSids):
...
+    """Pipeline filter indicating input term has data for a given window.
+    """
+
+    def __new__(cls,
+                inputs,
+                window_length,
+                mask=NotSpecified):
+
+        if isinstance(inputs[0], Filter):
+            raise ValueError(
```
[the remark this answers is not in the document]
ahgnaw: It makes more sense to check in _validate, will change it. Also changing the error to `TypeError`.

zipline/pipeline/filters/filter.py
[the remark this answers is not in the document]
ahgnaw: `CustomFilter` already has `PositiveWindowLengthMixin`, I'll add `StandardOutputs`.

I agree, a positive assertion is is more easily read if negated.

zipline/pipeline/filters/filter.py
richafrank: Would moving the negation inside work? `np_all(~value.is_missing(), axis=0, out=out)`

ahgnaw: That wouldn't work in the case where an asset has all filled in data.

For the case` value = [[3.], [4.], [5.]]`, the inner expression resolves to [[F], [F], [F]]. The function `np.all` would return `False` for that asset, even though it is not missing any values.

ahgnaw: That doesn't work in the case where some of the asset's values are missing, but `np.all` would.

ahgnaw: Ah, nvm, didn't see that @richafrank suggested `np.all` already.

tests/pipeline/test_filter.py
ssanderson: The place that data from loaders is normally converted into a LabelArray is in `zipline.lib.adjusted_array._normalize_array`, but that's not happening here because this test suite doesn't use any pipeline loaders, it just pre-populates the graph with the values provided by the test suite.

ssanderson: Turns out this **is** happening, in `PipelineEngine._inputs_for_term`.

docs/source/appendix.rst
```diff
@@ -251,6 +251,30 @@ Built-in Factors
 .. autoclass:: zipline.pipeline.factors.WeightedAverageValue
    :members:
 
+Built-in Filters
```
ssanderson: :+1:

zipline/pipeline/filters/filter.py
```diff
@@ -533,3 +535,27 @@ class StaticAssets(StaticSids):
...
+        if isinstance(self.inputs[0], Filter):
+            raise TypeError(
+                "Input to filter `AllPresent` cannot be a Filter."
+            )
+
+        return super(AllPresent, self)._validate()
+
+    def compute(self, today, assets, out, value):
+        if isinstance(value, LabelArray):
+            np_all(~value.is_missing(), axis=0, out=out)
```
ssanderson: Small perf note: I think we can avoid making an extra copy of the full window here by rewriting this expression as `~np.any(value.is_missing, ...)`.  Lifting the negation to the outer scope means that we only do the negation on the final reduced row, rather than on the whole window.

zipline/pipeline/filters/filter.py
```diff
@@ -533,3 +535,27 @@ class StaticAssets(StaticSids):
...
         sids = frozenset(asset.sid for asset in assets)
         return super(StaticAssets, cls).__new__(cls, sids)
+
+
+class AllPresent(CustomFilter, SingleInputMixin, StandardOutputs):
+    """Pipeline filter indicating input term has data for a given window.
+    """
+
+    def _validate(self):
+
```
ssanderson: minor style note: I'd probably remove the blank line here.

zipline/pipeline/filters/filter.py
ahgnaw: 👍

PR #1971, opened 2017-10-02
DOC: Distinct warnings for benchmark and treasury fetchers

JoaoAparicio:
The core of data/loader.py is the function load_market_data which
in turns calls functions ensure_benchmark_data and
ensure_treasury_data. These two functions are similar. Their
logger.info and logger.exception messages are slightly different
but their logger.warn messages are the same.

Having two different functions outputing the same warning can
be needlessly confusing when debugging.

This commit makes these two warnings explicitely different.

### DIFF
--- a/zipline/data/loader.py
+++ b/zipline/data/loader.py
@@ -234,7 +234,7 @@ def ensure_benchmark_data(symbol, first_date, last_date, now, trading_day,
         logger.exception('Failed to cache the new benchmark returns')
         raise
     if not has_data_for_dates(data, first_date, last_date):
-        logger.warn("Still don't have expected data after redownload!")
+        logger.warn("Still don't have expected benchmark data after redownload!")
     return data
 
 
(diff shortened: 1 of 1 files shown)


JoaoAparicio:
No idea what failed, my changes were extremely minimal.

FreddieV4:
Hi @JoaoAparicio it looks like there was some style warnings coming from flake8:

image

You'll probably want to just make that log string into a multi-line string, which you can do with parentheses. I'd also probably add in the dates there. An example can be found here: https://github.com/JoaoAparicio/zipline/blob/b7bd52fdcdf1bf0c4bd7522b976324fa77a0103b/zipline/data/loader.py#L218

JoaoAparicio:
@FreddieV4 You were right, I had forgotten to run flake8.

Made the fixes to mentioned. In addition added dates to yet another treasury fetcher logger.info message.

FreddieV4:
LGTM 👍 Thanks!

Issue #1957, opened 2017-09-20
KeyError: 'the label [2000-01-03 00:00:00+00:00] is not in the [index]'


kerwinxu:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

* Operating System: (win 10`)
* Python Version: python35
* Python Bitness: `$ python -c 'import math, sys;print(int(math.log(sys.maxsize + 1, 2) + 1))'`
* How did you install Zipline: (conda)
* Python packages: `$ pip freeze` or `$ conda list`

Now that you know a little about me, let me tell you about the issue I am
having:

# Description of Issue

* What did you expect to happen?
* What happened instead?

Here is how you can reproduce this issue on your machine:

## Reproduction Steps

1.i install "conda install -n python35 -c Quantopian zipline"
2.zipline ingest
3.zipline run -f dual_moving_average.py --start 2011-1-1 --end 2012-1-1 -o dma.pickle
4.error:
[2017-09-20 02:40:15.276265] WARNING: Loader: Refusing to download new benchmark data because a download succeeded at 2017-09-20 02:19:52.057758+00:00.
Traceback (most recent call last):
  File "d:\Anaconda3\envs\python35\lib\site-packages\pandas\core\indexing.py", line 1395, in _has_valid_type
    error()
  File "d:\Anaconda3\envs\python35\lib\site-packages\pandas\core\indexing.py", line 1390, in error
    (key, self.obj._get_axis_name(axis)))
KeyError: 'the label [2000-01-03 00:00:00+00:00] is not in the [index]'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "d:\Anaconda3\envs\python35\Scripts\zipline-script.py", line 11, in <module>
    load_entry_point('zipline==1.1.1', 'console_scripts', 'zipline')()
  File "d:\Anaconda3\envs\python35\lib\site-packages\click\core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "d:\Anaconda3\envs\python35\lib\site-packages\click\core.py", line 697, in main
    rv = self.invoke(ctx)
  File "d:\Anaconda3\envs\python35\lib\site-packages\click\core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "d:\Anaconda3\envs\python35\lib\site-packages\click\core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #1973, opened 2017-10-04
ENH: --force-redownload optional parameter.

JoaoAparicio:
zipline run calls function load_market_data. This function has a
cooldown of 1 hour. If it downloaded data less than 1 hour ago it won't
download again.

This has an inconvenient side-effect: if zipline run is executed,
data is not cached, data is downloaded but for some reason data is
still not present after download (like what happens with issue #1965),
and if then the issue is fixed, zipline run can still not be executed until
the cooldown time has expired. This can then lead users to open
new issues which seem related to the original one, but are not (see
for example issue #1957 and discussion therein).

This commit adds an optional --force-redownload flag to zipline run.
The default is set to false.

### DIFF
--- a/zipline/__main__.py
+++ b/zipline/__main__.py
@@ -178,6 +178,13 @@ def ipython_only(option):
     default=False,
     help='Print the algorithm to stdout.',
 )
+@click.option(
+    '--force-redownload',
+    is_flag=True,
+    default=False,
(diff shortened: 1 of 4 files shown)


andycwang:
***@***.***邮箱联系我，谢谢！

vangus11:
> ***@***.***邮箱联系我，谢谢！

全部屏蔽了

PR #1974, opened 2017-10-04
ENH: Enable the ability to compute multiple pipelines

lianga888:


zipline/errors.py
```diff
@@ -639,6 +639,16 @@ class NoSuchPipeline(ZiplineError, KeyError):
     )
 
 
+class DuplicatePipelineName(ZiplineError):
+    """
+    Raised when a user tries to attach a pipeline with a name that already
+    exists for another attached pipeline.
+    """
+    msg = (
+        "Pipeline name '{name}' already exists for another pipeline."
```
ssanderson: You can also use `{name!r}` here instead of having quotes around the format unit.

ssanderson: Might be helpful to include a suggestion of how to fix the issue here, e.g. "Tried to attach a pipeline named {name}, but there was already an attached pipeline with that name. Please use a different name for this pipeline."

tests/pipeline/test_pipeline_algo.py
```diff
@@ -344,6 +349,72 @@ def handle_data(context, data):
...
+                # Assets should appear iff they exist today and yesterday.
+                exists_today = self.exists(date, asset)
+                existed_yesterday = self.exists(date - self.trading_day, asset)
+                if exists_today and existed_yesterday:
+                    self.assertEqual(
+                        closes.loc[asset, 'close'],
+                        self.expected_close(date, asset)
+                    )
+                    self.assertEqual(
+                        volumes.loc[asset, 'close'],
```
ssanderson: I think this column name is a typo.  I'm surprised that this test currently passes.

tests/pipeline/test_pipeline_algo.py
```diff
@@ -344,6 +349,72 @@ def handle_data(context, data):
...
+                        self.expected_close(date, asset)
```
ssanderson: Should this be `expected_volume`?

tests/pipeline/test_pipeline_algo.py
```diff
@@ -181,6 +183,9 @@ def expected_close(self, date, asset):
             lookup = self.adj_closes
         return lookup.loc[date, asset]
 
+    def expected_volume(self, date, asset):
```
ssanderson: I would expect this function to have to account for the split in the test data used by this test.

lianga888:
@ssanderson this PR is ready for another pass, at your convenience

zipline/utils/cache.py
```diff
@@ -143,8 +174,14 @@ def get(self, key, dt):
         try:
             return self._cache[key].unwrap(dt)
         except Expired:
-            del self._cache[key]
-            raise KeyError(key)
+            # We can't handle safely handle the exception in this block
```
ssanderson: If we're going to re-raise an exception, then we don't need to explicitly clear the exception in the cleanup function.  The important thing is just that by the time we exit this function, the exception referencing the cached object is no longer in `sys.exc_info`.

zipline/utils/cache.py
```diff
@@ -119,7 +148,7 @@ def __init__(self, cache=None):
         else:
             self._cache = {}
 
-    def get(self, key, dt):
+    def get(self, key, dt, cleanup=basic_cleanup):
```
ssanderson: I would probably make this a parameter to `ExpiringCache` instead of to `get`.

zipline/utils/cache.py
```diff
@@ -86,6 +88,33 @@ def _unsafe_get_value(self):
...
+    # 1. self._pipeline_cache holds a reference.
+    # 2. The dataframe itself holds a reference via cached .iloc/.loc
+    #    accessors.
+    # 3. The traceback held in sys.exc_info includes stack frames in
+    #    which self._pipeline_cache is a local variable.
+
+    # We remove the above sources of references in reverse order:
+
+    # 3. Clear the traceback.  This is no-op in Python 3.
+    exc_clear()
```
ssanderson: I don't think this is necessary anymore.  See note below.

zipline/utils/cache.py
```diff
@@ -86,6 +88,33 @@ def _unsafe_get_value(self):
...
+def basic_cleanup(cache, key):
+    del cache[key]
+
+
+def dataframe_gc_cleanup(cache, key):
+    # Try to deterministically garbage collect the previous result by
+    # removing any references to it. There are at least three sources
+    # of references:
+
+    # 1. self._pipeline_cache holds a reference.
```
ssanderson: This comment is referencing variables from TradingAlgorithm, which doesn't really make sense if this is a standalone function.

zipline/utils/cache.py
```diff
@@ -86,6 +88,33 @@ def _unsafe_get_value(self):
...
+    # We remove the above sources of references in reverse order:
+
+    # 3. Clear the traceback.  This is no-op in Python 3.
+    exc_clear()
+
+    # 2. Clear the .loc/.iloc caches.
+    clear_dataframe_indexer_caches(cache[key]._unsafe_get_value())
+
+    # 1. Clear the reference to the `CachedObject`
+    del cache[key]
```
ssanderson: I don't think the cleanup function should have to be responsible for clearing the cache key.  That should happen unconditionally no matter what.

I think the right signature for the cleanup function is to just take the value wrapped in the `CachedObject`.

ssanderson:
LGTM

PR #1980, opened 2017-10-07
ENH: Load futures from bundles.

JoaoAparicio:
Currently only equities can be loaded from bundles. There are at least two
reasons for this. First, on write `zipline/data/bundles/core.py` ingest()
does not distinguish between equities and futures; instead all ingestions
are made into folders `daily_equities.bcolz` and `minute_equities.bcolz`.
Second, on read only equities readers are being passed into DataPortal.

This commit addresses the second issue.

This works well for data bundles containing futures exclusively. It does
not fix the bigger issue of supporting bundles which contain both equities
and futures.

### DIFF
--- a/zipline/data/bundles/core.py
+++ b/zipline/data/bundles/core.py
@@ -145,6 +145,7 @@ RegisteredBundle = namedtuple(
 BundleData = namedtuple(
     'BundleData',
     'asset_finder equity_minute_bar_reader equity_daily_bar_reader '
+    'future_minute_bar_reader future_daily_bar_reader '
     'adjustment_reader',
 )
 
@@ -527,6 +528,12 @@ def _make_bundle_core():
                 minute_equity_path(name, timestr, environ=environ),
             ),
             equity_daily_bar_reader=BcolzDailyBarReader(
+                daily_equity_path(name, timestr, environ=environ),
+            ),
+            future_minute_bar_reader=BcolzMinuteBarReader(
+                minute_equity_path(name, timestr, environ=environ),
+            ),
+            future_daily_bar_reader=BcolzDailyBarReader(
                 daily_equity_path(name, timestr, environ=environ),
             ),
             adjustment_reader=SQLiteAdjustmentReader(
--- a/zipline/data/bundles/core.py
+++ b/zipline/data/bundles/core.py
@@ -54,6 +54,20 @@ def daily_equity_path(bundle_name, timestr, environ=None):
     )
 
 
+def minute_future_path(bundle_name, timestr, environ=None):
+    return pth.data_path(
+        minute_equity_relative(bundle_name, timestr, environ),
+        environ=environ,
+    )
+
+
+def daily_future_path(bundle_name, timestr, environ=None):
+    return pth.data_path(
+        daily_equity_relative(bundle_name, timestr, environ),
+        environ=environ,
+    )
+
+
 def adjustment_db_path(bundle_name, timestr, environ=None):
     return pth.data_path(
         adjustment_db_relative(bundle_name, timestr, environ),
@@ -82,6 +96,14 @@ def daily_equity_relative(bundle_name, timestr, environ=None):
 
 def minute_equity_relative(bundle_name, timestr, environ=None):
     return bundle_name, timestr, 'minute_equities.bcolz'
+
+
+def daily_future_relative(bundle_name, timestr, environ=None):
+    return bundle_name, timestr, 'daily_futures.bcolz'
+
+
+def minute_future_relative(bundle_name, timestr, environ=None):
+    return bundle_name, timestr, 'minute_futures.bcolz'
 
 
 def asset_db_relative(bundle_name, timestr, environ=None, db_version=None):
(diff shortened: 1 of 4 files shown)


yankees714:
Hey @JoaoAparicio, thanks, this looks great! Would you be willing to tackle the first issue as well to make this a complete solution?

I believe the main fix would be to update the ingest step [here](https://github.com/quantopian/zipline/blob/master/zipline/data/bundles/core.py#L338) to have futures analogs for the directories and writers created for equities, as well as updating the ingest functions for the existing bundles, (e.g. https://github.com/quantopian/zipline/blob/master/zipline/data/bundles/quandl.py#L280-L281).

JoaoAparicio:
@yankees714 Please see https://github.com/quantopian/zipline/pull/1982

JoaoAparicio:
@yankees714 Could you clarify how you expect this to be done? You linked to the `ingest` function in `core.py`. Now look at [this](https://github.com/quantopian/zipline/blob/master/zipline/data/bundles/core.py#L440) line, where this function actually calls `bundle.ingest`. Two writers (`minute_bar_writer` and `daily_bar_writer`) are passed in that call. So, should we be passing two writers per asset class?

I see two options.

A) If for now we assume that a bundle contains only one asset class (either equities or futures) then the current commit works fine as it is, and the folder name where the data is written to is irrelevant, and I say we merge this as it is.

B) On the other hand, for a bundle to handle multiple asset classes then we have to modify that function call to pass 4 writers. However I'm expecting that a lot more code would have to be changed elsewhere for the backtester to support multiple asset classes simultaneously. Consider that even in quantopian you can't do both equities and futures simultaneously (when backtesting you have to either pick equities calendar, or futures calendar).

yankees714:
@JoaoAparicio The backtester supports both equities and futures in the same backtest – the calendar specifies when you can trade, but doesn't impact what you can trade. On Quantopian, we disallow trading futures on the equities calendar since there's not really a good reason to do it, but zipline shouldn't have an issue.

The part of this current solution that concerns me is that we're feeding the same data into both the futures and equities readers ([here](https://github.com/quantopian/zipline/pull/1980/files#diff-244661a28cbf53d7bbc167793fe7d138R527)), which I think could cause confusion.

I wouldn't expect B) to be a huge chunk of work, just, as you say, expanding the ingest step to have 2 writers per asset class. Think you'd be up for taking a look at that?

JoaoAparicio:
@yankees714 Yes, I'll do it.

JoaoAparicio:
@yankees714 I think I've addressed everything we talked about. The only thing remaining is getting the tests to pass. Let me know if everything is OK before I start working on the tests.

zipline/data/bundles/core.py
```diff
@@ -54,6 +54,20 @@ def daily_equity_path(bundle_name, timestr, environ=None):
     )
 
 
+def minute_future_path(bundle_name, timestr, environ=None):
+    return pth.data_path(
+        minute_equity_relative(bundle_name, timestr, environ),
```
yankees714: Should use `minute_future_relative`

zipline/data/bundles/core.py
```diff
@@ -54,6 +54,20 @@ def daily_equity_path(bundle_name, timestr, environ=None):
...
+        environ=environ,
+    )
+
+
+def daily_future_path(bundle_name, timestr, environ=None):
+    return pth.data_path(
+        daily_equity_relative(bundle_name, timestr, environ),
```
yankees714: Should use `daily_future_relative`

JoaoAparicio:
@yankees714 Fixed the typos you mentioned and rebased-out spurious commits. Gonna work on the tests now.

JoaoAparicio:
Modified `tests.data` so that it's consistent with ingest() taking 2 writers per asset class (total 4 writers).

Next need to modify `tests.test_example` to do the same.

JoaoAparicio:
@yankees714 Should a bundle that uses equities data but not futures data create empty futures folders?

yankees714:
I think that sounds fine, that's what you have now, correct?

JoaoAparicio:
No. `data/bundles/core.py` and `quantopian-quandl` have futures writers on the ingest function, but for some reason the `quantopian-quandl` bundle doesn't create empty futures folders. I expected it to, and apparently so does the `test.test_example` which looks at what writers the bundle has to know what folders to expect and is currently failing.

I think this is some idiosyncrasy specific to the `quantopian-quandl` bundle, the reason for which I haven't been able to identify so far. In comparison, my own futures-only bundle (that I also shared in a different PR) has no equities, but automatically zipline knows that it should create empty equities folders.

yankees714:
I think the issue here is not that the directories don't exist, but that [here](https://github.com/quantopian/zipline/pull/1980/files#diff-244661a28cbf53d7bbc167793fe7d138R576) we're attempting to open bcolz file with futures pricing data, and _that_ file doesn't exist. That makes sense, as in the past, the example have only ingested equities data.

Want to try adding a check to `zipline.data.bundles.core.load` that the futures minute/daily files exist before trying to construct readers around them? If the files don't exist, I think we can just set the reader to `None`.

ernestoeperez88:
Closing since it looks like this PR has gone stale. Please feel free to reopen if this is something you are still interested in. To be completely transparent, improving futures support is currently not in our short term roadmap.

PR #1982, opened 2017-10-12
ENH: Create futures bundle from CSV

JoaoAparicio:
For a user who has access to their own futures data it can be difficult
to figure out how to use that data with zipline. That is because currently
there is no way for a user to ingest futures CSV data into a bundle.

This commit provides that.

This bundle does NOT know how to ingest equities and futures data together.
It takes care of futures data only. In particular it uses the equities binary
writer, but correctly writes futures metadata into the bundle database. The
end result is the desired: a user with futures CSV data will be able to
backtest their futures strategies on zipline.

### DIFF
--- a/zipline/data/bundles/futurescsv.py
+++ b/zipline/data/bundles/futurescsv.py
@@ -1,0 +1,81 @@
+import pandas as pd
+import os
+
+# ASSUMPTIONS
+# This ingest function makes some assumptions.
+# The user's home folder should have have a folder named "futures".
+# Inside that folder there must be one file named meta.csv.
+# The file meta.csv must be a csv with the following columns: symbol,
+# root_symbol, notice_date, expiration_date, tick_size, multiplier, and
+# auto_close_date.
+# The "futures" folder must in addition contain a bunch of .csv files
+# containing actual price data. The data files must contain the following
+# columns: time, open_price, high, low, close_price, volume.
+# The symbol column in the meta.csv file will be used to generate the
+# names of the data files (e.g. symbol CLG16 will be expected to be in file
+# CLG16.csv).
+# The entire meta.csv will be read as a dataframe, and the index will be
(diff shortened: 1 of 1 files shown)


bartosh:
Would it make sense to add this functionality to CSVdir bundle #1860 ?

JoaoAparicio:
@bartosh I don't know. Personally my focus is currently on futures, so I will keep a futures-only bundle for myself. If I want to ingest new futures I don't want to every time be downloading equities which I won't be using anyway.

But more generally, if you want to have bundles with both equities and futures we need [this](https://github.com/quantopian/zipline/pull/1980/commits/3f4841b3343ffd43d9ae470ae10401813a978b14) to pass first. And if it does, both your csvdir and my futurescsv will have to be updated. So maybe we should focus on merge that first. From what I can tell the only thing missing is updating the tests.

What do you think?

bartosh:
@JoaoAparicio > `If I want to ingest new futures I don't want to every time be downloading equities which I won't be using anyway.`
I thought about making the bundle flexible enough to support ingesting of only equities, only futures or both. Not insisting on this though. Let's wait and see how it goes.

JoaoAparicio:
@bartosh I just want to understand whether you're not we're duplicating work. From what I can tell what your code does is allow the backtests to read data directly from CSV. Is that correct? That's not what my code does. My code uses the data bundle ingest function to read from CSV and write binaries in zipline's bundle format. Am I missing something?

bartosh:
@JoaoAparicio As far as I can see the functionality of my and yours code is the same - we both implemented ingesting csv price data into the bundle.

freddiev4:
Hi @JoaoAparicio thanks for opening this PR. I think we're going to close this in favor of both #1980 and #1860, as the machinery you'd need will already be there and you can build on top of the csvdir bundle 🙂

PR #1984, opened 2017-10-12
ENH: Retrieve session at offset.

kglowinski:
Adds a functions which allows a session label to be passed it alongside an offset, and retrieves the session label at that offset.

### DIFF
--- a/zipline/utils/calendars/trading_calendar.py
+++ b/zipline/utils/calendars/trading_calendar.py
@@ -333,7 +333,7 @@ class TradingCalendar(with_metaclass(ABCMeta)):
         Returns
         -------
         pd.Timestamp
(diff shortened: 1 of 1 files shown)


zipline/utils/calendars/trading_calendar.py
```diff
@@ -512,6 +512,34 @@ def sessions_in_range(self, start_session_label, end_session_label):
             )
         ]
 
+    def get_session_indices(self, session_label, offset):
```
llllllllll: why does this need to take an offset. To me, a more natural API is `get_session_index`. You can add the offset to that yourself.

llllllllll:
Could we get a test for this?

PR #1986, opened 2017-10-14
Mark slow tests

ssanderson:
Adds a marker that makes it easier to skip slow tests.  Slow tests are still run by default, but can be skipped by passing `-a '!slow'` to nosetests.

Currently the skipped tests take a total of around 30-40s, out of 280s
on my machine. Not sure if that's valuable enough to bother...


llllllllll:
I like this, could we have one for really slow which isn't run by default like the label array test?

PR #1989, opened 2017-10-17
improve error message when timestamp before asof_date

llllllllll:
I made this change when debugging but it is probably worth keeping.


ssanderson:
LGTM

PR #1990, opened 2017-10-17
ENH: Update quandl_bundle to use Quandl API v3.

ernestoeperez88:


zipline/data/bundles/quandl.py
```diff
@@ -13,265 +12,140 @@
...
-            symbol=symbol,
-            query=urlencode(query_params),
-        )
-    )
+    symbols = data['symbol'].unique()
+    asset_metadata = {}
 
+    asset_metadata['symbol'] = symbols
+    data.set_index('symbol', inplace=True)
+    asset_metadata['start_date'] = \
```
prsutherland: This might be something that can be replaced a `groupby` on a DataFrame, something like:

```
asset_metadata = (data.ix[:, ['symbol', 'date']]
           .groupby(by='symbol',)
           .agg({'date': [np.min, np.max]})
           .reset_index())
asset_metadata['start_date'] = asset_metadata.date.amin
asset_metadata['end_date'] = asset_metadata.date.amax
del asset_metadata['date']
asset_metadata.columns = asset_metadata.columns.get_level_values(0)
```

tests/resources/quandl_samples/rebuild_samples.py
```diff
@@ -2,62 +2,77 @@
...
-        url = format_wiki_url(
-            api_key=None,
-            symbol=sym,
-            start_date=start_date,
-            end_date=end_date,
-        )
-        print('Fetching from %s' % url)
-        response = requests.get(url)
-        response.raise_for_status()
+    api_key = 'Replace with Quandl API key.'
```
prsutherland: Might be worth documenting or linking to how to get an API key. On a less important note, parameterizing this so that the user can pass the API key from the commandline would be helpful.

tests/resources/quandl_samples/rebuild_samples.py
```diff
@@ -2,62 +2,77 @@
...
-from operator import methodcaller
 
-import pandas as pd
 import requests
+from io import BytesIO
+from zipfile import ZipFile
+from six.moves.urllib.parse import urlencode
+from zipline.testing import test_resource_path, write_compressed
 
+QUANDL_DATA_URL = (
```
prsutherland: Should we just import this from zipline/data/bundles/quandl.py?

tests/data/bundles/test_quandl.py
```diff
@@ -184,46 +173,42 @@ def expected_dividend_adjustment(idx, symbol):
...
-                    'metadata-%d.csv.gz' % n,
+                    'QUANDL_ARCHIVE.zip'), 'rb') as quandl_response:
+
+            self.responses.add(
+                self.responses.GET,
+                'https://file_url.mock.quandl',
+                body=quandl_response.read(),
+                content_type='application/zip',
+                status=200
+            )
```
richafrank: FWIW I don't think you need to indent all the rest of this, once you've read the `quandl_response`. You can close the file here, and leave the rest of the indentation unchanged.

zipline/data/bundles/quandl.py
```diff
@@ -391,6 +297,10 @@ def quantopian_quandl_bundle(environ,
                              cache,
                              show_progress,
                              output_dir):
+    log.warn(
+        'quantopian-quandl has been deprecated and '
```
richafrank: Should we use the `warnings` module for this?

richafrank: Also, if this is now deprecated, probably we should change the default `ingest` bundle to be the `quandl` bundle itself (in `zipline/__main__.py`.

tests/data/bundles/test_quandl.py
```diff
@@ -2,78 +2,60 @@
...
 from zipline.lib.adjustment import Float64Multiply
 from zipline.testing import (
     test_resource_path,
     tmp_dir,
     patch_read_csv,
 )
-from zipline.testing.fixtures import ZiplineTestCase
+from zipline.testing.fixtures import (
+    ZiplineTestCase,
+    WithResponses
```
richafrank: Not a blocker, but just to give a style recommendation, you might include a trailing comma on the final item in a list (such as multiple imports) so that the next person to add an element to the list doesn't see a change on this line as well.

zipline/data/bundles/quandl.py
```diff
@@ -13,265 +13,165 @@
...
-                except Exception:
-                    pass
-            else:
-                raise ValueError(
-                    'Failed to download metadata page %d after %d'
-                    ' attempts.' % (page_number, retries),
-                )
-
-            cache[key] = raw
+warnings.simplefilter('once', DeprecationWarning)
```
richafrank: Does this mean that all DeprecationWarnings from any module will only print once? I don't think we want to decide that for users - I think we probably don't want to set up any filters for any modules. What do you think?

zipline/data/bundles/quandl.py
```diff
@@ -13,265 +13,165 @@
...
+        'sid': pd.factorize(data.symbol)[0]
+    }).dropna()
+    df['record_date'] = df['declared_date'] = df['pay_date'] = pd.NaT
+    return df
+
+
+def parse_pricing_and_vol(data,
+                          sessions,
+                          symbol_map):
+    for asset_id, symbol in symbol_map.iteritems():
```
richafrank: I think `iteritems` is gone on python 3.

richafrank:
Oh, also could you add [commit prefixes](http://www.zipline.io/development-guidelines.html#commit-messages) to your commit messages?

tests/data/bundles/test_quandl.py
ernestoeperez88: Thanks for the tip. I had noticed this in other places, but didn't know what the purpose of it was.

zipline/data/bundles/quandl.py
ernestoeperez88: I think I misunderstood the correct use of the module on my first pass through the documentation. I found a good explanation [here](http://www.lesinskis.com/python_deprecation_tutorial.html).

richafrank:
LGTM!

VitalyErmilov:
I tried to use the new version of quandl.py, unfortunаtely, it went out of memory on 4GB ram amazon instance when reading df from downloaded cvs file. On 8GB instance everything worked ok. Cаn it be optimized? Mаybe by loаding csv into df in pаrts.

richafrank:
Thanks @VitalyErmilov - I opened #2022 to track it.

westurner:
Dask provides a similar API to Pandas and supports datasets larger than can fit into RAM.

From https://dask.pydata.org/en/latest/dataframe.html :

> A Dask DataFrame is a large parallel dataframe composed of many smaller Pandas dataframes, split along the index. These pandas dataframes may live on disk for larger-than-memory computing on a single machine, or on many different machines in a cluster.

- Docs: https://dask.pydata.org/en/latest/
- Src: https://github.com/dask/dask
- Pypi: https://pypi.org/project/dask/
- CondaPkg: ``dask``

ernestoeperez88:
@VitalyErmilov can you please test the latest master branch on your 4GB instance? and maybe post your findings in https://github.com/quantopian/zipline/issues/2022

PR #1991, opened 2017-10-18
ENH: Add suggestions when creating CustomTerms with incorrect dtypes

llllllllll:
Today I tried to make a CustomFactor with dtype=object and had no idea why that was failing.


dmichalowicz:
Nice!

Issue #1995, opened 2017-10-25
Error in graph.py when running momentum_pipeline.py example algorithm.  


VitalyErmilov:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

* Operating System: Ubuntu 16.04
* Python Version: 2.7 
* How did you install Zipline:  pip 
* List of packages:
alembic (0.9.6)
backports.functools-lru-cache (1.4)
bcolz (0.12.1)
Bottleneck (1.2.1)
certifi (2017.7.27.1)
chardet (3.0.4)
click (6.7)
contextlib2 (0.5.5)
cycler (0.10.0)
cyordereddict (1.0.0)
Cython (0.27.2)
decorator (4.1.2)
empyrical (0.3.2)
idna (2.6)
intervaltree (2.1.0)
Logbook (1.1.0)
lru-dict (1.1.6)
Mako (1.0.7)
MarkupSafe (1.0)
matplotlib (2.1.0)
multipledispatch (0.4.9)
networkx (2.0)
numexpr (2.6.4)
numpy (1.13.3)
pandas (0.18.1)
pandas-datareader (0.5.0)
patsy (0.4.1)
pip (9.0.1)
pkg-resources (0.0.0)
pyparsing (2.2.0)
python-dateutil (2.6.1)
python-editor (1.0.3)
pytz (2017.2)
requests (2.18.4)
requests-file (1.4.2)
requests-ftp (0.3.1)
scipy (0.19.1)
setuptools (36.6.0)
six (1.11.0)
sortedcontainers (1.5.7)
SQLAlchemy (1.1.14)
statsmodels (0.8.0)
subprocess32 (3.2.7)
tables (3.4.2)
toolz (0.8.2)
urllib3 (1.22)
v (3.0.0)
wheel (0.30.0)
zipline (1.1.1)

Now that you know a little about me, let me tell you about the issue I am
having:

When I run the momentum_pipeline.py from example algorithms by using the following command: 

**zipline run -f momentum_pipeline.py --start 2017-02-02 --end 2017-10-20 -o results_mom.pickle**

it results in the following error:

Traceback (most recent call last):
  File "/home/ubuntu/venv-zipline/bin/zipline", line 11, in <module>
    load_entry_point('zipline==1.1.1', 'console_scripts', 'zipline')()
  File "/home/ubuntu/venv-zipline/local/lib/python2.7/site-packages/click/core.py", line 722, in __call__
[...]

PR #1996, opened 2017-10-26
MAINT: networkx 2 changes the behavior of out_degree

richafrank:
For https://github.com/quantopian/zipline/issues/1995


ssanderson:
This looks like the right change for now.  I tried just updating the zipline logic that interacts with networkx (which is pretty minimal), but it looks like there's some upstream breakage in odo with the latest networkx as well, so :+1: .

ssanderson:
xref https://github.com/blaze/odo/issues/579 for the odo breakage.

richafrank:
Oh, thanks for looking - that's too bad re: odo.

PR #1997, opened 2017-10-26
ENH: Add default index names to pipeline results

ahgnaw:
Improve legibility of pipeline result manipulations by adding names to the multi-index.

### DIFF
--- a/zipline/pipeline/engine.py
+++ b/zipline/pipeline/engine.py
@@ -32,6 +32,11 @@ from zipline.utils.pandas_utils import categorical_df_concat
 from zipline.utils.sharedoc import copydoc
 
 
+DATE_INDEX_NAME = 'dates'
+SID_INDEX_NAME = 'sid'
+PIPELINE_INDEX_NAMES = (DATE_INDEX_NAME, SID_INDEX_NAME)
+
+
 class PipelineEngine(with_metaclass(ABCMeta)):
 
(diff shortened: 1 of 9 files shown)


zipline/pipeline/engine.py
```diff
@@ -32,6 +32,11 @@
 from zipline.utils.sharedoc import copydoc
 
 
+DATE_INDEX_NAME = 'dates'
```
ssanderson: It bothers me that `dates` is plural here but `sid` is singular.  It's probably not worth breaking all the consumers of this to make the names consistent though :(.

zipline/pipeline/engine.py
```diff
@@ -32,6 +32,11 @@
...
+SID_INDEX_NAME = 'sid'
+PIPELINE_INDEX_NAMES = (DATE_INDEX_NAME, SID_INDEX_NAME)
```
ssanderson: I'd probably put these in `zipline.pipeline.common` instead of in engine.

ahgnaw: Good point, moving them.

zipline/pipeline/engine.py
ahgnaw: I'll change it to `sids`. All of the downstream consumers currently change the name of the index from (None, None), to (`dates`, `sid`).

PR #2000, opened 2017-10-27
ENH: Fail better on AssetFinder(nonexistent_path).

ssanderson:
Make the AssetFinder fail with a clearer error if you construct and
AssetFinder pointing to a file path that doesnt exist.

The current implementation fails with:

```
InvalidRequestError: Could not reflect: requested table(s) not available in sqlite:///<filepath>
```

which makes it look like the error was that the db was corrupted, when
the actual error was just that the file didn't even exist. This is made
worse by the fact that sqlite3 will actually create an empty file when
you call `connect`, which further propagates the appearance of having
previously had a bad file.

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -19,6 +19,7 @@ Tests for the zipline.assets package
 from contextlib import contextmanager
 from datetime import timedelta
 from functools import partial
+import os
 import pickle
 import sys
 from types import GetSetDescriptorType
@@ -82,6 +83,7 @@ from zipline.testing.fixtures import (
     WithAssetFinder,
     ZiplineTestCase,
     WithTradingCalendars,
+    WithTmpDir,
 )
 from zipline.utils.range import range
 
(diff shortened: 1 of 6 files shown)


tests/test_assets.py
```diff
@@ -1585,3 +1587,18 @@ def test_fuzzy(self):
             results,
             [af.lookup_symbol(sym, dt, fuzzy=True) for sym in syms],
         )
+
+
+class TestAssetFinderPreprocessors(WithTmpDir, ZiplineTestCase):
+
+    def test_asset_finder_doesnt_silently_create_useless_empty_files(self):
+        nonexistent_path = self.tmpdir.getpath('nothing_here')
```
richafrank: What do you think about using a unique name instead of 'nothing_here' to prevent any conflict between tests in this class?

zipline/assets/asset_writer.py
```diff
@@ -358,7 +358,7 @@ class AssetDBWriter(object):
     """
     DEFAULT_CHUNK_SIZE = SQLITE_MAX_VARIABLE_NUMBER
 
-    @preprocess(engine=coerce_string_to_eng)
+    @preprocess(engine=coerce_string_to_eng(require_exists=False))
```
richafrank: Yup, this looks like the one place we don't want to require it exists. Do we ever append, as in, would we want to require that it does NOT exist?

ssanderson: I'm not sure about this offhand.  @yankees714 do you know if we ever do inplace updates internally?  My inclination is to say that even if we don't expect to update an asset db inplace, it's possible that another user might, so I don't feel compelled to disallow that by construction.

yankees714: I'm not aware of any place where we append, but I agree that it probably makes sense to leave this flexible in case others want to.

tests/test_assets.py
ssanderson: updated to prepend `self.id()`

richafrank: 👍

PR #2008, opened 2017-11-08
ENH: Add daily returns factor.

ssanderson:


tests/pipeline/test_factor.py
```diff
@@ -1283,3 +1287,24 @@ class F(Factor):
             f.to_workspace_value(pipeline_output, pd.Index([0, 1])),
             column_data,
         )
+
+
+class TestSpecialCases(WithEquityPricingPipelineEngine,
+                       ZiplineTestCase):
+
+    def check_equivalent_terms(self, terms):
```
llllllllll: should you just check that DailyReturns() is Return(window_length=2)?

llllllllll:
one comment about testing, otherwise lgtm

tests/pipeline/test_factor.py
ssanderson: The `is` won't be true, b/c `DailyReturns` is a subclass.

ssanderson: This would be an interesting use-case for graph rewrite rules if we ever add them.

PR #2009, opened 2017-11-08
Conda updates

richafrank:
Updated versions of conda tooling

Install newer miniconda - This fixes the failing appveyor build on master.

Removed python 3.4 builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ cache:
     - $HOME/.cache/.pip/
 
 before_install:
-  - if [ ${CONDA_ROOT_PYTHON_VERSION:0:1} == "2" ]; then wget https://repo.continuum.io/miniconda/Miniconda-3.7.0-Linux-x86_64.sh -O miniconda.sh; else wget https://repo.continuum.io/miniconda/Miniconda3-3.7.0-Linux-x86_64.sh -O miniconda.sh; fi
+  - if [ ${CONDA_ROOT_PYTHON_VERSION:0:1} == "2" ]; then wget https://repo.continuum.io/miniconda/Miniconda2-4.3.30-Linux-x86_64.sh -O miniconda.sh; else wget https://repo.continuum.io/miniconda/Miniconda3-4.3.30-Linux-x86_64.sh -O miniconda.sh; fi
   - chmod +x miniconda.sh
   - ./miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: false
 fast_finish: true
 python:
   - 2.7
-  - 3.4
   - 3.5
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,8 @@ before_install:
   - export PATH="$HOME/miniconda/bin:$PATH"
 install:
   - conda info -a
-  - conda install conda=4.3.25 conda-build=3.0.15 anaconda-client=1.6.3 --yes
+  - conda install conda=4.3.30 conda-build=3.0.28 anaconda-client=1.6.3 --yes -q
+  - conda list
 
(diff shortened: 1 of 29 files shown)


conda/parameterized/meta.yaml
```diff
@@ -0,0 +1,97 @@
+{% set name = "parameterized" %}
+{% set version = "0.6.1" %}
+{% set file_ext = "tar.gz" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "caf58e717097735de0d7e15386a46ffa5ce25bb6a13a43716a8854a8d34841e2" %}
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
```
freddiev4: Where does `lower` come from? Does that signify that when the packages are built then conda will read the `name` as lowercase?

ci/make_conda_packages.py
```diff
@@ -31,6 +31,7 @@ def main(env, do_upload):
                "--python", env['CONDA_PY'],
                "--numpy", env['CONDA_NPY'],
                "--skip-existing",
+               "--old-build-string",
```
freddiev4: 👍 for adding this arg

appveyor.yml
```diff
@@ -78,7 +68,8 @@ install:
   - sed -i "s/scipy==.*/scipy==%SCIPY_VERSION%/" etc/requirements.txt
 
   - conda info -a
-  - conda install conda=4.1.11 conda-build=1.21.11 anaconda-client=1.5.1 --yes -q
+  - conda install conda=4.3.30 conda-build=3.0.28 anaconda-client=1.6.3 --yes -q
```
freddiev4: Are there any many main differences/reasons for using these newer versions, aside from fixing the build?

.travis.yml
```diff
@@ -3,7 +3,6 @@ sudo: false
 fast_finish: true
 python:
   - 2.7
-  - 3.4
```
freddiev4: Should we still run tests for Py34?

conda/0_sortedcontainers/build.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
-
-$PYTHON setup.py install
-
-# Add more build steps here, if they are necessary.
-
-# See
-# http://docs.continuum.io/conda/build.html
-# for a list of environment variables that are set during the build process.
```
freddiev4: TIL you can move this into the `.yaml` files. That's cool!

conda/parameterized/meta.yaml
richafrank: > Where does `lower` come from? 

It's built in to the [jinja2](http://jinja.pocoo.org/docs/2.9/templates/#lower) interpreter used to handle the [templating](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#templating-with-jinja) here.

> Does that signify that when the packages are built then conda will read the `name` as lowercase?

Yea, the `name` variable (defined above statically as `"parameterized"`) will be lowered before being used as the name of the package. Given that static name, lowering is a noop here, but this is all autogenerated by `conda skeleton`.

conda/0_sortedcontainers/build.sh
richafrank: Yep, that's from the newer version of conda-build. I used `conda skeleton` to re-generate these recipes.

appveyor.yml
richafrank: There are definitely new features and bug fixes. Many are optional, like the new build strings and changes in how one pins dependencies (like the numpy C-API), which I haven't tried out. You noticed the consolidated recipes - I regenerated a bunch of ours.

The reason I updated in this PR was to fix our toolchain, in that the old miniconda could no longer work with the package index or update to newer conda versions.

.travis.yml
richafrank: I was seeing build failures from missing dependencies in py34, but we're no longer running anything on py34, so it's ok to remove.

freddiev4:
After reading through the responses, LGTM 👍

PR #2010, opened 2017-11-08
Updating Anaconda Tokens about to expire

taliarhodes:


richafrank:
Thanks @taliarhodes . Changes look good to me. Assuming no obvious issues in the PR build, I think the real test is CI being able to push to anaconda, which it'll attempt after merging to master.

Issue #1999, opened 2017-10-26
Pipeline factor.quantiles() method should set window_safe=True


luca-s:
Dear Zipline Maintainers,

I believe that factor.quantile(...) method and friends (quartiles, quintiles, etc.) should set window_safe=True.
I cannot think a reason why it shouldn't, what do you think?

Sincerely,
`$ whoami`

ssanderson:
hi @luca-s.  Sorry for the long delay responding here.  This seems like a reasonable change to me.

In general, the `window_safe` stuff is the part of the Pipeline API that I'm least happy with.  The idea behind it is that it's supposed to prevent you from making mistakes like repeatedly taking moving averages of split-adjusted quantities, but the line for what makes something window-safe or not isn't as cut-and-dried as I'd like, which has led me to wondering recently whether it's worth having the concept at all.

At any rate, given that we currently mark `rank` as window-safe, I don't see any reason to not do the same for `quantiles` and friends.

PR #2012, opened 2017-11-10
MAINT: Mark quantiles() and friends window_safe.

ssanderson:
This is at least as reasonable as marking `rank()` as window_safe, which
we allow.

xref: https://github.com/quantopian/zipline/issues/1999


h55nick:
Know this is old but any reason this isn't moving in?

PR #2013, opened 2017-11-10
move dict invert and remove unused or redundant functions

llllllllll:
nothing used `invert_unique`. anywhere where I would have wanted that I have just used a dict comprehension.


zipline/utils/control_flow.py
```diff
@@ -1,45 +0,0 @@
-"""
```
ssanderson: rip

ssanderson:
:+1: from me.

Issue #2016, opened 2017-11-16
Mac OSX 10.13 always get installing Zipline stuck 


burkaslarry:
Dear Zipline Maintainers,

I  need help on installation
1 Always stuck on 

  Using cached zipline-1.1.1.tar.gz 

pip installed as 8.1.0

burkaslarry:
<img width="1166" alt="screenshot 2017-11-16 22 47 54" src="https://user-images.githubusercontent.com/5709615/32897332-41dd2262-cb20-11e7-8b57-6bba5169bead.png">
<img width="1197" alt="screenshot 2017-11-16 22 48 00" src="https://user-images.githubusercontent.com/5709615/32897333-4214972e-cb20-11e7-9fa7-f50a31bf9d7f.png">

<img width="1158" alt="screenshot 2017-11-16 22 48 47" src="https://user-images.githubusercontent.com/5709615/32897364-581e8aac-cb20-11e7-807c-1b5091f814c5.png">



while   `pip install --verbose zipline`

richafrank:
Hi @burkaslarry , I've definitely seen slowness while setuptools is easy_installing setup dependencies. Maybe it's progressing but slowly? 

Either way, want to try `pip install numpy cython` first, and then try `pip install zipline` after? Then it can avoid the easy_install.

burkaslarry:
Usually how long does it take for setup.py to be finished ? I have tried using verbose, got repeating numpy  warnings 

```
Installing collected packages: zipline
  Running setup.py install for zipline ... /
```

burkaslarry:
like this one. repeatedly 

```
_DEPRECATED_API NPY_1_7_API_VERSION" [-W#warnings]
    #warning "Using deprecated NumPy API, disable it by " \
     ^
    1 warning generated.
    In file included from zipline/assets/_assets.c:526:
    In file included from /Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/numpy/core/include/numpy/arrayobject.h:4:
    In file included from /Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/numpy/core/include/numpy/ndarrayobject.h:18:
    In file included from /Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/numpy/core/include/numpy/ndarraytypes.h:1809:
[...]

PR #2020, opened 2017-11-21
Pip install setup docs

richafrank:
From https://github.com/quantopian/zipline/issues/2016#issuecomment-344957172

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -23,8 +23,17 @@ single tool to install Python and non-Python dependencies, or if you're already
 using `Anaconda <http://continuum.io/downloads>`_ as your Python distribution,
 you can skip to the :ref:`Installing with Conda <conda>` section.
 
+Since building Zipline's C extensions requires numpy's CPython C API as well as
+`Cython <http://cython.org/>`_, it's safest to pip install these before pip
+installing zipline itself:
+
+.. code-block:: bash
+
+   $ pip install numpy Cython
+
+
 Once you've installed the necessary additional dependencies (see below for
-your particular platform), you should be able to simply run
+your particular platform), you should be able to simply run:
 
 .. code-block:: bash
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -135,10 +135,10 @@ channel:
 
 Managing ``conda`` environments
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-It is recommended to install Zipline in an isolated ``conda`` environment. 
-Installing Zipline in ``conda`` environments will not interfere your default 
-Python deployment or site-packages, which will prevent any possible conflict 
-with your global libraries. For more information on ``conda`` environment, see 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -33,51 +33,6 @@
 import versioneer
 
 
-class LazyBuildExtCommandClass(dict):
```
ssanderson: Can we check for numpy/Cython being installed and fail with a clear error here if they're not?

ssanderson: We should remove the `setup_requires` for numpy/Cython as well, since they're the main thing that causes breakage.

richafrank: Hmm, as it currently is, `setup.py develop` is broken. Will need to dig into it if we want to remove even more.

richafrank: Ok, updated.

setup.py
```diff
@@ -33,94 +34,76 @@
...
-        'zipline.data._resample',
-        ['zipline/data/_resample.pyx']
-    ),
-    Extension(
-        'zipline.pipeline.loaders.blaze._core',
-        ['zipline/pipeline/loaders/blaze/_core.pyx'],
-        depends=['zipline/lib/adjustment.pxd'],
-    ),
-]
+    ext_modules = [
```
ssanderson: Wow, we have a lot of extensions.  Probably not for this PR, but I wonder if we could consolidate these somehow.  That might speed up some of our CI build times significantly I think.

richafrank: Yea, the `pip install -e .` of the builds takes 90+ seconds currently.

ssanderson: oof

docs/source/install.rst
```diff
@@ -23,8 +23,17 @@ single tool to install Python and non-Python dependencies, or if you're already
 using `Anaconda <http://continuum.io/downloads>`_ as your Python distribution,
 you can skip to the :ref:`Installing with Conda <conda>` section.
 
+Since building Zipline's C extensions requires numpy's CPython C API as well as
+`Cython <http://cython.org/>`_, pip install these before pip installing zipline
+itself:
+
+.. code-block:: bash
+
+   $ pip install numpy Cython
```
ssanderson: This will give people the latest version of numpy.  Do we support that?

richafrank: AFAIK we do. If we don't, then we should update setup.py's `REQ_UPPER_BOUNDS`.

freddiev4:
Hi @richafrank anything I can do to help get this branch merged?

richafrank:
> Hi @richafrank anything I can do to help get this branch merged?

I think we should get agreement with the team that the install workflow changes here are what we want, and that there aren't any broken scenarios.

ssanderson:
I'm :+1: on merging these changes as-is. The only thing we might want to add to this is an entry to the release notes telling people that they need to pre-install numpy now if they're using pip.

rapetter94:
@ssanderson @richafrank should we merge once we add requirement to install numpy to release notes (per Scott's comment^)?

PR #2021, opened 2017-11-21
PERF: Add a new factor that just computes beta.

ssanderson:
This is a little over 1100x faster (!) than using
`RollingLinearRegressionOfReturns` on my machine.

Profiling output for a 1-month pipeline using both terms for with a
90-day lookback:

```
Tue Nov 21 02:05:48 2017    pipebench/perf/betas.stats

         57724856 function calls (57689155 primitive calls) in 92.342 seconds

   Ordered by: cumulative time
   List reduced from 1212 to 3 due to restriction <'statistical.py'>

       21    0.612    0.029   95.461    4.546 statistical.py:194(compute)
   172201    0.407    0.000   94.843    0.001 statistical.py:201(regress)
       21    0.048    0.002    0.082    0.004 statistical.py:500(compute)
```

- The `compute` function that takes 95 seconds is `RollingLinearRetressionOfReturns`.
- The `compute` function that takes 0.082 seconds is the new implementation.

### DIFF
--- a/tests/pipeline/test_statistical.py
+++ b/tests/pipeline/test_statistical.py
@@ -1,6 +1,7 @@
 """
 Tests for statistical pipeline terms.
 """
+import numpy as np
 from numpy import (
     arange,
     full,
@@ -17,6 +18,8 @@ from pandas import (
 from pandas.util.testing import assert_frame_equal
 from scipy.stats import linregress, pearsonr, spearmanr
 
+from empyrical.stats import beta_aligned as empyrical_beta
+
 from zipline.assets import Equity
 from zipline.errors import IncompatibleTerms, NonExistentAssetInTimeFrame
 from zipline.pipeline import CustomFactor, Pipeline
@@ -30,6 +33,7 @@ from zipline.pipeline.factors import (
     RollingSpearmanOfReturns,
     SimpleBeta,
 )
+from zipline.pipeline.factors.statistical import vectorized_beta
 from zipline.pipeline.loaders.frame import DataFrameLoader
 from zipline.pipeline.sentinels import NotSpecified
 from zipline.testing import (
@@ -47,6 +51,7 @@ from zipline.testing.fixtures import (
 )
 from zipline.testing.predicates import assert_equal
 from zipline.utils.numpy_utils import (
+    as_column,
     bool_dtype,
     datetime64ns_dtype,
     float64_dtype,
@@ -820,3 +825,42 @@ class StatisticalMethodsTestCase(WithSeededRandomPipelineEngine,
                 columns=assets,
             )
             assert_frame_equal(output_result, expected_output_result)
+
+
+class VectorizedBetaTestCase(ZiplineTestCase):
+
+    def compare_with_empyrical(self, dependents, independent):
+        result = vectorized_beta(dependents, independent)
+        expected = np.array([
+            empyrical_beta(dependents[:, i].ravel(), independent.ravel())
+            for i in range(dependents.shape[1])
+        ])
+        assert_equal(result, expected, array_decimal=7)
+        return result
+
+    @parameter_space(seed=[1, 2, 3])
+    def test_vectorized_beta_matches_empyrical_beta_aligned(self, seed):
+        rand = np.random.RandomState(seed)
+        true_betas = np.array([-0.5, 0.0, 0.5, 1.0, 1.5])
+        independent = as_column(np.linspace(-5., 5., 20))
+        noise = as_column(rand.uniform(-.1, .1, 20))
+        dependents = 1.0 + true_betas * independent + noise
+
+        result = self.compare_with_empyrical(dependents, independent)
+        self.assertTrue((np.abs(result - true_betas) < 0.01).all())
+
+    @parameter_space(seed=[1, 2, 3])
+    def test_vectorized_beta_nan_handling_matches_empyrical(self, seed):
+        rand = np.random.RandomState(seed)
+        true_betas = np.array([-0.5, 0.0, 0.5, 1.0, 1.5])
+        independent = as_column(np.linspace(-5., 5., 20))
+        noise = as_column(rand.uniform(-.1, .1, 20))
+        dependents = 1.0 + true_betas * independent + noise
+
+        # Fill 20% of the input arrays with nans randomly.
+        dependents[rand.uniform(0, 1, dependents.shape) < 0.2] = nan
+        independent[rand.uniform(0, 1, independent.shape) < 0.2] = nan
+
+        result = self.compare_with_empyrical(dependents, independent)
+        self.assertTrue(not np.isnan(result).any())
+        self.assertTrue((np.abs(result - true_betas) < 0.01).all())
--- a/tests/pipeline/test_statistical.py
+++ b/tests/pipeline/test_statistical.py
@@ -331,6 +331,19 @@ class StatisticalBuiltInsTestCase(WithTradingEnvironment, ZiplineTestCase):
         )
         assert_equal(results['simple'], results['complex'], check_names=False)
 
+    def test_simple_beta_allowed_missing_calculation(self):
+        for percentage, expected in [(0.651, 65),
+                                     (0.659, 65),
+                                     (0.66, 66),
+                                     (0.0, 0),
+                                     (1.0, 100)]:
+            beta = SimpleBeta(
+                target=self.my_asset,
+                regression_length=100,
+                allowed_missing_percentage=percentage,
+            )
+            self.assertEqual(beta.params['allowed_missing_count'], expected)
+
     def test_correlation_and_regression_with_bad_asset(self):
         """
         Test that `RollingPearsonOfReturns`, `RollingSpearmanOfReturns` and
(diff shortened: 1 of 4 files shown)


ssanderson:
I think the only significant decision that needs to be made here is what we want to do if there are NaNs in the input.  The current implementation uses `nanmean` everywhere possible, which means it produces values more often than the existing `RollingLinearRegressionOfReturns`, (which internally calls scipy.stats.linregress, which is, as it turns out, not particularly efficient when called 172,000 times).

cc @Jstauth @jmccorriston @ahgnaw @twiecki for thoughts on how we want to handle missing data here.  Note that I'm not trying to port our downstream shrinkage beta (I think we should also do that, but the use-case here is to just provide the simplest and fastest reasonable implementation of beta, in particular with an eye toward use in portfolio optimization.)

zipline/pipeline/factors/statistical.py
```diff
@@ -148,11 +155,6 @@ class RollingLinearRegression(CustomFactor, SingleInputMixin):
         The factor/slice whose columns are the predictor/independent variable
         of each regression with `dependent`. If `independent` is a Factor,
         regressions are computed asset-wise.
-    independent : zipline.pipeline.Term with a numeric dtype
```
ssanderson: This is an unrelated change to the rest of the PR.  We had this documented twice.

ssanderson:
Hilarious performance note: in the current implementation more than half of the total time is spent constructing namedtuple types:

```
Function                                called...
                                            ncalls  tottime  cumtime
_stats_mstats_common.py:11(linregress)  ->  172201    3.918   15.788  _distn_infrastructure.py:1720(sf)
                                            172201   44.965   58.895  collections.py:305(namedtuple)
                                            344402    0.445    5.133  fromnumeric.py:2796(mean)
                                            172201    2.672   10.022  function_base.py:2298(cov)
                                            344402    0.209    0.272  numeric.py:414(asarray)
                                            172201    0.028    0.028  {len}
```

ahgnaw:
In `empyrical.beta_aligned()` we remove any date-pairs with `np.nan` in them. Here we might want to add a minimum number of non-nan date pairs required to calculate beta.

jmccorriston:
What is the current behavior of `RollingLinearRegressionOfReturns` with regards to nan values?

zipline/pipeline/factors/statistical.py
```diff
@@ -475,3 +477,43 @@ def __new__(cls,
             regression_length=regression_length,
             mask=mask,
         )
+
+
+class SimpleBeta(CustomFactor, StandardOutputs):
+    """
+    Factor producing the slope of a regression line between each asset's daily
+    returns the daily returns of a single "target" asset.
```
dmichalowicz: "...each asset's daily returns _and_ the daily returns of..." (missing "and")

ssanderson:
> What is the current behavior of RollingLinearRegressionOfReturns with regards to nan values?

Any asset with a NaN at any point in the lookback window ends up NaN.

ssanderson:
Pushed a change that makes the behavior here match empyrical exactly.  Handling NaNs imposes around a 2.5x performance penalty (from 0.9 seconds per year to 2.2 on my machine).  That's still a huge win compared to 90+ seconds **per month** with the old implementation.

**Before:**
```
        1    0.001    0.001    3.544    3.544 engine.py:445(compute_chunk)
        2    0.027    0.013    1.469    0.735 mixins.py:189(_compute)
      253    0.518    0.002    0.917    0.004 statistical.py:512(compute)
```

**After:**
```
        1    0.001    0.001    4.909    4.909 engine.py:445(compute_chunk)
        2    0.028    0.014    2.844    1.422 mixins.py:189(_compute)
      253    0.003    0.000    2.252    0.009 statistical.py:515(compute)
```

ssanderson:
Hrm, though, thinking about this a bit more, I think we either need to require a minimum number of data points if we're handling nans in this way (as @ahgnaw suggests above).  If we don't, we'll produce crazy values for regressions where we only have 2-3 data points.

jmccorriston:
Oh, that's not great. I like Ana's idea of a minimum number or minimum percentage of non-nan date pairs.

dmichalowicz:
@ssanderson Let me know if this needs another round of review

zipline/pipeline/factors/statistical.py
```diff
@@ -475,3 +480,142 @@ def __new__(cls,
...
+
+    def __repr__(self):
+        return "{}(window_length={}, allowed_missing={})".format(
+            type(self).__name__,
+            self.window_length,
+            self.params['allowed_missing'],
+        )
+
+
+def vectorized_beta(dependents, independent, allowed_missing, out=None):
```
ahgnaw: I couldn't find another place where `vectorized_beta` might be useful at the moment, however at a later date we might want to move it to `empyrical`.

tests/pipeline/test_statistical.py
```diff
@@ -802,3 +879,139 @@ def test_factor_regression_method_two_factors(self, regression_length):
...
+                expect_nan = num_nans[i] > allowed_missing
+                true_beta = true_betas[i]
+                if expect_nan:
+                    self.assertTrue(np.isnan(result))
+                else:
+                    self.assertTrue(np.abs(result - true_beta) < 0.01)
+
+    def test_allowed_missing_doesnt_double_count(self):
+        # Test that allowed_missing only counts a row as missing one
+        # observation if it's missing in both the dependent and independent
```
ahgnaw: should "both" be "either"? Dates for which the dependent or independent (or both) data is nan cannot be included in the beta calculation.

ssanderson:
@dmichalowicz yeah, I think this is ready for another pass if you can take another look.

zipline/pipeline/factors/statistical.py
ssanderson: I don't think empyrical would use the `vectorized` part of this anywhere right now, but I suspect that the direct calculation of covariance and variance here is faster than what we currently do in empyrical, because it avoids making copies of the data and it avoids calculating parts of the covariance matrix that we don't need.  I'm not super worried about porting this in the near term, because I don't think anything is currently bottlenecked on empyrical's beta calculations, but if that becomes a bottleneck this is probably a good place to look to speed things up.

tests/pipeline/test_statistical.py
ssanderson: I think this wording is correct; the thing that this test is checking is that if a row has a NaN in both the dependent and independent, then we only count that row as adding 1 toward the number of required missing values.

zipline/pipeline/factors/statistical.py
```diff
@@ -475,3 +480,146 @@ def __new__(cls,
...
+    # Cython.
+
+    # shape: (N, M)
+    independent = np.where(
+        isnan(dependents),
+        nan,
+        independent,
+    )
+
+    # Calculate beta as Cov(X, Y) / Cov(Y, Y).
```
dmichalowicz: I think you mean Cov(X, Y) / Cov(X, X)  (The calculation below is correct, just this comment is off)

zipline/pipeline/factors/statistical.py
```diff
@@ -475,3 +480,146 @@ def __new__(cls,
...
+    # shape: (M,)
+    covariances = nanmean(ind_residual * dep_residual, axis=0)
+
+    # We end up with different variances in each column here because each
+    # column may have a different subset of the data dropped due to missing
+    # data in the corresponding dependent column.
+    # shape: (M,)
+    independent_variances = nanmean(ind_residual ** 2, axis=0)
+
+    # shape: (M,)
```
dmichalowicz: Thanks for these shape comments!

tests/pipeline/test_statistical.py
```diff
@@ -802,3 +879,139 @@ def test_factor_regression_method_two_factors(self, regression_length):
...
+        noise = as_column(rand.uniform(-.1, .1, 30))
+        dependents = 1.0 + true_betas * independent + noise
+
+        # Fill 20% of the input arrays with nans randomly.
+        dependents[rand.uniform(0, 1, dependents.shape) < 0.2] = nan
+        independent[rand.uniform(0, 1, independent.shape) < 0.2] = nan
+
+        # Sanity check that we actually inserted some nans.
+        self.assertTrue(np.count_nonzero(np.isnan(dependents)) > 0)
+        self.assertTrue(np.count_nonzero(np.isnan(independent)) > 0)
```
dmichalowicz: It's a very small chance, but doesn't this mean this test could fail randomly?

tests/pipeline/test_statistical.py
```diff
@@ -802,3 +879,139 @@ def test_factor_regression_method_two_factors(self, regression_length):
...
+                     np.array([False, True, True, True, False]))
+
+        # With four allowed missing values, everything but the middle column
+        # should produce a value. The middle column will have 5 nans because
+        # the dependent nans have no overlap with the independent nans.
+        result5 = vectorized_beta(dependents, independent, allowed_missing=4)
+        assert_equal(np.isnan(result5),
+                     np.array([False, False, True, False, False]))
+
+        # With six allowed missing values, everything should produce a value.
```
dmichalowicz: With _five_ allowed...

zipline/pipeline/factors/statistical.py
```diff
@@ -475,3 +480,146 @@ def __new__(cls,
...
+    #
+    # PERF_TODO: We could probably avoid the space blowup by doing this in
+    # Cython.
+
+    # shape: (N, M)
+    independent = np.where(
+        isnan(dependents),
+        nan,
+        independent,
+    )
```
dmichalowicz: Should you do a similar operation on the `dependent` matrix? Otherwise it will be using rows in its residual calculation that the independent variable does not use.

tests/pipeline/test_statistical.py
ssanderson: We're seeding the rng, so this is still deterministic.

zipline/pipeline/factors/statistical.py
ssanderson: Centering dependent turns out not to matter.  See 6bacafab65f8278a3730592b334fdd6dda953dea for a lengthy description.  (We discussed this in person, but putting the link here for posterity.)

PR #2026, opened 2017-11-28
MAINT: Pass calendars to DataPortal

### LINKED ISSUE #2018: Error running zipline backtest with custom trading calendar
From @vonpupp in https://github.com/quantopian/zipline/pull/1800#issuecomment-344724240:

Hi @freddiev4 ,

I am new to zipline. I am trying to use a 24/7 calendar, so I am trying to use the feature of this PR already to do it in a proper way. I am using zipline installed via pip directly from github's egg (`pip install git+git://github.com/quantopian/zipline')

Unfortunately I am not able to use this feature properly, here is what I am doing:

I edited the `extensions.py` as follows:

```python
# A bunch of imports removed for the sake of clarity

class TwentyFourSevenCalendar(TradingCalendar):
    """
    Exchange calendar for 24/7 trading.

    Open Time: 12am, UTC
    Close Time: 11:59pm, UTC

    """
    @property
    def name(self):
        return "TWENTYFOURSEVEN"

    @property
    def tz(self):
        return timezone("UTC")

    @property
    def open_time(self):
        return time(0, 0)

    @property
    def close_time(self):
        return time(23,59)

    @lazyval
    def day(self):
        return CustomBusinessDay(
            weekmask='Mon Tue Wed Thu Fri Sat Sun',
        )

start_session = pd.Timestamp('2014-01-07', tz='utc')
end_session = pd.Timestamp('2017-11-13', tz='utc')
assets = ['BTC_USD']

register_calendar(
    'TWENTYFOURSEVEN',
    TwentyFourSevenCalendar(
        start=start_session,
        end=end_session
    )
)
register(
    'custom-csvdir-bundle',
    csvdir_equities(["daily"], '/home/av/repos/ziplinetest/csvdir'),
    calendar_name='TWENTYFOURSEVEN',
    start_session=start_session,
    end_session=end_session
)
```
As you can see I am using csvdir bundle which has recently been merged. The ingestion process seems to work fine (no errors and I also verified the directory and the sqlite files exist, so I assume the ingestion is ok). I conclude therefore that the 24/7 calendar works properly during ingestion.

When I try to execute with:
```shell
zipline run -f csvdir_example.py \
    -b custom-csvdir-bundle \
    --start 2016-1-1 \
    --end 2017-1-1 \
    --data-frequency daily \
    --trading-calendar TWENTYFOURSEVEN \
    --capital-base 100000 \
    -o csvdir_example.pickle
```
I get:

```
Error: Invalid value for "--trading-calendar": invalid choice: TWENTYFOURSEVEN. (choose from BMF, CFE, CME, ICE, LSE, NYSE, TSX, us_futures)
```

Could you please help me?

Thanks a lot.

freddiev4:
Following the posts in #2018, should fix that issue

When we enter the `DataPortal()` setup in `zipline/zipline/data/data_portal.py(296)`, and call

```
# Get the first trading minute
self._first_trading_minute, _ = (
    self.trading_calendar.open_and_close_for_session(
    self._first_trading_day
)
```

We error at this point, if we're using a custom `TradingCalendar` because we actually only pass the `NYSE` calendar to the `DataPortal()` in `zipline/utils/run_algo.py` , so I changed the line to pass the calendar received from the CLI:

```
--- a/zipline/utils/run_algo.py
+++ b/zipline/utils/run_algo.py
@@ -134,7 +134,8 @@ def _run(handle_data,
         first_trading_day =\
             bundle_data.equity_minute_bar_reader.first_trading_day
         data = DataPortal(
-            env.asset_finder, get_calendar("NYSE"),
+            env.asset_finder,
+            trading_calendar=trading_calendar,
             first_trading_day=first_trading_day,
             equity_minute_reader=bundle_data.equity_minute_bar_reader,
             equity_daily_reader=bundle_data.equity_daily_bar_reader,
```

### DIFF
--- a/zipline/__main__.py
+++ b/zipline/__main__.py
@@ -7,10 +7,7 @@ import pandas as pd
 from six import text_type
 
 from zipline.data import bundles as bundles_module
-from zipline.utils.calendars.calendar_utils import (
-    get_calendar,
-    default_calendar_names
-)
+from zipline.utils.calendars.calendar_utils import get_calendar
 from zipline.utils.compat import wraps
 from zipline.utils.cli import Date, Timestamp
 from zipline.utils.run_algo import _run, load_extensions
@@ -179,7 +176,6 @@ def ipython_only(option):
 @click.option(
     '--trading-calendar',
     metavar='TRADING-CALENDAR',
-    type=click.Choice(default_calendar_names),
     default='NYSE',
     help="The calendar you want to use e.g. LSE. NYSE is the default."
 )
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -54,7 +54,6 @@ from zipline.data.history_loader import (
     MinuteHistoryLoader,
 )
 from zipline.data.us_equity_pricing import NoDataOnDate
-
 from zipline.utils.math_utils import (
     nansum,
     nanmean,
@@ -155,6 +154,7 @@ class DataPortal(object):
                  daily_history_prefetch_length=_DEF_D_HIST_PREFETCH):
 
         self.trading_calendar = trading_calendar
+
         self.asset_finder = asset_finder
 
         self._adjustment_reader = adjustment_reader
--- a/zipline/utils/run_algo.py
+++ b/zipline/utils/run_algo.py
@@ -113,6 +113,9 @@ def _run(handle_data,
         else:
             click.echo(algotext)
 
+    if trading_calendar is None:
+        trading_calendar = get_calendar('NYSE')
+
     if bundle is not None:
         bundle_data = load(
             bundle,
@@ -134,7 +137,8 @@ def _run(handle_data,
         first_trading_day =\
             bundle_data.equity_minute_bar_reader.first_trading_day
         data = DataPortal(
-            env.asset_finder, get_calendar("NYSE"),
+            env.asset_finder,
+            trading_calendar=trading_calendar,
             first_trading_day=first_trading_day,
             equity_minute_reader=bundle_data.equity_minute_bar_reader,
             equity_daily_reader=bundle_data.equity_daily_bar_reader,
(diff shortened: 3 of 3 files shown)


zipline/data/data_portal.py
```diff
@@ -154,7 +154,11 @@ def __init__(self,
                  minute_history_prefetch_length=_DEF_M_HIST_PREFETCH,
                  daily_history_prefetch_length=_DEF_D_HIST_PREFETCH):
 
-        self.trading_calendar = trading_calendar
+        if trading_calendar is not None:
```
richafrank: It looks like we do this check in `run_algo._run` [as well](https://github.com/quantopian/zipline/pull/2026/files#diff-ec1e1e05a956fe81da0703490710af21L159). Since that's the only call site that now might pass a `None` trading_calendar, can we do the check just there instead of here too?

zipline/__main__.py
```diff
@@ -179,7 +176,6 @@ def _(*args, **kwargs):
 @click.option(
     '--trading-calendar',
     metavar='TRADING-CALENDAR',
-    type=click.Choice(default_calendar_names),
```
yankees714: I kind of like including all of the options here. Looking at `TradingCalendarDispatcher`, it seems like we're actually indirectly defining what we want here in the `has_calendar` method:
https://github.com/quantopian/zipline/blob/49a3a4a2785a4a75dc21f04389094681eb408579/zipline/utils/calendars/calendar_utils.py#L92-L100

What would you think of splitting out an `available_calendar_names` property that just returns the union of what we check against in `has_calendar`? Then we could use that in the definition of `has_calendar`, and also expose it for the CLI.

richafrank: Have we loaded all the user extensions by this point in time, e.g. at zipline import time?

yankees714: Oh good point, not necessarily 😞 

Probably worth just leaving this out then.

freddiev4:
@richafrank should be good to go now (?)

PR #2027, opened 2017-11-28
MAINT: Move _shift_dates to pipeline utils as shift_dates

yankees714:
So that it can be used by other loaders that need to shift the dates
of pipeline outputs.

### DIFF
--- a/zipline/pipeline/loaders/utils.py
+++ b/zipline/pipeline/loaders/utils.py
@@ -2,6 +2,7 @@ import datetime
 
 import numpy as np
 import pandas as pd
+from zipline.errors import NoFurtherDataError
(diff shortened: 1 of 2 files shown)


zipline/pipeline/loaders/utils.py
```diff
@@ -405,3 +406,48 @@ def ffill_across_cols(df, columns, name_map):
             df[column_name] = df[
                 column_name
             ].fillna(column.missing_value).astype(column.dtype)
+
+
+def shift_dates(dates, start_date, end_date, shift):
```
ssanderson: I might add a docstring explaining the expected usage of this, otherwise LGTM.

PR #2032, opened 2017-11-30
DEP: Remove old Yahoo code

freddiev4:
We don't use this code anywhere that I'm aware of (grepped through Zipline in emacs)

### DIFF
--- a/zipline/data/loader.py
+++ b/zipline/data/loader.py
@@ -13,13 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import os
-from collections import OrderedDict
 
 import logbook
 import pandas as pd
-from pandas_datareader.data import DataReader
-import pytz
-from six import iteritems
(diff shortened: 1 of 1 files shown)


zipline/data/loader.py
```diff
@@ -410,39 +410,6 @@ def _load_raw_yahoo_data(indexes=None, stocks=None, start=None, end=None):
...
-    :type stocks: list
-    :param start: Retrieve prices from start date on.
-    :type start: datetime
-    :param end: Retrieve prices until end date.
-    :type end: datetime
-    :param adjusted: Adjust the price for splits and dividends.
-    :type adjusted: bool
-
-    """
-    data = _load_raw_yahoo_data(indexes, stocks, start, end)
```
prsutherland: You might be able to ax `_load_raw_yahoo_data` too.

freddiev4: Oooh good call

PR #2031, opened 2017-11-30
ENH: Use IEX Trading data instead of pandas-datareader

freddiev4:
Got fed up with the breaking/removal of Google & Yahoo Finance APIs, and so since we've been using pandas-datareader to get data from those APIs I removed that dependency for when we want to get data for a specific benchmark (in this case `SPY`).

 I changed our data source to IEX which will give us up to 5 years worth of data. It has really great documentation and seems like its a lot more reliable.

Should fix:
- https://github.com/quantopian/zipline/issues/2029
- https://github.com/quantopian/zipline/issues/2024
- https://github.com/quantopian/zipline/issues/1965
- https://github.com/quantopian/zipline/issues/2002
- https://github.com/quantopian/zipline/issues/2036
- https://github.com/quantopian/zipline/issues/2030

And should be better than PR https://github.com/quantopian/zipline/pull/1972

PR #2044, opened 2017-12-06
MAINT: Sync and fill benchmarks through latest trading day

ChrisPappalardo:
It's analytically superior to have a trading system synced to the latest market close than to lag the market waiting for treasury benchmark data.

This PR removes a hard-wired 2-day lag and combines the timeseries of the market and treasury benchmarks before filling data forwards and backwards along the timeseries.  This PR could be improved by averaging the treasury curve across points before backfilling the timeseries.

See comments to [#2031](https://github.com/quantopian/zipline/pull/2031) for more information.

### DIFF
--- a/zipline/data/loader.py
+++ b/zipline/data/loader.py
@@ -139,21 +139,8 @@ def load_market_data(trading_day=None, trading_days=None, bm_symbol='SPY',
     first_date = trading_days[0]
     now = pd.Timestamp.utcnow()
 
-    # We expect to have benchmark and treasury data that's current up until
-    # **two** full trading days prior to the most recently completed trading
-    # day.
-    # Example:
-    # On Thu Oct 22 2015, the previous completed trading day is Wed Oct 21.
-    # However, data for Oct 21 doesn't become available until the early morning
-    # hours of Oct 22.  This means that there are times on the 22nd at which we
-    # cannot reasonably expect to have data for the 21st available.  To be
-    # conservative, we instead expect that at any time on the 22nd, we can
-    # download data for Tuesday the 20th, which is two full trading days prior
-    # to the date on which we're running a test.
-
-    # We'll attempt to download new data if the latest entry in our cache is
-    # before this date.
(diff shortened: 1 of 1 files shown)


zipline/data/loader.py
```diff
@@ -172,6 +159,12 @@ def load_market_data(trading_day=None, trading_days=None, bm_symbol='SPY',
         now,
         environ,
     )
+
+    # combine dt indices and reindex using ffill then bfill
+    all_dt = br.index.union(tc.index)
+    br = br.reindex(all_dt, method='ffill').fillna(method='bfill')
+    tc = tc.reindex(all_dt, method='ffill').fillna(method='bfill')
```
freddiev4: The tradeoff here that I'm concerned about is that we'll end up having benchmark returns for much further back than 5 years (which is the max amount we can get from IEX), except the returns of that benchmark will just be the same b/c we're `bfill`ing.

What is it that you're trying to get out of reindexing here?

ChrisPappalardo: I'm trying to sync the two time series.  

The Federal Reserve H15 report is not released on a timely basis with market closes, so there will always be missing data on the front-end of the treasury curve time series.  At the same time, there are dates missing from both time series relative to each other.  Joining the time series, re-indexing, and then filling the missing data solves both issues.

If the concern is filling data back past 5 years on the benchmark, you could add a line to drop all dates older than the most recent start of the two respective time series.

freddiev4: OK. That makes sense to me and sounds fair. Also just did some quick profiling and wasn't what I had thought it would be like.

Can you you `amend` your commit to use a [Commit Prefix](http://www.zipline.io/development-guidelines.html#commit-messages)? I think `MAINT:` would be appropriate here.

ChrisPappalardo: @freddiev4 this PR is good to go

freddiev4:
Hmm I thought I had merged this after seeing the email notification... 🤔 

Anyways awesome work @ChrisPappalardo. Thanks for the PR 😃

PR #2047, opened 2017-12-07
ENH: add fixed percent slippage model

vikram-narayan:
add `FixedBasisPointsSlippage` which models slippage as a fixed percentage of fill price

### DIFF
--- a/tests/finance/test_slippage.py
+++ b/tests/finance/test_slippage.py
@@ -1152,27 +1152,26 @@ class OrdersStopTestCase(WithSimParams,
 class FixedBasisPointsSlippageTestCase(WithCreateBarData,
                                        ZiplineTestCase):
 
-    START_DATE = pd.Timestamp('2006-01-05 14:31', tz='utc')
-    END_DATE = pd.Timestamp('2006-01-05 14:36', tz='utc')
+    START_DATE = pd.Timestamp('2006-01-05', tz='utc')
+    END_DATE = pd.Timestamp('2006-01-05', tz='utc')
 
     ASSET_FINDER_EQUITY_SIDS = (133,)
-    minutes = pd.DatetimeIndex(
-        start=START_DATE,
-        end=START_DATE,
-        freq='1min'
+
+    first_minute = (
+        pd.Timestamp('2006-01-05 9:31', tz='US/Eastern').tz_convert('UTC')
     )
 
     @classmethod
     def make_equity_minute_bar_data(cls):
         yield 133, pd.DataFrame(
             {
-                'open': [3.00],
+                'open': [2.9],
                 'high': [3.15],
                 'low': [2.85],
                 'close': [3.00],
                 'volume': [200],
             },
-            index=[cls.minutes[0]],
+            index=[cls.first_minute],
         )
 
     @classmethod
@@ -1181,12 +1180,28 @@ class FixedBasisPointsSlippageTestCase(WithCreateBarData,
         cls.ASSET133 = cls.asset_finder.retrieve_asset(133)
 
     @parameterized.expand([
+        # Volume limit of 10% on an order of 100 shares. Since the bar volume
+        # is 200, we should hit the limit and only fill 20 shares.
         ('5bps_over_vol_limit', 5, 0.1, 100, 3.0015, 20),
+        # Same as previous, but on the short side.
+        ('5bps_negative_over_vol_limit', 5, 0.1, -100, 2.9985, -20),
+        # Volume limit of 10% on an order of 10 shares. We should fill the full
+        # amount.
         ('5bps_under_vol_limit', 5, 0.1, 10, 3.0015, 10),
+        # Same as previous, but on the short side.
         ('5bps_negative_under_vol_limit', 5, 0.1, -10, 2.9985, -10),
+        # Change the basis points value.
+        ('10bps', 10, 0.1, 100, 3.003, 20),
+        # Change the volume limit points value.
+        ('20pct_volume_limit', 5, 0.2, 100, 3.0015, 40),
     ])
-    def test_fixed_bps_slippage(self, name, basis_points, volume_limit,
-                                order_amount, expected_price, expected_amount):
+    def test_fixed_bps_slippage(self,
+                                name,
+                                basis_points,
+                                volume_limit,
+                                order_amount,
+                                expected_price,
+                                expected_amount):
 
         slippage_model = FixedBasisPointsSlippage(basis_points=basis_points,
                                                   volume_limit=volume_limit)
@@ -1201,7 +1216,7 @@ class FixedBasisPointsSlippageTestCase(WithCreateBarData,
         ]
 
         bar_data = self.create_bardata(
-            simulation_dt_func=lambda: self.minutes[0],
+            simulation_dt_func=lambda: self.first_minute
         )
 
         orders_txns = list(slippage_model.simulate(
@@ -1228,7 +1243,11 @@ class FixedBasisPointsSlippageTestCase(WithCreateBarData,
         self.assertEquals(expected_txn, txn.__dict__)
 
     @parameterized.expand([
+        # Volume limit for the bar is 20. We've ordered 10 total shares.
+        # We should fill both orders completely.
         ('order_under_limit', 9, 1, 9, 1),
+        # Volume limit for the bar is 20. We've ordered 21 total shares.
+        # The second order should have one share remaining after fill.
         ('order_over_limit', -3, 18, -3, 17),
     ])
     def test_volume_limit(self, name,
@@ -1251,7 +1270,7 @@ class FixedBasisPointsSlippageTestCase(WithCreateBarData,
         ]
 
         bar_data = self.create_bardata(
-            simulation_dt_func=lambda: self.minutes[0],
+            simulation_dt_func=lambda: self.first_minute,
         )
 
         orders_txns = list(slippage_model.simulate(
@@ -1266,3 +1285,23 @@ class FixedBasisPointsSlippageTestCase(WithCreateBarData,
         _, second_txn = orders_txns[1]
         self.assertEquals(first_txn['amount'], first_order_fill_amount)
         self.assertEquals(second_txn['amount'], second_order_fill_amount)
+
+    def test_broken_constructions(self):
+        with self.assertRaises(ValueError) as e:
+            FixedBasisPointsSlippage(basis_points=0)
+
+        self.assertEqual(
+            str(e.exception),
+            "FixedBasisPointsSlippage() expected a value strictly "
+            "greater than 0 for argument 'basis_points', but got 0 instead."
+        )
+
+        with self.assertRaises(ValueError) as e:
+            FixedBasisPointsSlippage(volume_limit=0)
+
+        self.assertEqual(
+            str(e.exception),
+            "FixedBasisPointsSlippage() expected a value strictly "
+            "greater than 0 for argument 'volume_limit', but got 0 instead."
+        )
+
--- a/tests/finance/test_slippage.py
+++ b/tests/finance/test_slippage.py
@@ -1178,7 +1178,7 @@ class FixedBasisPointsSlippageTestCase(WithCreateBarData,
     @classmethod
     def init_class_fixtures(cls):
         super(FixedBasisPointsSlippageTestCase, cls).init_class_fixtures()
-        cls.ASSET133 = cls.env.asset_finder.retrieve_asset(133)
+        cls.ASSET133 = cls.asset_finder.retrieve_asset(133)
 
     def test_fixed_bps_slippage(self):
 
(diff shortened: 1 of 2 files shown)


zipline/finance/slippage.py
```diff
@@ -502,3 +502,30 @@ def get_simulated_impact(self,
...
+class FixedBasisPointsSlippage(SlippageModel):
+    """
+    Model slippage as a fixed percentage of fill price.
+
+    Parameters
+    ----------
+    bps : int, optional
+        basis points to apply
+    """
+    def __init__(self, bps=5):
```
ssanderson: I would call this `basis_points` to match the `__repr__`.

zipline/finance/slippage.py
```diff
@@ -502,3 +502,30 @@ def get_simulated_impact(self,
     def get_txn_volume(self, data, order):
         volume = data.current(order.asset, 'volume')
         return volume * self.volume_limit
+
+
+class FixedBasisPointsSlippage(SlippageModel):
```
ssanderson: We should add tests for this to tests/finance/test_slippage.py.

ssanderson: `VolumeShareSlippageTestCase` looks like it would be a reasonable model to look at for adding these.

zipline/finance/slippage.py
```diff
@@ -502,3 +502,30 @@ def get_simulated_impact(self,
...
+    bps : int, optional
+        basis points to apply
+    """
+    def __init__(self, bps=5):
+        super(FixedBasisPointsSlippage, self).__init__()
+        self.bps = bps
+
+    def __repr__(self):
+        return '{class_name}(basis_points={bps})'.format(
+            class_name=self.__class__.__name__, spread=self.bps,
```
dmichalowicz: I think you meant `bps=self.bps` here.

zipline/finance/slippage.py
```diff
@@ -502,3 +502,30 @@ def get_simulated_impact(self,
...
+        )
+
+    def process_order(self, data, order):
+        price = data.current(order.asset, "close")
+
+        return (
+            price + price * (self.bps / 10000.0 * order.direction),
```
ssanderson: Small perf note: there's no reason to do this division on every order we process; we can just do this once at construction time.

zipline/finance/slippage.py
```diff
@@ -502,3 +502,30 @@ def get_simulated_impact(self,
     def get_txn_volume(self, data, order):
         volume = data.current(order.asset, 'volume')
         return volume * self.volume_limit
+
+
+class FixedBasisPointsSlippage(SlippageModel):
+    """
+    Model slippage as a fixed percentage of fill price.
```
ssanderson: This should probably also mention that it executes the full order immediately.

zipline/finance/slippage.py
ssanderson: I'm wondering if we should be capping the percentage of the bar that we fill here.  I know that part of the goal of adding this model is to give a reasonable estimate without simulating a huge number of transactions, but it seems pretty inaccurate to fill a 1000 share order in a minute where only 10 shares transacted.

zipline/finance/slippage.py
vikram-narayan: added a test for this slippage model (may change the test if we change the model)

zipline/finance/slippage.py
vikram-narayan: whoops, fixed

zipline/finance/slippage.py
vikram-narayan: - added `self.percentage` to avoid doing division on every order
- will change the slippage model depending on what we decide about that

zipline/finance/slippage.py
vikram-narayan: updated the docstring

zipline/finance/slippage.py
vikram-narayan: fixed

vikram-narayan:
@dmichalowicz @ssanderson added 10% volume limit, mind taking a quick look?

tests/finance/test_slippage.py
```diff
@@ -1146,3 +1147,85 @@ def test_orders_stop(self, name, order_data, event_data, expected):
...
+        ))
+
+        self.assertEquals(len(orders_txns), 1)
+        _, txn = orders_txns[0]
+
+        expected_txn = {
+            'price': float(3.0015),
+            'dt': datetime.datetime(
+                2006, 1, 5, 14, 31, tzinfo=pytz.utc),
+            'amount': int(20),
```
ssanderson: `int` here seems uncessary?

tests/finance/test_slippage.py
```diff
@@ -1146,3 +1147,85 @@ def test_orders_stop(self, name, order_data, event_data, expected):
...
+            bar_data,
+            self.ASSET133,
+            open_orders,
+        ))
+
+        self.assertEquals(len(orders_txns), 1)
+        _, txn = orders_txns[0]
+
+        expected_txn = {
+            'price': float(3.0015),
```
ssanderson: `float` here seems unecessary?

tests/finance/test_slippage.py
```diff
@@ -1146,3 +1147,85 @@ def test_orders_stop(self, name, order_data, event_data, expected):
...
+        )
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(FixedBasisPointsSlippageTestCase, cls).init_class_fixtures()
+        cls.ASSET133 = cls.env.asset_finder.retrieve_asset(133)
+
+    def test_fixed_bps_slippage(self):
+
+        slippage_model = FixedBasisPointsSlippage()
```
ssanderson: We should test that this actually respects the basis points value that it's given.  One way to do that would be to use a parameterized test, either with `parameter_space` or with `nose.parameterized`.

tests/finance/test_slippage.py
```diff
@@ -1146,3 +1147,85 @@ def test_orders_stop(self, name, order_data, event_data, expected):
...
+                'close': [3.00],
+                'volume': [200],
+            },
+            index=[cls.minutes[0]],
+        )
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(FixedBasisPointsSlippageTestCase, cls).init_class_fixtures()
+        cls.ASSET133 = cls.env.asset_finder.retrieve_asset(133)
```
ssanderson: Just use `self.asset_finder` here.  `TradingEnvironment` is pretty vestigial, and any tests that are still using are just doing so by historical accident.

tests/finance/test_slippage.py
```diff
@@ -1146,3 +1147,85 @@ def test_orders_stop(self, name, order_data, event_data, expected):
...
+    ASSET_FINDER_EQUITY_START_DATE = pd.Timestamp('2006-01-05', tz='utc')
+    ASSET_FINDER_EQUITY_END_DATE = pd.Timestamp('2006-01-07', tz='utc')
+    minutes = pd.DatetimeIndex(
+        start=START_DATE,
+        end=END_DATE - pd.Timedelta('1 minute'),
+        freq='1min'
+    )
+
+    @classproperty
+    def CREATE_BARDATA_DATA_FREQUENCY(cls):
```
ssanderson: Since we're not actually using sim params in the test here, I would probably just set this attribute directly at class scope and remove the dependency on sim_params entirely.

tests/finance/test_slippage.py
```diff
@@ -1146,3 +1147,85 @@ def test_orders_stop(self, name, order_data, event_data, expected):
...
+        cls.ASSET133 = cls.env.asset_finder.retrieve_asset(133)
+
+    def test_fixed_bps_slippage(self):
+
+        slippage_model = FixedBasisPointsSlippage()
+
+        open_orders = [
+            Order(
+                dt=datetime.datetime(2006, 1, 5, 14, 30, tzinfo=pytz.utc),
+                amount=100,
```
ssanderson: We should probably have a test case for negative amounts to ensure that we're adjusting in the correct direction on shorts.

zipline/finance/slippage.py
```diff
@@ -502,3 +502,45 @@ def get_simulated_impact(self,
...
+""".strip().format(
+            class_name=self.__class__.__name__,
+            basis_points=self.basis_points,
+            volume_limit=self.volume_limit,
+        )
+
+    def process_order(self, data, order):
+
+        volume = data.current(order.asset, "volume")
+        max_volume = self.volume_limit * volume
```
ssanderson: I think this also needs to account for the fact that we might have already filled a different order for this asset.  `VolumeShareSlippage` uses `self.volume_for_bar` (which is maintained by the base class) to track this.

tests/finance/test_slippage.py
ssanderson: This would also be a reasonable way to test different amounts.

zipline/finance/slippage.py
```diff
@@ -502,3 +502,45 @@ def get_simulated_impact(self,
...
+
+class FixedBasisPointsSlippage(SlippageModel):
+    """
+    Model slippage as a fixed percentage of fill price. Executes the full
+    order immediately.
+
+    Parameters
+    ----------
+    basis_points : int, optional
+        basis points to apply
```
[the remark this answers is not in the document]
ssanderson: We should document `volume_limit` as well now.

tests/finance/test_slippage.py
dmichalowicz: In that vein, probably worth having a test for an amount less than the volume limit.

tests/finance/test_slippage.py
```diff
@@ -1146,3 +1147,85 @@ def test_orders_stop(self, name, order_data, event_data, expected):
...
+                                       ZiplineTestCase):
+
+    START_DATE = pd.Timestamp('2006-01-05 14:31', tz='utc')
+    END_DATE = pd.Timestamp('2006-01-05 14:36', tz='utc')
+    SIM_PARAMS_CAPITAL_BASE = 1.0e5
+    SIM_PARAMS_DATA_FREQUENCY = 'minute'
+    SIM_PARAMS_EMISSION_RATE = 'daily'
+
+    ASSET_FINDER_EQUITY_SIDS = (133,)
+    ASSET_FINDER_EQUITY_START_DATE = pd.Timestamp('2006-01-05', tz='utc')
```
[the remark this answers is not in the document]
vikram-narayan: removed start and end date for asset finder

tests/finance/test_slippage.py
```diff
@@ -1146,3 +1147,85 @@ def test_orders_stop(self, name, order_data, event_data, expected):
...
+    ASSET_FINDER_EQUITY_END_DATE = pd.Timestamp('2006-01-07', tz='utc')
+    minutes = pd.DatetimeIndex(
```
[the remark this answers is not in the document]
vikram-narayan: changed this to be an array of length 1

tests/finance/test_slippage.py
vikram-narayan: removed SimParams

tests/finance/test_slippage.py
vikram-narayan: 👍

tests/finance/test_slippage.py
vikram-narayan: added these testes

tests/finance/test_slippage.py
vikram-narayan: removed the `float`

tests/finance/test_slippage.py
vikram-narayan: removed this

zipline/finance/slippage.py
```diff
@@ -502,3 +502,45 @@ def get_simulated_impact(self,
...
+
+
+class FixedBasisPointsSlippage(SlippageModel):
+    """
+    Model slippage as a fixed percentage of fill price. Executes the full
+    order immediately.
+
+    Parameters
+    ----------
+    basis_points : int, optional
```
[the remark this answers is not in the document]
vikram-narayan: true, updated docstring for this to be a float, and changed the order amount to be castto an `int`

zipline/finance/slippage.py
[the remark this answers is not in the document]
vikram-narayan: updated docstring

zipline/finance/slippage.py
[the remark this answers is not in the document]
vikram-narayan: truncated this

vikram-narayan: updated to subtract `volume_for_bar` from order amount

tests/finance/test_slippage.py
vikram-narayan: added this test

vikram-narayan:
added more tests, updated docstring, and used `self.volume_for_bar` to account for previous orders

ssanderson:
LGTM

PR #2051, opened 2017-12-14
MAINT: set FixedBasisPointsSlippage as default

vikram-narayan:
set `FixedBasisPointsSlippage` as default slippage model

### DIFF
--- a/tests/finance/test_blotter.py
+++ b/tests/finance/test_blotter.py
@@ -30,6 +30,7 @@ from zipline.finance.order import ORDER_STATUS, Order
 from zipline.finance.slippage import (
     DEFAULT_EQUITY_VOLUME_SLIPPAGE_BAR_LIMIT,
     FixedSlippage,
+    VolumeShareSlippage,
 )
 from zipline.gens.sim_engine import BAR, SESSION_END
 from zipline.testing.fixtures import (
@@ -261,7 +262,8 @@ class BlotterTestCase(WithCreateBarData,
         status indication. When a fill happens, the order should switch
         status to OPEN/FILLED as necessary
         """
-        blotter = Blotter(self.sim_params.data_frequency)
+        blotter = Blotter(self.sim_params.data_frequency,
+                          equity_slippage=VolumeShareSlippage())
         # Nothing happens on held of a non-existent order
         blotter.hold(56)
         self.assertEqual(blotter.new_orders, [])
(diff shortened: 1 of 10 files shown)


zipline/algorithm.py
```diff
@@ -320,6 +321,7 @@ def __init__(self, *args, **kwargs):
         if not self.blotter:
             self.blotter = Blotter(
                 data_frequency=self.data_frequency,
+                equity_slippage=FixedBasisPointsSlippage(),
```
ssanderson: How were we setting the default slippage model before if we weren't passing it here?  Should we be doing this there instead?

vikram-narayan:
changed the default setting to be in `blotter.py`, and added params to make equity and future slippage models configurable

zipline/algorithm.py
vikram-narayan: it was set in blotter before, changed the default to be set in blotter, but made it configurable from `TradingAlgorithm`

ssanderson:
@vikram-narayan changes here look good to me, but I think you've still got some real test failures from updating the slippage model.

vikram-narayan:
@ssanderson made the slippage model return `None, None` if the number of shares to fill is 0. Appveyor is passing, still seeing the weird issue with travis builds

zipline/algorithm.py
```diff
@@ -250,6 +250,12 @@ def __init__(self, *args, **kwargs):
             identifiers : List
                 Any asset identifiers that are not provided in the
                 equities_metadata, but will be traded by this TradingAlgorithm
+            equity_slippage : SlippageModel
```
ssanderson: Is there a reason to prefer making this an arg vs. using the existing `set_slippage` API function?  It seems a little confusing to have two different ways to do this.

zipline/examples/__init__.py
```diff
@@ -77,6 +78,8 @@ def run_example(example_name, environ):
         analyze=getattr(mod, 'analyze', None),
         bundle='test',
         environ=environ,
+        # example test data use the VolumeShareSlippage model
+        equity_slippage=VolumeShareSlippage(),
```
ssanderson: Setting this here makes the examples a little bit confusing; if you ran the example code exactly as defined in the example file, you'd get a different result than what we produce here.  It looks like we call `set_commission` in most of the examples for this reason.

ssanderson:
LGTM

PR #2057, opened 2017-12-19
BUG: raise LiquidityExceeded if no shares to fill

vikram-narayan:


zipline/finance/slippage.py
```diff
@@ -558,6 +558,9 @@ def process_order(self, data, order):
         shares_to_fill = min(abs(order.open_amount),
                              max_volume - self.volume_for_bar)
 
+        if shares_to_fill == 0:
```
ssanderson: Any reason not to use the `raise LiquidityExceeded()` pattern we use in `VolumeShareSlippage` for this?

ssanderson: Also, as pointed out by @jbredeche, we should add a test for this case.

vikram-narayan: just changed it to use `LiquidityExceeded`, i had seen the `return None, None` pattern in `VolumeShareSlippage` as well so I used that

vikram-narayan: `simulate` in slippage.py catches the `LiquidityExceeded` exception, so the test just checks that a transaction doesn't get created

PR #2062, opened 2017-12-26
BUG: cal attribute of aggregate rules was not being propagated

richafrank:
Otherwise algos crash with
```python
AttributeError: 'BeforeClose' object has no attribute 'cal' 
```
if a function is scheduled with, e.g.
```python
schedule_function(
    func=my_func,
    date_rule=date_rules.every_day(),
    time_rule=time_rules.market_close(minutes=30) & time_rules.every_minute(),
)
```

### DIFF
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -1064,7 +1064,7 @@ class TradingAlgorithm(object):
 
         return csv_data_source
 
-    def add_event(self, rule=None, callback=None):
+    def add_event(self, rule, callback):
         """Adds an event to the algorithm's EventManager.
 
         Parameters
--- a/zipline/utils/events.py
+++ b/zipline/utils/events.py
@@ -223,7 +223,7 @@ class Event(namedtuple('Event', ['rule', 'callback'])):
     with the current algorithm context, data, and datetime only when the rule
     is triggered.
     """
-    def __new__(cls, rule=None, callback=None):
+    def __new__(cls, rule, callback=None):
         callback = callback or (lambda *args, **kwargs: None)
         return super(cls, cls).__new__(cls, rule=rule, callback=callback)
 
@@ -236,6 +236,18 @@ class Event(namedtuple('Event', ['rule', 'callback'])):
 
 
 class EventRule(six.with_metaclass(ABCMeta)):
+    # Instances of EventRule are assigned a calendar instance when scheduling
+    # a function.
+    _cal = None
+
+    @property
+    def cal(self):
+        return self._cal
+
+    @cal.setter
+    def cal(self, value):
+        self._cal = value
+
     @abstractmethod
     def should_trigger(self, dt):
         """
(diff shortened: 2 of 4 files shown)


ssanderson:
Just to make sure I understand, `time_rules.market_close(minutes=30) & time_rules.every_minute(),` should be functionally equivalent to just `time_rules.market_close(minutes=30)`, right?

zipline/algorithm.py
```diff
@@ -1063,7 +1063,7 @@ def fetch_csv(self,
 
         return csv_data_source
 
-    def add_event(self, rule=None, callback=None):
+    def add_event(self, rule, callback=None):
```
ssanderson: Is there a reason `callback` is allowed to be None here?  I would think failing to pass a callback is almost always a bug.

zipline/utils/events.py
```diff
@@ -546,11 +557,14 @@ class StatefulRule(EventRule):
     def __init__(self, rule=None):
         self.rule = rule or Always()
 
-    def new_should_trigger(self, callable_):
```
ssanderson: Why was this removed?

richafrank:
Yup, semantically there's no reason there for the second term; this was just to show that each of those terms doesn't get assigned a `cal` attribute.

zipline/utils/events.py
```diff
@@ -236,6 +236,8 @@ def handle_data(self, context, data, dt):
 
 
 class EventRule(six.with_metaclass(ABCMeta)):
+    cal = None  # A calendar instance is set when scheduling a function.
```
ssanderson: Maybe worth clarifying that the calendar is set on instances of the rule, as opposed to the class?

ssanderson: Should we make `cal` an abstract property to ensure that all the subclasses implement setters?

richafrank:
Agreed. I went through the same thought process, and settled on this PR.

zipline/utils/events.py
richafrank: I didn't find any call sites.

zipline/algorithm.py
richafrank: I'm guessing it was convenient for tests. Maybe @llllllllll knows more.

zipline/utils/events.py
richafrank: I thought about that as well. My only concern was about the performance of the properties inside the `should_trigger` implementation, but I didn't validate that, and `@lazyval` could also solve that.

ssanderson: I think for simple attribute access `property `will actually be faster than `lazyval`.  lazyval has more complicated machinery in exchange for memoizing the result, but if the fetching the attribute is trivial, then I think property will probably have less overhead.

zipline/algorithm.py
llllllllll: Yeah, I don't think there are good defaults for either. We should make them required.

PR #2064, opened 2017-12-28
Create MinLeverage control

srklonaris:
Similar to `MaxLeverage` create a `MinLeverage` control.  

`MinLeverage` takes a float as well as a deadline date.  `MinLeverage` validations will only happen after the algorithm has reached the deadline date.  This allows algorithms enough time to reach their min leverage because it may take time for them to have a non zero leverage.

### DIFF
--- a/tests/test_algorithm.py
+++ b/tests/test_algorithm.py
@@ -137,6 +137,7 @@ from zipline.test_algorithms import (
     SetAssetRestrictionsAlgorithm,
     SetMultipleAssetRestrictionsAlgorithm,
     SetMaxLeverageAlgorithm,
+    SetMinLeverageAlgorithm,
     api_algo,
     api_get_environment_algo,
     api_symbol_algo,
@@ -3531,6 +3532,46 @@ class TestAccountControls(WithDataPortal, WithSimParams, ZiplineTestCase):
                                        env=self.env)
         self.check_algo_succeeds(algo, handle_data)
 
+    def test_set_min_leverage(self):
+        def handle_data(algo, data):
+            algo.order_target_percent(algo.sid(self.sidint), .5)
+            algo.record(latest_time=algo.get_datetime())
+
+        # Set min leverage to 1.
+        # The algorithm will succeed because it doesn't run for more
+        # than 10 days.
+        offset = pd.Timedelta('10 days')
+        algo = SetMinLeverageAlgorithm(1, offset, sim_params=self.sim_params,
+                                       env=self.env)
+        self.check_algo_succeeds(algo, handle_data)
+
+        # The algorithm will fail because it doesn't reach a min leverage of 1
+        # after 1 day.
+        offset = pd.Timedelta('1 days')
+        algo = SetMinLeverageAlgorithm(1, offset, sim_params=self.sim_params,
+                                       env=self.env)
+        self.check_algo_fails(algo, handle_data)
+        self.assertEqual(
+            algo.recorded_vars['latest_time'],
+            pd.Timestamp('2006-01-04 21:00:00', tz='UTC'),
+        )
+
+        # Increase the offset to 2 days, and the algorithm fails a day later
+        offset = pd.Timedelta('2 days')
+        algo = SetMinLeverageAlgorithm(1, offset, sim_params=self.sim_params,
+                                       env=self.env)
+        self.check_algo_fails(algo, handle_data)
+        self.assertEqual(
+            algo.recorded_vars['latest_time'],
+            pd.Timestamp('2006-01-05 21:00:00', tz='UTC'),
+        )
+
+        # Set the min_leverage to .0001 and the algorithm succeeds.
+        algo = SetMinLeverageAlgorithm(.0001,
+                                       offset,
+                                       sim_params=self.sim_params,
+                                       env=self.env)
+        self.check_algo_succeeds(algo, handle_data)
 
 # FIXME re-implement this testcase in q2
 # class TestClosePosAlgo(TestCase):
--- a/zipline/algorithm.py
+++ b/zipline/algorithm.py
@@ -70,6 +70,7 @@ from zipline.finance.controls import (
     MaxOrderSize,
     MaxPositionSize,
     MaxLeverage,
+    MinLeverage,
     RestrictedListOrder
 )
 from zipline.finance.execution import (
@@ -2232,6 +2233,22 @@ class TradingAlgorithm(object):
             be no maximum.
         """
         control = MaxLeverage(max_leverage)
+        self.register_account_control(control)
+
+    @api_method
+    def set_min_leverage(self, min_leverage, grace_period):
+        """
+        Set a limit on the minimum leverage of the algorithm.
+
+        Parameters
+        ----------
+        min_leverage : float
+            The minimum leverage for the algorithm.
+        grace_period : pd.Timedelta
+            The offset from the start date used to enforce a minimum leverage.
+        """
+        deadline = self.sim_params.start_session + grace_period
+        control = MinLeverage(min_leverage, deadline)
         self.register_account_control(control)
 
     ####################
--- a/zipline/finance/controls.py
+++ b/zipline/finance/controls.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 import abc
 import logbook
+from datetime import datetime
 
 import pandas as pd
 
@@ -23,6 +24,11 @@ from zipline.errors import (
     AccountControlViolation,
     TradingControlViolation,
 )
+from zipline.utils.input_validation import (
+    expect_bounded,
+    expect_types,
+)
+
 
 log = logbook.Logger('TradingControl')
 
@@ -424,4 +430,45 @@ class MaxLeverage(AccountControl):
         Fail if the leverage is greater than the allowed leverage.
         """
         if _account.leverage > self.max_leverage:
+            self.fail()
+
+
+class MinLeverage(AccountControl):
+    """AccountControl representing a limit on the minimum leverage allowed
+    by the algorithm after a threshold period of time.
+
+    Parameters
+    ----------
+    min_leverage : float
+        The gross leverage in decimal form.
+    deadline : datetime
+        The date the min leverage must be achieved by.
+
+    For example, min_leverage=2 limits an algorithm to trading at minimum
+    double the account value by the deadline date.
+    """
+
+    @expect_types(
+        __funcname='MinLeverage',
+        min_leverage=(int, float),
+        deadline=datetime
+    )
+    @expect_bounded(__funcname='MinLeverage', min_leverage=(0, None))
+    def __init__(self, min_leverage, deadline):
+        super(MinLeverage, self).__init__(min_leverage=min_leverage,
+                                          deadline=deadline)
+        self.min_leverage = min_leverage
+        self.deadline = deadline
+
+    def validate(self,
+                 _portfolio,
+                 account,
+                 algo_datetime,
+                 _algo_current_data):
+        """
+        Make validation checks if we are after the deadline.
+        Fail if the leverage is less than the min leverage.
+        """
+        if algo_datetime > self.deadline and \
+           account.leverage < self.min_leverage:
             self.fail()
--- a/zipline/test_algorithms.py
+++ b/zipline/test_algorithms.py
@@ -485,6 +485,13 @@ class SetMaxLeverageAlgorithm(TradingAlgorithm):
         self.set_max_leverage(max_leverage=max_leverage)
 
 
+class SetMinLeverageAlgorithm(TradingAlgorithm):
+    def initialize(self, min_leverage, grace_period):
+        self.set_min_leverage(
+            min_leverage=min_leverage, grace_period=grace_period
+        )
+
+
 ############################
 # TradingControl Test Algos#
 ############################
--- a/zipline/api.pyi
+++ b/zipline/api.pyi
@@ -633,6 +633,20 @@ def set_max_leverage(max_leverage):
         be no maximum.
     """
 
+def set_min_leverage(min_leverage, timedelta):
+
+    """
+    Set a limit on the minimum leverage of the algorithm.
+
+    Parameters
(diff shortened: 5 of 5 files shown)


zipline/finance/controls.py
```diff
@@ -425,3 +425,48 @@ def validate(self,
...
+
+class MinLeverage(AccountControl):
+    """
+    AccountControl representing a limit on the minimum leverage allowed
+    by the algorithm after a threshold period of time.
+    """
+
+    def __init__(self, min_leverage, deadline):
+        """
+        max_leverage is the gross leverage in decimal form. For example,
```
ssanderson: Docstring here is still copied from `MaxLeverage`.  We've also standardized on [numpy-style docstrings](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) in zipline since `MaxLeverage` was written.  I don't think we need to go back and change existing code, but for new docstrings we should still use the style we use elsewhere in Zipline.

zipline/finance/controls.py
```diff
@@ -425,3 +425,48 @@ def validate(self,
...
+        2, limits an algorithm to trading at most double the account value.
+        """
+        super(MinLeverage, self).__init__(min_leverage=min_leverage,
+                                          deadline=deadline)
+        self.min_leverage = min_leverage
+        self.deadline = deadline
+
+        if min_leverage is None:
```
ssanderson: You might want to look at using the decorators in `zipline.utils.input_validation` to implement these checks. I think `expect_types(min_leverage=(int, float), deadline=datetime)` and `expect_strictly_bounded(min_leverage=(0, None))` would capture the behavior we want here.  If you do end up using those functions, you probably want to pass `__funcname='MinLeverage'` so that the generated error messages reference `MinLeverage` instead of `__init__`.

zipline/test_algorithms.py
```diff
@@ -485,6 +485,11 @@ def initialize(self, max_leverage=None):
         self.set_max_leverage(max_leverage=max_leverage)
 
 
+class SetMinLeverageAlgorithm(TradingAlgorithm):
+    def initialize(self, min_leverage=None, deadline=None):
```
ssanderson: Should we just make these required if it's invalid to pass `None` for these values to `set_min_leverage`?

zipline/algorithm.py
```diff
@@ -2234,6 +2235,22 @@ def set_max_leverage(self, max_leverage):
...
+    def set_min_leverage(self, min_leverage, deadline):
+        """
+        Set a limit on the minimum leverage of the algorithm.
+
+        Parameters
+        ----------
+        min_leverage : float
+            The minimum leverage for the algorithm. If not provided there will
+            be no minimum.
+        deadline : datetime
```
ssanderson: This might be more nicely expressed as a timedelta from the start of the algorithm instead of a datetime.

zipline/algorithm.py
```diff
@@ -2234,6 +2235,22 @@ def set_max_leverage(self, max_leverage):
...
 
+    @api_method
+    def set_min_leverage(self, min_leverage, deadline):
+        """
+        Set a limit on the minimum leverage of the algorithm.
+
+        Parameters
+        ----------
+        min_leverage : float
+            The minimum leverage for the algorithm. If not provided there will
```
ssanderson: Is the second sentence here actually true?  If not, should we just remove it?

zipline/finance/controls.py
```diff
@@ -425,3 +425,48 @@ def validate(self,
...
+                "min_leverage must be positive"
+            )
+
+        if deadline is None:
+            raise ValueError(
+                "Must supply deadline"
+            )
+
+    def validate(self,
+                 _portfolio,
```
ssanderson: The underscores here are a weird circa-2014 scott-ism: the contention elsewhere in this file is that the paramters start with an underscore if they're unused in the function, which was a convention I'd used in a previous codebase.  We don't do that anywhere else in zipline though.  I'd either follow the existing convention in this file (in which case `algo_datetime` and `account` shouldn't have underscores), or just make them all normal-looking parameters.

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,42 @@ def handle_data(algo, data):
                                        env=self.env)
         self.check_algo_succeeds(algo, handle_data)
 
+    def test_set_min_leverage(self):
+        # Algo date is not past deadline so algorithm succeeds
+        deadline = pd.Timestamp('2010-01-05', tz='UTC')
+
+        # Set min leverage to 0 so buying one share fails.
+        def handle_data(algo, data):
```
ssanderson: I think this `handle_data` definition is sufficient for all the tests in this function.  Seems like it'd be simpler and clearer to just always use this instead of re-defining it for each test?

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,42 @@ def handle_data(algo, data):
                                        env=self.env)
         self.check_algo_succeeds(algo, handle_data)
 
+    def test_set_min_leverage(self):
+        # Algo date is not past deadline so algorithm succeeds
+        deadline = pd.Timestamp('2010-01-05', tz='UTC')
+
+        # Set min leverage to 0 so buying one share fails.
```
ssanderson: This comment doesn't match the test case.

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,42 @@ def handle_data(algo, data):
                                        env=self.env)
         self.check_algo_succeeds(algo, handle_data)
 
+    def test_set_min_leverage(self):
+        # Algo date is not past deadline so algorithm succeeds
```
ssanderson: What is "algo date" referring to in this comment?

zipline/test_algorithms.py
```diff
@@ -485,6 +485,11 @@ def initialize(self, max_leverage=None):
         self.set_max_leverage(max_leverage=max_leverage)
 
 
+class SetMinLeverageAlgorithm(TradingAlgorithm):
+    def initialize(self, min_leverage=None, timedelta=None):
```
ssanderson: Should these just be required if passing them as None will isn't actually supported?

zipline/algorithm.py
```diff
@@ -2234,6 +2235,23 @@ def set_max_leverage(self, max_leverage):
         control = MaxLeverage(max_leverage)
         self.register_account_control(control)
 
+    @api_method
+    def set_min_leverage(self, min_leverage, timedelta):
```
ssanderson: I would still call this `deadline`.  `timedelta` is shadowing the name of a built-in type.

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,42 @@ def handle_data(algo, data):
...
+        offset = pd.Timedelta('10 days')
+
+        # Set min leverage to 0 so buying one share fails.
```
ssanderson: This comment doesn't match the test case.

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,42 @@ def handle_data(algo, data):
...
+        def handle_data(algo, data):
+            algo.order(algo.sid(self.sidint), 1)
+            algo.record(latest_time=algo.get_datetime())
+
+        algo = SetMinLeverageAlgorithm(1, offset, sim_params=self.sim_params,
+                                       env=self.env)
+        self.check_algo_fails(algo, handle_data)
+        self.assertEqual(
+            algo.recorded_vars['latest_time'],
+            pd.Timestamp('2006-01-04 21:00:00', tz='UTC'),
```
ssanderson: Can we also test that this fails at 01-05 if we increase the offset delta to 2 days?  (I might just break out those two cases into separate test methods instead of having all the cases at once).

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,42 @@ def handle_data(algo, data):
                                        env=self.env)
         self.check_algo_succeeds(algo, handle_data)
 
+    def test_set_min_leverage(self):
+        # Algo date is not past deadline so algorithm succeeds
+        offset = pd.Timedelta('10 days')
+
+        # Set min leverage to 0 so buying one share fails.
+        def handle_data(algo, data):
```
ssanderson: All the `handle_data` functions we define in this method are functionally the same.  Is there a reason we're defining a separate one each time?

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,42 @@ def handle_data(algo, data):
...
+            pd.Timestamp('2006-01-04 21:00:00', tz='UTC'),
+        )
+
+        # Set min leverage to 0 so buying one share passes
+        def handle_data(algo, data):
+            algo.order(algo.sid(self.sidint), 1)
+
+        algo = SetMinLeverageAlgorithm(0, offset, sim_params=self.sim_params,
+                                       env=self.env)
+        self.check_algo_succeeds(algo, handle_data)
```
ssanderson: I think we need a test case where we set a minimum leverage amount and the algorithm has a leverage above that limit after the deadline and doesn't crash.  Bonus points if the algorithm then goes below the leverage limit and crashes.

ssanderson: This might be easier to implement using `order_target_percent` to hit fixed leverage amounts instead of ordering fixed numbers of shares.

zipline/algorithm.py
srklonaris: What about `offset`, since `deadline` is what is computed and passed into `MinLeverage`

ssanderson: As a consumer of this function, I don't think `offset` would be a very clear name for this parameter.  My immediate question would be "offset from what"?  Maybe `grace_period`?

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,46 @@ def handle_data(algo, data):
                                        env=self.env)
         self.check_algo_succeeds(algo, handle_data)
 
+    def test_set_min_leverage(self):
+        def handle_data(algo, data):
+            algo.order_target_percent(algo.sid(self.sidint), .5)
+            algo.record(latest_time=algo.get_datetime())
+
+        # Set min leverage to 1.
```
ssanderson: I don't think this comment matches the test case?

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,46 @@ def handle_data(algo, data):
...
+        # The algorithm will succeed because it doesn't run for more
+        # than 10 days.
+        offset = pd.Timedelta('10 days')
+        algo = SetMinLeverageAlgorithm(0, offset, sim_params=self.sim_params,
+                                       env=self.env)
+        self.check_algo_succeeds(algo, handle_data)
+
+        # The algorithm will fail because it doesn't reach a min leverage of 1
```
ssanderson: :+1:

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,46 @@ def handle_data(algo, data):
...
+    def test_set_min_leverage(self):
+        def handle_data(algo, data):
+            algo.order_target_percent(algo.sid(self.sidint), .5)
+            algo.record(latest_time=algo.get_datetime())
+
+        # Set min leverage to 1.
+        # The algorithm will succeed because it doesn't run for more
+        # than 10 days.
+        offset = pd.Timedelta('10 days')
+        algo = SetMinLeverageAlgorithm(0, offset, sim_params=self.sim_params,
```
ssanderson: I think this should be a 1?

tests/test_algorithm.py
```diff
@@ -3531,6 +3532,46 @@ def handle_data(algo, data):
...
+        offset = pd.Timedelta('1 days')
+        algo = SetMinLeverageAlgorithm(1, offset, sim_params=self.sim_params,
+                                       env=self.env)
+        self.check_algo_fails(algo, handle_data)
+        self.assertEqual(
+            algo.recorded_vars['latest_time'],
+            pd.Timestamp('2006-01-04 21:00:00', tz='UTC'),
+        )
+
+        # Increase the offset to 2 days, and the algorithm fails a day later
```
ssanderson: This is nicely clear to follow as a reader :+1: .

zipline/algorithm.py
```diff
@@ -2234,6 +2235,22 @@ def set_max_leverage(self, max_leverage):
...
+    def set_min_leverage(self, min_leverage, grace_period):
+        """
+        Set a limit on the minimum leverage of the algorithm.
+
+        Parameters
+        ----------
+        min_leverage : float
+            The minimum leverage for the algorithm.
+        grace_period : pd.Timedelta
+            The offset from the start date used to enforce a minimum leverage
```
ssanderson: Missing period here.

zipline/finance/controls.py
```diff
@@ -425,3 +430,40 @@ def validate(self,
...
+class MinLeverage(AccountControl):
+    """
+    AccountControl representing a limit on the minimum leverage allowed
+    by the algorithm after a threshold period of time.
+    """
+
+    @expect_types(
+        __funcname='MinLeverage',
+        min_leverage=(int, float),
+        deadline=pd.datetime
```
ssanderson: `pd.datetime` is just an alias of the standard library `datetime.datetime` (presumably b/c something in the top level pandas module does `from datetime import datetime`).  We should use the standard library location here.

zipline/finance/controls.py
```diff
@@ -425,3 +430,40 @@ def validate(self,
...
+    )
+    @expect_bounded(__funcname='MinLeverage', min_leverage=(0, None))
+    def __init__(self, min_leverage, deadline):
+        """
```
ssanderson: We generally don't put docstrings on `__init__` functions; we use the class docstring documenting the parameters to a class and describing what it's for.

zipline/finance/controls.py
```diff
@@ -425,3 +430,40 @@ def validate(self,
         """
         if _account.leverage > self.max_leverage:
             self.fail()
+
+
+class MinLeverage(AccountControl):
+    """
```
ssanderson: This should probably also be a numpy-style docstring.

ssanderson:
LGTM

PR #2071, opened 2018-01-02
PERF: Speedup RollingPearson

ssanderson:
Use the same techniques used in `SimpleBeta` to re-implement
RollingPearson (and transitively, RollingPearsonOfReturns).

For a 1 year window length, this provides about a 60x speedup on my
machine:

```
pipebench/perf/pearson.stats% stats statistical.py
Mon Jan  1 20:00:31 2018    pipebench/perf/pearson.stats

         9326197 function calls (9325541 primitive calls) in 8.407 seconds

   Ordered by: cumulative time
   List reduced from 766 to 3 due to restriction <'statistical.py'>

       # Old implementation.
       21    0.243    0.012    7.130    0.340 statistical.py:92(compute)
       # New implementation.
       21    0.000    0.000    0.253    0.012 statistical.py:105(compute)
       21    0.170    0.008    0.253    0.012 statistical.py:685(vectorized_pearson_r)
```

The new `vectorized_pearson_r` also has the same support for missing
data that was implemented in `vectorized_beta`, but I haven't pushed that
to logic up to `RollingPearson` yet.  I added a fast path for the allowed_missing=0 
case, since we can be significantly faster there, and it's likely to stay the default
for backwards compatibility reasons.

### DIFF
--- a/tests/pipeline/test_statistical.py
+++ b/tests/pipeline/test_statistical.py
@@ -33,7 +33,10 @@ from zipline.pipeline.factors import (
     RollingSpearmanOfReturns,
     SimpleBeta,
 )
-from zipline.pipeline.factors.statistical import vectorized_beta
+from zipline.pipeline.factors.statistical import (
+    vectorized_beta,
+    vectorized_pearson_r,
+)
 from zipline.pipeline.loaders.frame import DataFrameLoader
 from zipline.pipeline.sentinels import NotSpecified
 from zipline.testing import (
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_statistical.py
```diff
@@ -1051,3 +1054,84 @@ def test_allowed_missing_doesnt_double_count(self):
...
+            right_col = right_col[~missing]
+            r, pvalue = func(left_col, right_col)
+            out[col] = r
+
+        return out
+
+    def naive_columnwise_pearson(self, left, right):
+        return self.naive_columnwise_func(pearsonr, left, right)
+
+    def naive_columnwise_spearman(self, left, right):
```
freddiev4: Is this used anywhere?

tests/pipeline/test_statistical.py
```diff
@@ -1051,3 +1054,84 @@ def test_allowed_missing_doesnt_double_count(self):
...
+        def do_check(independent):
+            result = vectorized_pearson_r(
+                dependent, independent, allowed_missing=0
+            )
+            assert_equal(result, np.array([1.0, 1.0, -1.0]))
+
+        # We should get the same result from passing a N x 1 array or an N x 3
+        # array with the column tiled 3 times.
+        do_check(_independent)
+        do_check(np.tile(_independent, 3))
```
freddiev4: TIL about `np.tile()`. That's cool 🙂

tests/pipeline/test_statistical.py
```diff
@@ -1051,3 +1054,84 @@ def test_allowed_missing_doesnt_double_count(self):
...
+        seed=[1, 2, 42],
+        nan_offset=[-1, 0, 1],
+        nans=['dependent', 'independent', 'both'],
+        __fail_fast=True,
+    )
+    def test_produce_nans_when_too_much_missing_data(self,
+                                                     seed,
+                                                     nans,
+                                                     nan_offset):
+        rand = np.random.RandomState(seed)
```
freddiev4: What's the advantage of using this over `np.random.seed()` and then `np.random.uniform()`? Looking at the other test cases it seems like they all use this as well.

h55nick:
would be great to get this locked in! Use the RollingPearson factor a bunch

chunky123:
Would it be worth it to get the RollingSpearman stuff too? I find it hard to use that as well.

tests/pipeline/test_statistical.py
llllllllll: np.random.seed sets the global random state's seed, which is mutating shared global state. This creates a new random state which is isolated from other tests or callers.

PR #2074, opened 2018-01-04
BUG: Fixed daily price ffill using today's close

richafrank:
when we haven't seen any minute volume yet today

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from collections import OrderedDict
 from textwrap import dedent
 
 from nose_parameterized import parameterized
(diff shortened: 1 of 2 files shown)


richafrank:
Will look into the py3 failures.

zipline/data/data_portal.py
```diff
@@ -924,22 +924,29 @@ def get_history_window(self,
...
 
             assets_with_leading_nan = np.where(isnull(df.iloc[0]))[0]
 
             history_start, history_end = df.index[[0, -1]]
+            if ffill_data_frequency == 'daily' != data_frequency:
+                # When we're looking for a daily value, but we haven't seen any
+                # volume in today's minute bars yet, we need to use the
+                # previous day's ffilled daily price. Using today's daily price
+                # could yield a value from later today.
+                history_start -= pd.Timedelta(days=1)
```
ehebert: Is there a reason this isn't a trading calendar day?

richafrank: Since the most recent close as of the weekend is the same as the most recent close as of Friday, I didn't bother to figure out which trading calendar to use here, especially if there are multiple assets. Do you think this breaks under certain conditions?

ehebert: I don't think it breaks here.

Where it would be relevant is if we had a calendar with a session spanning multiple days, but in practice, I do not think we have that concern.

tests/test_history.py
```diff
@@ -1592,6 +1592,49 @@ def test_daily_history_blended_gaps(self, field):
...
+                    bar_count,
+                    '1d',
+                    'price',
+                    'minute',
+                )[self.ASSET3]
+
+                self.assertEqual(len(window), bar_count)
+
+                if not day_idx and idx + 1 < 10:
+                    self.assertTrue(np.isnan(window[-1]))
```
ehebert: I wouldn't block on it, but I would recommend picking a few specific times and expected values, crafted to test each condition we expect. That would 1) speed up the test time 2) make the narrative of what is being tested and why more clear.

yankees714: Seconded 👍

richafrank: Good idea! Updated.

ehebert: Thanks. In most cases I'd recommend the specific Timestamps, i.e. hardcoding `pd.Timestamp('2017-01-09 10:44'`); however since this test is iterating over multiple days, I agree with using the index to select the time.

richafrank: Yea, in this case, I saw the important information as being the open or close or the temporal relationship to the open, etc, as opposed to exact times, so I represented them that way.

PR #2078, opened 2018-01-09
MAINT: Provides better error message when QUANDL_API_KEY isn't set

ernestoeperez88:
Quandl's new API for the WIKI Prices dataset requires an API key.
Omitting the API key when running 'zipline ingest' caused:
HTTPError: HTTP Error 400: Bad Request

The new message asks for QUANDL_API_KEY to be set before
running the ingest process again.

### DIFF
--- a/zipline/data/bundles/quandl.py
+++ b/zipline/data/bundles/quandl.py
@@ -197,8 +197,14 @@ def quandl_bundle(environ,
     For more information on Quandl's API and how to obtain an API key,
     please visit https://docs.quandl.com/docs#section-authentication
     """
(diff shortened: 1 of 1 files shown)


zipline/data/bundles/quandl.py
```diff
@@ -105,6 +106,14 @@ def fetch_data_table(api_key,
                 show_progress=show_progress,
             )
 
+        except HTTPError as err:
+            if err.code == 400:
```
llllllllll: Can we just check this condition when we read the QUANDL_API_KEY out of the environment on line 210?

llllllllll:
lgtm, thanks

PR #2087, opened 2018-01-22
BUG: get_last_traded_dt expects a trading day

richafrank:
Otherwise we get back NaN.

The consumer of `history_start` is the later call to `self.get_last_traded_dt`, which returns `NaN` for non-trading days.

We were so close to catching this in the [previous PR](https://github.com/quantopian/zipline/pull/2074#discussion_r159711938)...

@ehebert @yankees714 Is `DataPortal.trading_calendar` good enough here, or do we need to look up a calendar per asset (and/or per reader)?

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -1599,8 +1599,16 @@ class MinuteEquityHistoryTestCase(WithHistory, ZiplineTestCase):
 
         # January 5 2015 is the first day, and there is volume only every
         # 10 minutes.
+
+        # January 6 has the same volume pattern and is used here to ensure we
+        # ffill correctly from the previous day when there is no volume yet
+        # today.
+
+        # January 12 is a Monday, ensuring we ffill correctly when the previous
+        # day is not a trading day.
         for day_idx, day in enumerate([pd.Timestamp('2015-01-05', tz='UTC'),
-                                       pd.Timestamp('2015-01-06', tz='UTC')]):
+                                       pd.Timestamp('2015-01-06', tz='UTC'),
+                                       pd.Timestamp('2015-01-12', tz='UTC')]):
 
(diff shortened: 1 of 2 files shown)


tests/test_history.py
```diff
@@ -1641,8 +1649,17 @@ def test_daily_history_minute_gaps_price_ffill(self, test_name, bar_count):
...
+                self.assertEqual(
+                    len(window),
+                    bar_count,
+                    "Unexpected window length at {}. Expected {}, but was {}."
+                        .format(minute, bar_count, len(window))
+                )
+                np.testing.assert_allclose(
+                    window[-1],
+                    expected,
+                    err_msg="at minute {}".format(minute),
```
richafrank: These changes to the assertions are just adding more informative messages.

ehebert:
The reader calendars are aligned to the `DataPortal.trading_calendar`, so I believe that your use of the `trading_calendar` member is correct.

tests/test_history.py
```diff
@@ -1600,7 +1600,8 @@ def test_daily_history_minute_gaps_price_ffill(self, test_name, bar_count):
         # January 5 2015 is the first day, and there is volume only every
         # 10 minutes.
         for day_idx, day in enumerate([pd.Timestamp('2015-01-05', tz='UTC'),
-                                       pd.Timestamp('2015-01-06', tz='UTC')]):
+                                       pd.Timestamp('2015-01-06', tz='UTC'),
+                                       pd.Timestamp('2015-01-12', tz='UTC')]):
```
ehebert: This date is crafted to include a day for which the previous day (when using Timedelta) is not a trading day, correct?
Might want to put a comment here about that, so than any future edits do not lose the intent of the test.

richafrank: Yes - good call.

PR #2089, opened 2018-01-24
Daily adjustments

richafrank:
BUG: On daily data, yesterday's adjustments were being applied
to yesterday's data from today's perspective, but no adjustment
happened between yesterday and today. They happened before yesterday's
open.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -545,9 +545,6 @@ class DataPortal(object):
             The timestamp for the desired value.
         perspective_dt : pd.Timestamp
             The timestamp from which the data is being viewed back from.
-        data_frequency : str
-            The frequency of the data to query; i.e. whether the data is
-            'daily' or 'minute' bars
 
         Returns
         -------
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -568,7 +568,7 @@ class DataPortal(object):
                 asset, self._splits_dict, "SPLITS"
             )
             for adj_dt, adj in split_adjustments:
-                if dt <= adj_dt <= perspective_dt:
+                if dt < adj_dt <= perspective_dt:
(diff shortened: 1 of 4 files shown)


zipline/data/data_portal.py
```diff
@@ -643,6 +640,9 @@ def get_adjusted_value(self, asset, field, dt,
                                                  data_frequency)
 
         if isinstance(asset, Equity):
+            if data_frequency == 'daily':
+                dt = self.trading_calendar.session_open(dt)
```
ehebert: When in `daily` data frequency, I don't believe we should be introducing a dt which is not date normalized.

I think this bug shows that we should be adding a `data_frequency` parameter to `get_adjustments`, and making the query for `get_adjustments` aware of the `data_frequency`.

ssanderson: I disagree with this recommendation.  `data_frequency` is a algorithm-level concept; we shouldn't be pushing that down to the layers of the stack that are purely about data retention and manipulation.  It's straightforward and context-independent to reason about what it means to say "give me data from time t0 adjusted as of time t1", whereas I have no intuition for what it means to say "give me data from time t0 at 'daily' data_frequency".

ehebert: @ssanderson you are right about keeping `DataPortal` free needing knowledge from the algorithm/backtest level data frequency. It had been a while since I've looked at this code, and had forgotten how important that is. Thanks for the admonition of keeping the levels separate.

However, I still believe that this is the wrong place to fix the application of adjustments when the dt and the perspective dt are on the same day.
The bug is that `get_adjustments` returns an adjustment when . I'll follow up with a suggestion how to fix that method.

richafrank:
@ehebert Updated as we discussed. I think the test cases here cover the scenarios we talked through.

PR #2091, opened 2018-01-26
Don't dispatch unneeded pipeline terms and fix repr bug in Maximum

ssanderson:


zipline/pipeline/graph.py
```diff
@@ -109,9 +109,7 @@ def ordered(self):
 
     @lazyval
     def loadable_terms(self):
-        return tuple(
-            term for term in self.graph if isinstance(term, LoadableTerm)
-        )
+        return {term for term in self.graph if isinstance(term, LoadableTerm)}
```
llllllllll: given that we memoize this as a lazyval, could we make it a frozenset to preserve the immutability?

ssanderson: We could, though FWIW we have a bunch of other lazyvals on this class that are dicts, so I think we're already operating under the assumption that no one is mutating these objects.

PR #2092, opened 2018-01-26
MAINT: Refactor who has to know about LoadableTerm.

ssanderson:
The engine already cares about LoadableTerm vs. ComputableTerm, so we
can remove the need to have that knowedge in TermGraph.


richafrank:
@ssanderson Is this still something we should try to pull in?

PR #2096, opened 2018-01-30
BUG: call get_account(True)

vikram-narayan:
In `handle_minute_close` and `handle_market_close`,
`get_account` with `performance_needs_update=False`
is called, so the account values are not fully
updated.


vikram-narayan:
Closing this in favor of the upcoming metrics refactor.

PR #2098, opened 2018-02-03
MAINT: bump empyrical to 0.3.4

vikram-narayan:
bump empyrical to 0.3.4 to fix rolling metrics and add perf attrib to empyrical


ssanderson:
the empyrical changes look good to me once we figure out the build

ssanderson:
LGTM when the build passes.

PR #2101, opened 2018-02-05
labelarray boundary condition

llllllllll:
fix an issue where labelarray construction checked the size of the categories to compute the word size before injecting the missing value.


yankees714:
LGTM

Issue #2102, opened 2018-02-09
zipline cli errors due to pandas-datareader conflict


ybryan:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment 

* Operating System: (Windows Version or `$ uname --all`) 
**1. MacOS 10.13.3**
**2. Linux 0af9b3940091 4.9.60-linuxkit-aufs #1 SMP Mon Nov 6 16:00:12 UTC 2017 x86_64 GNU/Linux**

* Python Version: **Python 3.5.4**
* Python Bitness: **64**
* How did you install Zipline: (`pip`, `conda`, or `other (please explain)`)
>1. docker from 70f8d14
>2. virtualenv + `pip install zipline` (from latest master 70f8d14)
* Python packages: `$ pip freeze` or `$ conda list` 
*Mac OS virtualenv 3.5.4*
```alembic==0.9.7
bcolz==0.12.1
Bottleneck==1.2.1
certifi==2018.1.18
chardet==3.0.4
click==6.7
contextlib2==0.5.5
cyordereddict==1.0.0
Cython==0.27.3
decorator==4.2.1
empyrical==0.3.4
idna==2.6
intervaltree==2.1.0
Logbook==1.2.1
lru-dict==1.1.6
lxml==4.1.1
Mako==1.0.7
MarkupSafe==1.0
multipledispatch==0.4.9
networkx==1.11
numexpr==2.6.4
numpy==1.14.0
pandas==0.18.1
pandas-datareader==0.6.0
patsy==0.5.0
python-dateutil==2.6.1
python-editor==1.0.3
pytz==2017.3
requests==2.18.4
requests-file==1.4.3
requests-ftp==0.3.1
scipy==1.0.0
six==1.11.0
sortedcontainers==1.5.9
SQLAlchemy==1.2.2
statsmodels==0.8.0
tables==3.4.2
toolz==0.9.0
urllib3==1.22
wrapt==1.10.11
zipline==1.1.1+240.g70f8d147
```
Now that you know a little about me, let me tell you about the issue I am
having:

# Description of Issue

`zipline ingest ` and all zipline cli commands fail due to pandas version conflict. Really due to pandas-datareader dependency on pandas 0.19.2

* What did you expect to happen?
Progress bar on zipline ingest and bcolz data input
* What happened instead?
```
Traceback (most recent call last):
[...]

PR #2104, opened 2018-02-09
MAINT: Latest pandas-datareader requires a newer pandas version

richafrank:
than zipline supports.

For https://github.com/quantopian/zipline/issues/2102


setup.py
```diff
@@ -269,6 +270,7 @@ def setup_requirements(requirements_path, module_names, strict_bounds,
         )
     return module_lines
 
+
```
freddiev4: I'm guessing this was added because of a `flake8` error?

richafrank: Yup, exactly

PR #2107, opened 2018-02-09
ENH: Move benchmark logic to bundle

freddiev4:
**Problem:**
Previously, we made calls to services such as Yahoo and Google Finance for data about `SPY`, used as our default benchmark, and made these requests every time we ran a backtest. Eventually, we switched to IEX Trading after the other two services broke on several occasions (more details back in https://github.com/quantopian/zipline/pull/2031)

**Moving Benchmark Logic:**
To prevent further errors from the service we're getting `SPY` data from (i.e.the service goes down) and to reduce the number of requests we're making, this PR makes it so that we only make a request to IEX when running `zipline ingest`.

**Alternative Measure:**
This also adds a "fake" asset called `ZPLN`, which is in OHLCV format and filled with zeroes.

If we want to skip that request and just have zeroes for benchmark metrics, we can call `set_benchmark('ZPLN')` in our algos

OR

If the request to IEX fails, then the backtest will use `ZPLN` instead of `SPY`.

Hopefully no company IPOs with `ZPLN` as their symbol anytime soon 🙂 

**Bundle Core:**
This adds a `MetadataReader` and `MetadataWriter` to `zipline.data.bundles.core`, which are used to store the `default_benchmark_sid` that we want to use for our bundles.


freddiev4:
Expecting a lot of test failures 🙂

yiorgosn:
Freddie,

Don't use ZPLN as a ticker, try NO_BCHMK or some variant of this so it is evident also in any further reports that there is NO benchmark being used.  I would also throw a message when the NO_BCHMK  is being used.

Thank you for the great work.

Best,

PR #2110, opened 2018-02-14
PERF: Variable-size minutely cache.

ssanderson:
Use a variable-size cache for minutely pricing data.

Increase the default cache size for close to 3000, since close prices
are used in many places in the simulation as the best-known price of
assets. This dramatically speeds up algorithms that read the prices of
many assets without ordering them.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -906,7 +906,8 @@ class BcolzMinuteBarReader(MinuteBarReader):
         'low': 1550,
         'volume': 1550,
     }
-    assert set(FIELDS) == set(DEFAULT_MINUTELY_SID_CACHE_SIZES)
+    assert set(FIELDS) == set(DEFAULT_MINUTELY_SID_CACHE_SIZES), \
+        "FIELDS should match DEFAULT_MINUTELY_SID_CACHE_SIZES keys"
 
     # Wrap the defaults in proxy so that we don't accidentally mutate them in
     # place in the constructor. If a user wants to change the defaults, they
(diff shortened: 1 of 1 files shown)


zipline/data/minute_bars.py
```diff
@@ -898,8 +899,22 @@ class BcolzMinuteBarReader(MinuteBarReader):
...
     """
     FIELDS = ('open', 'high', 'low', 'close', 'volume')
+    DEFAULT_MINUTELY_SID_CACHE_SIZES = {
+        'close': 3000,
+        'open': 1550,
+        'high': 1550,
+        'low': 1550,
+        'volume': 1550,
+    }
+    assert set(FIELDS) == set(DEFAULT_MINUTELY_SID_CACHE_SIZES)
```
llllllllll: put a message on this

zipline/data/minute_bars.py
```diff
@@ -898,8 +899,22 @@ class BcolzMinuteBarReader(MinuteBarReader):
     zipline.data.minute_bars.BcolzMinuteBarWriter
     """
     FIELDS = ('open', 'high', 'low', 'close', 'volume')
+    DEFAULT_MINUTELY_SID_CACHE_SIZES = {
+        'close': 3000,
```
llllllllll: are we sure we are okay with doubling this?

ssanderson: I believe so.  https://github.com/quantopian/zipline/pull/2108 shrinks the expected size of the volume cache significantly for most algorithms, so we have some wiggle room.  For Quantopian, even if we were fully saturating all the caches, the extra memory cost of increasing just this field isn't that big of a deal, and the performance cost of thrashing this cache is bad enough that I think any zipline user who is doing so would almost certainly prefer the extra memory cost here in exchange for the substantil (~50% or more in local testing) speedup this provides.

PR #2112, opened 2018-02-16
TST: Add compression to QUANDL_ARCHIVE.zip

adegtyarev:
By default, the contents of 'zipfile' archive are not compressed. This change sets compression mode of the generated sample thus reducing file size by ~70%.


ssanderson:
This looks reasonable to me.  @ernestoeperez88 any reason we shouldn't be compressing this file?

ernestoeperez88:
@adegtyarev thanks for catching this, and for the PR.

PR #2120, opened 2018-03-07
TST: add instance_of wrapper

llllllllll:


zipline/testing/predicates.py
```diff
@@ -84,7 +84,37 @@ def __ne__(other):
...
+    def __eq__(self, other):
+        if self.exact:
+            return type(other) in self.types
+
+        return isinstance(other, self.types)
+
+    def __repr__(self):
+        return '<%s: %r%s>' % (
+            type(self).__name__,
+            self.types[0] if len(self.types) == 1 else self.types,
```
ssanderson: This is going to crash on the `len` call if `types` is a single type.

zipline/testing/predicates.py
```diff
@@ -84,7 +84,37 @@ def __ne__(other):
...
+        self.exact = exact
+
+    def __eq__(self, other):
+        if self.exact:
+            return type(other) in self.types
+
+        return isinstance(other, self.types)
+
+    def __repr__(self):
+        return '<%s: %r%s>' % (
```
ssanderson: Is there a reason not to just make this match exactly what the call would look like? `instance_of(str)` seems pretty readable to me?

zipline/testing/predicates.py
ssanderson: Oh, nvm, you're converting at construction time.

llllllllll: it is always a tuple, check __init__

zipline/testing/predicates.py
```diff
@@ -84,7 +84,37 @@ def __ne__(other):
...
+
+    Parameters
+    ----------
+    types : type or tuple[type]
+        The types to compare equal to.
+    exact : bool, optional
+        Only compare equal to exact instances, not instances of subclasses?
+    """
+    def __init__(self, types, exact=False):
+        if not isinstance(types, tuple):
```
ssanderson: Might be worth asserting `isinstance(types, type)` here?

PR #2124, opened 2018-03-12
MAINT: Bump empyrical.

ssanderson:


richafrank:
Version and hash bump look right to me.

PR #2126, opened 2018-03-13
BUG: fix session_distance for flipped ranges

llllllllll:


tests/calendars/test_trading_calendar.py
```diff
@@ -661,8 +661,17 @@ def test_sessions_window(self):
     def test_session_distance(self):
         sessions = self._get_session_block()
 
-        self.assertEqual(2, self.calendar.session_distance(sessions[0],
-                                                           sessions[-1]))
+        forward_distance = self.calendar.session_distance(
```
ahgnaw: can we add a case for the zero distance?

ahgnaw:
LGTM

PR #2128, opened 2018-03-19
make it easy to register new assert_equal dispatches

llllllllll:
Adds a small helper to raise consistent looking error messages. I am not going to port the other functions in here to use this because I don't think it is worth refactoring the existing errors, I just want this to register an assertion somewhere else.


ssanderson:
Oh.  More significant issue, this always raises in the base dispatch case...

llllllllll:
updated the name, I agree that it looks confusing in dir

ssanderson:
LGTM (round two)

PR #2136, opened 2018-04-03
MAINT: Update Python stub files

freddiev4:
From: http://www.zipline.io/release-process.html#updating-the-python-stub-files


freddiev4:
I think I'm just gonna go ahead and merge this once CI passes if that's alright (considering we're just updating the docstrings here)

freddiev4:
Thanks David!

PR #2143, opened 2018-04-09
DOC: add import to beginner tutorial to fix NameError

aivensong:
The DMA cross-over working example was missing an import of `matplotlib.pyplot` which led to a `NameError: global name 'plt' is not defined` when running in an iPython notebook.


freddiev4:
@aivensong good catch! Thanks a lot 🙂

PR #2155, opened 2018-04-26
ENH: don't copy the sids and values in the core loader

llllllllll:
small win, about a 1/5 seconds for 1 year of market cap data


llllllllll:
@freddiev4 can you take a look at this one also?

freddiev4:
@llllllllll looking now

PR #2164, opened 2018-05-01
BUG: Prevent negative lookup in history.

ehebert:
Raise a value error on negative bar counts to prevent lookups ahead of
simulation time.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -642,6 +642,20 @@ class MinuteEquityHistoryTestCase(WithHistory, ZiplineTestCase):
         with self.assertRaises(HistoryInInitialize):
             test_algo.initialize()
 
+    def test_negative_bar_count(self):
+        """
+        Negative bar counts leak future information.
+        """
+        with self.assertRaises(ValueError):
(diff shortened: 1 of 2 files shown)


tests/test_history.py
```diff
@@ -642,6 +642,20 @@ def handle_data(context, data):
         with self.assertRaises(HistoryInInitialize):
             test_algo.initialize()
 
+    def test_negative_bar_count(self):
+        """
+        Negative bar counts leak future information.
+        """
+        with self.assertRaises(ValueError):
```
ssanderson: Should we assert that this raises the ValueError we expect? I could imagine messing up the test setup here in a way that would cause a different ValueError to be raised.

ehebert: Good point. I'll change over to use a `assertRaisesRegexp`.

ehebert: @ssanderson, here is the change you requested, https://github.com/quantopian/zipline/pull/2164/commits/a1ae9a32cc356747751a385736f81ee538894dda

PR #2165, opened 2018-05-01
Use fixtures in test algo

ssanderson:


tests/pipeline/test_pipeline_algo.py
```diff
@@ -85,13 +85,20 @@ def rolling_vwap(df, length):
...
     sids = 1, 2, 3
     START_DATE = pd.Timestamp('2014-01-01', tz='utc')
     END_DATE = pd.Timestamp('2014-02-01', tz='utc')
     dates = date_range(START_DATE, END_DATE, freq=get_calendar("NYSE").day,
                        tz='utc')
 
+    SIM_PARAMS_DATA_FREQUENCY = 'daily'
+    DATA_PORTAL_USE_MINUTE_DATA = False
+
+    # FIXME: This currently uses benchmark returns from the trading
```
ssanderson: This is in service of a larger PR that aims to eventually kill `TradingEnvironment` entirely, so in the interest of changing as little as possible here, I'm not changing this behavior.

ehebert: :+1:

PR #2166, opened 2018-05-01
MAINT: Update before_trading_start timing tests.

ssanderson:


tests/test_tradesimulation.py
```diff
@@ -17,92 +17,76 @@
...
-
-FREQUENCIES = {'daily': 0, 'minute': 1}  # daily is less frequent than minute
-
-
-class TestTradeSimulation(WithTradingEnvironment, ZiplineTestCase):
-
-    def fake_minutely_benchmark(self, dt):
-        return 0.01
-
-    def test_minutely_emissions_generate_performance_stats_for_last_day(self):
```
ssanderson: I'm not sure what this test was originally supposed to test, but it looks like for a while it's just beeing testing the length of `sim_params.sessions` in a very confusing way...

ehebert: Going back in the logs this had originally been testing the value of `algo.perf_tracker.day_count`. `day_count` was removed with the risk rewrite, so I believe this test can be safely removed.

PR #2171, opened 2018-05-03
Update test_algorithm.py to use WithMakeAlgo.

ssanderson:
Updated about half of `test_algorithm.py` to use `WithMakeAlgo`.

Most of these changes are straightforward no-op changes. A notable exception is `TestTransformAlgorithm`, which had evolved organically enough over the years that I decided to just scrap it and re-write the useful tests from the suite in a modern style. Those new tests live in `test_ordering.py`


tests/test_algorithm.py
```diff
@@ -197,12 +163,27 @@
...
-    ASSET_FINDER_EQUITY_SIDS = 133,
+class TestRecord(zf.WithMakeAlgo, zf.ZiplineTestCase):
+    ASSET_FINDER_EQUITY_SIDS = (133,)
+    SIM_PARAMS_DATA_FREQUENCY = 'daily'
+    DATA_PORTAL_USE_MINUTE_DATA = False
 
     def test_record_incr(self):
-        algo = RecordAlgorithm(sim_params=self.sim_params, env=self.env)
-        output = algo.run(self.data_portal)
+
```
ssanderson: A general theme of this PR is that I tried to "inline" algorithms into test sites if they were small and if there were assertions in the test body that depended on the behavior of the algorithm.

For tests that were just "run algorithm X", I didn't see as much value in doing the inlining.

tests/test_algorithm.py
```diff
@@ -214,16 +195,16 @@ def test_record_incr(self):
...
-                           WithDataPortal,
-                           ZiplineTestCase):
+class TestMiscellaneousAPI(zf.WithMakeAlgo, zf.ZiplineTestCase):
 
     START_DATE = pd.Timestamp('2006-01-03', tz='UTC')
     END_DATE = pd.Timestamp('2006-01-04', tz='UTC')
     SIM_PARAMS_DATA_FREQUENCY = 'minute'
     sids = 1, 2
 
+    # FIXME: Pass a benchmark source instead of this.
```
ssanderson: The default behavior of `WithMakeAlgo` is to use the first equity sid as a benchmark, but that often doesn't work, because if you run an algorithm from START_DATE to END_DATE, then you don't have enough days of historical data to calculate benchmark returns (in particular, you need the close from the day before START_DATE to compute a daily benchmark return for the first day of the backtest.

Setting BENCHMARK_SID to None causes us to use benchmark returns retrieved from the TradingEnvironment, which uses a static set of data that's checked into git.  I'm not sure what we want long-term for benchmarks in testing, but this is what we were already doing in all of these tests, and I didn't want to layer in more changes than were necessary.  Once we have all these suites creating algorithms in a uniform way, it will be easier to make broad changes to how we handle benchmarks.

tests/test_algorithm.py
```diff
@@ -782,151 +734,96 @@ def test_future_symbol(self):
...
+            self.assertEqual(zipline.api.symbol('DUP').sid, self.sids[1])
 
             with self.assertRaises(UnsupportedDatetimeFormat):
-                algo.set_symbol_lookup_date('foobar')
+                set_symbol_lookup_date('foobar')
+
+        self.run_algorithm(initialize=initialize)
 
 
-class TestTransformAlgorithm(WithLogger,
```
ssanderson: `TestTransformAlgorithm` was kind of a mess. Several of the tests in this suite turned out to be complicated no-ops (e.g., the tests for ordering futures ran an algorithm that only ordered equities).  Also, none of these tests were about the old `transform` API, which is what the name of this suite originally referenced.  It ultimately seemed easier and clearer to just break this up and re-write the tests worth keeping.  Most of the re-written tests now live in `test_ordering.py`.

tests/test_algorithm.py
```diff
@@ -935,186 +832,16 @@ def test_run_twice(self):
...
-        sim_params = factory.create_simulation_parameters(
-            num_days=4, data_frequency='minute')
-
-        algo = TestRegisterTransformAlgorithm(
-            sim_params=sim_params,
-            env=self.env,
-        )
-        self.assertEqual(algo.sim_params.data_frequency, 'minute')
-
-    def test_order_rounding(self):
```
ssanderson: I moved this into its own suite.

tests/test_algorithm.py
```diff
@@ -1167,22 +894,51 @@ def make_future_minute_bar_data(cls):
         )
         return ((sid, frame) for sid in sids)
 
-    def test_empty_portfolio(self):
-        algo = EmptyPositionsAlgorithm(self.asset_finder.equities_sids,
-                                       sim_params=self.sim_params,
-                                       env=self.env)
-        daily_stats = algo.run(self.data_portal)
+    def test_portfolio_exited_position(self):
```
ssanderson: This is another algorithm that was inlined.

tests/test_algorithm.py
```diff
@@ -1284,13 +1038,13 @@ def make_equity_minute_bar_data(cls):
             yield sid, create_minute_df_for_asset(
                 cls.trading_calendar,
                 cls.data_start,
-                cls.sim_params.end_session,
+                cls.END_DATE,
```
ssanderson: Many of the `make_*_bar_data` methods used `sim_params` to construct data, but since there's no ordering dependency between `WithSimParams` and `With*BarReader`, `cls.sim_params` isn't guaranteed to exist when this method is called.  The right thing to do here is usually to use the calendar, which is a direct dependency of `With*BarReader`, and is therefore guaranteed to be initialized by the time this method is called.

tests/test_algorithm.py
```diff
@@ -1669,69 +1395,52 @@ def make_equity_info(cls):
...
-                    [100] * days,
-                    timedelta(days=1),
-                    cls.sim_params,
-                    cls.trading_calendar)
-            },
-            index=cls.equity_daily_bar_days,
-        )
+        cal = cls.trading_calendars[Equity]
+        sessions = cal.sessions_in_range(cls.START_DATE, cls.END_DATE)
+        frame = pd.DataFrame({
```
ssanderson: The original impetus for rewriting this method was removing the dependency on `cls.sim_params` (see note above), but I think this also makes it much clearer what the price data is in this suite.

tests/test_algorithm.py
```diff
@@ -4755,10 +4366,11 @@ def make_equity_info(cls):
             orient='index',
         )
 
-    @classmethod
-    def init_class_fixtures(cls):
-        super(TestOrderAfterDelist, cls).init_class_fixtures()
-        cls.data_portal = FakeDataPortal(cls.env)
+    # XXX: This suite doesn't use the data in its DataPortal; it uses a
+    # FakeDataPortal with different mock data.
+    def init_instance_fixtures(self):
```
ssanderson: This needs to be in `init_instance_fixtures` because `WithDataPortal` sets `self.data_portal` as a instance attribute, which takes precedence over a class-level attribute.

tests/test_algorithm.py
ehebert: Thank you. I agree with the inlining. While debugging (or just reading tests), this will reduce cognitive load.

tests/test_algorithm.py
ehebert: Thanks for cleaning this one up.

ahgnaw:
I prodded around for a bit and couldn't find anything, lgtm

ssanderson:
@ehebert I think it probably makes sense to remove FakeDataPortal, but I don't have time to do that before pycon and I don't want this to go stale while I'm gone.

ehebert:
> @ehebert I think it probably makes sense to remove FakeDataPortal, but I don't have time to do that before pycon and I don't want this to go stale while I'm gone.

I agree with merging this without removing `FakeDataPortal`. I may have not been clear, I was offering to do the removal as a follow up to this PR.

PR #2188, opened 2018-05-16
MAINT: Remove unused code.

ssanderson:
Removes unused `serialization_utils.py`. Parts of this are still used
downstream at Quantopian, but there are no more zipline consumers, so
I've moved this to the downstream consumer.


richafrank:
Looks good to me! (assuming consumers are updated)

Issue #2132, opened 2018-03-31
Update to support Pandas 0.22


YongBeomKim:
Dear Zipline Maintainers,
# Environment
* Operating System: (ubuntu mate 17.04)
* Python Version: 3.6.4
* How did you install Zipline: `pip`,

# Description of Issue
https://github.com/quantopian/zipline/issues/2043 [link](https://github.com/quantopian/zipline/issues/2043)
@nateGeorge  is fixed  to use in Pandas 0.22 (New Version)
please chack & update on your Source

mickeydonkey:
there is a branch updating the pandas and other relevant packages, but it has been dominant for several weeks.

freddiev4:
Hi @YongBeomKim thanks for opening the issue. There is a branch open that's a WIP for supporting a newer pandas version. It'll take some time to finish as the remaining work is for the Pipeline API.

richafrank:
Linking to previous PR and ongoing discussion: https://github.com/quantopian/zipline/pull/2086#issuecomment-370858680

twiecki:
Pandas 0.23.0 is the most recent.

richafrank:
Closing - See https://github.com/quantopian/zipline/pull/2194 and https://github.com/quantopian/zipline/issues/2243

PR #2194, opened 2018-05-23
Pandas 0.22 upgrade

richafrank:
Makes zipline compatible with pandas 0.19 and 0.22, but lacks support for pipeline categoricals.
Part of https://github.com/quantopian/zipline/issues/2132.


appveyor.yml
```diff
@@ -27,6 +27,16 @@ environment:
     - PYTHON_VERSION: "2.7"
     - PYTHON_VERSION: "3.5"
 
+    - PYTHON_VERSION: "3.5"
+      PANDAS_VERSION: "0.19.2"
+
+    - PYTHON_VERSION: "3.5"
+      PANDAS_VERSION: "0.22.0"
```
twiecki: I think 0.23.0 is the most recent.

setup.py
```diff
@@ -171,8 +171,7 @@ def _filter_requirements(lines_iter, filter_names=None,
 
 REQ_UPPER_BOUNDS = {
     'bcolz': '<1',
-    'pandas': '<0.19',
-    'pandas-datareader': '<0.6',  # 0.6.0 requires pandas >=0.19.2
+    'pandas': '<=0.22',
```
aseyboldt: Shouldn't this be "~=0.22" (or hopefully "~=0.23" if that works as well)? Otherwise bugfix releases of pandas will not be accepted.

richafrank: Without further changes, 0.23 didn't work, so I didn't add it to this PR (would love not to add more scope to this instead of merging what we have). 

We could make it `<0.23`, if we think it's safe to support the micro versions. We have seen in the past that micro versions of pandas have changed behavior and broken things...

zipline/resources/market_data/SPY_benchmark.csv
```diff
@@ -5097,1030 +5097,1261 @@
...
-2013-05-29 00:00:00+00:00,-0.006494287432351253
-2013-05-30 00:00:00+00:00,0.003692046967679552
-2013-05-31 00:00:00+00:00,-0.01441234999698493
-2013-06-03 00:00:00+00:00,0.005567792462065624
-2013-06-04 00:00:00+00:00,-0.004806814724672925
-2013-06-05 00:00:00+00:00,-0.014000978234287031
-2013-06-06 00:00:00+00:00,0.00905314069572749
-2013-06-07 00:00:00+00:00,0.01272045719904158
+2013-05-01 00:00:00+00:00,-0.008768026837593923
+2013-05-02 00:00:00+00:00,0.009287480470740572
```
ssanderson: Why the changes here?

zipline/utils/numpy_utils.py
```diff
@@ -321,7 +324,7 @@ def rolling_window(array, length):
 
 # Sentinel value that isn't NaT.
 _notNaT = make_datetime64D(0)
-iNaT = NaTns.view(int64_dtype)
+iNaT = int(NaTns.view(int64_dtype))
```
ssanderson: I would have expected this to be an int64. What's the reason for converting to a python int here?

zipline/utils/pandas_utils.py
```diff
@@ -12,6 +12,14 @@
 from distutils.version import StrictVersion
 
 pandas_version = StrictVersion(pd.__version__)
+new_pandas = pandas_version >= StrictVersion('0.19')
+skip_pipeline_new_pandas = \
+    'Pipeline categoricals are not yet compatible with pandas >=0.19'
+
+if pandas_version >= StrictVersion('0.20'):
```
ssanderson: I think this function is supposed to be private. Could we just define our own version of this using the `normalize` method of timestamps? E.g.:

```
def normalize_date(dt):
    return dt.normalize()
```

zipline/utils/run_algo.py
```diff
@@ -148,7 +148,13 @@ def _run(handle_data,
                 "invalid url %r, must begin with 'sqlite:///'" %
                 str(bundle_data.asset_finder.engine.url),
             )
-        env = TradingEnvironment(asset_db_path=connstr, environ=environ)
+        env = TradingEnvironment(
```
ssanderson: Why the changes here? (I'm about to remove this class entirely, so I'm surprised to see nontrivial changes involving it.)

tests/__init__.py
```diff
@@ -0,0 +1 @@
+from zipline import setup, teardown  # noqa For nosetests
```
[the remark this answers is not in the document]
richafrank: I put them in `zipline.__init__` so that doctests running on code in `zipline` would apply the compat.

zipline/utils/numpy_utils.py
richafrank: The only usage of this now (it was unused before, except in the assert below) is to represent a missing `first_trading_day` attribute in the bcolz ctable for us equity pricing, which requires json-serializability. We used to use `pandas.tslib.iNaT` (a [`long`](https://github.com/numpy/numpy/blob/master/numpy/core/include/numpy/npy_common.h#L420-L421)) for this.

tests/__init__.py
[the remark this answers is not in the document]
ssanderson: Ah. Can we add a comment explaining that?

zipline/resources/market_data/SPY_benchmark.csv
richafrank: I needed to rebuild the example data for use in test_examples with the new pandas version. That required downloading benchmarks, but our current benchmarks downloading process doesn't go back as far as the current data. I merged in data here from `resources/market_data_SPY_benchmark.csv`, which had a different range of dates, but also some floating point differences where they overlap (such as here). Another option might be to change the start/end dates of the test, but I was hoping to avoid that.

zipline/utils/pandas_utils.py
richafrank: Sounds good to me.

zipline/utils/run_algo.py
richafrank: This was also for test_examples. Previously we ignore the cached benchmark data if it doesn't cover the requested date range, but that date range was always the entire trading calendar (`trading_calendar.schedule.index`). Now it's filtered to the range of the backtest, so we can use the cached benchmarks.

tests/__init__.py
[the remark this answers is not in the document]
richafrank: 👍

ssanderson:
@richafrank this looks good to me. We should probably open a tracking issue to document the known failures with categoricals in old pandas.

It also might make sense to do a release with these changes and the `trading_calendars` refactor before we merge the breaking changes from https://github.com/quantopian/zipline/pull/2192.

twiecki:
🎉 👍 🥇

twiecki:
Should we release a new version with the updated pandas dependency?

richafrank:
@twiecki Great idea 😄 . [zipline 1.3.0](https://github.com/quantopian/zipline/releases/tag/v1.3.0) has this change.

PR #2199, opened 2018-05-31
Added loaders parameter to run_algorithm() in run_algo.py

calmitchell617:
Added an optional parameter to run_algorithm() that takes a dictionary of PipelineLoader objects. Allows you to load Columns of data that aren't from USEquityLoader while running a backtest.

This is my first pull request on any meaningful project, constructive criticism is very welcome.


zipline/utils/run_algo.py
```diff
@@ -168,9 +169,12 @@ def _run(handle_data,
         def choose_loader(column):
             if column in USEquityPricing.columns:
                 return pipeline_loader
-            raise ValueError(
-                "No PipelineLoader registered for column %s." % column
-            )
+            elif column in loaders:
```
ssanderson: I think this will crash in the current implementation if the user doesn't supply any custom loaders (because `loaders` will be None, so `column in loaders` will barf).

calmitchell617:
@ssanderson, I changed some language to show that this argument is used for loading dataframes, and changed the ```choose_loader()``` function so it doesn't freak out when loaders is None:
```
def choose_loader(column):
            if column in USEquityPricing.columns:
                return pipeline_loader
            elif data_frame_loaders and column in data_frame_loaders:
                return data_frame_loaders[column]
            else:
                raise ValueError(
                    "No PipelineLoader registered for column %s." % column
                )
```
I will try to implement your suggestions. 

Just playing devil's advocate, can you think of any cases where it would be better to have separate, optional arguments for different types of loaders, in the ```run_algorithm()``` function? (eg: an argument for Dataframe loaders, and a different argument for Dataset loaders).

zipline/utils/run_algo.py
```diff
@@ -71,7 +71,8 @@ def _run(handle_data,
          print_algo,
          metrics_set,
          local_namespace,
-         environ):
+         environ,
+         data_frame_loaders):
```
llllllllll: I don't think there is anything special about this being a dataframe loader, right?

zipline/utils/run_algo.py
```diff
@@ -344,6 +349,9 @@ def run_algorithm(start,
     environ : mapping[str -> str], optional
         The os environment to use. Many extensions use this to get parameters.
         This defaults to ``os.environ``.
+    loaders : iterable{PipelineLoader}, optional
```
llllllllll: the docstring and parameter name don't match here.

jnazaren:
I've created PR #2246 with one of the solutions suggested above

PR #2200, opened 2018-05-31
ENH: Custom blotter support

jnazaren:
Adds support for custom blotter registration, and an argument in run_algorithm where a string mapping to a custom blotter class can be specified.


jnazaren:
@llllllllll I have a feeling this will look somewhat different with @ssanderson's updates

zipline/finance/blotter/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    register,
+    unregister,
+    load,
+)
+
+__all__ = ['SimulatedBlotter', 'unregister', 'blotters', 'load']
+
+
+@register('default')
+def default_blotter():
```
llllllllll: Is there a reason to wrap this in a function that returns the class instead of just registering the class directly?

zipline/finance/blotter/blotter.py
```diff
@@ -0,0 +1,188 @@
...
+
+        self.current_dt = None
+
+    def __repr__(self):
+        raise NotImplementedError
+
+    def set_date(self, dt):
+        self.current_dt = dt
+
+    @expect_types(asset=Asset)
```
llllllllll: Given that we are not calling these functions, this decorator won't do anything. The docstring will be enough to communicate the types.

zipline/finance/blotter/blotter.py
```diff
@@ -0,0 +1,188 @@
...
+# limitations under the License.
+from zipline.finance.cancel_policy import NeverCancel
+from zipline.assets import Asset
+from zipline.utils.input_validation import expect_types
+
+
+class Blotter(object):
+
+    def __init__(self, data_frequency, cancel_policy=None):
+
```
llllllllll: Can you remove the empty newlines between the assignments here?

zipline/finance/blotter/blotter.py
```diff
@@ -0,0 +1,188 @@
...
+        self.data_frequency = data_frequency
+
+        self.cancel_policy = cancel_policy if cancel_policy else NeverCancel()
+
+        self.current_dt = None
+
+    def __repr__(self):
```
llllllllll: all objects are repr-able by default, so we don'tneed to put it in the interface.

zipline/finance/blotter/blotter.py
```diff
@@ -0,0 +1,188 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from zipline.finance.cancel_policy import NeverCancel
+from zipline.assets import Asset
+from zipline.utils.input_validation import expect_types
+
+
+class Blotter(object):
```
llllllllll: We should make this an ABC (Abstract Base Class). For an example, look at `zipline/data/history_loader.py:HistoryLoader`

zipline/finance/blotter/blotter.py
```diff
@@ -0,0 +1,188 @@
...
+        amount > 0 :: Buy/Cover
+        amount < 0 :: Sell/Short
+        Market order:    order(asset, amount)
+        Limit order:     order(asset, amount, style=LimitOrder(limit_price))
+        Stop order:      order(asset, amount, style=StopOrder(stop_price))
+        StopLimit order: order(asset, amount, style=StopLimitOrder(limit_price,
+                               stop_price))
+        """
+
+        raise NotImplementedError
```
llllllllll: it is best practice to raise an instance of a class instead of the class itself, for example: `NotImplementedError()` vs `NotImplmentedError`. I also prefer mentioning which method was not implemented like:
```python
raise NotImplemented('order')
```

zipline/finance/blotter/core.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        try:
+            del _blotters[name]
+        except KeyError:
+            raise ValueError(
+                'blotter class %r was not already registered' % name,
+            )
+
+    def load(name):
+        """Return an instance of the blotter class registered
```
llllllllll: This doesn't return an instance, just the class itself, right?

zipline/finance/blotter/core.py
```diff
@@ -0,0 +1,123 @@
...
+        with the given name.
+
+        Returns
+        -------
+        blotter class : type
+            A new instance of the metrics set.
```
llllllllll: s/metrics set/blotter class/g

zipline/finance/blotter/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    blotters,
+    register,
+    unregister,
+    load,
+)
+
+__all__ = ['SimulatedBlotter', 'unregister', 'blotters', 'load']
+
+
+@register('default')
```
llllllllll: should we just decorate the definition of the class itself?

zipline/finance/blotter/simulatedblotter.py
```diff
@@ -19,6 +19,7 @@
 from six import iteritems
```
llllllllll: Can you rename this to `simulated_blotter.py`?

zipline/finance/blotter/blotter.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def batch_order(self, order_arg_lists):
```
llllllllll: +1 on having this part of the ABC, it makes it easier to implement.

zipline/finance/blotter/core.py
```diff
@@ -0,0 +1,123 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from functools import partial
+
+from zipline.finance.blotter.blotter import Blotter
+from zipline.utils.compat import mappingproxy
+
+
+def _make_blotters_core():
```
ssanderson: This looks like it's copying the pattern of the `make_bundle_core` functionality, so this is probably more of a question for @llllllllll than for @jnazaren, but is there a reason we're using this odd javascript-style closure construct instead of just defining a class?

zipline/finance/blotter/__init__.py
jnazaren: I wanted to wrap this so that like with metrics, we would store each blotter class as a function (and load it by calling that function), which would allow people to use decorators in general. It was basically just for consistency purposes, because otherwise I don't exactly know how we could use decorators.

zipline/finance/blotter/core.py
ssanderson: An alternative (and I would argue, much clearer and simpler) way to build a similar registration API would be to have a global instance of something like a `BlotterRegistry` class.  For example, we do something analogous for trading calendars here: https://github.com/quantopian/zipline/blob/d4de30860b15e18aa5814167f8bceb98a71a15b5/zipline/utils/calendars/calendar_utils.py#L39-L267

llllllllll: Originally it was just a few functions in a module with a global dict, and then I wanted to test it in an isolated way so I just copied them into a closure. Making it a class with methods is fine, but I don't have a strong opinion on this. If you have even a slight preference it's fine to change it.

ssanderson: I'm definitely on board with encapsulating the global state in a way that makes isolated testing possible. My thought here is just that having a closure that defines a dictionary and a bunch of functions that share the state of that dictionary is essentially a roundabout way of creating an instance of a class.

zipline/finance/blotter/blotter.py
jnazaren: Ok, I've taken it out

zipline/finance/blotter/blotter.py
jnazaren: Yep

zipline/finance/blotter/blotter.py
jnazaren: took it out

zipline/finance/blotter/blotter.py
jnazaren: Done. I've also added @abstractmethod decorators to the unimplemented functions

zipline/finance/blotter/blotter.py
jnazaren: Changed that

zipline/finance/blotter/core.py
jnazaren: Yes, the functions registered are meant to return classes rather than instances. See the return value of defualt_blotter() in __init__.py - that's what you're looking for, correct?

jnazaren: Yeah sorry about that, just realized I wasn't really paying attention to the comments when I copied them over

zipline/finance/blotter/__init__.py
jnazaren: Yes, that could be another option - but again, similarly to the first comment at the very top, if we switch to handling classes getting passed in from decorators, that's fine, but I'm not sure we would be able to support user-defined functions getting passed in at the same time... unless we don't want to do that or if we distinguish by type of thing being passed in, that is - does that sound like something we would want to have?

zipline/finance/blotter/simulatedblotter.py
jnazaren: Done

zipline/finance/blotter/core.py
jnazaren: fixed

zipline/finance/blotter/blotter.py
jnazaren: 👍

jnazaren: Also let me know if there are any similar constraint fields/methods I might want to implement

zipline/finance/blotter/__init__.py
llllllllll: You can decorate a class directly, just like you would a function. For example:

```python
@register('my-blotter')
class MyBlotter(Blotter):
    ...
```

tests/blotter/test_core.py
```diff
@@ -0,0 +1,173 @@
...
+        msg = "blotter class 'ayy-lmao' is already registered"
+        with assert_raises_str(ValueError, msg):
+            @self.register('ayy-lmao')
+            class Fake(object):
+                pass
+
+        msg = "The class specified is not a subclass of Blotter"
+        with assert_raises_str(TypeError, msg):
+            @self.register('something-different')
+            class ImproperDummyBlotter(object):
```
llllllllll: why does this raise `TypeError` but `Fake` raised a `ValueError`

tests/blotter/test_core.py
```diff
@@ -0,0 +1,173 @@
...
+        with assert_raises_str(ValueError, msg):
+            self.load('ayy-lmao')
+
+        msg = "blotter class 'ayy-lmao' was not already registered"
+        with assert_raises_str(ValueError, msg):
+            self.unregister('ayy-lmao')
+
+    def test_register_non_decorator(self):
+
+        class ProperDummyBlotter(Blotter):
```
llllllllll: To reduce the number of things that need to change if we change the blotter, can we just use an empty subclass of `SimulatedBlotter` here?

zipline/finance/blotter/core.py
llllllllll: @jnazaren Can you do this?

tests/blotter/test_core.py
jnazaren: It's testing 2 different types of issues that can come up when registering blotter classes - namely one is testing a class getting registered with a duplicate name, while the other is testing the registration of a class that isn't a subclass of Blotter. I can change this, or just rename the classes to make it more obvious - let me know what works better.

tests/blotter/test_core.py
jnazaren: Sure

tests/blotter/test_core.py
llllllllll: Oh, I misread this and thought the error messages were the same. that's my bad.

zipline/finance/blotter/core.py
jnazaren: Ok, will this need a 'factories' storage like the one with calendars for lazy instantiation?

zipline/finance/blotter/blotter_utils.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        self._blotter_factories.clear()
+
+
+# Global blotter class dispatcher
+global_blotter_class_dispatcher = BlotterClassDispatcher(
+    blotter_factories={}
+)
+
+get_blotter_class = global_blotter_class_dispatcher.get_blotter_class
```
llllllllll: sorry for all the churn on this, but could we rename these free functions and methods to drop the `_blotter_class` suffix. The expected use is something like:

```
from zipline import blotter


@blotter.register
class ...
```
in which case we have a prefix to disambiguate. This looks more like the other user extension points like bundles and metrics sets too.

zipline/utils/run_algo.py
```diff
@@ -5,6 +5,10 @@
 import warnings
 
 import click
+
+from zipline.finance.blotter.blotter import Blotter
```
llllllllll: we should import these things both from `zipline.finance.blotter` directly instead of going to the submodules. The reason to reexport the interface through the `__init__` is that consumers (like this file) don't need to know about the layout of the `blotter` directory.

zipline/finance/blotter/blotter_utils.py
```diff
@@ -0,0 +1,112 @@
+from functools import partial
```
llllllllll: Could we call this something like `dispatcher.py` so it is more clear what this contains?

llllllllll:
I just ran a sample and it worked, so maybe we are trampling the registered set somewhere in the tests.

zipline/finance/blotter/blotter_utils.py
jnazaren: Ok, I'll do that

zipline/utils/run_algo.py
jnazaren: Done

tests/blotter/test_core.py
```diff
@@ -0,0 +1,119 @@
...
+from zipline.utils.compat import mappingproxy
+
+bcd = BlotterClassDispatcher(blotter_factories={})
+
+
+class BlotterCoreTestCase(ZiplineTestCase):
+
+    def init_instance_fixtures(self):
+        super(BlotterCoreTestCase, self).init_instance_fixtures()
+        bcd.clear()
```
llllllllll: instead of clearing this every time, why don't we just create a new one to be safe?

zipline/finance/blotter/core.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    A class for registering and dispatching custom blotter classes.
+
+    Method of a global instance of this class are provided by
+    zipline.utils.calendar_utils.
+
+    Parameters
+    ----------
+    blotter_factories : dict[str -> function]
+        Factories for lazy blotter creation.
```
llllllllll: Are these actually factories and not just the classes themselves?
This dict should be `dict[str, type]` right?

zipline/finance/blotter/core.py
```diff
@@ -0,0 +1,111 @@
...
+            raise ValueError("no blotter class registered as %r, options "
+                             "are: %r" % (name, sorted(self._blotter_factories
+                                                       )))
+
+    def class_exists(self, name):
+        """
+        Whether or not the global list of blotter classes contains the
+        class with the specified name
+
+        :param name: The name of the blotter class
```
llllllllll: we use numpy doc style instead of normal sphinx docstrings. This would look like:

```
Parameters
----------
name : str
    The name of the blotter class.

Returns
-------
exists : bool
    Is this name registered to a blotter class?
```

zipline/finance/blotter/core.py
```diff
@@ -0,0 +1,111 @@
...
+        -------
+        blotter_class : zipline.finance.blotter.Blotter
+            The desired blotter class.
+        """
+        try:
+            return self._blotter_factories[name]
+        except KeyError:
+            raise ValueError("no blotter class registered as %r, options "
+                             "are: %r" % (name, sorted(self._blotter_factories
+                                                       )))
```
llllllllll: Normally when we need to break a line to fit into the line limit, we would indent after the function call and list every argument on a new line, for example:

```python
            raise ValueError(
                "no blotter class registered as %r, options are: %r" % (
                    name,
                    sorted(self._blotter_factories),
                ),
            )
```

Most people have ample vertical space, and this make the argument and expression boundaries very clear.

zipline/finance/blotter/blotter.py
```diff
@@ -0,0 +1,190 @@
...
+        Notes
+        -----
+        This is required for `Blotter` subclasses to be able to place a batch
+        of orders, instead of being passed the order requests one at a time.
+        """
+
+        return [self.order(*order_args) for order_args in order_arg_lists]
+
+    @abstractmethod
+    def cancel(self, order_id, relay_status=True):
```
llllllllll: Should we add a docstring to cancel?

zipline/finance/blotter/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    blotter_classes,
+)
+
+__all__ = ['SimulatedBlotter',
+           'Blotter',
+           'register',
+           'unregister',
+           'blotter_classes',
+           'load',
+           ]
```
llllllllll: The trailing bracket here seems misaligned to me. I would find this easier to read as:
```python
__all__ = [
    'SimulatedBlotter',
    'Blotter',
    'register',
    'unregister',
    'blotter_classes',
    'load',
]
```

freddiev4:
Is there anything from this PR we want to merge? (saw conversation in #2201)

jnazaren:
@freddiev4 I think I combined this one into the larger PR (#2210) as well

PR #2201, opened 2018-06-04
ENH: Custom command line argument support

jnazaren:
Adds an '-x' option to the top-level entry point, which allows a user to define multiple custom args which would then be under __main__.extension_args. Also adds a function under __main__ that parses these args, making sure that they are entered in key=value form.

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -1,11 +1,13 @@
 import zipline.__main__ as main
+import zipline
 from zipline.testing import ZiplineTestCase
 from zipline.testing.predicates import (
     assert_equal,
     assert_raises_str,
+    assert_is,
 )
-from mock import patch
 from click.testing import CliRunner
+from zipline.extensions import ExtensionArgs
 
 
 class CmdLineTestCase(ZiplineTestCase):
@@ -14,25 +16,48 @@ class CmdLineTestCase(ZiplineTestCase):
         super(CmdLineTestCase, self).init_instance_fixtures()
 
         # make sure this starts empty
-        assert_equal(main.extension_args, {})
+        # TODO: decide how this needs to be reset
+        zipline.extension_args = ExtensionArgs([])
 
     def test_parse_args(self):
-        with patch.dict(main.extension_args, {}):
-            main.parse_extension_arg('arg1=test1')
-            main.parse_extension_arg('arg2=test2')
 
-            assert_equal(main.extension_args,
-                         {'arg2': 'test2', 'arg1': 'test1'})
-            msg = 'invalid extension argument 1=test3, ' \
-                  'must be in key=value form'
-            with assert_raises_str(ValueError, msg):
-                main.parse_extension_arg('1=test3')
-            msg = 'invalid extension argument arg4 test4, ' \
-                  'must be in key=value form'
-            with assert_raises_str(ValueError, msg):
-                main.parse_extension_arg('arg4 test4')
+        e = ExtensionArgs(['arg1=test1', 'arg2=test2'])
+        assert_equal(e.extension_args,
+                     {'arg2': 'test2', 'arg1': 'test1'})
+        assert_equal(e.arg1, 'test1')
+        assert_equal(e.arg2, 'test2')
+
+        msg = 'invalid extension argument 1=test3, ' \
+              'must be in key=value form'
+        with assert_raises_str(ValueError, msg):
+            e.parse_extension_arg('1=test3')
+        msg = 'invalid extension argument arg4 test4, ' \
+              'must be in key=value form'
+        with assert_raises_str(ValueError, msg):
+            e.parse_extension_arg('arg4 test4')
+
+    def test_parse_namespaces(self):
+
+        e = ExtensionArgs(["first.second.a=blah1",
+                           "first.second.b=blah2",
+                           "first.third=blah3", ])
+        assert_equal(e.first.second.a, 'blah1')
+        assert_equal(e.first.second.b, 'blah2')
+        assert_equal(e.first.third, 'blah3')
+
+        msg = "Conflicting assignments at namespace level 'second'"
+        with assert_raises_str(ValueError, msg):
+            e = ExtensionArgs(["first.second.a=blah1",
+                               "first.second.b=blah2",
+                               "first.second=blah3", ])
 
     def test_user_input(self):
-        with patch.dict(main.extension_args, {}):
-            runner = CliRunner()
-            result = runner.invoke(main.main, [])
+        runner = CliRunner()
+        result = runner.invoke(main.main, [    '-xfirst.second.a=blah1',
+                                       '-xfirst.second.b=blah2',
+                                       '-xfirst.third=blah3',
+                                  'bundles', ])
+
+        assert_equal(zipline.extension_args.first.second.a, 'blah1')
+        assert_equal(zipline.extension_args.first.second.b, 'blah2')
+        assert_equal(zipline.extension_args.first.third, 'blah3')
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -6,7 +6,11 @@ from zipline.testing.predicates import (
     assert_raises_str,
 )
 from click.testing import CliRunner
-from zipline.extensions import ExtensionArgs
+from zipline.extensions import (
+    Namespace,
+    create_args,
+    parse_extension_arg,
+)
 
 
 class CmdLineTestCase(ZiplineTestCase):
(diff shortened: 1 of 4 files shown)


zipline/__main__.py
```diff
@@ -17,6 +18,20 @@
 except NameError:
     __IPYTHON__ = False
 
+extension_args = {}
+
+
+def _parse_extension_arg(arg):
```
llllllllll: If we want to test this individually then we can remove the leading underscore

zipline/__main__.py
```diff
@@ -17,6 +18,20 @@
 except NameError:
     __IPYTHON__ = False
 
+extension_args = {}
```
llllllllll: I think we should make this dictionary available at either the top level like `zipline.extension_args` or somewhere easy like `zipline.extensions.extension_args`.

@ssanderson do you have any ideas?

llllllllll: How do you feel about making this an object with attribute access so that we can encourage users to namespace their keys like `zipline_live.variable`?

tests/test_cmdline.py
```diff
@@ -0,0 +1,31 @@
...
+    assert_raises_str,
+)
+
+
+class CmdLineTestCase(ZiplineTestCase):
+
+    def init_instance_fixtures(self):
+        super(CmdLineTestCase, self).init_instance_fixtures()
+
+        main.extension_args = {}
```
llllllllll: We shouldn't permanently mutate mutate this. Instead, we can use a patch context manager: https://docs.python.org/3/library/unittest.mock.html.

Note: because we are using a python2 compatibility layer, `mock` is not a submodule of `unittest`, it is a top-level package.

tests/test_cmdline.py
```diff
@@ -0,0 +1,31 @@
+import zipline.__main__ as main
+from zipline.testing import ZiplineTestCase
+from zipline.testing.predicates import (
+    assert_equal,
+    assert_raises_str,
+)
+
+
+class CmdLineTestCase(ZiplineTestCase):
```
llllllllll: we should add some tests that actually invoke to the entry point to make sure we parse the args correctly when running. Click has some helpers for testing applications which we can take advantage of: http://click.pocoo.org/5/testing/

zipline/__main__.py
```diff
@@ -17,6 +18,20 @@
...
+def _parse_extension_arg(arg):
+    match = re.match(r'^(([^\d\W]\w*)(\.[^\d\W]\w*)*)=(.*)$', arg)
+    if match is None:
+        raise ValueError(
+            "invalid extension argument %s, must be in key=value form" % arg
+        )
+
+    name = match.group(1)
+    value = match.group(4)
+    extension_args[name] = value
```
ssanderson: Rather than mutating a global value in place, I would probably write this function as:

```python
def parse_extension_args(args):
    out = {}
    for arg in args:
        name, value = parse_single_extension_arg(arg)
        out[name] = value
    return out
```

Then, in `main`, we can do:

```python
args = parse_extension_args(x)
set_global_extension_args(args)  # Put the args somewhere user-accessible.
```

The benefit of this approach is that its easier to test `parse_extension_args` in isolation, because it doesn't interact with any global state: it's just responsible for the business logic of converting a list of command line flags into a dictionary.

zipline/__main__.py
ssanderson: I don't have a super strong preference about whether this is a a dictionary or a namespace object. Do we think that most extensions are going to know ahead of time what args they're interested in? If so then a namespace object probably makes sense. If we think people are going to need to dynamically scan the args, then a dict is probably preferable.

zipline/__main__.py
ssanderson: As far as setting the actual global state goes, there are a bunch of options. The simplest option is to just set a dictionary (or namespace object) into a global variable somewhere public and documented like `zipline.extensions.extension_args` as @llllllllll suggested in https://github.com/quantopian/zipline/pull/2201/files#r192841333.

Slightly more complex would be make the exact location of the storage private (e.g., `zipilne.extensions._global_extension_args`, and expose a public function like `get_extension_args` that retrieves the global variable. For testing, it might be useful to provide a [contextmanager](https://docs.python.org/2/library/contextlib.html#contextlib.contextmanager) for contextually setting and re-setting the global state of the extension args.

Here's an example of what this might look like:

```python
# zipline.extension.extension.py
from contextlib import contextmanager


_extension_args = None


def get_extension_args():
    """Useful user-facing docstring.
    """
    if _extension_args is None:
        raise NotInAZiplineExtension("stahp")
    return _extension_args


def _set_extension_args(args):
    """Private, internal-facing docstring.
    """
    global extension_args
    # Return the old args so that they can be restored.
    old_args = extension_args
    extension_args = args
    return old_args


@contextmanager
def _extension_args_context(args):
    """Private, internal-facing docstring.
    """
    try:
        old = _set_extension_args(args)
        yield
    finally:
        _set_extension_args(old)
```

With this version, our `main()` would look something like:

```
def main(..., x):
    with _extension_args_context(parse_extension_args(x)):
        load_extensions(...)
    do_stuff(...)
```

and user code would interact with this system by doing:

```python
from zipline.extensions import get_extension_args

args = get_extension_args()
do_stuff_with(args)
```

zipline/extensions.py
```diff
@@ -0,0 +1,44 @@
...
+                    raise ValueError("Conflicting assignments at namespace"
+                                     " level '%s'" % path[0])
+                self.get_namespace(getattr(obj, path[0]), path[1:], name)
+            else:
+                a = Arg()
+                setattr(obj, path[0], a)
+                self.get_namespace(getattr(obj, path[0]), path[1:], name)
+
+
+class Arg(object):
```
llllllllll: Could we call this `Namespace` or something to make it more clear in a repr.

zipline/extensions.py
```diff
@@ -0,0 +1,44 @@
+import re
+
+
+class ExtensionArgs(object):
```
llllllllll: imo we should not have the parsing code and consuming code in the same place. If a user is looking to see which extension arguments have been set, right now they will see `extension_args`, `parse_extension`, and `get_namespace` in addition to whatever was set. This makes it harder for us to maintain because we will be making this part of the public interface. Instead, we could make the root group an `Arg` or `Namespace`, and in the `main`, parse into this structure and that make this empty object available as `zipline.extension_args`.

tests/test_cmdline.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_parse_args(self):
+
+        e = ExtensionArgs(['arg1=test1', 'arg2=test2'])
+        assert_equal(e.extension_args,
+                     {'arg2': 'test2', 'arg1': 'test1'})
+        assert_equal(e.arg1, 'test1')
+        assert_equal(e.arg2, 'test2')
+
+        msg = 'invalid extension argument 1=test3, ' \
```
llllllllll: don't use `\` to continue lines, you can wrap the expression in parentheses like:
```python
msg = (
    'first part of the string'
    ' second part of the string'
)
```

tests/test_cmdline.py
```diff
@@ -0,0 +1,62 @@
...
+        with assert_raises_str(ValueError, msg):
+            e = ExtensionArgs(["first.second.a=blah1",
+                               "first.second.b=blah2",
+                               "first.second=blah3", ])
+
+    def test_user_input(self):
+        runner = CliRunner()
+        runner.invoke(main.main, ['-xfirst.second.a=blah1',
+                                  '-xfirst.second.b=blah2',
+                                  '-xfirst.third=blah3',
```
llllllllll: can we add some other top level keys, and maybe a non-nested key like `a=value`

tests/test_cmdline.py
```diff
@@ -0,0 +1,62 @@
...
+class CmdLineTestCase(ZiplineTestCase):
+
+    def init_instance_fixtures(self):
+        super(CmdLineTestCase, self).init_instance_fixtures()
+
+        # make sure this starts empty
+        # TODO: decide how this needs to be reset
+        zipline.extension_args = ExtensionArgs([])
+
+    def test_parse_args(self):
```
llllllllll: we should add some tests for nested keys, keys that have numbers in them, keys with underscores, and any other interesting names you can think of.

tests/test_cmdline.py
```diff
@@ -0,0 +1,62 @@
...
+
+        msg = "Conflicting assignments at namespace level 'second'"
+        with assert_raises_str(ValueError, msg):
+            e = ExtensionArgs(["first.second.a=blah1",
+                               "first.second.b=blah2",
+                               "first.second=blah3", ])
+
+    def test_user_input(self):
+        runner = CliRunner()
+        runner.invoke(main.main, ['-xfirst.second.a=blah1',
```
llllllllll: we should assert that this exits with success

tests/test_cmdline.py
jnazaren: I'll add some tests here for weird names, but currently I keep all of the tests for nested vs non-nested separate (no particular reason, just the way I had started doing it earlier), so I'll add nested ones to `test_parse_namespaces()`.

ssanderson:
@jnazaren @llllllllll is there still stuff here that we want to merge, or was this replaced by https://github.com/quantopian/zipline/pull/2210 ?

jnazaren:
This PR was replaced by #2210 -- sorry should have closed it earlier

PR #2206, opened 2018-06-08
ENH: Improve short reprs

jnazaren:
Improves the representation of some nodes in the pipeline graph

### DIFF
--- a/zipline/pipeline/expression.py
+++ b/zipline/pipeline/expression.py
@@ -14,6 +14,8 @@ from numpy import (
 
 from zipline.pipeline.term import Term, ComputableTerm
 from zipline.utils.numpy_utils import bool_dtype
+from decimal import Decimal
+import re
 
 
 _VARIABLE_NAME_RE = re.compile("^(x_)([0-9]+)$")
@@ -328,7 +330,12 @@ class NumericalExpression(ComputableTerm):
 
     def short_repr(self):
         """Short repr to use when rendering Pipeline graphs."""
-        return "Expression: {expr}".format(
+        final = self._expr
+        num_list = re.findall(r"[-+]?\d*\.\d+", self._expr)
+        for i in num_list:
+            final = final.replace(i, '%.2E' % Decimal(i))
+
+        return "Expression:\n{expr}".format(
             typename=type(self).__name__,
-            expr=self._expr,
+            expr=final,
         )
--- a/zipline/pipeline/expression.py
+++ b/zipline/pipeline/expression.py
@@ -330,12 +330,9 @@ class NumericalExpression(ComputableTerm):
 
     def short_repr(self):
         """Short repr to use when rendering Pipeline graphs."""
-        final = self._expr
-        num_list = re.findall(r"[-+]?\d*\.\d+", self._expr)
-        for i in num_list:
-            final = final.replace(i, '%.2E' % Decimal(i))
-
-        return "Expression:\n{expr}".format(
-            typename=type(self).__name__,
-            expr=final,
+        final = re.sub(r"[-+]?\d*\.\d+",
+                       lambda x: format(float(x.group(0)), '.2E'),
+                       self._expr)
+        return "Expression:\l{}\l".format(
+            final,
         )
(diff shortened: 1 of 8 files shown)


zipline/pipeline/expression.py
```diff
@@ -328,7 +330,12 @@ def __repr__(self):
 
     def short_repr(self):
         """Short repr to use when rendering Pipeline graphs."""
-        return "Expression: {expr}".format(
+        final = self._expr
+        num_list = re.findall(r"[-+]?\d*\.\d+", self._expr)
```
ssanderson: Is the purpose of this to replace numbers with scientific notation? If so, do we need to use `Decimal` here (vs. say, `float`)?

zipline/pipeline/factors/factor.py
```diff
@@ -1411,6 +1411,13 @@ def __repr__(self):
             mask=self.mask,
         )
 
+    def short_repr(self):
+        return "Rank({})\nmethod={}\nmask={}".format(
```
ssanderson: It feels a little odd to me to have some of the parameters to `Rank` be inside parentheses, but to have method and mask outside. What do you think about something more yaml-ish like this?

```
Rank:
  method: 'ordinal'
  mask: SomeFilter 
```

zipline/pipeline/filters/filter.py
```diff
@@ -481,6 +488,12 @@ def _compute(self, arrays, dates, assets, mask):
         data = arrays[0]
         return params['op'](data, *params['opargs']) & mask
 
+    def short_repr(self):
+        return "{}\nop={}.{}()".format(type(self).__name__,
```
ssanderson: flake8 is going to complain about the formatting here. I would format this as

```python
"...".format(
    type(self).__name__,
    ...,
    ...,
)
```

zipline/pipeline/expression.py
```diff
@@ -328,7 +330,12 @@ def __repr__(self):
...
+        for i in num_list:
+            final = final.replace(i, '%.2E' % Decimal(i))
```
richafrank: How does the loop compare to `re.sub`? I could see it being potentially more efficient, but I don't know if it matters here.

jnazaren: Oh I'll use that instead - str.replace() is an old habit I have from Java

llllllllll: small tip: you can get access to the `%` formatting behavior for a single object with the `format` builtin:

```python
In [6]: format(Decimal(1), '.2E')
Out[6]: '1.00E+0'
```

zipline/pipeline/expression.py
jnazaren: No, that was just from the original example of shortening numbers I found - I've changed it to `float`

zipline/pipeline/factors/factor.py
jnazaren: Ok, yeah that makes sense

zipline/pipeline/expression.py
```diff
@@ -328,7 +326,9 @@ def __repr__(self):
 
     def short_repr(self):
         """Short repr to use when rendering Pipeline graphs."""
-        return "Expression: {expr}".format(
-            typename=type(self).__name__,
-            expr=self._expr,
+        final = re.sub(r"[-+]?\d*\.\d+",
```
ssanderson: Can we add a comment explaining what's happening here?

zipline/pipeline/expression.py
```diff
@@ -328,7 +326,9 @@ def __repr__(self):
...
+                       lambda x: format(float(x.group(0)), '.2E'),
+                       self._expr)
+        return "Expression:\l{}\l".format(
```
ssanderson: In our other uses of `\l` it looks like we're putting two leading spaces after the line break. Any reason not to do that here too"

jnazaren: Sure, I can add that in

zipline/pipeline/expression.py
jnazaren: Yep

ssanderson:
LGTM, feel free to merge when ready.

richafrank:
Nice! Before merging, would be good to clean up the WIP commit history.

PR #2207, opened 2018-06-08
ENH: prevent zero capital base from zipline

PrestonYadegar:


PrestonYadegar:
Just checked that conflict. Not sure where that code for __fail_fast (on master) is coming from/what it's purpose is

kglowinski:
When you made your branch, did you clone off of head of master, or head of the branch that qexec is currently pointing to? You might try rebasing against master, and see if that Just Fixes (TM) the issue.

tests/test_algorithm.py
```diff
@@ -313,6 +314,47 @@ def handle_data(context, data):
         with self.assertRaises(TypeError):
             algo.run()
 
+    @parameterized.expand([
+        (-1000, 'invalid_base'),
```
kglowinski: 👍 on multiple tests.

tests/test_algorithm.py
```diff
@@ -313,6 +314,47 @@ def handle_data(context, data):
...
+            capital_base=cap_base,
+            data_frequency="minute",
+            trading_calendar=self.trading_calendar
+        )
+
+        with self.assertRaises(ZeroCapitalError) as exc:
+            # make_algo will trace to TradingAlgorithm,
+            # where the exception will be raised
+            algo = self.make_algo(script=algo_text, sim_params=sim_params)
+        # It shouldn't be able to run the algorithm
```
kglowinski: Comment indenting got weird here. When you say "it shouldn't" do you mean that you shouldn't be able to get to this line of code? Or that the return into `output` should be `None`?

tests/test_algorithm.py
```diff
@@ -313,6 +314,47 @@ def handle_data(context, data):
...
+            output = algo.run()
+            # If capital base was less than zero, the none of the orders
```
kglowinski: s/less than/less than or equal to

zipline/algorithm.py
```diff
@@ -253,6 +254,7 @@ def __init__(self, *args, **kwargs):
                 Any asset identifiers that are not provided in the
                 equities_metadata, but will be traded by this TradingAlgorithm
         """
+
```
kglowinski: Cruft

zipline/algorithm.py
```diff
@@ -304,6 +306,12 @@ def __init__(self, *args, **kwargs):
                 trading_calendar=self.trading_calendar,
             )
 
+        # Ensure that capital base hasn't been overridden to an invalid value
+        # This test duplicates that from qexec/algoproxy, but targets users
```
kglowinski: Is this comment still true?

tests/test_algorithm.py
```diff
@@ -313,6 +314,47 @@ def handle_data(context, data):
...
+            # should have been filled so starting/ending cash should be
+            # equal
+            expected_cash = [cap_base for x in range(4)]
+            self.assertEqual(expected_cash, list(output['ending_cash']))
+            # Make sure the correct error was raised
+            error = exc.exception
+            self.assertEqual(str(error), 'The initial capital base is set \
```
kglowinski: This assertion doesn't match the error being raised in the error class down below. Is this test actually passing?

richafrank: It probably passes because `self.make_algo(...)` raises (as expected), so none of these lines afterwards, but inside the context manager, are executed.

kglowinski: Ah, we probably want to outdent this block, and since we save exc as a variable, we can do the checks after.

tests/test_algorithm.py
PrestonYadegar: Changed the indentation and made the comment more clear. What I meant was that since self.make_algo() will raise the zero capital error, make_algo() may not finish executing, so algo.run() may then end up raising some errors

tests/test_algorithm.py
PrestonYadegar: Fixed

zipline/algorithm.py
PrestonYadegar: Fixed

tests/test_algorithm.py
kglowinski: So it looks like you changed the indentation of the comment, but the lines below it are still indented (thus they won't be run after that error is raised). So this test will still pass, no matter what assertions are made below and their correctness.

PrestonYadegar: I realized this yesterday. I scrapped the assertion for starting and ending capital base (to be equal) because I could no longer execute algo.run(). The only assertion I left is that the text of the error matches the ZeroCapitalError text. (I outdented the assertion this time)

tests/test_algorithm.py
PrestonYadegar: Saw this and took it into account for test_invalid_capital_base. Removed the starting and ending capital base assertion and took the other one out of the context manager

zipline/algorithm.py
PrestonYadegar: No, good catch. Forgot that I removed the algoproxy code

docs/deploy.py
```diff
@@ -75,5 +75,6 @@ def main():
     print("Updated documentation branch in directory %s" % ZIPLINE_ROOT)
     print("If you are happy with these changes, commit and push to gh-pages.")
 
+
```
kglowinski: Cruft

versioneer.py
```diff
@@ -419,6 +419,7 @@ def get(parser, name):
 class NotThisMethod(Exception):
     pass
 
+
```
kglowinski: Cruft.

tests/test_algorithm.py
```diff
@@ -313,6 +314,40 @@ def handle_data(context, data):
...
+            capital_base=cap_base,
+            data_frequency="minute",
+            trading_calendar=self.trading_calendar
+        )
+
+        with self.assertRaises(ZeroCapitalError) as exc:
+            # make_algo will trace to TradingAlgorithm,
+            # where the exception will be raised
+            algo = self.make_algo(script=algo_text, sim_params=sim_params)
+            algo.run()
```
kglowinski: So since we're going to hit your error in the line above, this line is never being run, and probably shouldn't be included in the test. If the idea is to make flake8 not yell at you about unused assignment, you can just not assign for the `make_algo` call above.

zipline/algorithm.py
kglowinski: Line is still there?

tests/test_algorithm.py
PrestonYadegar: Yeah I added that to comply with flake8, just fixed it

PrestonYadegar:
Everything seem good to go here? @kglowinski

jbredeche:
@PrestonYadegar can you squash the commits?

PrestonYadegar:
Sure @jbredeche

zipline/algorithm.py
```diff
@@ -409,6 +410,12 @@ def noop(*args, **kwargs):
         if 'data_frequency' in kwargs:
             self.data_frequency = kwargs.pop('data_frequency')
 
+        capital_base = self.sim_params.capital_base
+        if capital_base <= 0:
+            assert capital_base <= 0, \
```
kglowinski: So this is probably unnecessary, since the message inside the error that you're raising is the same as this. Either you probably want that assert, in which case, you're going to get an `AssertionError` with that message, or you want to raise the `ZeroCapitalError`, in which case you could probably pass the text that you wanted displayed to the error class, if you had wanted that to be variable in this context (the thing you're doing with your assert). In this case the assertion is basically serving the same purpose as the if statement, and you don't really need both.

PrestonYadegar: I had added the assert because of some logging issues (I wasn't seeing the error) but it resolved. I'll update and remove that

PR #2209, opened 2018-06-10
Include LICENSE file in wheels

toddrme2178:
Although not strictly required by the license, having a copy of the license file is preferred by some groups such as Linux distros that repackage the software.  This makes sure the license is included in the wheels.  See the wheel documentation [here](https://wheel.readthedocs.io/en/stable/#including-the-license-in-the-generated-wheel-file) for more information.


richafrank:
Thanks @toddrme2178!

PR #2214, opened 2018-06-17
Correcting market open time

YasserHassan:
The first minute of the regular trading hours is 9:30, not 9:31. You can confirm that by checking the time of the first candle of stocks like MSFT or AAPL on a 1 minute candlestick chart during regular tading hours on Yahoo Finance or TradingView.com


ssanderson:
@YasserHassan it's true that NYSE opens at 9:30, but Zipline uses the convention that a 1-minute bar is labelled with the timestamp as of the **end** of the bar, so the first minute of the NYSE day has a label of 9:31.

PR #2218, opened 2018-06-21
BUG: Fix factor.top(1) with no groupby.

ssanderson:


tests/pipeline/test_engine.py
```diff
@@ -1529,3 +1529,30 @@ def test_run_chunked_pipeline(self):
...
+                            ZiplineTestCase):
+    ASSET_FINDER_EQUITY_SIDS = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+
+    def test_no_groupby_maximum(self):
+        # This is a regression test for a bug where factor.top(1) would fail
+        # when not passed a groupby parameter.
+
+        factor = TestingDataSet.float_col.latest
+        maximum = factor.top(1)
+        pipe = Pipeline({'factor': factor, 'maximum': maximum})
```
dmichalowicz: Any reason for including the actual factor in the pipeline?

zipline/pipeline/filters/filter.py
```diff
@@ -593,6 +595,10 @@ class MaximumFilter(Filter, StandardOutputs):
     window_length = 0
 
     def __new__(cls, factor, groupby, mask):
+        from zipline.pipeline.classifiers import Everything
+        if groupby is NotSpecified:
+            groupby = Everything()
```
yankees714: If this import is going to be lazy anyway, is there a reason not to put it inside the if block?

tests/pipeline/test_engine.py
ssanderson: The line below that does `result.groupby(level=0).max()` is using the factor to re-compute the max to verify the result that came out of pipeline.

zipline/pipeline/filters/filter.py
```diff
@@ -593,6 +595,10 @@ class MaximumFilter(Filter, StandardOutputs):
     window_length = 0
 
     def __new__(cls, factor, groupby, mask):
+        from zipline.pipeline.classifiers import Everything
```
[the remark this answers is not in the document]
ssanderson: We can't import classifiers at module scope here because the `classifiers` module imports `filters`, because a bunch of the Classifier methods return Filter instances. I don't think there's a straightforward way to break that circular dependency more cleanly than this.

zipline/pipeline/filters/filter.py
ssanderson: Probably not. I can move this down.

PR #2220, opened 2018-06-25
ENH: Add function for computing all pipelines separately

jnazaren:
This provides a function for computing all pipelines separately from the `pipeline_output` call in `before_trading_start`.


zipline/algorithm.py
```diff
@@ -2416,7 +2421,7 @@ def set_long_only(self, on_error='fail'):
         name=string_types,
         chunks=(int, Iterable, type(None)),
     )
-    def attach_pipeline(self, pipeline, name, chunks=None):
+    def attach_pipeline(self, pipeline, name, chunks=None, is_eager=True):
```
ssanderson: I would call this parameter `eager` rather than `is_eager`. `is_eager` suggests that the parameter is a property of the pipeline, whereas `eager` (at least to me) more clearly connotes that this flag describes how we want the pipeline to be computed in the future.

zipline/algorithm.py
```diff
@@ -608,6 +608,11 @@ def _calculate_universe(self):
         # our universe is all the assets passed into `run`.
         return self._assets_from_source
 
+    def _compute_pipelines(self):
+        for name, p in self._pipelines.items():
+            if p[2]:
```
ssanderson: Using `[2]` here is a bit obscure. We might want to use a [namedtuple](https://docs.python.org/2/library/collections.html#collections.namedtuple) for registered pipelines if we're going to interact with individual fields.

zipline/algorithm.py
```diff
@@ -608,6 +608,11 @@ def _calculate_universe(self):
         # our universe is all the assets passed into `run`.
         return self._assets_from_source
 
+    def _compute_pipelines(self):
```
ssanderson: Couple notes on this method:

1. Since we intend to override this to be called in a subclass downstream, we should make this a public method and document it. 
2. I would probably name this something like `compute_eager_pipelines` to make it clearer that this only computes pipelines that were attached with `eager=True`.
3. It doesn't look like anything is actually calling this in Zipline right now. I would expect that we want to call this at the start of `TradingAlgorithm.before_trading_start`. Note that that function exits early if `self._before_trading_start` is None, but we should still run eager pipelines even if the user hasn't defined a BTS, since part of the purpose of this PR is to remove the need for superfluous BTS implementations that just call pipeline_output and copy results onto context. 
4. We need a test for the new behavior here. Testing this is a bit tricky, since the values exposed to the user shouldn't change; only when they get computed. One way to test that things are computed in the correct order would be to create an algorithm that logs in its `before_trading_start` and during the execution of a `CustomFactor`. You can then capture the logs using a logbook TestHandler and verify that the logs from the CustomFactor fire before the logs from before_trading_start. (You could also do something similar by appending to a list on context instead of logging).

zipline/testing/fixtures.py
```diff
@@ -1540,7 +1540,7 @@ def make_adjustment_db_conn_str(cls):
         return cls.adjustments_db_path
 
 
-class WithSeededRandomPipelineEngine(WithTradingSessions, WithAssetFinder):
+class WithSeededRandomPipelineEngine(WithAssetFinder, WithTradingSessions):
```
prsutherland: Is there a mro issue with `WithSeededRandomPipelineEngine`?

tests/pipeline/test_pipeline_algo.py
```diff
@@ -738,3 +738,62 @@ def before_trading_start(context, data):
...
+            def compute(self, today, assets, out):
+                trace.append("CustomFactor call")
+
+        def initialize(context):
+            pipeline = attach_pipeline(Pipeline(), 'my_pipeline')
+            test_factor = TestFactor()
+            pipeline.add(test_factor, 'test_factor')
+
+        def before_trading_start(context, data):
+            pipeline_output('my_pipeline')
```
prsutherland: probably worth swapping the `pipeline_output` with the `trace` so that "BTS call" will appear first in the trace if there is an error.

tests/pipeline/test_pipeline_algo.py
```diff
@@ -738,3 +738,62 @@ def before_trading_start(context, data):
...
+                end_session=pd.Timestamp('2014-12-31', tz='utc'),
+                data_frequency='daily',
+                emission_rate='daily',
+                trading_calendar=self.trading_calendar,
+            )
+        )
+
+        # All pipeline computation calls should occur before any BTS calls,
+        # and the algorithm is being run for 3 days, so the first 3 calls
+        # should be to the custom factor and the next 3 calls should be to BTS
```
prsutherland: Why is it three and then three and not interleaved?

zipline/testing/fixtures.py
jnazaren: No, there was and I fixed it, but according to @ssanderson `WithSeededRandomPipelineEngine` isn't even necessary here. I just needed to create a custom factor with no inputs.

prsutherland: I don't fully understand the change of inheritance order, but if it isn't needed, should we undo the change?

tests/pipeline/test_pipeline_algo.py
jnazaren: I believe this is because all three of the dates are included in the chunk, and therefore, they are all computed in one for loop -- see https://github.com/quantopian/zipline/blob/f61d3dfb141b12e14c9b4de332bdd7b7de620eb1/zipline/pipeline/mixins.py#L203

tests/pipeline/test_pipeline_algo.py
jnazaren: Ok

zipline/testing/fixtures.py
jnazaren: Sure, sorry my previous answer was pretty ambiguous because I wasn't actually talking about this specific place - there was an mro error when I had used `WithSeededRandomPipelineEngine` in my test code, but I no longer need it in my tests, so I will revert this change.

PR #2223, opened 2018-06-26
[BUG] Fix invalid data loading in csvdir bundle

ASzot:
Calculating the name of CSV file from the symbol name is not done corretly. The current method would match a filename to a symbol if it ended in the symbol concatenated with `.csv`. This means that if there are two stocks Ford (symbol F) and Regions Financial Corporation (symbol RF). Then the calculated filename to load for Ford could be RF.csv which is clearly not the data file for Ford. 

This is clearly a bug.

### DIFF
--- a/zipline/data/bundles/csvdir.py
+++ b/zipline/data/bundles/csvdir.py
@@ -175,10 +175,13 @@ def _pricing_iter(csvdir, symbols, metadata, divs_splits, show_progress):
         for sid, symbol in enumerate(it):
             logger.debug('%s: sid %s' % (symbol, sid))
 
+            fname = symbol + '.csv'
             try:
-                fname = [fname for fname in files
-                         if '%s.csv' % symbol in fname][0]
-            except IndexError:
+                dfr = read_csv(os.path.join(csvdir, fname),
+                               parse_dates=[0],
+                               infer_datetime_format=False,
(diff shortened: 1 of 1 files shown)


zipline/data/bundles/csvdir.py
```diff
@@ -175,10 +175,13 @@ def _pricing_iter(csvdir, symbols, metadata, divs_splits, show_progress):
         for sid, symbol in enumerate(it):
             logger.debug('%s: sid %s' % (symbol, sid))
 
+            fname = symbol + '.csv'
             try:
-                fname = [fname for fname in files
-                         if '%s.csv' % symbol in fname][0]
-            except IndexError:
+                dfr = read_csv(os.path.join(csvdir, fname),
```
freddiev4: Should we remove the second call to `read_csv()` below on line 185? We're already calling it here and exiting early if we don't have the file we want.

zipline/data/bundles/csvdir.py
```diff
@@ -175,10 +175,13 @@ def _pricing_iter(csvdir, symbols, metadata, divs_splits, show_progress):
...
+                               parse_dates=[0],
+                               infer_datetime_format=False,
```
freddiev4: Why wouldn't we want to infer the datetime format?

Issue #1277, opened 2016-06-14
instant_fill was removed.


FayolChang:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

* Operating System: (Darwin  15.5.0 Darwin Kernel Version 15.5.0: Tue Apr 19 18:36:36 PDT 2016; root:xnu-3248.50.21~8/RELEASE_X86_64 x86_64)
* Python Version: `2.7.11`
* Python Bitness: `64`
* How did you install Zipline: (source code)
* Python packages: ``

Now that you know a little about me, let me tell you about the issue I am
having:

# Description of Issue
instant_fill was removed from TradingAlgorithm, every_bar in transform of tradesimulation module now only process orders from last bar.
* What did you expect to happen?
tradesimulation can process orders from current bar,  as instant_fill = True in 0.8.4.
* What happened instead?
just as descripted
Here is how you can reproduce this issue on your machine:

## Reproduction Steps

1.
2.
3.
...

## What steps have you taken to resolve this already?

...

# Anything else?
It would be nice if portfolio be updated instantly after a transaction generated. 
I need a feature that the transaction should be rejected if there is not enough cash. negative ending_cash
is not allowed.
Sincerely,
Phil Zhang

FreddieV4:
Hi @FayolChang we're now on version `1.1.1` of zipline; is this issue still relevant to you for the latest version?

frosk1:
Is there even an alternative way of getting orders filled at the same day?

ayxemma:
also have the same question, is there any answer on this?

PR #2232, opened 2018-07-02
DOC: Remove instant_fill docstring reference from TradingAlgorithm

freddiev4:
Are there any places where we allow orders to be filled on the same day / work in the same way that `instant_fill` used to work?


freddiev4:
x-ref https://github.com/quantopian/zipline/issues/1277

freddiev4:
@llllllllll mind taking a look at this? Just a small doc change (assuming I didn't miss anything)

llllllllll:
lgtm, grep shows that this isn't used anywhere

PR #2233, opened 2018-07-02
MAINT: reexport more calendar stuff for backwards compat

vikram-narayan:


vikram-narayan:
talked offline with @ssanderson - going to change the imports in the places that require it, instead of reexporting the entire modules

PR #2234, opened 2018-07-03
ENH: Store more information about exchanges

llllllllll:
Use an object to store more attributes of an exchange. This info is stored in the `exchange` table in the assets db. We will need this to be able to store international equities in the same assets db.

This PR also adds `country_code` as a parameter to `lookup_symbol` and `lifetimes`. This PR does not allow us to trade international equities yet, it just sets up the infrastructure to reference equities listed in different countries.


zipline/algorithm.py
```diff
@@ -1299,7 +1310,15 @@ def symbols(self, *args):
         --------
         :func:`zipline.api.set_symbol_lookup_date`
         """
-        return [self.symbol(identifier) for identifier in args]
+        country_code = kwargs.pop('country_code', None)
```
ssanderson: My kingdom for keyword-only args...

zipline/algorithm.py
```diff
@@ -1275,13 +1240,59 @@ def continuous_future(self,
...
+        SymbolNotFound
+            Raised when the symbols was not held on the current lookup date.
+
+        See Also
+        --------
+        :func:`zipline.api.set_symbol_lookup_date`
+        """
+        # If the user has not set the symbol lookup date,
+        # use the end_session as the date for symbol->sid resolution.
+        _lookup_date = self._symbol_lookup_date \
```
ssanderson: Could we simplify this by initializing `self._symbol_lookup_date` to `self.sim_params.end_session` at construction time?

zipline/algorithm.py
ssanderson: Actually, is there a reason to prefer this over forwarding `**kwargs` to `self.symbol`? I guess it's slightly nicer to have the traceback refer to `symbols` instead of `symbol`?

zipline/assets/_assets.pyx
```diff
@@ -37,12 +37,6 @@ cimport numpy as np
 from trading_calendars import get_calendar
 
 
-# IMPORTANT NOTE: You must change this template if you change
```
ssanderson: Was this just not true?

zipline/assets/_assets.pyx
```diff
@@ -160,14 +157,15 @@ cdef class Asset:
             'auto_close_date': self.auto_close_date,
             'exchange': self.exchange,
             'exchange_full': self.exchange_full,
+            'exchange_info': self.exchange_info,
```
ssanderson: Do we want to include all three of `exchange`, `exchange_full` and `exchange_info`?

zipline/assets/asset_db_migrations.py
```diff
@@ -313,3 +360,163 @@ def _downgrade_v5(op):
...
+        'exchanges',
+        sa.Column(
+            'exchange',
+            sa.Text,
+            unique=True,
+            nullable=False,
+            primary_key=True,
+        ),
+        sa.Column('timezone', sa.Text),
+        selection_string="exchange, 'US/Eastern'",
```
ssanderson: Is this just unconditionally setting the 'timezone' to US/Eastern for all exchanges in the old schema?

zipline/assets/asset_db_migrations.py
```diff
@@ -10,6 +10,53 @@
...
+        raise TypeError(
+            'alter_columns received extra arguments: %r' % sorted(kwargs),
+        )
+    if selection_string is None:
+        selection_string = ', '.join(column.name for column in columns)
+
+    tmp_name = '_alter_columns_' + name
+    op.rename_table(name, tmp_name)
+
+    for column in columns:
```
ssanderson: Can we add a comment explaining what's happening in this loop? It looks like we're assuming that all columns might have indices on them, and just deleting any of them. I think that means this would break if we used a different index naming scheme from sqlalchemy's default?

zipline/assets/asset_db_migrations.py
ssanderson: If so, a comment here might be helpful.

zipline/testing/fixtures.py
```diff
@@ -351,11 +355,13 @@ class WithAssetFinder(WithDefaultDateBounds):
     ASSET_FINDER_EQUITY_SIDS = ord('A'), ord('B'), ord('C')
     ASSET_FINDER_EQUITY_SYMBOLS = None
     ASSET_FINDER_EQUITY_NAMES = None
+    ASSET_FINDER_EQUITY_EXCHANGE = 'TEST'
     ASSET_FINDER_EQUITY_START_DATE = alias('START_DATE')
     ASSET_FINDER_EQUITY_END_DATE = alias('END_DATE')
+    ASSET_FINDER_COUNTRY_CODE = '??'
```
ssanderson: Should this be `ASSET_FINDER_DEFAULT_COUNTRY_CODE`? I expect us to want to have more than one country code shortly, and it seems like this primarily configures the default.

zipline/testing/fixtures.py
```diff
@@ -316,6 +316,10 @@ class WithAssetFinder(WithDefaultDateBounds):
         ``END_DATE``.
     ASSET_FINDER_EQUITY_NAMES: iterable[str]
         The default names to use for the equities.
+    ASSET_FINDER_EQUITY_EXCHANGE : str
```
ssanderson: Same question as below re: calling this `ASSET_FINDER_DEFAULT_EQUITY_EXCHANGE` or similar.

zipline/assets/assets.py
```diff
@@ -1248,27 +1391,34 @@ def map_identifier_index_to_sids(self, index, as_of_date):
...
+                        equities_cols.start_date,
+                        equities_cols.end_date,
+                    )).where(
+                        (self.exchanges.c.exchange == equities_cols.exchange) &
+                        (self.exchanges.c.country_code.in_(country_codes))
+                    ).execute(),
+                ),
+                dtype='f8',  # use doubles so we get NaNs
+            )
+        else:
```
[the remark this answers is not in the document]
llllllllll: It was a case I thought up when testing. I think that a 0 column dataframe is the natural answer for "what are the companies in the empty set of countries. It makes calling code that is already generic over `country_codes` easier to support this case.

zipline/assets/assets.py
```diff
@@ -823,17 +928,34 @@ def lookup_symbol(self, symbol, as_of_date, fuzzy=False):
...
         if fuzzy:
-            return self._lookup_symbol_fuzzy(symbol, as_of_date)
-        return self._lookup_symbol_strict(symbol, as_of_date)
+            f = self._lookup_symbol_fuzzy
+            mapping = self._which_fuzzy_symbol_ownership_map(country_code)
+        else:
+            f = self._lookup_symbol_strict
+            mapping = self._which_symbol_ownership_map(country_code)
+
+        if mapping is None:
```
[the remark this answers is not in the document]
llllllllll: it seems like that should raise a consistent error that says, "I don't know about this symbol country pair". I am imagining a user who has multi-country symbols to map, and the assets db only knows about a subset of the symbols. If we add a new error, they now need to care about these different failed lookups instead of handling it the same way.

zipline/assets/assets.py
```diff
@@ -776,13 +853,21 @@ def _lookup_symbol_fuzzy(self, symbol, as_of_date):
         if len(options) == 1:
             return self.retrieve_asset(sid_keys[0])
 
+        exact_options = []
```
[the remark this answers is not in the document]
llllllllll: maybe, do you think that will make it much easier to read. Right now you can just follow this relatively simple algorithm from top to bottom.

zipline/assets/assets.py
```diff
@@ -624,10 +681,16 @@ def _retrieve_asset_dicts(self, sids, asset_tbl, querying_equities):
 
         if querying_equities:
             def mkdict(row,
+                       exchanges=self.exchange_info,
                        symbols=self._lookup_most_recent_symbols(sids)):
-                return merge(row, symbols[row['sid']])
+                d = dict(row)
```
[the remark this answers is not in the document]
llllllllll: we do call this function on every single symbol->sid record. This code is already super slow, but not because we have one single hot function, we are just bleeding small losses all over. I could make it a non-local but idk if it makes it any less readable this way.

zipline/assets/assets.py
```diff
@@ -757,7 +834,7 @@ def _lookup_symbol_fuzzy(self, symbol, as_of_date):
             # there are more than one exact match for this fuzzy symbol
             raise MultipleSymbolsFound(
                 symbol=symbol,
-                options=set(options),
+                options=[self.retrieve_asset(owner.sid) for owner in owners],
```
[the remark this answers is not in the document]
llllllllll: no, this is just following an old pattern

zipline/assets/exchange_info.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @property
+    def calendar(self):
+        return get_calendar(self.canonical_name)
+
+    def __eq__(self, other):
+        if not isinstance(other, ExchangeInfo):
+            return NotImplemented
+
+        return all(
```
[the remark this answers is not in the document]
llllllllll: I tend to not like to use vars unless I truly want to scan the attributes regardless of semantic meaning. I guess it is a style thing. idk.

zipline/testing/fixtures.py
llllllllll: it's not really default, it is the value that will be used in `make_equity_info` just like `ASSET_FINDER_EQUITY_NAMES` is the value used in the same function.

zipline/testing/fixtures.py
llllllllll: If you need more than one country you should just implement `make_exchange_info` just like when you need interesting assets and need to implement `make_equity_info`

zipline/assets/assets.py
```diff
@@ -310,37 +350,43 @@ def reload_symbol_maps(self):
...
-            pass
-        try:
-            del type(self).equity_supplementary_map[self]
-        except KeyError:
-            pass
-        try:
-            del type(self).equity_supplementary_map_by_sid[self]
-        except KeyError:
-            pass
+        for value in vars(type(self)).items():
```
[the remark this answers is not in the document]
llllllllll: good point, I just wish this wasn't so repetitive

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+        exchanges = pd.DataFrame({
+            'exchange': ['EXCHANGE %d' % n for n in range(num_exchanges)],
+            'country_code': [
+                self.country_code(n) for n in range(num_exchanges)
+            ]
+        })
+        self.write_assets(equities=metadata, exchanges=exchanges)
+        finder = self.asset_finder
+
+        # note: these assertions walk forward in time, starting at assertions
```
ssanderson: :+1: this is super helpful.

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+                self.country_code(0),
+                self.country_code(1),
+                self.country_code(2),
+            )
+        )
+        self.assertEqual(str(e.exception), expected_error_msg)
+
+    def test_endless_multiple_resolves(self):
+        """
+        Situation:
```
ssanderson: :+1:

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+                    'end_date': T('2014-01-05'),
+                },
+                {
+                    'symbol': 'A',  # claiming the unused symbol 'A'
+                    'asset_name': 'Asset A',
+                    'start_date': T('2014-01-06'),
+                    'end_date': T('2014-01-10'),
+                },
+            ] * num_exchanges,
+            index=np.repeat(np.arange(num_exchanges * 2), 2),
```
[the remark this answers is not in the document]
ssanderson: Ah, is this representing ticker changes? That might be a useful comment to put on the second records.

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+            with self.assertRaises(SymbolNotFound):
+                finder.lookup_symbol(
+                    'C',
+                    T('2013-12-31'),
+                    country_code=self.country_code(n),
+                )
+
+        for asof in pd.date_range('2014-01-01', '2014-01-05', tz='utc'):
+            # from 01 through 05 the first sid on the exchange held 'A'
+            with self.assertRaises(MultipleSymbolsFound):
```
ssanderson: Why does this raise for multiple symbols if the first sid on the exchange held A?

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+                )
+
+        for n in range(num_exchanges):
+            check_sid(n * 3, 'PRTY_HRD', n)
+            check_sid(n * 3 + 1, 'BRKA', n)
+            check_sid(n * 3 + 2, 'BRK_A', n)
+
+    def test_lookup_symbol_change_ticker(self):
+        T = partial(pd.Timestamp, tz='utc')
+        num_exchanges = 3
```
ssanderson: One thing that I've found a bit confusing while reading through these tests is that they're generally written with reference to the number of exchanges, but it seems like for the most part the thing we actually care about is the country code..

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+                    'A',
+                    asof,
+                    country_code=self.country_code(n),
+                )
+                assert_equal(
+                    A_result,
+                    finder.retrieve_asset(n * 2),
+                    msg=str(asof),
+                )
+                # The symbol and asset_name should always be the last held
```
ssanderson: These short comments giving informal reasoning/plain english descriptions of the assertions are super helpful.

tests/test_assets.py
ssanderson: Ah, I see, this is a conflict between exchanges. I wonder if it might be helpful (both here and generally for a consumer) to make a subclass of `MultipleSymbolsFound` that's specific to the case of "multiple symbols found in different countries". Making it a subclass means that calling code can still catch `MultipleSymbolsFound` and have it do the right thing, but it might be easier to track down an error and/or figure out a fix if you know that the conflict is cross-market rather than cross-time.

zipline/algorithm.py
llllllllll: I just moved this, idk if that change belongs in this pr.

zipline/algorithm.py
llllllllll: oh, I guess I could do that

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+                assert_equal(B_result.asset_name, 'Asset B')
+
+            # from 06 through 10 sid 1 held 'A'
+            # we test through the 11th because sid 1 is the last to hold 'A'
+            # so it should ffill
+            with self.assertRaises(MultipleSymbolsFound):
+                finder.lookup_symbol('A', asof)
+
+            for n in range(num_exchanges):
+                A_result = finder.lookup_symbol(
```
ssanderson: It seems like there are a few patterns of assertions in this function that could be factored out to make them read more clearly, e.g.:

```python
def assert_resolves(symbol, asof, country_to_sid, expected_symbol, expected_name):
    with self.assertRaises(MultipleSymbolsFound):
        finder.lookup_symbol(symbol, asof)
    for n in range(num_exchanges):
        result = finder.lookup_symbol(symbol, asof, country_code=self.country_code(n))
        ...  # Other assertions
    
```

zipline/assets/_assets.pyx
llllllllll: There is nothing that references this that I could find in zipline or our downstream consumers

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+                )
+                assert_equal(
+                    A_result,
+                    finder.retrieve_asset(n * 2 + 1),
+                    msg=str(asof),
+                )
+                assert_equal(A_result.symbol, 'A')
+                assert_equal(A_result.asset_name, 'Asset A')
+
+    def test_lookup_symbol(self):
```
ssanderson: It seems like basically all the tests in this suite are tests for lookup_symbol. Is this testing something more specific?

zipline/assets/_assets.pyx
llllllllll: We need the exchange_info to be be able to round trip from_dict and to_dict. I left the others for backwards compat because users may be using `to_dict()` and then reading these fields. The keys are constant so I could imagine someone assuming it is present.

tests/test_assets.py
```diff
@@ -940,63 +951,126 @@ def test_map_identifier_index_to_sids(self):
...
+            ],
+            ignore_index=True,
         )
-        self.write_assets(equities=frame)
+        # make every symbol unique
+        equities['symbol'] = [chr(ord('A') + i) for i in range(len(equities))]
+
+        # shuffle up the sids so they are not contiguous per exchange
+        sids = np.arange(len(equities))
+        np.random.RandomState(1337).shuffle(sids)
```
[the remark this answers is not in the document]
llllllllll: This is list length 10, I could also just hard code out a permutation which we know shuffles it up.

tests/test_assets.py
```diff
@@ -940,63 +951,126 @@ def test_map_identifier_index_to_sids(self):
...
-            assert_frame_equal(result, expected_with_start)
-
-            expected_no_start = pd.DataFrame(
-                data=expected_no_start_raw,
-                index=dates,
-                columns=frame.index.values,
-            )
-            result = finder.lifetimes(dates, include_start_date=False)
-            assert_frame_equal(result, expected_no_start)
+            for country_codes in powerset(exchanges.country_code.unique()):
```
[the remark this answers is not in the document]
llllllllll: I don't really want to duplicate this too much because this test is _really_ slow. I think it is probably better to more thoroughly explain this with comments.

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+            finder.lookup_symbol('PRTYHRD', dt)
+
+        for n in range(num_exchanges):
+            with self.assertRaises(SymbolNotFound):
+                finder.lookup_symbol(
+                    'PRTYHRD',
+                    None,
+                    country_code=self.country_code(n),
+                )
+                with self.assertRaises(SymbolNotFound):
```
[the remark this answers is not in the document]
llllllllll: yeah, we aren't hitting this

tests/test_assets.py
llllllllll: yeah, I should rename this

tests/test_assets.py
```diff
@@ -1376,6 +1450,538 @@ def test_error_message_plurality(self,
...
+            [
+                {
+                    'sid': n,
+                    'symbol':  'TEST.A',
+                    'company_name': "company %d" % n,
+                    'start_date': as_of.value,
+                    'end_date': as_of.value,
+                    'exchange': 'EXCHANGE %d' % n,
+                }
+                for n in range(num_assets)
```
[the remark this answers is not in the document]
llllllllll:
```
[self.country_code(n) for n in range(num_assets)],
```
is not about the sids though, I actually just want incrementing values for, one per sid.

tests/test_assets.py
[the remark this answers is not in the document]
llllllllll: I already have:
```
# second sid per country
```
what did you have in mind?

tests/test_assets.py
```diff
@@ -1627,159 +1629,131 @@ def test_lookup_symbol_change_ticker(self):
...
-                # values
-                assert_equal(C_result.symbol, 'A')
-                assert_equal(C_result.asset_name, 'Asset A')
+        # note: these assertions walk forward in time, starting at assertions
+        # about ownership before the start_date and ending with assertions
+        # after the end_date; new assertions should be inserted in the correct
+        # locations
+
+        # no one held 'A' before 01
+        assert_doesnt_resolve('A', T('2013-12-31'))
```
ssanderson: This is much easier to follow now.

zipline/assets/assets.py
```diff
@@ -350,7 +351,8 @@ def reload_symbol_maps(self):
         symbol maps.
         """
         # clear the lazyval caches, the next access will requery
-        for value in vars(type(self)).items():
+        for attr in dir(type(self)).items():
```
ssanderson: `dir` returns a list of strings, so I don't think `dir(type(self)).items()` can be correct here.

llllllllll: nothing even calls this function the test suite. nice!

zipline/assets/assets.py
```diff
@@ -912,7 +973,9 @@ def lookup_symbol(self,
             shareclass of ``BRK`` as ``BRK.A``, where others could write
             ``BRK_A``.
         country_code : str or None, optional
-            The country to limit searches to.
+            The country to limit searches to. If not provided, the search will
+            span all countries which increases the likely hood of an ambiguous
```
ssanderson: likelihood

zipline/assets/assets.py
ssanderson: FWIW, the only call-site of this is removed on https://github.com/quantopian/zipline/pull/2192

llllllllll: Okay, I removed the .items() because that is obviously wrong. I don't think we should invest any more time into this method if it is going away anyway.

tests/pipeline/test_statistical.py
```diff
@@ -400,7 +403,10 @@ def test_correlation_and_regression_with_bad_asset(self):
                 )
 
     def test_require_length_greater_than_one(self):
-        my_asset = Equity(0, exchange="TEST")
+        my_asset = Equity(
+            0,
+            exchange_info=ExchangeInfo('TEST', 'TEST FULL', 'US'),
```
dmichalowicz: Here (and a couple other tests it looks like) the order of `'TEST'` and `'TEST FULL'` look backwards.

PR #2244, opened 2018-07-11
OSX CI on Travis

richafrank:
OSX testing and packages


.travis.yml
```diff
@@ -17,30 +17,59 @@ env:
     #   See https://docs.travis-ci.com/api#external-apis.
     - secure: "qQ3vMSXUybUyKljJjgcVT9YaWcZ0PkdLZSFj6BVuo82SWO7e4L+hC8pqsa1Sm/EjyetVwfKk5lkvruPpro++8hDPLq3ahNOMmm4axfcWvACpW+acD/VCA5D8Dxz2ohrvpwsqo7BmIV6OUtIO5JeePHP1SqaIdVZFJ0pIAZToEWQ="
     - CONDA_ROOT_PYTHON_VERSION: "2.7"
-    - NUMPY_VERSION=1.11.1
-    - PANDAS_VERSION=0.18.1
-    - SCIPY_VERSION=0.17.1
+  matrix:
+    - PANDAS_18=1
```
ssanderson: What do you think about calling this `OLD_PANDAS` and `NEW_PANDAS` instead of encoding versions? I think the long term model we'd like to have is to support a "stable" version of pandas for use on Quantopian, and a relatively new release for general consumption.

.travis.yml
```diff
@@ -17,30 +17,59 @@ env:
...
+      language: generic
+      env: TRAVIS_PYTHON_VERSION=3.5 PANDAS_22=1
+
 cache:
   directories:
     - $HOME/.cache/.pip/
 
 before_install:
-  - if [ ${CONDA_ROOT_PYTHON_VERSION:0:1} == "2" ]; then wget https://repo.continuum.io/miniconda/Miniconda2-4.3.30-Linux-x86_64.sh -O miniconda.sh; else wget https://repo.continuum.io/miniconda/Miniconda3-4.3.30-Linux-x86_64.sh -O miniconda.sh; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then MINICONDA_OS=MacOSX; else MINICONDA_OS=Linux; fi
```
ssanderson: Thoughts on moving this into a script like `etc/travis_before_install.sh`? Writing bash in a yaml list seems strictly worse to me than writing it in a regular file that we invoke from the yaml.

.travis.yml
```diff
@@ -17,30 +17,59 @@ env:
...
+        sed -i "$@"
+      }
+    fi
+  - sed_inplace "s/numpy==.*/numpy==$NUMPY_VERSION/" etc/requirements.txt
+  - sed_inplace "s/pandas==.*/pandas==$PANDAS_VERSION/" etc/requirements.txt
+  - sed_inplace "s/scipy==.*/scipy==$SCIPY_VERSION/" etc/requirements.txt
+  - if [ -n "$PANDAS_DATAREADER_VERSION" ]; then sed_inplace "s/pandas-datareader==.*/pandas-datareader==$PANDAS_DATAREADER_VERSION/" etc/requirements.txt; fi
+  - if [ -n "$DASK_VERSION" ]; then sed_inplace "s/dask\[dataframe\]==.*/dask\[dataframe\]==$DASK_VERSION/" etc/requirements_blaze.txt; fi
   - cat etc/requirements.txt
+
```
ssanderson: Same question here for moving this into its own file instead of writing a bunch of bash inline.

.travis.yml
freddiev4: +1 on this

.travis.yml
```diff
@@ -17,30 +17,59 @@ env:
     #   See https://docs.travis-ci.com/api#external-apis.
     - secure: "qQ3vMSXUybUyKljJjgcVT9YaWcZ0PkdLZSFj6BVuo82SWO7e4L+hC8pqsa1Sm/EjyetVwfKk5lkvruPpro++8hDPLq3ahNOMmm4axfcWvACpW+acD/VCA5D8Dxz2ohrvpwsqo7BmIV6OUtIO5JeePHP1SqaIdVZFJ0pIAZToEWQ="
     - CONDA_ROOT_PYTHON_VERSION: "2.7"
-    - NUMPY_VERSION=1.11.1
-    - PANDAS_VERSION=0.18.1
-    - SCIPY_VERSION=0.17.1
+  matrix:
+    - PANDAS_18=1
+    - PANDAS_22=1
```
freddiev4: What does the syntax mean here? Does`PANDAS_22` to `1`  generate an environment variable?

richafrank: Yup, exactly. It sets that env var.

.travis.yml
richafrank: I had this internal debate as well. I don't have a super strong opinion, but the two thoughts I had before going with the proposed vars were:

1. When I go to the travis build matrix, I actually find it useful to know what versions are being tested, without having to click through to see that NEW=0.22.
1. Will there be transition periods where we support three versions briefly, or do you think the cut-overs will cleanly update the NEW version? If we ever have three, then again, I think explicit versions will be clearer than OLD, NEW, NEWER.

.travis.yml
richafrank: What do you think about breaking out the different pieces of this section into `ci/travis/install_miniconda.sh` and `ci/travis/overwrite_requirements.sh` as opposed to one bigger script?

.travis.yml
ssanderson: > Will there be transition periods where we support three versions briefly, or do you think the cut-overs will cleanly update the NEW version?

I don't think it's going to be feasible for us to realistically support more than two build configurations given the resources we currently have. I think the most likely model for transitions will be that we'll either update `OLD_PANDAS` to match `NEW_PANDAS`, in which case they'll be the same for a bit (and maybe we comment out the `NEW_PANDAS` builds temporarily) or we update `OLD_PANDAS` to some newer version that's still not as new as the newest (e.g., if we want to bump the version on Quantopian but can't upgrade all the way to the latest for some reason).

> When I go to the travis build matrix, I actually find it useful to know what versions are being tested, without having to click through to see that NEW=0.22.

I feel like we update pandas versions infrequently enough that it's not too hard to keep track of the OLD/NEW versions.

Overall, my instinct is that making it easier to just increase the `NEW_PANDAS` version will make it more likely that we actually keep the new version up to date, and that seems more valuable to me than the points you raised. I don't have a super strong opinion about this though.

.travis.yml
ssanderson: That seems fine to me as well. Having a few lines in the yaml doesn't bother me. It just seems a bit silly to have a full blown bash script here.

.travis.yml
richafrank: Sounds good, and we can always change it later.

richafrank:
Updated.

PR #2246, opened 2018-07-11
ENH: Add extension point for dataset-loader associations

jnazaren:
Implements a possible solution according to what @ssanderson suggested in #2199

### DIFF
--- a/tests/pipeline/test_dispatcher.py
+++ b/tests/pipeline/test_dispatcher.py
@@ -1,0 +1,138 @@
+import os
+
+from zipline.data import bundles
+from zipline.pipeline import USEquityPricingLoader
+from zipline.pipeline.data import (
+    Column,
+    DataSet,
+    BoundColumn,
+    USEquityPricing,
+)
+from zipline.pipeline.dispatcher import PipelineDispatcher
+from zipline.pipeline.loaders.base import PipelineLoader
+from zipline.pipeline.sentinels import NotSpecified
+from zipline.testing import ZiplineTestCase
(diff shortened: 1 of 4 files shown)


zipline/utils/run_algo.py
```diff
@@ -166,12 +171,13 @@ def _run(handle_data,
             bundle_data.adjustment_reader,
         )
 
+        # we register our default loader last, after any loaders from users
+        # have been registered via extensions
+        register_pipeline_loader(USEquityPricing, pipeline_loader)
+
         def choose_loader(column):
```
ssanderson: We don't need to create a new function here anymore. `global_pipeline_dispatcher` is already a callable with the right signature.

tests/pipeline/test_dispatcher.py
```diff
@@ -0,0 +1,140 @@
...
+        super(PipelineDispatcherTestCase, cls).init_class_fixtures()
+        cls.default_pipeline_loader = USEquityPricingLoader(
+            cls.bcolz_equity_daily_bar_reader,
+            cls.adjustment_reader,
+        )
+
+        cls.add_class_callback(clear_all_associations)
+
+    def test_load_not_registered(self):
+        fake_col_instance = FakeColumn(
```
ssanderson: I haven't looked carefully at these tests yet, but in general, it's not expected that anyone should ever construct BoundColumn instances explicitly. The usual way to get a bound column is to do:

```
class SomeDataSet(DataSet):
    column = Column(...)

SomeDataSet.column  # This evaluates to a BoundColumn because Column implements the descriptor protocol.
```

PR #2247, opened 2018-07-12
BUG: Fix crash in repr for non-BoundColumn inputs.

ssanderson:


llllllllll:
+1

ssanderson:
Just using `short_repr` isn't quite right actually. `short_repr` should really only be used for graphviz output now, because it contains graphviz-specific escapes like `\l` (which means left-aligned newline).

I've updated this to rename `short_repr` to `graph_repr`, and added a new `recursive_repr` to be used in contexts where we want a short repr for recursive rendering. The default recursive_repr is just "TypeName". ComputableTerms have a recursive repr of "TypeName(...)", and BoundColumns have a recursive repr of "DataSet.column_name".

ssanderson:
@jnazaren @llllllllll I'm gonna merge this on pass unless there are any objections.

zipline/pipeline/term.py
```diff
@@ -348,8 +348,16 @@ def dependencies(self):
...
+    def graph_repr(self):
+        """A short repr to use when rendering GraphViz graphs.
+        """
+        # Default graph_repr is just the name of the type.
+        return type(self).__name__
+
+    def recursive_repr(self):
+        """A short repr to use when recursively rendering terms with inputs.
+        """
+        # Default graph_repr is just the name of the type.
```
llllllllll: this comment should be updated

zipline/pipeline/term.py
```diff
@@ -679,10 +687,13 @@ def __repr__(self):
             "{type}([{inputs}], {window_length})"
         ).format(
             type=type(self).__name__,
-            inputs=', '.join(i.name for i in self.inputs),
+            inputs=', '.join(i.recursive_repr() for i in self.inputs),
             window_length=self.window_length,
         )
 
+    def recursive_repr(self):
```
llllllllll: why not have the `...` in the repr but have in in the short repr?

zipline/pipeline/visualize.py
```diff
@@ -174,8 +174,8 @@ def writeln(f, s):
 
 def fmt(obj):
     if isinstance(obj, Term):
-        if hasattr(obj, 'short_repr'):
-            r = obj.short_repr()
+        if hasattr(obj, 'graph_repr'):
+            r = obj.graph_repr()
```
llllllllll: don't all terms have this because it is defined on `Term`

ssanderson: Other stuff runs through this function as well (e.g., non-term parameters).

llllllllll: `if isinstance(obj, Term):`

ssanderson: Ah. idk then. This change was just made with find-replace. This is probably just a holdover from before I put a default impl on short_repr.

zipline/pipeline/visualize.py
```diff
@@ -174,10 +174,7 @@ def writeln(f, s):
...
 def fmt(obj):
     if isinstance(obj, Term):
-        if hasattr(obj, 'short_repr'):
-            r = obj.short_repr()
-        else:
-            r = type(obj).__name__
+        r = obj.graph_repr()
     else:
         r = obj
     return '"%s"' % r
```
llllllllll: This formatting is broken for strings. If you want to repr strings with quotes use `%r` I wouldn't block on this comment though.

PR #2253, opened 2018-07-16
Last minute release changes

ssanderson:


docs/source/release-process.rst
```diff
@@ -126,23 +126,20 @@ Edit your ``~/.pypirc`` file to look like:
    [distutils]
    index-servers =
        pypi
-       pypitest
 
    [pypi]
    username:
    password:
 
-   [pypitest]
```
ssanderson: This seems to not work anymore. The only way I could get pypitest to work was using the incantation below, which came from https://packaging.python.org/guides/using-testpypi/.

zipline/data/loader.py
```diff
@@ -295,9 +295,21 @@ def ensure_treasury_data(symbol, first_date, last_date, now, environ=None):
...
     if resource_name == 'benchmark':
-        from_csv = pd.Series.from_csv
+        # We expect a series here.
+        def from_csv(path):
+            return pd.read_csv(
+                path,
+                parse_dates=[0],
+                index_col=0,
+                header=None,
+            ).tz_localize('UTC').iloc[:, 0]
```
llllllllll: I think you just want to pass `squeeze=True` to `read_csv` instead of slicing here

docs/source/release-process.rst
```diff
@@ -126,23 +126,20 @@ Edit your ``~/.pypirc`` file to look like:
...
-   username:
-   password:
-
 after that, run:
 
 .. code-block:: bash
 
-   $ python setup.py sdist upload -r pypitest
+   $ pip install twine
+   $ twine upload --repository-url https://test.pypi.org/legacy/ dist/zipline-<version-number>.tar.gz
```
llllllllll: given this, we don't need to edit the pypirc at all

ssanderson: You still need it if you're going to include pypi creds there for the non-test upload I think?

llllllllll: it just prompts you

ssanderson:
image

PR #2257, opened 2018-07-18
ENH: Create extension point for data portals

jnazaren:
Creates an extension point for `DataPortal`, which currently defaults to `HistoricDataPortal`, the renamed version of `DataPortal`. This PR follows the method used in #2210.


zipline/utils/run_algo.py
```diff
@@ -179,6 +171,24 @@ def choose_loader(column):
         except ValueError as e:
             raise _RunAlgoError(str(e))
 
+    data = HistoricDataPortal(
```
richafrank: Would it make sense to create this only in an `else` branch below? A bunch of branches below ignore this value.

tests/test_algorithm.py
```diff
@@ -4301,7 +4301,7 @@ def make_equity_info(cls):
             orient='index',
         )
 
-    # XXX: This suite doesn't use the data in its DataPortal; it uses a
+    # XXX: This suite doesn't use the data in its HistoricDataPortal; it uses a
```
richafrank: Is this talking specifically about the implementation, or the interface?

richafrank:
Will be curious what @ehebert thinks of this class hierarchy being a public interface...

richafrank:
Do you expect sibling classes to the HistoricalDataPortal, or will they be subclasses? Makes me wonder if Historical is the right name.

jnazaren:
@richafrank I agree - I called it HistoricDataPorta because it made sense in terms of the type of data it was providing, but since I do imagine there would be subclasses to this implementation in the future, there's probably a better name for it. Would BasicDataPortal or DefaultDataPortal be more accurate?

jnazaren:
Closing in favor of #2269, which opens up the possibility of making readers an extension point instead.

PR #2258, opened 2018-07-20
Error message for Custom Factors that don't define compute message

leohentschker:
Adding in more descriptive error message for CustomFactors that don't define a compute method


zipline/pipeline/mixins.py
```diff
@@ -144,7 +144,9 @@ def compute(self, today, assets, out, *arrays):
         """
         Override this method with a function that writes a value into `out`.
         """
-        raise NotImplementedError()
+        raise NotImplementedError(
+            "Custom Factors must define a \"compute\" method."
```
ssanderson: This error will also get raised for CustomFilter and CustomClassifier, so we probably shouldn't just hard-code `Factor` in the message here.

PR #2259, opened 2018-07-20
BUG: Updating window safety validation with window_length > 1

leohentschker:
Updates `ComputableTerm` so that it doesn't raise `NonWindowSafeInput` when it has `window_length` 1.


tests/pipeline/test_slice.py
```diff
@@ -221,7 +221,7 @@ def compute(self, today, assets, out, col):
         my_unsafe_factor_slice = my_unsafe_factor[my_asset]
 
         class UsesSlicedInput(CustomFactor):
-            window_length = 1
+            window_length = 2
```
ssanderson: We should update `MySafeFactor` below as well so that it's still testing that we allow a potentially-unsafe factor to be run over a window-safe slice.

leohentschker: @ssanderson fixed in most recent commit

PR #2261, opened 2018-07-25
MAINT: Adding in validation for ComputableTerm input

leohentschker:
Make `ComputableTerm` check that the inputs are of type `Term`

1. Adds in a check to the `_validate` method
2. Adds in a new error `TermInputInvalid`
3. Adds a test ensuring the error is raised when an input of incorrect type is passed


ssanderson:
:+1: LGTM

PR #2262, opened 2018-07-26
International Pipelines

ssanderson:
This is an implementation of the work outlined in https://github.com/quantopian/zipline/issues/2265.

It includes:

- A new `zipline.pipeline.domain` module with a new `Domain` class.
- Generic pipeline expressions and limited support for specialization (see below).
- Updates to all USEquityPricing-based terms to use EquityPricing instead.
- Updates to all existing loaders to accept a new `domain` arg. As part of this change, I've added [`interface`](https://github.com/ssanderson/interface).
- New tests in `tests/pipeline/test_international_markets.py` and `tests/pipeline/test_domain_specialization.py`.
- Lots of updates to existing tests (mostly fixing bugs where we needed to apply specialization, or where we were failing to find any assets in the domain of the pipeline after applying specialization). In general, I updated tests that were already using the NYSE calendar to us the USEquities domain, and I updated tests that were specifying custom session indices to use EquitySessionDomain.

#### Domains and Calendars

In my original design, a domain was always guaranteed to be associated with a proper `trading_calendars.TradingCalendar`.  Much of the pipeline machinery, however, was already parametric on an arbitrary DatetimeIndex, which was used to compute output labels, and consequently there a bunch of places in the test suite where we were providing an index that doesn't naturally permit representation as a calendar (we could implement a "calendar" that just takes a DatetimeIndex, but at that point it's not really clear what use the calendar abstraction is providing on top of just having a DatetimeIndex).

In light of the above, the easiest way to get everything working was to change the interface of Domain to provide an `all_session` index instead of a full TradingCalendar. This has the advantage of requiring very few changes to the existing code, and after working with it I'm inclined to think that it's actually a cleaner interface for domains. I'd be curious to hear if people can think of cases where it'd be important to have access to a full calendar instead of just a sessions index. One case where it might be relevant is handling non-8:45 cutoff dates (see below), though we could just as easily add a `data_availability_cutoff` attribute to the domain interface.

#### Testing Gaps?

I've called out a few places in the test suite that seem like they would probably benefit from having some new tests written with domain parametricity in mind, but I'm sure there are some places I've missed.

#### Module Moves

I've split up `zipline.data.us_equity_pricing` into `zipline.data.adjustments` and `zipline.data.bcolz_daily_bars`.

#### Specialization only for `LoadableTerm`s

Currently, this PR currently only implements specialization for `LoadableTerm`s. Avoiding general-purpose specialization makes some things simpler (in particular, we don't have to figure out how to specialize all the different Filter/Factor/Classifier subclasses with non-uniform constructors), but it also adds some compliexity to `ExecutionPlan` and `PipelineEngine`, which are now responsible for specializing all `LoadableTerms` prior to:
  1. Passing terms to `get_loader`.
  2. Passing terms to any `load_adjusted_array` functions.
  
For the most part I think this branch manages to keep the complexity of specialization low by strategically specializing at the right time in ExecutionPlan (earlier drafts of this PR did the specialization in PipelineEngine, which was much more complicated).

#### Generic Terms and `populate_initial_workspace`

There's currently a big TODO block in SimplePipelineEngine about whether/how we want to support generic ComputableTerms in initial workspaces. My initial reasoning is that we shouldn't allow generics in an initial workspace, because by definition a generic term doesn't correspond to any concrete result until it's been specialized to a domain, but since we don't (yet) support specialization for generic terms, there isn't any way for `populate_initial_workspace` to signal that it can pre-compute a generic term unless it's allowed to provide a generic.

I think there are two reasonable ways we can go here:

1. Disallow any generic terms in initial workspaces, with the rationale being that a generic term never corresponds to a concrete output. This could eventually be fixed by adding support for specialization to all Pipeline terms.
2. Pass a domain to `populate_initial_workspace`, and have the contract be that generic terms should be populated with the specialization for the given domain. I think this is reasonable in isolation (and is probably the easiest thing to implement in the short term), but it also creates an awkward asymmetry when compared to `load_adjusted_array`, whose contract is currently that it always gets passed specialized terms.

#### (Not) Handling Non-8:45 Cutoff Times

One important piece of functionality missing from this PR is a system for handling different choices of cutoff time for intraday data availability. Happily, our first target market, Canada, has the same start time as NYSE, so I don't think we need to solve that problem for a first deliverable. 

Currently, only the `BlazeLoader` supports any notion of a variable cutoff time for intraday data bucketing, but the cutoff time is specified as a constructor parameter to the loader. I wonder if it might be more natural to associate the cutoff time with the domain? On the one hand, this seems reasonable because the domain is what knows about market-specific timing information (e.g., when the market opens). On the other hand, the loader should be responsible for actually knowing when data was ingested and should be made available. Maybe the blaze loader should be given a map from domain to availability time instead of a single static time?

#### Naming Things

Reprs, strs, and names are in a bit of an odd place at the moment on this branch.

Datasets currently repr as follows:

```
In [8]: EquityPricing
Out[8]: <DataSet: 'EquityPricing', domain=sentinel('NotSpecified')>

In [9]: USEquityPricing
Out[9]: <DataSet: 'EquityPricing', domain=EquityCountryDomain('US', 'NYSE')>
```

BoundColumns, however, repr like this:

```
In [10]: EquityPricing.close
Out[10]: EquityPricing.close::float64

In [11]: USEquityPricing.close
Out[11]: EquityPricing<US>.close::float64
```

The syntax here is inspired by Java/C++/Rust generics. Technically it's a bit misleading to have just the country code in the `<>` instead of a representation of a full domain, but I think in most real-world contexts domains are going to be primarily associated with countries (at least in the near term), and I don't want to make the BoundColumn repr a lot longer. (As is, I'm somewhat tempted to drop the type annotation).

These reprs seem individually fine, but it feels a little weird that we're rendering the dataset fairly differently in different contexts.


ssanderson:
The CI failures are from missing new dependencies. They pass for me locally. I'll update requirements/travis/appveyor tomorrow.

zipline/country.py
```diff
@@ -0,0 +1,14 @@
...
+
+class CountryCode(object):
+    """A simple namespace of iso3166 alpha2 country codes.
+    """
+    CANADA = countries_by_name['CANADA'].alpha2
+    GERMANY = countries_by_name['GERMANY'].alpha2
+    UNITED_KINGDOM = countries_by_name[
+        'UNITED KINGDOM OF GREAT BRITAIN AND NORTHERN IRELAND'
+    ].alpha2
+    UNITED_STATES = countries_by_name['UNITED STATES OF AMERICA'].alpha2
```
llllllllll: should this be comprehensive?

zipline/testing/core.py
```diff
@@ -1095,7 +1096,10 @@ def get_loader(column):
         yield SimplePipelineEngine(get_loader, calendar, finder)
 
 
-def parameter_space(__fail_fast=False, **params):
+_FAIL_FAST_DEFAULT = bool(os.environ.get('PARAMETER_SPACE_FAIL_FAST'))
```
llllllllll:
```python
In [2]: os.environ['A']
Out[2]: '0'

In [3]: bool(os.environ['A'])
Out[3]: True
```

llllllllll: maybe:
```python
def envvar_bool(key):
    value = os.environ.get(key)
    if value is None:
        return False
    return bool(json.loads(value.lower()))
```

zipline/utils/pandas_utils.py
```diff
@@ -320,3 +320,30 @@ def empty_dataframe(*columns):
...
+    indexes : iterable[pd.Index]
+        Iterable of indexes to check.
+
+    Raises
+    ------
+    ValueError
+        If the indexes are not all the same.
+    """
+    iterator = iter(indexes)
+    first = next(iterator)
```
llllllllll: should we support vacuous truth here?

zipline/testing/core.py
```diff
@@ -1594,6 +1598,37 @@ def prices_generating_returns(returns, starting_price):
...
+    ----------
+    starting_price : float
+        The first price of the series.
+    count : int
+        Number of price observations to return.
+    tick_size : float
+        Minimum unit of price movement between observations.
+    tick_range : (int, int)
+        Pair of lower/upper bounds for number of ticks observed between
+        observations. Samples w
```
llllllllll: I think this was clipped.

zipline/pipeline/pipeline.py
```diff
@@ -6,8 +7,10 @@
     optional,
 )
 
-from .graph import ExecutionPlan, TermGraph
+from .domain import Domain, infer_domain
+from .graph import ExecutionPlan, TermGraph, _SCREEN_NAME
 from .filters import Filter
+from .sentinels import NotSpecified, NotSpecifiedType
```
llllllllll: the pattern Python uses for None and Pandas uses for NaT is just to use `type(sentinel)` instead of importing the type directly.

zipline/pipeline/data/dataset.py
```diff
@@ -160,6 +167,29 @@ def _static_identity(cls, dataset, name, doc, metadata, *args, **kwargs):
...
+        return type(self)(
+            dtype=self.dtype,
+            missing_value=self.missing_value,
+            dataset=self._dataset.specialize(domain),
+            name=self._name,
+            doc=self.__doc__,
+            metadata=self._metadata,
+        )
+
+    def unspecialize(self):
```
llllllllll: should we call this `generalize`?

zipline/data/adjustments.py
```diff
@@ -0,0 +1,599 @@
+from collections import namedtuple
```
llllllllll: do I need to look over any particular changes in here or was this purely moved?

zipline/data/adjustments.py
```diff
@@ -0,0 +1,599 @@
...
+from ._adjustments import load_adjustments_from_sqlite
+
+log = Logger(__name__)
+
+
+SQLITE_ADJUSTMENT_TABLENAMES = frozenset(['splits', 'dividends', 'mergers'])
+
+UNPAID_QUERY_TEMPLATE = """
+SELECT sid, amount, pay_date from dividend_payouts
+WHERE ex_date=? AND sid IN ({0})
```
llllllllll: why is this a string template and not a query param?

zipline/assets/assets.py
```diff
@@ -1532,13 +1502,16 @@ def lifetimes(self, dates, include_start_date, country_codes):
         numpy.putmask
         zipline.pipeline.engine.SimplePipelineEngine._compute_root_mask
         """
-        # normalize this to a cache-key
+        # TODO_SS: Should we just accept a single string here?
+        if isinstance(country_codes, string_types):
```
llllllllll: I wrote a very similar thing the other day, string being an iterable is frustrating.

zipline/pipeline/graph.py
```diff
@@ -207,53 +221,69 @@ class ExecutionPlan(TermGraph):
...
                        term,
                        all_dates,
                        start_date,
                        end_date,
                        min_extra_rows):
-        """
-        Compute ``extra_rows`` for transitive dependencies of ``root_terms``
-        """
+        # Specialize any loadable terms before adding extra rows.
+        term = maybe_specialize(term, self.domain)
```
llllllllll: I call specialize in a lot of distinct places. Could we lift this to a single location?

zipline/pipeline/graph.py
```diff
@@ -165,7 +179,7 @@ def _decref_depencies_recursive(self, term, refcounts, garbage):
             # workspace to conserve memory.
             if refcounts[parent] == 0:
                 garbage.add(parent)
-                self._decref_depencies_recursive(parent, refcounts, garbage)
+                self._decref_dependencies_recursive(parent, refcounts, garbage)
```
llllllllll: classic

zipline/pipeline/engine.py
```diff
@@ -478,25 +487,45 @@ def compute_chunk(self, graph, dates, assets, initial_workspace):
...
+        # together requests for LoadableTerms if they are provided by the same
+        # loader and they require the same number of extra rows.
+        #
+        # The extra rows condition is a simplification: we don't currently have
+        # a mechanism for asking a loader to fetch different windows of data
+        # for different terms, so we only batch requests together when they're
+        # going to produce data for the same set of dates. That may change in
+        # the future if we find a loader that can still benefit significantly
+        # from batching unequal-length requests.
+        def loader_group_key(term):
```
llllllllll: this change is just a nop, right?

zipline/pipeline/graph.py
```diff
@@ -16,6 +15,14 @@ class CyclicDependency(Exception):
     pass
 
 
+# This sentinel value is uniquely-generated at import time so that we can
+# guarantee that it never conflicts with a user-provided column name.
+#
+# (Yes, technically, a user can import this file and pass this as the name of a
```
llllllllll: can we make this impossible?

zipline/pipeline/engine.py
```diff
@@ -452,17 +461,17 @@ def _inputs_for_term(term, workspace, graph):
...
         """
         Compute the Pipeline terms in the graph for the requested start and end
         dates.
 
+        This is where we do the actual work of running a pipeline.
+
         Parameters
         ----------
-        graph : zipline.pipeline.graph.TermGraph
+        graph : zipline.pipeline.graph.ExecutionPlan
```
llllllllll: the variable name `graph` seems to get used for both of these types of objects.

zipline/pipeline/engine.py
```diff
@@ -478,25 +487,45 @@ def compute_chunk(self, graph, dates, assets, initial_workspace):
...
+            return loader, extra_rows
+
+        # Only produce loader groups for the terms we expect to load.  This
+        # ensures that we can run pipelines for graphs where we don't have a
+        # loader registered for an atomic term if all the dependencies of that
+        # term were supplied in the initial workspace.
+        will_be_loaded = graph.loadable_terms - viewkeys(workspace)
         loader_groups = groupby(
             loader_group_key,
-            # Only produce loader groups for the terms we expect to load.  This
```
llllllllll: was this comment wrong before?

zipline/pipeline/graph.py
```diff
@@ -207,53 +221,69 @@ class ExecutionPlan(TermGraph):
...
+                 domain,
                  terms,
-                 all_dates,
                  start_date,
                  end_date,
                  min_extra_rows=0):
         super(ExecutionPlan, self).__init__(terms)
 
+        # Specialize loadable terms.
+        # NOTE: This isn't using self.loadable_terms because we don't want to
```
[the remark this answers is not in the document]
ssanderson: At this point in the construction the graph still contains un-specialized loadable terms. This is the point where we're actually going through and specializing all of them. We don't want to cache `self.loadable_terms` until after we've specialized them.

zipline/country.py
dmichalowicz: Also could this just be a dictionary instead of a class?

zipline/data/adjustments.py
ssanderson: This was just moved.

zipline/pipeline/graph.py
ssanderson: I don't think there's any plausible scenario in which this will matter unless a user is intentionally trying to break something, in which case I don't feel bad if this breaks.

zipline/pipeline/engine.py
ssanderson: `ExecutionPlan` is a subclass of `TermGraph`, so `graph` is accurate for both of them.

zipline/pipeline/engine.py
ssanderson: Yeah, I think previously we unconditionally called `loader_group_key` on loadable terms even if they were pre-supplied in the workspace.

zipline/data/adjustments.py
ssanderson: I think it's because we use pandas.read_sql_query to execute this, which doesn't support query params, but I'm not sure. This was here previously.

zipline/utils/pandas_utils.py
ssanderson: I can't think of a case where I'd want to call this with no inputs (or, for that matter, with one input, though that does currently work). I think this is different from `all` and `any` (which do support an empty input) because this is checking that a binary relation holds between elements in the input, vs. all/any which are checking a simple predicate on each input.

etc/requirements_blaze.txt
```diff
@@ -1,5 +1,7 @@
--e git://github.com/quantopian/datashape.git@bf06a41dc0908baf7c324aeacadba8820468ee78#egg=datashape-dev
--e git://github.com/quantopian/odo.git@da7f26d87702f5d293763e8ed54c7e25fd3af386#egg=odo-dev
+-e
```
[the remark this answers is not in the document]
ssanderson: I think this is just cruft. Not sure how it got here.

zipline/pipeline/engine.py
```diff
@@ -361,39 +365,38 @@ def _compute_root_mask(self, start_date, end_date, extra_rows):
...
         lifetimes = finder.lifetimes(
-            calendar[start_idx - extra_rows:end_idx],
+            sessions[start_idx - extra_rows:end_idx],
             include_start_date=False,
-            # TODO: update this when we add domains.
-            country_codes={'??', 'US'},
+            country_codes=(domain.country_code,),
         )
 
-        if lifetimes.index[extra_rows] != start_date:
```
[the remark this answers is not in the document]
ssanderson: These checks are equivalent to the checks up above that do `if start/end_date not in sessions`, and the message about "did you forget to align your inputs to the trading calendar" isn't as helpful now that the calendar is variable. I could see maybe putting these back in as assertions to sanity check that the lifetimes matrix calculation matches what we expect.

zipline/pipeline/loaders/frame.py
```diff
@@ -174,7 +175,13 @@ def load_adjusted_array(self, columns, dates, assets, mask):
...
-                adjustments=self.format_adjustments(dates, assets),
+                adjustments=self.format_adjustments(dates, sids),
                 missing_value=column.missing_value,
             ),
         }
+
+    def _validate_input_column(self, column):
+        """Make sure a passed column is our column.
+        """
+        if column != self.column and column.unspecialize() != self.column:
```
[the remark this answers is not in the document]
ssanderson: By unspecializing here, we allow a user to register a single DataFrame loader that can serve queries across multiple domains if the loader is constructed with a generic column. 

The `load_adjusted_array` implementation here does the right thing if you give it data for assets on multiple domains (it filters down to just the requested assets and calendars at call time). 

I think something like this will be necessary eventually to support dynamic domains, so given that this "just works" with what I think are the right semantics, I'm inclined to not force people to statically register a bunch of specializations if we don't need to.

zipline/pipeline/pipeline.py
ssanderson: I used `NotSpecifiedType` because that was the pattern we already had in a bunch of places. I could see an argument for updating this, but I'd rather do that in a separate atomic PR instead of rolling more miscellaneous changes into this.

zipline/pipeline/graph.py
ssanderson: We've talked about this a bunch in person. The general pattern in this implementation is that anything that comes out of `TermGraph` should be correctly specialized already, but anything that's being fetched from a term (in particular, `term.inputs` and `term.dependencies`) needs to be wrapped in a `maybe_specialize` call.

I think we agree that the right long term design for this is to specialize the entire term graph in the place where we currently specialize loadable terms. Implementing that either requires adding custom `specialize` implementations to 10 or so classes, or it requires refactoring `Term` a fair amount to allow reliably making functional copies of any term. I think I see a path toward the second option, but I'd prefer to push that into a follow-up, since I'm fairly confident that this implementation is correct, and I don't want to block downstream work on this.

tests/pipeline/base.py
```diff
@@ -95,24 +92,27 @@ def run_graph(self, graph, initial_workspace, mask=None):
         results : dict
             Mapping from termname -> computed result.
         """
+        def get_loader(c):
```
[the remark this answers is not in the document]
ssanderson: This differs because we now raise if we even call `get_loader` with a column. Previously we allowed calling `get_loader`, but would then return an ExplodingObject.

tests/pipeline/test_domain_specialization.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class MixedGenericsTestCase(zf.WithSeededRandomPipelineEngine,
+                            zf.ZiplineTestCase):
+    START_DATE = pd.Timestamp('2014-01-02', tz='utc')
+    END_DATE = pd.Timestamp('2014-01-31', tz='utc')
+    ASSET_FINDER_EQUITY_SIDS = (1, 2, 3, 4, 5)
+    ASSET_FINDER_COUNTRY_CODE = 'US'
+
+    def create_equivalent_sums(self, length):
```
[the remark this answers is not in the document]
ssanderson: Yeah, this is unused. Will remove.

tests/pipeline/test_blaze.py
```diff
@@ -863,6 +872,8 @@ def test_custom_query_time_tz(self):
                                   Equity(66 [B])          3    3.0
                                   Equity(67 [C])          4    4.0
         """
+        dates = self.dates
```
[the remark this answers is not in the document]
ssanderson: It's used in a run_pipeline call below. I updated this to be clearer.

tests/pipeline/test_dataset.py
```diff
@@ -0,0 +1,60 @@
+"""Tests for zipline.pipeline.data.DataSet and friends.
```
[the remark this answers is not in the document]
ssanderson: I merged this into `test_domain_specialization`

tests/pipeline/test_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+            # original column.
+            self.assertIsNot(original, new)
+            self.assertIs(new, original.specialize(USEquities))
+
+            # Columns should be bound to their respective datasets.
+            self.assertIs(original.dataset, MyDataSet)
+            self.assertIs(new.dataset, specialized)
+
+            # The new column should have the domain of the specialization.
+            assert_equal(new.domain, USEquities)
```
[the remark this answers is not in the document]
ssanderson: I think we only use identity in this file at least for dataset/column checks. We don't currently depend on identity for domains anywhere that I know of (we use it as part of the static identity when constructing new terms, but that just depends on equality).

tests/pipeline/test_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+
+            # Names, dtypes, and missing_values should match.
+            assert_equal(original.name, new.name)
+            assert_equal(original.dtype, new.dtype)
+            assert_equal(original.missing_value, new.missing_value)
+
+    def test_unspecialize(self):
```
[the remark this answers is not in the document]
ssanderson: Updated

tests/pipeline/test_downsampling.py
```diff
@@ -726,6 +730,8 @@ class TestDownsampledRowwiseOperation(WithAssetFinder, ZiplineTestCase):
 
     dates = pd.date_range(START_DATE, END_DATE)
 
+    ASSET_FINDER_COUNTRY_CODE = '??'
```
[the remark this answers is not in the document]
ssanderson: I kept this as `'??'` because this suite actually just uses `date_range` as its calendar. I don't want to use US for this since that would signal to me that I should expect the US calendar. Do you have thoughts on a better code to use for "just testing"/"synthetic"?

Issue #2265, opened 2018-07-30
Design Document: Pipeline Domains


ssanderson:
This issue is a work-in-progress public design document for a series of upcoming changes to Zipline that move us toward the goal of supporting international equities. Specifically, this document outlines the design and motivation for adding international equity support to the Pipeline API via the concept of **domains**.

### Tasks:

- [x] Domain Objects (https://github.com/quantopian/zipline/pull/2262)
- [x] Dataset and Column Specialization (https://github.com/quantopian/zipline/pull/2262)
- [x] PipelineEngine Updates (https://github.com/quantopian/zipline/pull/2262)
- [x] Built-in PipelineLoader Updates (https://github.com/quantopian/zipline/pull/2262)
- [x] Variable Data Availability Times (https://github.com/quantopian/zipline/pull/2288)

### Open Questions
- The design for specialization proposed here complicates the responsibilities of the `get_loader` function used to determine what loader should be used to service a given LoadableTerm. Previously, most `get_loader` functions could be constructed using a simple dictionary from `BoundColumn -> PipelineLoader`. With this design, however, users will need to take care to call `specialize` ahead of time if they intend to produce a `get_loader` function for pipelines running on a particular domain. 

  Alternatively, users may want to implement a fallback mechanism for looking up specialized columns in terms in a dictionary that may contain the generic version of the associated column (e.g., I might want to register a loader for all `EquityPricing` queries, and have dispatching behave as expected when a lookup happens for a concrete specialization. Something like @jnazaren's work in https://github.com/quantopian/zipline/pull/2246 might help simplify the work required here.)

# TL;DR

We're adding a new idea of a "domain" to the Pipeline API. The domain of a Pipeline determines the top-level set of assets that the pipeline computes over, as well as the calendar used to align and bucket data observations. Concretely, these correspond to the column and row labels of the 2D arrays passed to CustomFactor.compute (as well as to PipelineLoader.load_adjusted_array, and various other functions internal to the pipeline machinery).

- In the near term, we intend to use domains to support the creation of pipelines and pipeline expressions that are parametric over a user's choice of a single market. As part of this work, we're adding the ability to expression pipeline computations that are **generic**, in the sense that they can be re-used without changes on any market for which the user has provided data.
- In the medium term, we may extend domains to support set-like operations, allowing users to compute over unions of domains (e.g., US + Japan + Germany).
- In the long term, we might consider generalizing further to support pipeline computations over non-asset domains like companies, countries or sectors.

[...]

PR #2266, opened 2018-07-30
ENH: Domains, Inference, and Specialization

ssanderson:
This is the first in a series of PRs that aim to implement support for **domains** in the Pipeline API. See https://github.com/quantopian/zipline/issues/2265 for WIP design document.

In particular, this PR implements initial support for Domain objects, along with the machinery for **specialization** of DataSets and BoundColumns to domains.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -80,3 +80,6 @@ tables==3.3.0
 
 # For trading calendars
 trading-calendars==1.0.1
+
+# Interface definitions
+python-interface==1.4.0
--- a/tests/pipeline/test_dataset.py
+++ b/tests/pipeline/test_dataset.py
@@ -1,6 +1,6 @@
 """Tests for zipline.pipeline.data.DataSet and friends.
 """
-from zipline.pipeline.domain import USEquities
+from zipline.pipeline.domain import USEquities, CanadaEquities
 from zipline.pipeline.data.dataset import Column, DataSet
 import zipline.testing.fixtures as zf
 from zipline.testing.predicates import assert_equal
@@ -45,3 +45,24 @@ class DataSetTestCase(zf.ZiplineTestCase):
             assert_equal(original.name, new.name)
             assert_equal(original.dtype, new.dtype)
             assert_equal(original.missing_value, new.missing_value)
+
+    def test_repr(self):
+        class Data(DataSet):
+            col1 = Column(dtype=float)
+            col2 = Column(dtype=int, missing_value=100)
+            col3 = Column(dtype=object, missing_value="")
+
+        expected = "<DataSet: 'Data'>"
+        self.assertEqual(repr(Data), expected)
+
+        specialized = Data.specialize(USEquities)
+        expected_specialized = "<DataSet: 'Data_US', domain={}>".format(
+            str(USEquities),
+        )
+        self.assertEqual(repr(specialized), expected_specialized)
+
+        specialized_CA = Data.specialize(CanadaEquities)
+        expected_specialized_CA = "<DataSet: 'Data_CA', domain={}>".format(
+            str(CanadaEquities),
+        )
+        self.assertEqual(repr(specialized_CA), expected_specialized_CA)
--- a/tests/pipeline/test_dataset.py
+++ b/tests/pipeline/test_dataset.py
@@ -1,0 +1,47 @@
+"""Tests for zipline.pipeline.data.DataSet and friends.
+"""
+from zipline.pipeline.domain import USEquities
+from zipline.pipeline.data.dataset import Column, DataSet
+import zipline.testing.fixtures as zf
+from zipline.testing.predicates import assert_equal
+
+
+class DataSetTestCase(zf.ZiplineTestCase):
+
+    def test_specialize(self):
+        class MyDataSet(DataSet):
+            col1 = Column(dtype=float)
+            col2 = Column(dtype=int, missing_value=100)
+            col3 = Column(dtype=object, missing_value="")
+
+        specialized = MyDataSet.specialize(USEquities)
+
+        # Specializations should be memoized.
+        self.assertIs(specialized, MyDataSet.specialize(USEquities))
+
+        # Specializations should have the same name, but prefixed with the
+        # country code of the new domain.
+        assert_equal(specialized.__name__, "MyDataSet_US")
+        self.assertIs(specialized.domain, USEquities)
+
+        for attr in ('col1', 'col2', 'col3'):
+            original = getattr(MyDataSet, attr)
+            new = getattr(specialized, attr)
+
+            # We should get a new column from the specialization, which should
+            # be the same object that we would get from specializing the
+            # original column.
+            self.assertIsNot(original, new)
+            self.assertIs(new, original.specialize(USEquities))
+
+            # Columns should be bound to their respective datasets.
+            self.assertIs(original.dataset, MyDataSet)
+            self.assertIs(new.dataset, specialized)
+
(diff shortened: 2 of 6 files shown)


zipline/pipeline/domain.py
```diff
@@ -0,0 +1,182 @@
...
+        """Get all trading sessions for the calendar of this domain.
+        """
+
+
+# Create a base class so that we can do type-based validation of user input.
+Domain = implements(IDomain)
+Domain.__name__ = 'Domain'
+
+
+# TODO: Better name for this?
```
vikram-narayan: `CountryDomain`, or somesuch might be more descriptive if we have iterate through subclasses to get a list of domains?

zipline/pipeline/domain.py
```diff
@@ -0,0 +1,182 @@
...
+    def __repr__(self):
+        return "StandardDomain(name={!r}, country={!r}, calendar={!r})".format(
+            self._name,
+            self.country_code,
+            self._calendar_name,
+        )
+
+
+# TODO: Is this the casing convention we want for domains?
+USEquities = StandardDomain('USEquities', CountryCode.UNITED_STATES, 'NYSE')
```
vikram-narayan: would it be helpful to namespace domains?

tests/pipeline/test_dataset.py
```diff
@@ -0,0 +1,68 @@
...
+        class Data(DataSet):
+            col1 = Column(dtype=float)
+            col2 = Column(dtype=int, missing_value=100)
+            col3 = Column(dtype=object, missing_value="")
+
+        expected = "<DataSet: 'Data'>"
+        self.assertEqual(repr(Data), expected)
+
+        specialized = Data.specialize(USEquities)
+        expected_specialized = "<DataSet: 'Data_US', domain={}>".format(
```
vikram-narayan: is it necessary to append `_Country` if the domain is also in the repr?

zipline/pipeline/domain.py
vikram-narayan: nvm, realized you can just do `StandardDomain.__subclasses__()`

etc/requirements.txt
```diff
@@ -80,3 +80,6 @@ tables==3.3.0
 
 # For trading calendars
 trading-calendars==1.0.1
+
```
vikram-narayan: should iso3166 be added here as well?

zipline/pipeline/domain.py
```diff
@@ -0,0 +1,182 @@
...
+    Returns
+    -------
+    inferred : Domain or NotSpecified
+
+    Raises
+    ------
+    AmbiguousDomain
+        Raised if more than one concrete domain is present in the input terms.
+    """
+    domains = {NotSpecified}
```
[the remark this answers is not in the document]
ssanderson: It makes the logic below simpler. If we initialized this to an empty set, then we'd have to check the `len(domains) == 0` case to handle an empty input, and in the `len(domains) == 1` or `len(domains) == 2` cases, we'd have to have sub-branches depending on whether or not `NotSpecified` was in the inputs (e.g., it's an error to have two different non-NotSpecified domains in the input).

I think a reasonable alternative implementation would be:

```python
def infer_domain(terms):
    domains = {}
    for t in terms:
        domains.update(t.domain)

    try:
        domains.remove(NotSpecified)
    except KeyError:
        pass

    if len(domains) == 0:
        return NotSpecified
    elif len(domains) == 1:
        return domains.pop()
    else:
        raise AmbiguousDomain(...)
```

These feel like they have about the same amount of complexity in them. Maybe the latter is slightly better since it only requires one call to `remove`.

zipline/pipeline/domain.py
ssanderson: I'm not sure we'd expect/want a user to be doing that, but the `domain` module itself is also potentially a perfectly good namespace.

zipline/country.py
```diff
@@ -0,0 +1,14 @@
+"""Canonical definitions of country code constants.
+"""
+from iso3166 import countries_by_name
+
+
+class CountryCode(object):
```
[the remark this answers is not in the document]
ssanderson: I found this useful for ones where I didn't know the name offhand, or where the name was surprising to me. In particular, I kept wanting to use `'UK'` instead of `'GB'`. Going through iso3166 seemed like a nice way to make sure we're actually using the correct code for any given country.

zipline/pipeline/domain.py
ssanderson: CountryDomain seems not quite right because, at least right now, all domains have a country code (that might change in the future, but it will be a pretty significant change if/when it happens). 

Really the difference here is whether we get our dates via a named calendar or via a simple DatetimeIndex.

zipline/pipeline/data/dataset.py
```diff
@@ -248,8 +276,58 @@ def __new__(mcls, name, bases, dict_):
...
+            self._domain_specializations[domain] = new_type
+            return new_type
+
+    def _create_specialization(self, domain):
+        assert self.domain is NotSpecified, \
+            "Can't specialize non-generic dataset!"
+
+        # Create a new subclass of ``self`` with the given domain whose name is
+        # our name prefixed with domain.country_code.
+        name = '_'.join((self.__name__, domain.country_code))
```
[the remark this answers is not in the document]
ssanderson: The main reason I added this was that the dataset name shows up in the repr for BoundColumn. and I didn't want to make that repr any longer by including a domain, so suffixing the dataset with a country code seemed like a simple way to make that information available somewhere. I'm open to suggestions for other ideas on this.

zipline/pipeline/domain.py
```diff
@@ -0,0 +1,182 @@
...
+
+from trading_calendars import get_calendar
+
+from zipline.country import CountryCode
+from zipline.utils.memoize import lazyval
+
+from .sentinels import NotSpecified
+
+
+class IDomain(Interface):
```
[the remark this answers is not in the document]
ssanderson: I think this is useful for keeping the testing implementations in sync with the main implementation. I've already gone through at least one major revision of what methods I needed here. Initially these provided a calendar instead of just sessions, but that made it hard to port some of the blaze core loader tests (see note on https://github.com/quantopian/zipline/pull/2266/files/0a6405a1b65461752e10d8651447b070879a1e66#diff-2e4c20f8c11194598b15ac870e83ff15R31), so I changed to have this just provide a session index, which is (at least right now) all we need.

The one thing I'm not super excited about on this is that we actually care about types here because we want to be able to validate user input, and at least on quantopian I don't think we want to support duck-typed domains. But I think even with the slightly awkward workaround the extra strictness of interface makes it a better `abc` than `abc` for what we want here.

zipline/pipeline/data/dataset.py
```diff
@@ -160,6 +166,28 @@ def _static_identity(cls, dataset, name, doc, metadata, *args, **kwargs):
             frozenset(sorted(metadata.items(), key=first)),
         )
 
+    def specialize(self, domain):
+        """Specialize ``self`` to a concrete domain.
+        """
+        if self.domain == domain:
```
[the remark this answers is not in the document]
ssanderson: Most of the time when we're specializing stuff, our logic is essentially "take a mix of generic and already-specialized terms, and make sure they're all specialized to an actual domain", so having this work allows us to just write:

`[term.specialize(domain) for term in terms]`

instead of something more verbose like:

`[term.specialize(domain) if term.domain is not domain else term for term in terms]`

ssanderson:
Closing this in favor of https://github.com/quantopian/zipline/pull/2262.

PR #2282, opened 2018-08-29
DOC: Add local installation in dev guide.

### LINKED ISSUE #2281: Local installation of pip packages
Hi Zipline maintainers!

I was trying to locally build and install Zipline from Git in my machine.
The documentation is really great, very well explained.

It was easy to download, build but it was lacking instructions on how to install the package.
Half an hour later I found in a forum that a `python setup.py install`do the job.
It's quite obvious and even a bit stupid that this is the first thing to try, but would be nice to add that to the documentation (http://www.zipline.io/development-guidelines.html).

Kind regards,
--
Willian Paixao

willianpaixao:
Fixes #2281

### DIFF
--- a/docs/source/development-guidelines.rst
+++ b/docs/source/development-guidelines.rst
@@ -36,11 +36,18 @@ The following section assumes you already have virtualenvwrapper and pip install
    $ pip install -r ./etc/requirements_dev.txt
    $ pip install -r ./etc/requirements_blaze.txt 
 
-Finally, you can build the C extensions by running:
+You can build the C extensions by running:
 
 .. code-block:: bash
 
    $ python setup.py build_ext --inplace
+
+In case you want to install the package locally and use ``zipline`` wrapper script:
+
+.. code-block:: bash
+
+   $ pip install -e .
+   $ zipline --help
 
 To finish, make sure `tests`__ pass.
 


docs/source/development-guidelines.rst
```diff
@@ -36,12 +36,19 @@ The following section assumes you already have virtualenvwrapper and pip install
...
 
 .. code-block:: bash
 
    $ python setup.py build_ext --inplace
 
+In case you want to install the package locally and use ``zipline`` wrapper script:
+
+.. code-block:: bash
+
+   $ pip install .
```
ssanderson: What do you think about recommending [`pip install -e .`](https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs) here instead? Using an editable install means that if you make changes in the source tree, those changes will have an effect when you run zipline. Since this doc is in part targeting users that want to contribute to Zipline, I could imagine someone being confused if they do `pip install .`, make changes, and then don't see those changes have any effect.

ssanderson:
The travis failure here is spurious (we need to bump an outdated package version).

docs/source/development-guidelines.rst
willianpaixao: Agree. Just updated the commit.

ssanderson:
I'll merge this once we get the builds sorted out.

PR #2288, opened 2018-09-05
ENH: move the data_query_time to the domain

llllllllll:


zipline/pipeline/domain.py
```diff
@@ -100,27 +123,53 @@ class EquityCalendarDomain(Domain):
         ISO-3166 two-letter country code of the domain
     calendar_name : str
         Name of the calendar, to be looked by by trading_calendar.get_calendar.
+    data_query_offset : datetime.timedelta
+         The offset from market open when data should no longer be considered
+         available for a session.
```
ssanderson: An example might be helpful here, e.g. "For example, a data_query_offset of -timedelta(minutes=45) signifies that data must have been available 45 minutes prior to market open for it to appear in pipeline inputs for a given day."

zipline/pipeline/domain.py
```diff
@@ -100,27 +123,53 @@ class EquityCalendarDomain(Domain):
...
-        return self.calendar.all_sessions
-
     @lazyval
     def calendar(self):
         return get_calendar(self._calendar_name)
 
+    def all_sessions(self):
+        return self.calendar.all_sessions
+
+    def data_query_cutoff_for_sessions(self, sessions):
```
ssanderson: Can we add unit tests to exercise the success and fail cases here?

zipline/pipeline/domain.py
```diff
@@ -100,27 +123,53 @@ class EquityCalendarDomain(Domain):
...
     @expect_types(
         country_code=str,
         calendar_name=str,
         __funcname='EquityCountryDomain',
     )
-    def __init__(self, country_code, calendar_name):
+    def __init__(self,
+                 country_code,
+                 calendar_name,
+                 data_query_offset=-datetime.timedelta(minutes=45)):
```
ssanderson: Can we add a test somewhere that passes this with a non-default value and verifies that it affects results?

zipline/pipeline/domain.py
```diff
@@ -211,19 +260,42 @@ class EquitySessionDomain(Domain):
...
     @property
     def country_code(self):
         return self._country_code
 
     def all_sessions(self):
         return self._sessions
+
+    def data_query_cutoff_for_sessions(self, sessions):
+        return days_at_time(
+            sessions.tz_localize('UTC'),
```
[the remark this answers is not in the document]
llllllllll: In the tests, it appears to never be localized, meaning we are passing unlocalized dts to pipeline loaders.

zipline/pipeline/domain.py
```diff
@@ -100,27 +123,53 @@ class EquityCalendarDomain(Domain):
...
     @lazyval
     def calendar(self):
         return get_calendar(self._calendar_name)
 
+    def all_sessions(self):
+        return self.calendar.all_sessions
+
+    def data_query_cutoff_for_sessions(self, sessions):
+        try:
+            opens = self.calendar.opens.loc[sessions]
```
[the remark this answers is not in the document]
llllllllll: it just does broadcasted lookups, index is array-like. I don't think this should require continuous dates, and in the past I have used this for non continuous dates.

zipline/pipeline/domain.py
```diff
@@ -52,6 +55,21 @@ def country_code(self):
...
         """
 
+    def data_query_cutoff_for_sessions(self, sessions):
+        """Compute the data query cutoff time for the given sessions.
+
+        Parameters
+        ----------
+        sessions : pd.DatetimeIndex
+            The sessions to get the data query cutoff times for. This index
+            will contain all midnight UTC values.
```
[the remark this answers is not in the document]
llllllllll: I don't believe these docstrings make it to the implementations of these methods, so I had written them with the implementer of the interface in mind. These are preconditions that can be assumed when writing the function. Idk if this is a good pattern.

zipline/pipeline/loaders/utils.py
```diff
@@ -132,165 +138,15 @@ def previous_event_indexer(all_dates,
     return out
 
 
-def normalize_data_query_time(dt, time, tz):
```
ssanderson: RIP

zipline/pipeline/loaders/utils.py
```diff
@@ -112,11 +114,15 @@ def previous_event_indexer(all_dates,
...
+    out = np.full(
+        (len(data_query_cutoff_times), len(all_sids)),
+        -1,
+        dtype=np.int64,
+    )
 
     eff_dts = np.maximum(event_dates, event_timestamps)
     sid_ixs = all_sids.searchsorted(event_sids)
-    dt_ixs = all_dates.searchsorted(eff_dts)
+    dt_ixs = data_query_cutoff_times.searchsorted(eff_dts, side='right')
```
ssanderson: Reasoning through this change:

`data_query_cutoff_times` is the index into which we're aligning our input data. For each event record `i`, `dt_ix[i]` should be the index into the cutoff times of the first date on which we should show the event's data. The only case where `searchsorted` produces different values is when the value being looked up appears in the searched data. so this is effectively saying that if an event has a timestamp that's exactly on the data_query_cutoff_time, we push it forward to the next day.

Did I get that right? If so, what prompted this change? This seems like a semi-arbitrary distinction.

llllllllll: we use side=right in the blaze core loader. I wanted to consistently handle this case wherever we care about the cutoff time.

PR #2289, opened 2018-09-06
BUG: fix support for non unique sid->exchange mappings.

llllllllll:
We now take the most recent.


zipline/assets/asset_writer.py
```diff
@@ -355,10 +355,10 @@ def _split_symbol_mappings(df, exchanges):
     mappings['sid'] = mappings.index
     mappings.reset_index(drop=True, inplace=True)
 
-    asset_exchange = df[['exchange']]
-    asset_exchange['sid'] = df.index
-    asset_exchange.drop_duplicates(inplace=True)
-    asset_exchange = asset_exchange['exchange']
+    # take the most recent sid->exchange mapping based on end date
```
ssanderson: Is there a place we can add a quick test for this behavior?

PR #2290, opened 2018-09-06
TST: add test for resolving overloaded sid->exchange

llllllllll:


tests/test_assets.py
```diff
@@ -2289,11 +2290,51 @@ def test_read_from_asset_finder(self):
...
+                for n, date in enumerate(dates)
+                for sid in sids
+            ]
+        )
+        self.writer.write(equities=df)
+
+        reader = self.new_asset_finder()
+        equities = reader.retrieve_all(reader.sids)
+
+        expected_exchange = 'EXCHANGE-%d' % (len(dates) - 1)
```
ssanderson: Out of paranoia, should we give each (or at least one of) the assets a different exchange to make sure that we're correctly deduplicating only within each asset?

PR #2291, opened 2018-09-11
MAINT: Silence warning about setting with copy.

ssanderson:
Pandas raises a warning here because we're assigning to a column of a
subset of a dataframe. This warning is intended to prevent people from
doing things like::

    `df[df.col1 > 10]['col1'] = <value>`

which has no effect because the first [] creates a copy, which then gets
mutated, so the original frame remains unchanged.

In this case, however, we've bound the copy to a name that we're going to
continue to use, so we don't care about the warning.


llllllllll:
LGTM

PR #2295, opened 2018-09-17
HDF5 daily bars

yankees714:
- Adds `HDF5DailyBarWriter`, which writes daily pricing in a new format as an HDF5 file. Each OHLCV field is stored as a 2D array in a chunked HDF5 dataset, with a row per sid and a column per day. The file also supports multiple countries.
- Add `HDF5DailyBarReader`, which implements the `BarReader` interface and can read files written by `HDF5DailyBarWriter`


zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+    def _compute_date_range_slice(self, start_date, end_date):
+        # Get the index of the start of dates for ``start_date``.
+        start_ix = self.dates.searchsorted(start_date)
+
+        # Get the index of the start of the first date **after** end_date.
+        end_ix = self.dates.searchsorted(end_date, side='right')
+
+        return slice(start_ix, end_ix)
+
+    def _requesteddates(self, start_date, end_date):
```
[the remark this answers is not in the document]
yankees714: I think this isn't actually used anymore

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+        """
+        return pd.Timestamp(self.dates[0], tz='UTC')
+
+    @property
+    def sessions(self):
+        """
+        Returns
+        -------
+        sessions : DatetimeIndex
+           All session labels (unionining the range for all assets) which the
```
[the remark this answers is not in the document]
yankees714: The BcolzDailyBarReader has this implemented as a lazyval, I'll update this to do the same

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+                        columns,
+                        start_date,
+                        end_date,
+                        assets):
+        """
+        Parameters
+        ----------
+        columns : list of str
+           'open', 'high', 'low', 'close', or 'volume'
+        start_dt: Timestamp
```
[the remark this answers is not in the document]
yankees714: Ah, this was copied over from the docstring for BarReader.load_raw_arrays. I'll update both of them

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+    def first_trading_day(self):
+        """
+        Returns
+        -------
+        dt : pd.Timestamp
+            The first trading day (session) for which the reader can provide
+            data.
+        """
+        return pd.Timestamp(
+            min(reader.dates[0] for reader in self._readers.values()),
```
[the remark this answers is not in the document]
yankees714: Nope there's not, will change

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+    date_chunk_size : int
+        The number of days per chunk in the HDF5 file. If this is
+        greater than the number of days in the data, the chunksize will
+        match the actual number of days.
+
+    See Also
+    --------
+    zipline.data.hdf5_daily_bars.HDF5DailyBarReader
+    """
+    def __init__(self, filename, date_chunk_size, driver=None):
```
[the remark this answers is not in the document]
yankees714: Ah, I was playing around with this in case we wanted to use a different driver to make testing easier. Going to remove this for now, should be easy enough to add back if we do want it.

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+                dataset.attrs['scaling_factor'] = scaling_factors[field]
+
+                log.debug(
+                    'Writing dataset {} to file {}',
+                    dataset.name, self._filename
+                )
+
+        return self._h5_file(mode='r')
+
+    def write_from_sid_df_pairs(self, country_code, iterable):
```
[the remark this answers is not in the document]
yankees714: This is intended to mimic the write method on the BcolzDailyBarWriter, where we call this `data`, so I'll update this to match.

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+        dt : pd.Timestamp
+            The minute at which to start searching for the last traded minute.
+
+        Returns
+        -------
+        last_traded : pd.Timestamp
+            The dt of the last trade for the given asset, using the input
+            dt as a vantage point.
+        """
+        pass
```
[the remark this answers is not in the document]
yankees714: I think this can just be implemented using `asset_end_dates` now?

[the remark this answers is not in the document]
yankees714: Oh, BcolzDailyBarReader implements this by finding the last day with positive volume.

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,668 @@
...
+FIELDS = (OPEN, HIGH, LOW, CLOSE, VOLUME)
+
+DAY = 'day'
+SID = 'sid'
+
+START_DATE = 'start_date'
+END_DATE = 'end_date'
+
+
+scaling_factors = {
```
llllllllll: shouldn't we read this out of the file's metadata?

yankees714: These are the factors that get written to the file (this is hardcoded right now but I plan to make it a parameter with these as the defaults). In the reader, they are read out of the file [here](https://github.com/quantopian/zipline/pull/2295/files/d754fb93b29f6069be49fff1a6f99b6f63aa34ff#diff-1a70086405577f2f9433c7624909ec30R261).

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+    Returns
+    -------
+    start_date_ixs : np.array[int64]
+        The index of the first date with non-nan values, for each sid.
+    end_date_ixs : np.array[int64]
+        The index of the last date with non-nan values, for each sid.
+    """
+    # Build a 2D array (dates x sids), where an entry is True if all
+    # fields are nan for the given day and sid.
+    is_null_matrix = reduce(
```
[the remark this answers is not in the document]
yankees714: TIL. I'm not too worried about this part, so I'll switch to use the built-in and at least avoid the extra import. We can change it back if it causes problems.

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,634 @@
...
+        if not frame_0.index.equals(frame_1.index):
+            raise ValueError('Frames have mistmatched days.')
+
+        if not frame_0.columns.equals(frame_1.columns):
+            raise ValueError('Frames have mismatched sids.')
+
+    return frame_0.index.values, frame_0.columns.values
+
+
+class HDF5DailyBarWriter(object):
```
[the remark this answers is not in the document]
yankees714: @ssanderson Added in cead3eb

zipline/testing/fixtures.py
```diff
@@ -837,6 +841,11 @@ def init_class_fixtures(cls):
 
         cls.equity_daily_bar_days = days
 
+        if cls.EQUITY_DAILY_BAR_COUNTRY_CODES is None:
+            cls.EQUITY_DAILY_BAR_COUNTRY_CODES = tuple(
+                cls.asset_finder.symbol_ownership_maps_by_country_code.keys()
```
ssanderson: No need for `.keys()` here.

zipline/testing/fixtures.py
```diff
@@ -837,6 +841,11 @@ def init_class_fixtures(cls):
 
         cls.equity_daily_bar_days = days
 
+        if cls.EQUITY_DAILY_BAR_COUNTRY_CODES is None:
```
ssanderson: Thoughts on making this a classproperty rather than having the default logic live here?

zipline/testing/fixtures.py
```diff
@@ -1001,15 +1011,31 @@ def make_bcolz_daily_bar_rootdir_path(cls):
     def init_class_fixtures(cls):
         super(WithBcolzEquityDailyBarReader, cls).init_class_fixtures()
 
+        if cls.BCOLZ_DAILY_BAR_COUNTRY_CODE is None:
```
ssanderson: Same Q here re: using a classproperty for this.

zipline/testing/fixtures.py
```diff
@@ -1192,10 +1218,21 @@ def init_class_fixtures(cls):
...
-            cls.HDF5_DAILY_BAR_COUNTRY_CODE,
-            cls.make_equity_daily_bar_data(),
-        )
+        writer = HDF5DailyBarWriter(p, date_chunk_size=30)
+
+        for country_code in cls.HDF5_DAILY_BAR_COUNTRY_CODES:
+            sids = [
+                sid for sid in cls.asset_finder.equities_sids
+                if (
+                    cls.asset_finder.retrieve_asset(sid).country_code
```
ssanderson: What do you think about moving this method into the asset finder as something like `equity_sids_for_country_code`. Even if we keep the current impl for now, in the future we can probably make this faster by going straight to the sqlite db.

ssanderson: (we actually might be able to just piggyback on the country-parametric lifetimes machinery for this).

zipline/testing/fixtures.py
ssanderson: I'd probably call this `BCOLZ_EQUITY_DAILY_BAR_COUNTRY_CODE` to distinguish from the futures case.

yankees714: The `BCOLZ_DAILY_BAR_` prefix keeps things consistent with the existing attributes on this class, do you think it's worth changing all of them?

zipline/testing/fixtures.py
yankees714: Great idea 👍

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,758 @@
...
+    date_chunk_size : int
+        The number of days per chunk in the HDF5 file. If this is
+        greater than the number of days in the data, the chunksize will
+        match the actual number of days.
+
+    See Also
+    --------
+    zipline.data.hdf5_daily_bars.HDF5DailyBarReader
+    """
+    def __init__(self, filename, date_chunk_size):
```
ssanderson: One thing we might want to consider here is having this take a fully-formed object rather than just a file path. Doing so would allow us to do things like write to an in-memory h5 file, which we can't express in the current design. We could still have this take a filepath and use `coerce_types` (or just code in the constructor) to convert to the relevant object (this is how AssetFinder works, for example). Is there a reason that we want/need this to defer creation until a `write` call? One issue might be that we'd probably want to make this a context manager to ensure that the file gets closed, which would be a fairly different API from the rest of our writers.

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,758 @@
...
+        end = end_date.asm8
+
+        sid_selector = self.sids.searchsorted(assets)
+
+        date_slice = self._compute_date_range_slice(start, end)
+        nrows = date_slice.stop - date_slice.start
+        out = []
+        for column in columns:
+            dataset = self._country_group[DATA][column]
+            ncols = dataset.shape[0]
```
ssanderson: I think `ncols` is also just `len(self.sids)`, right? That would let us pull this shape construction out of the loop, which means we could allocate `buf` once at the top of the loop and re-use it by just zeroing it at the start of each loop iteration. Avoiding that large allocation in each loop iteration might be a meaningful win here.

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,758 @@
...
+            shape = (ncols, nrows)
+            buf = np.full(shape, 0, dtype=np.uint32)
```
ssanderson: Minor optimization note: we should use `np.zeros` instead of `np.full` here. Most modern operating systems have fast paths for returning zeroed pages, so I'd expect to get a small win here.

zipline/data/hdf5_daily_bars.py
ssanderson: My instinct is probably to leave this as-is for now given the worry about inconsistency with other writers. I could be convinced otherwise though.

zipline/assets/assets.py
```diff
@@ -1521,6 +1521,10 @@ def lifetimes(self, dates, include_start_date, country_codes):
 
         return pd.DataFrame(mask, index=dates, columns=lifetimes.sid)
 
+    def equity_sids_for_country_code(self, country_code):
+        sids = self._compute_asset_lifetimes([country_code]).sid
+        return tuple(sids.tolist())
```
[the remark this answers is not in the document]
yankees714: This is to get a tuple of Python `int` types, without that we get a tuple of `int64`. Maybe it doesn't matter?

I.e.
```
(Pdb++) type(tuple(sids.tolist())[0])
<class 'int'>
```
vs.
```
(Pdb++) type(tuple(sids)[0])
<class 'numpy.int64'>
```

zipline/testing/fixtures.py
```diff
@@ -1128,6 +1170,86 @@ def _trading_days_for_minute_bars(calendar,
...
+    def make_hdf5_daily_bar_path(cls):
+        return cls.tmpdir.getpath(cls.HDF5_DAILY_BAR_PATH)
+
+    @classmethod
+    def init_class_fixtures(cls):
+        super(WithHDF5EquityDailyBarReader, cls).init_class_fixtures()
+
+        cls.hdf5_daily_bar_path = p = cls.make_hdf5_daily_bar_path()
+
+        writer = HDF5DailyBarWriter(p, date_chunk_size=30)
```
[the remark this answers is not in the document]
yankees714: Nope, good idea

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,758 @@
...
+
+    Raises
+    ------
+    ValueError
+        If the dataframes passed are not all indexed by the same days
+        and sids.
+    """
+    frame_0 = frames[0]
+
+    # Ensure the indices and columns all match.
```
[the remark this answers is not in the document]
yankees714: Wow, TIL

zipline/assets/assets.py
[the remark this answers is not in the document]
ssanderson: Gotcha. Is there a reason we specifically want a tuple here? (I'm not sure it matters much either way, this just struck me as surprising).

zipline/data/hdf5_daily_bars.py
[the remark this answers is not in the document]
ssanderson: I only  added this like two weeks ago.

zipline/data/hdf5_daily_bars.py
yankees714: Yeah this is just the number of sids. Good idea, I'll update this

zipline/assets/assets.py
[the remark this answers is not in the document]
yankees714: Just was trying to make this behave the same way as the `sids`, `equities_sids`, etc. properties.

zipline/data/hdf5_daily_bars.py
llllllllll: I actually came to this PR to make this same request because of how I want to use this in our writer service. If you want to make a reader over the recently written object, there is currently a risk of opening two handles to the same h5 file which is not totally safe.

zipline/data/hdf5_daily_bars.py
```diff
@@ -0,0 +1,762 @@
...
+        end_date: Timestamp
+           End of the window range.
+        assets : list of int
+           The asset identifiers in the window.
+
+        Returns
+        -------
+        list of np.ndarray
+            A list with an entry per field of ndarrays with shape
+            (minutes in range, sids) with a dtype of float64, containing the
```
llllllllll: this should be `days`, not minutes`

PR #2298, opened 2018-09-21
Vectorize dividend ratio calc

llllllllll:


zipline/testing/core.py
```diff
@@ -981,8 +981,19 @@ def wrapped(*args, **kwargs):
 
 
 class MockDailyBarReader(object):
+    def __init__(self, dates):
+        self.sessions = pd.DatetimeIndex(dates)
+
+    def load_raw_arrays(self, columns, start, stop, sids):
+        dates = self.sessions
+        output_dates = dates[(dates >= start) & (dates <= stop)]
```
ssanderson: Should this barf if `start` or `stop` is outside the duration of `self.sessions`?

zipline/data/adjustments.py
```diff
@@ -349,72 +350,62 @@ def calc_dividend_ratios(self, dividends):
...
+        dates = pricing_reader.sessions.values
+
+        close, = pricing_reader.load_raw_arrays(
+            ['close'],
+            pd.Timestamp(dates[0], tz='UTC'),
+            pd.Timestamp(dates[-1], tz='UTC'),
+            unique_sids,
+        )
+        date_ix = np.searchsorted(dates, dividends.ex_date.values)
+        mask = date_ix > 0
```
ssanderson: I don't think this can ever be False.

```python
In [6]: data = np.arange(5)

In [7]: data.searchsorted([-1, 3, 6])
Out[7]: array([0, 3, 5])
```

I think ^ also means we need to handle the case of dates that are outside the range or misaligned with the input index. The way I've written that lookup before is:

```python
def loc_as_of(dts, asof):
    """
    Find the position of the last entry in ``dts`` on or before ``as_of``.

    Parameters
    ----------
    dts : ndarray or DatetimeIndex
        Array of sorted datetime-like objects.
    as_of : pd.Timestamp
        Timestamp to find the position as of.
    """
    ix = dts.searchsorted(asof.asm8, side='right') - 1
    if ix < 0:
        raise IndexError("{} is before first value {}", asof, dts[0])
    return ix
```

I think this algorithm vectorizes straightforwardly, though I'm not sure the way it handles boundaries is what we want. If we get an `ex_date` in the future or too far in the past, what do we want to do? My guess is we probably just shouldn't compute a ratio for them.

zipline/data/adjustments.py
```diff
@@ -349,72 +350,62 @@ def calc_dividend_ratios(self, dividends):
...
+
+        date_ix = date_ix[mask]
+        sids_ix = sids_ix[mask]
+        input_dates = dividends.ex_date.values[mask]
+
+        # subtract one day to get the close on the day prior to the merger
+        previous_close = close[date_ix - 1, sids_ix]
```
ssanderson: I think this has the potential to wrap to -1?

zipline/data/adjustments.py
```diff
@@ -349,72 +350,62 @@ def calc_dividend_ratios(self, dividends):
...
+            log.warn(
+                "Couldn't compute ratio for dividend "
+                " sid={sid}, ex_date={ex_date!s}, amount={amount}",
+                sid=input_sids[ix],
+                ex_date=dates[date_ix[ix]],
+                amount=amount[ix],
+            )
+
+        positive_ratio_mask = ratio > 0
+        for ix in np.flatnonzero(~positive_ratio_mask):
```
ssanderson: Since `NaN > 0` is False, I think this is going to double-warn for NaN ratios.

zipline/data/adjustments.py
```diff
@@ -349,72 +350,62 @@ def calc_dividend_ratios(self, dividends):
...
+
+        # subtract one day to get the close on the day prior to the merger
+        previous_close = close[date_ix - 1, sids_ix]
+        input_sids = input_sids[mask]
+
+        amount = dividends.amount.values[mask]
+        ratio = 1.0 - amount / previous_close
+
+        non_nan_ratio_mask = ~np.isnan(ratio)
+        for ix in np.flatnonzero(~non_nan_ratio_mask):
```
ssanderson: Do we have test coverage for these warning cases? I think we could set that up relatively easily using an in-memory bar reader.

zipline/data/adjustments.py
llllllllll: I am not sure what your first example is showing, the first element of that would be false:
```
In [9]: data = np.arange(5)

In [10]: ix = data.searchsorted([-1, 3, 6])

In [11]: ix > 0
Out[11]: array([False,  True,  True], dtype=bool)
```

zipline/data/adjustments.py
llllllllll: that is why we masked to only values that are > 0

zipline/data/adjustments.py
ssanderson: Ah, yeah, sorry, I read it as `>=`.

PR #2301, opened 2018-09-25
ENH: add DE, CH, JP domains and use ISO exchange names

vikram-narayan:
Need to bump the version of trading_calendars before this is merged.


tests/pipeline/test_domain.py
```diff
@@ -154,7 +175,9 @@ def do_checks(cls, colnames):
         do_checks(MyData, ['col1', 'col2', 'col3'])
         do_checks(MyDataSubclass, ['col1', 'col2', 'col3', 'col4'])
 
-    @parameter_space(domain_param=[US_EQUITIES, CA_EQUITIES])
+    @parameter_space(domain_param=[BE_EQUITIES, CA_EQUITIES, CH_EQUITIES,
```
ssanderson: Any reason not to just use `DOMAINS` here?

tests/pipeline/test_domain.py
```diff
@@ -76,7 +97,7 @@ def run(ts):
 
 class SpecializeTestCase(zf.ZiplineTestCase):
 
-    @parameter_space(domain=[US_EQUITIES, CA_EQUITIES, GB_EQUITIES])
+    @parameter_space(domain=DOMAINS)
```
ssanderson: It's probably overkill to run these against every single domain we have. The purpose of these tests is to verify that `specialize` works with an arbitrary domain, but I don't think it's likely that this will pass on three domains but fail on a fourth. That said, these tests are super fast, so I'm not sure it matters.

tests/pipeline/test_domain.py
```diff
@@ -26,6 +33,20 @@
 from zipline.utils.pandas_utils import days_at_time
 
 
+DOMAINS = [
+    BE_EQUITIES,
```
ssanderson: Probably the most important thing for us to test for each new domain is that its calendar code resolves as expected. The test for the EquityCalendarDomain defaults at https://github.com/quantopian/zipline/pull/2301/files#diff-8dd6ea1d2db90f77c7c793ecc4704888R339 do a good job of exercising that functionality, but right now that requires someone adding a new domain to remember to add it to the list in that test. One way to make it more likely that that happens would be to move this list to `domain.py` and rename it to something like `BUILT_IN_DOMAINS`. Then in this test, we could define the expected cutoff time for each domain and assert we have a definition for each builtin, so that if a user adds a domain and doesn't update the test, we'd fail.

tests/pipeline/test_domain.py
```diff
@@ -314,15 +338,22 @@ def _test_equity_calendar_domain(self,
 
     def test_equity_calendar_domain(self):
         # test the defaults
-        self._test_equity_calendar_domain(US_EQUITIES, datetime.time(8, 45))
+        self._test_equity_calendar_domain(BE_EQUITIES, datetime.time(8, 15))
```
ssanderson: It might be worth moving this block into its own test named something like `test_built_in_equity_calendar_domain_defaults`. It's useful to have coverage of the default definition for all the built-in domains, but the original purpose of this test was to be a test for the `EquityCalendarDomain` class itself, not for each built-in instance, so this is kind of testing two different things at this point.

zipline/__main__.py
```diff
@@ -185,8 +185,8 @@ def _(*args, **kwargs):
 @click.option(
     '--trading-calendar',
     metavar='TRADING-CALENDAR',
-    default='NYSE',
-    help="The calendar you want to use e.g. LSE. NYSE is the default."
+    default='XNYS',
```
ssanderson: I'd probably leave this be for now. This is part of the public-facing API of the CLI, and more users are going to know NYSE than XNYS.

zipline/algorithm.py
```diff
@@ -2384,7 +2384,14 @@ def all_api_methods(cls):
 
 # Map from calendar name to default domain for that calendar.
 _DEFAULT_DOMAINS = {
-    'NYSE': domain.US_EQUITIES,
-    'TSX': domain.CA_EQUITIES,
-    'LSE': domain.GB_EQUITIES,
+    'XBRU': domain.BE_EQUITIES,
```
ssanderson: If we have a list of built-in domains defined in the domain module, then this can probably turn into something like:

```python
_DEFAULT_DOMAINS = {d.calendar_name: d for d in domain.BUILT_IN_DOMAINS}
```
(We should make `calendar_name` public if we do ^).

tests/pipeline/test_domain.py
vikram-narayan: I didn't use `DOMAINS` because it includes `GB_EQUITIES`, and the test uses GB_EQUITIES as a domain meant to be different from `domain_param`

ssanderson: Ah. Clearer here might be something like `BUILT_IN_DOMAINS - {GB_EQUITIES}` ? Or honestly we could probably just keep this to two or three examples. I think there are pretty hard diminishing returns on testing every single built-in domain for this.

Another option would be to use `EquitySessionDomain`s as test inputs here, since that would guarantee that we're not somehow depending on special behavior for the known built-in domains.

tests/pipeline/test_domain.py
```diff
@@ -312,17 +323,30 @@ def _test_equity_calendar_domain(self,
...
+            CA_EQUITIES: datetime.time(8, 45),
+            CH_EQUITIES: datetime.time(8, 15),
+            DE_EQUITIES: datetime.time(8, 15),
+            FR_EQUITIES: datetime.time(8, 15),
+            GB_EQUITIES: datetime.time(7, 15),
+            JP_EQUITIES: datetime.time(8, 15),
+            NL_EQUITIES: datetime.time(8, 15),
+            PT_EQUITIES: datetime.time(7, 15),
+            US_EQUITIES: datetime.time(8, 45),
+        }
```
ssanderson: Can we add an assertion here that the set of keys in this dict matches the set of entries in BUILT_IN_DOMAINS. I think the main thing we want from this test is to make sure that all the built-in domain definitions can correctly resolve their calendars names, so I want to make sure we're covering all of them.

vikram-narayan: updated tests

vikram-narayan:
squashed commits and rebased against master, will merge once travis passes

PR #2303, opened 2018-09-26
BUG: Fix bug calculating end offsets for hdf5 daily bars.

ssanderson:
I also highlighted a couple places where it looks like we probably need some test coverage.

### DIFF
--- a/tests/data/test_daily_bars.py
+++ b/tests/data/test_daily_bars.py
@@ -47,7 +47,7 @@ from zipline.pipeline.loaders.synthetic import (
     OHLCV,
     asset_start,
     asset_end,
-    expected_bar_value,
+    expected_bar_value_with_holes,
     expected_bar_values_2d,
     make_bar_data,
 )
@@ -383,17 +383,16 @@ class BcolzDailyBarTestCase(WithBcolzEquityDailyBarReader, _DailyBarsTestCase):
             multiplier = 1 if column == 'volume' else 1000
             for asset_id in self.assets:
                 for date in self.dates_for_asset(asset_id):
-                    if asset_id in HOLES and date in HOLES[asset_id]:
-                        expected = 0
-                    else:
-                        expected = expected_bar_value(
-                            asset_id,
-                            date,
-                            column,
-                        ) * multiplier
-
-                    self.assertEqual(data[idx], expected)
-
+                    self.assertEqual(
+                        data[idx],
+                        expected_bar_value_with_holes(
+                            asset_id=asset_id,
+                            date=date,
+                            colname=column,
+                            holes=HOLES,
+                            missing_value=0,
+                        ) * multiplier,
+                    )
                     idx += 1
             self.assertEqual(idx, len(data))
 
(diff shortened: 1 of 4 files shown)


tests/data/test_daily_bars.py
```diff
@@ -119,6 +119,8 @@
 
 TEST_QUERY_ASSETS = EQUITY_INFO.index
 
+HOLES = {3: (Timestamp('2015-06-17', tz='UTC'),)}
```
ssanderson: Can we add a note on the description for this equity saying that it has a hole on this day?)

zipline/data/hdf5_daily_bars.py
```diff
@@ -328,8 +328,12 @@ def compute_asset_lifetimes(frames):
         [frames[field].isnull().values for field in FIELDS],
     )
 
+    # Offset of the first null from the start of the input.
     start_date_ixs = is_null_matrix.argmin(axis=0)
-    end_date_ixs = is_null_matrix[::-1].argmin(axis=0)
+    # Offset of the *last null from the **end** of the input.
```
ssanderson: Extra * hanging out here.

zipline/pipeline/loaders/synthetic.py
```diff
@@ -250,6 +251,9 @@ def make_bar_data(asset_info, calendar):
         DataFrame with asset_id as index and 'start_date'/'end_date' columns.
     calendar : pd.DatetimeIndex
         The trading calendar to use.
+    holes : dict[int -> tuple[pd.Timestamps]]
```
ssanderson: We should add a note here that this is optional and that the default is no holes.

zipline/pipeline/loaders/synthetic.py
```diff
@@ -346,6 +355,12 @@ def expected_bar_values_2d(dates, asset_info, colname):
             # date.
             if not (start <= date <= end):
                 continue
+            if (
```
ssanderson: I tend to try to avoid in-line complex conditionals in if-statements if I can.

This might be clearer if we added an in-line `is_hole` function at the start of this function like:

```python
def is_hole(date, asset):
    return holes is not None and date in holes.get(asset, ())
```
or maybe better yet:

```python
if holes is not None:
    def is_hole(date, asset):
        return date in holes.get(asset, ())
else:
    def is_hole(date, asset):
        return False
```

ssanderson: Second thought on reading this: would it make sense to push this logic down to `expected_bar_value`? That would simplify the logic in https://github.com/quantopian/zipline/pull/2303/files#diff-c2524ef7178e306daad5af35e6180d3bR388.

tests/data/test_daily_bars.py
```diff
@@ -363,14 +382,17 @@ def test_write_ohlcv_content(self):
             multiplier = 1 if column == 'volume' else 1000
             for asset_id in self.assets:
                 for date in self.dates_for_asset(asset_id):
-                    self.assertEqual(
-                        expected_bar_value(
+                    if asset_id in HOLES and date in HOLES[asset_id]:
+                        expected = 0
+                    else:
+                        expected = expected_bar_value(
```
ssanderson: I noted this below as well: this might be clearer if we passed HOLES all the way down to `expected_bar_value`, since that's what the 2d version is calling anyway.

ssanderson:
LGTM

PR #2305, opened 2018-09-26
Add country_code parameter to lookup_generic.

ssanderson:


zipline/sources/test_source.py
```diff
@@ -113,65 +109,23 @@ class SpecificEquityTrades(object):
...
+                 trading_calendar,
+                 asset_finder,
+                 sids,
+                 start,
+                 end,
+                 delta,
+                 count=500):
 
         self.trading_calendar = trading_calendar
-        asset_finder = kwargs.pop('asset_finder')
```
ssanderson: None of this code was used as far as I can tell.

tests/test_assets.py
```diff
@@ -931,28 +993,6 @@ def test_security_dates_warning(self):
                 self.assertTrue(issubclass(warning.category,
                                            DeprecationWarning))
 
-    def test_map_identifier_index_to_sids(self):
```
ssanderson: I deleted the function this tested.

zipline/assets/assets.py
```diff
@@ -1297,133 +1279,106 @@ def _lookup_generic_scalar(self,
...
             )
 
         for obj in iterator:
-            if isinstance(obj, ContinuousFuture):
-                matches.append(obj)
-            else:
-                self._lookup_generic_scalar(obj, as_of_date, matches, missing)
-        return matches, missing
-
-    def map_identifier_index_to_sids(self, index, as_of_date):
```
ssanderson:
```
-*- mode: grep; default-directory: "~/zipline_repo/" -*-
Grep started at Wed Sep 26 14:42:44

git --no-pager grep -n -e map_identifier_index -- 

Grep finished with no matches found at Wed Sep 26 14:42:44
```

tests/test_assets.py
```diff
@@ -176,65 +209,89 @@ def build_lookup_generic_cases(asset_finder_type):
...
-            # ContinuousFuture and Equity
-            (finder, [cf, 0], None, [cf, assets[0]]),
+        all_assets = list(equities) + [fof14, cf]
+
+        dupe_old_start = dupe_old.start_date
+        dupe_new_start = dupe_new.start_date
+
+        for asset in list(equities) + [fof14, cf]:
+            # Looking up an asset object directly should yield itself.
+            yield case(asset, None, None, asset, )
```
llllllllll: weird trailing `, `

tests/test_assets.py
```diff
@@ -176,65 +209,89 @@ def build_lookup_generic_cases(asset_finder_type):
...
+            # Looking up an asset by sid should yield the asset.
+            yield case(asset.sid, None, None, asset)
+
+        # Duplicated US equity symbol with resolution date.
+        for country in ('US', None):
+            yield case('DUPLICATED_IN_US', dupe_old_start, country, dupe_old)
```
llllllllll: should we check a sample of days?

tests/test_assets.py
```diff
@@ -761,12 +818,16 @@ def test_lookup_generic(self):
...
-        with build_lookup_generic_cases(self.asset_finder_type) as cases:
-            for finder, symbols, reference_date, expected in cases:
-                results, missing = finder.lookup_generic(symbols,
-                                                         reference_date)
-                self.assertEqual(results, expected)
-                self.assertEqual(missing, [])
+        cases = build_lookup_generic_cases()
+        # Make sure we clean up temp resources in the generator if we don't
+        # consume the whole thing because of a failure.
+        self.add_instance_callback(cases.close)
```
llllllllll: weird, but makes sense

tests/test_assets.py
```diff
@@ -761,12 +818,16 @@ def test_lookup_generic(self):
...
+        for finder, inputs, reference_date, country, expected in cases:
+            results, missing = finder.lookup_generic(
+                inputs, reference_date, country,
+            )
+            self.assertEqual(results, expected)
+            self.assertEqual(missing, [])
```
llllllllll: should we check some cases that we expect to fail?

ssanderson: There's a separate test case for `test_lookup_generic_handle_missing` that exercises those code paths.

zipline/assets/assets.py
```diff
@@ -1268,27 +1267,10 @@ def _(self):
...
-            # lookup_future_symbol method does not use as_of date, since
-            # symbols are unique.
-            #
-            # Wrap the function in a lambda so that both methods share a
-            # signature, so that when the functions are iterated over
-            # the consumer can use the same arguments with both methods.
-            lambda symbol, _: self.lookup_future_symbol(symbol)
-        )
-
     def _lookup_generic_scalar(self,
```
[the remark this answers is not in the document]
ssanderson: I updated this to use explicit early returns. I'm not sure I think it's clearer, but I don't think it's any less clear either. The reason I structured this as an `if/elif` was to try to make it clear that the various cases were mutually exclusive (morally, this function really wants pattern matching).

zipline/assets/assets.py
```diff
@@ -1297,133 +1279,110 @@ def _lookup_generic_scalar(self,
...
-                result = self.retrieve_asset(int(asset_convertible))
+                result = self.retrieve_asset(int(obj))
+                matches.append(result)
             except SidsNotFound:
-                missing.append(asset_convertible)
-                return None
-            matches.append(result)
+                missing.append(obj)
+            finally:
+                return
```
llllllllll: this masks any exceptions in that block

tests/test_assets.py
ssanderson: I updated this to check the boundary condition of `old_start, new_start - minute, new_start, new_start + minute`. We already have very good coverage for `lookup_symbol`, which is ultimately what this function calls into, so I don't think it's important to check a sample of the interior here.

llllllllll: makes sense

tests/test_assets.py
llllllllll: found that, thanks

PR #2308, opened 2018-09-28
Small improvements for HDF5 daily bars and tests

yankees714:
- Raise an exception on requests for assets and/or dates not available in the daily bars
- Generalize remaining US-specific tests in _DailyBarsTestCase

### DIFF
--- a/zipline/data/hdf5_daily_bars.py
+++ b/zipline/data/hdf5_daily_bars.py
@@ -435,6 +435,7 @@ class HDF5DailyBarReader(SessionBarReader):
             (minutes in range, sids) with a dtype of float64, containing the
             values for the respective field over start and end dt range.
         """
+        self._validate_assets(assets)
         self._validate_timestamp(start_date)
         self._validate_timestamp(end_date)
 
@@ -477,6 +478,23 @@ class HDF5DailyBarReader(SessionBarReader):
         end_ix = self.dates.searchsorted(end_date, side='right')
 
         return slice(start_ix, end_ix)
+
+    def _validate_assets(self, assets):
+        """Validate that asset identifiers are contained in the daily bars.
+
+        Parameters
+        ----------
+        assets : array-like[int]
+           The asset identifiers to validate.
+
+        Raises
+        ------
+        NoDataForSid
+            If one or more of the provided asset identifiers are not
+            contained in the daily bars.
+        """
+        if not np.in1d(assets, self.sids).all():
+            raise NoDataForSid()
 
     def _validate_timestamp(self, ts):
         if ts.asm8 not in self.dates:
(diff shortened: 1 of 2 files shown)


zipline/data/hdf5_daily_bars.py
```diff
@@ -430,6 +435,10 @@ def load_raw_arrays(self,
             (minutes in range, sids) with a dtype of float64, containing the
             values for the respective field over start and end dt range.
         """
+        self._validate_assets(assets)
```
llllllllll: for `load_raw_arrays`, I could imagine just returning `nan` for the assets that don't exist. For code that wants to use this safely, they need to filter down the the known assets, which is hard to do because it is not part of the interface of a session bar reader. An example of a place where I ran into this was computing dividend ratios. When doing that, I needed to manually exclude any assets that we don't have pricing for, but if we just return all nan columns I can continue to vectorize the calculation and then strip nans in the final result.

cc @ssanderson what do you think about this?

zipline/data/hdf5_daily_bars.py
```diff
@@ -623,13 +656,16 @@ def __init__(self, readers):
...
-
-        if country_codes.isnull().any():
-            raise ValueError(
-                'Assets not contained in daily pricing file: {}'.format(
-                    list(country_codes[country_codes.isnull()].index)
-                )
+        try:
+            country_codes = self._country_map.loc[assets]
+        except KeyError as exc:
+            raise_from(
```
llllllllll: nice use of raise from

zipline/data/hdf5_daily_bars.py
```diff
@@ -470,6 +479,27 @@ def _compute_date_range_slice(self, start_date, end_date):
...
+           The asset identifiers to validate.
+
+        Raises
+        ------
+        NoDataForSid
+            If one or more of the provided asset identifiers are not
+            contained in the daily bars.
+        """
+        if not np.in1d(assets, self.sids).all():
+            raise NoDataForSid()
```
llllllllll: we could use `np.setdiff1d` instead of `in1d` and then report the sids that are missing. That might be useful for the vectorized case if only one or two assets are missing in a large array.

yankees714: Good idea, will update

zipline/data/hdf5_daily_bars.py
ssanderson: Looking at what the other BarReaders do here:

- InMemoryDailyBarReader raises a NoDataForSid on unexpected sids.
- BcolzDailyBarReader doesn't handle this case at all (looks like it would raise a KeyError from initial inspection).
- BcolzMinuteBarReader raises NoDataForSid
- The other readers all delegate to a sub-reader.

Given the current behavior, it seems like raising an error is the behavior that's most consistent with the rest of the system right now. I buy that in some contexts the nan-returning behavior might be preferable, but I don't think we actually have any use-cases where that's **required** right now, so the current erroring behavior also feels a bit safer to me (returning nans is useful if the caller knowingly gave us unknown sids, but it could also mask bugs where the caller expected us to have the requested sids). I also think that we should probably make `sids` part of the public interface for at least some of our readers, since that's a requirement for many natural composition patterns. Having readers expose their sids might lessen the value of the nan-returning behavior.

llllllllll:
LGTM

PR #2309, opened 2018-09-28
BUG: Round values before truncating when writing bcolz minute/day tables.

ernestoeperez88:
Currently, pricing values are scaled by `OHLC_RATIO` (1000) and truncated, before being written to bcolz minute/daily bar tables as `uint32` values.

For floats that don't have an exact representation, this means we could loose some precision. Here is an example using 130.23:
```
In [2]: arr = numpy.array([130.23], dtype='f8')
In [3]: arr[0]
Out[3]: 130.22999999999999
In [5]: arr[0] * 1000
Out[5]: 130229.99999999999
In [6]: (arr[0] * 1000).astype('uint32')
Out[6]: 130229
```
When reading back the value from the bcolz table, we will get 130.229.

This can be solved by rounding values after scaling, before truncating to `uint32`.

### DIFF
--- a/tests/data/test_minute_bars.py
+++ b/tests/data/test_minute_bars.py
@@ -142,6 +142,38 @@ class BcolzMinuteBarTestCase(WithTradingCalendars,
 
         self.assertEquals(50.0, volume_price)
 
+    def test_precision_after_scaling(self):
+        '''For numbers that don't have an exact float representation,
+        assert that scaling the value does not cause a loss in precision.
+        '''
+        minute = self.market_opens[self.test_calendar_start]
+        sid = 1
+        data = DataFrame(
+            data={
+                'open': [130.23],
+                'high': [130.23],
+                'low': [130.23],
+                'close': [130.23],
+                'volume': [1000]
+            },
+            index=[minute])
+        self.writer.write_sid(sid, data)
+
+        open_price = self.reader.get_value(sid, minute, 'open')
+        self.assertEquals(130.23, open_price)
+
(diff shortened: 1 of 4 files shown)


tests/data/test_minute_bars.py
```diff
@@ -142,6 +142,43 @@ def test_write_one_ohlcv(self):
...
+                'high': [130.23],
+                'low': [130.23],
+                'close': [130.23],
+                'volume': [1000]
+            },
+            index=[minute])
+        self.writer.write_sid(sid, data)
+
+        open_price = self.reader.get_value(sid, minute, 'open')
+
```
llllllllll: It might be easier to read without these newlines between the read and the assert because then they will fall into little groupings.

tests/test_data_portal.py
```diff
@@ -406,7 +406,7 @@ def test_get_adjustments(self, data_frequency, field):
         day = calendar.day
         dividend_date = self.trading_days[2]
 
-        prev_day_price = 1.005
+        prev_day_price = 1.006
```
llllllllll: is this also not perfectly representable?

ernestoeperez88: This value is fine, but we are trying to write 1.0055, which ends up being 1006 after scaling and rounding.

imkoukou: For 1.0055 to 1.006, cannot change the OHLC_RATIO ？I changed it to 100000 for 5 decimal FX csv data.

PR #2312, opened 2018-09-29
BUG: Drop orphaned records from asset db in v6 downgrade.

ssanderson:
Remove records that no longer refer to equities from equity_symbol_mappings and
asset_router.


ssanderson:
> the PR looks good but my mouse died so clicking accept is currently a challenge

PR #2315, opened 2018-10-02
BUG: Use calendar to disambiguate symbol lookups in fetcher.

ssanderson:


zipline/algorithm.py
```diff
@@ -2384,3 +2398,9 @@ def all_api_methods(cls):
 
 # Map from calendar name to default domain for that calendar.
 _DEFAULT_DOMAINS = {d.calendar_name: d for d in domain.BUILT_IN_DOMAINS}
+# Map from calendar name to default country code for that calendar.
+_DEFAULT_FETCH_CSV_COUNTRY_CODES = {
```
dmichalowicz: Maybe this could just be called `_DEFAULT_CALENDAR_COUNTRY_CODES` or somesuch, since it isn't necessarily tied to `fetch_csv`.

zipline/algorithm.py
```diff
@@ -2367,6 +2375,12 @@ def default_pipeline_domain(calendar):
         """
         return _DEFAULT_DOMAINS.get(calendar.name, domain.GENERIC)
 
+    @staticmethod
+    def default_fetch_csv_country_code(calendar):
+        """
+        """
+        return _DEFAULT_FETCH_CSV_COUNTRY_CODES.get(calendar.name)
```
dmichalowicz: Why does this need to be a `get` instead of normal key access? Do we ever expect this to fail, and if so what happens when it returns `None`?

tests/test_fetcher.py
```diff
@@ -61,48 +61,55 @@ def make_equity_info(cls):
...
                     'end_date': pd.Timestamp('2007-01-01', tz='UTC'),
                     'symbol': 'DELL',
-                    'asset_type': 'equity',
                     'exchange': 'nasdaq'
                 },
                 13: {
                     'start_date': pd.Timestamp('2006-01-01', tz='UTC'),
                     'end_date': pd.Timestamp('2010-01-01', tz='UTC'),
                     'symbol': 'NFLX',
-                    'asset_type': 'equity',
```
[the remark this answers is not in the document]
ssanderson: These were legacy cruft from an old asset finder version. (We used to just provide a single assets frame with a type column). Nothing looked at this field.

zipline/algorithm.py
ssanderson: I'm not sure that the choice I'm making here would be the right choices for any other API that needed a default country code. I think it **probably** would be, but since nothing else is using this I don't want to make this look more general than it is until we're confident it can be used generically.

dmichalowicz: Ok makes sense.

zipline/algorithm.py
dmichalowicz: Thanks for the explanation, I think what you have is good then.

PR #2316, opened 2018-10-02
checkpoint issue

llllllllll:
When resolving data points to an output row index, we do not include records with a timestamp of _exactly_ the data query cutoff time:

https://github.com/quantopian/zipline/blob/5cda2ad542cb1eab02a84a6685dfb3d2adabd44e/zipline/pipeline/loaders/blaze/_core.pyx#L566-L572

however, when deciding which checkpoint to retrieve we would select anything less than **or equal to** the query start timestamp, after it had been adjusted to the data query cutoff time:

https://github.com/quantopian/zipline/blob/5cda2ad542cb1eab02a84a6685dfb3d2adabd44e/zipline/pipeline/loaders/blaze/core.py#L1043-L1047

If we had a checkpoint at exactly the data query time, we would have selected it meaning all of the data would have a timestamp of exactly the data query time. We only select base rows between the checkpoint timestamp and the upper dt, so we would not have any base rows between the checkpoint and the first output index. When we go to slot these checkpoint rows into the output rows, we would shift them all forward one day leaving us with a leading missing row instead of the checkpoint data.

This change makes us select the most recent checkpoint strictly less than the `lower_dt` which correctly gives us the checkpoint data on the first row of the output when the query begins on a checkpoint boundary.

### DIFF
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -2179,19 +2179,31 @@ class BlazeToPipelineTestCase(WithAssetFinder, ZiplineTestCase):
         # provide two checkpoints, one before the data in the base table
         # and one after, these should not affect the value on the third
         asof_dates = pd.to_datetime(['2013-12-30', '2014-01-04'])
-        checkpoints = pd.DataFrame({
+        out_of_bounds = pd.DataFrame({
             'value': [-2, 2],
             'asof_date': asof_dates,
             'timestamp': asof_dates + pd.Timedelta(hours=23),
         })
 
-        self._test_checkpoints_macro(checkpoints)
+        # Add a single checkpoint on the query day with a timestamp of exactly
+        # the data query time. This should not get pulled to overwrite the
+        # expected data on the 3rd.
+        exact_query_time = pd.DataFrame({
+            'value': [1],
+            'asof_date': [pd.Timestamp('2014-01-02')],
+            'timestamp': [pd.Timestamp('2014-01-03')],
+        })
+
+        self._test_checkpoints_macro(
+            pd.concat([out_of_bounds, exact_query_time]),
+        )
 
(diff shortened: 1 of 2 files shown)


ssanderson:
Repeating the issue in my own words to confirm my understanding:

1. If we get a record whose timestamp is exactly the data query cutoff for day N, we slot it into day N + 1.
2. When we query for checkpoints, we currently (incorrectly) fetch data for the last checkpoint timestamp <= to the data query time of the first row of our window.
3. If the checkpoint time falls **exactly on** the data query cutoff of day 0 in our window, then we'll roll that data forward to day 1, leaving us with empty data for day 0.

Is that right?

llllllllll:
Yes, that is correct.

tests/pipeline/test_blaze.py
```diff
@@ -2179,19 +2179,31 @@ def test_checkpoints_out_of_bounds_macro(self):
...
         })
 
-        self._test_checkpoints_macro(checkpoints)
+        # Add a single checkpoint on the query day with a timestamp of exactly
+        # the data query time. This should not get pulled to overwrite the
+        # expected data on the 3rd.
+        exact_query_time = pd.DataFrame({
+            'value': [1],
+            'asof_date': [pd.Timestamp('2014-01-02')],
+            'timestamp': [pd.Timestamp('2014-01-03')],
```
ssanderson: It's not clear to me how these values are connected to the description in the comment above. What's enforcing that this is the leading data query cutoff time of the query we're going to make? Are these just the values that `test_checkpoints_*` uses by default? If so, it might help clarify to lift this value to a global constant that's shared with `test_checkpoints_*`

llllllllll:
yeah, your reasoning there sound correct to me. I will merge on pass.

PR #2318, opened 2018-10-02
DOC: Volatility of 'returns' in slippage

freddiev4:


freddiev4:
I'm pretty confident that we can just merge this so I'm gonna go ahead and do that

PR #2323, opened 2018-10-04
BUG: don't crash on dividend warning

vikram-narayan:
Lookup dates for warning message in input_dates.

### DIFF
--- a/tests/data/test_adjustments.py
+++ b/tests/data/test_adjustments.py
@@ -83,9 +83,12 @@ class TestSQLiteAdjustementsWriter(WithTradingCalendars,
             first_date_ix:first_date_ix + 3
         ]
 
-        one_day_past_pricing_data = self.trading_calendar.all_sessions[
-            first_date_ix + 3
-        ].tz_convert(None)
+        before_pricing_data = \
+            (dates[0] - self.trading_calendar.day).tz_convert(None)
+        one_day_past_pricing_data = \
+            (dates[-1] + self.trading_calendar.day).tz_convert(None)
+        ten_days_past_pricing_data = \
+            (dates[-1] + self.trading_calendar.day * 10).tz_convert(None)
 
         def T(n):
             return dates[n].tz_convert(None)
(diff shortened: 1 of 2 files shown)


tests/data/test_adjustments.py
```diff
@@ -167,6 +175,10 @@ def T(n):
             "Couldn't compute ratio for dividend sid=2, ex_date=1990-10-18,"
             " amount=10.000",
         ))
+        self.assertTrue(self.log_handler.has_warning(
```
ssanderson: For paranoia's sake, can we also check that we handle the case where this is well beyond the end of the pricing data (e.g., 10 days or something)? Since we're fiddling with indices in the code under test here, I could imagine a world where this worked for dates one past the end of the array, but failed if we were well after the end.

ssanderson: Might also be worth checking that we handle an ex_date that's before our inputs.

tests/data/test_adjustments.py
```diff
@@ -83,13 +83,17 @@ def test_calculate_dividend_ratio(self):
...
+        ].tz_convert(None)
+
         def T(n):
             return dates[n].tz_convert(None)
 
         close = pd.DataFrame(
             [[10.0, 0.5,   30.0],   # noqa
              [ 9.5, 0.4, np.nan],   # noqa
-             [15.0, 0.6,   35.0]],  # noqa
+             [15.0, 0.6,   np.nan]],  # noqa
```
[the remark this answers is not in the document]
vikram-narayan: we want the dividend ratio on `one_day_past_pricing_data` to be nan to test the warning logic, and the previous close needs to be nan for that to happen

tests/data/test_adjustments.py
```diff
@@ -83,13 +83,17 @@ def test_calculate_dividend_ratio(self):
             first_date_ix:first_date_ix + 3
         ]
 
+        one_day_past_pricing_data = self.trading_calendar.all_sessions[
```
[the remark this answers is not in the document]
vikram-narayan: 👍 fixed

tests/data/test_adjustments.py
vikram-narayan: added these cases

PR #2332, opened 2018-10-17
MAINT: Removed tornado, backports.ssl-match-hostname, and certifi

cmichelQT:
@richafrank As per discussion earlier

These are no longer required by the version of matplotlib we're installing.


richafrank:
Looks good to me, @cmichelQT .

PR #2348, opened 2018-10-26
MAINT: Make load_raw_arrays for daily bar readers handle unknown sids

yankees714:
The readers would previously error, they'll now return `nan`/`0` for these sids.

### DIFF
--- a/zipline/data/_equities.pyx
+++ b/zipline/data/_equities.pyx
@@ -213,6 +213,10 @@ cpdef _read_bcolz_data(ctable_t table,
 
             for asset in range(nassets):
                 first_row = first_rows[asset]
+                if first_row == -1:
+                    # This is an unknown asset, leave its slot empty.
+                    continue
+
                 last_row = last_rows[asset]
                 offset = offsets[asset]
                 out_start = offset
--- a/zipline/data/_equities.pyx
+++ b/zipline/data/_equities.pyx
@@ -100,11 +100,11 @@ cpdef _compute_row_slices(dict asset_starts_absolute,
         intp_t asset_end_calendar
 
     for i, asset in enumerate(requested_assets):
-        try:
-            asset_start_data = asset_starts_absolute[asset]
-        except KeyError:
+        if asset not in asset_starts_absolute:
             # This is an unknown asset, leave its slot empty.
             continue
+        else:
+            asset_start_data = asset_starts_absolute[asset]
 
         asset_end_data = asset_ends_absolute[asset]
         asset_start_calendar = asset_starts_calendar[asset]
--- a/zipline/data/_equities.pyx
+++ b/zipline/data/_equities.pyx
@@ -99,13 +99,18 @@ cpdef _compute_row_slices(dict asset_starts_absolute,
         intp_t asset_start_calendar
         intp_t asset_end_calendar
 
+        # Flag to check whether we should raise an error because we don't know
+        # about any of the requested sids.
+        uint8_t any_hits = 0
+
     for i, asset in enumerate(requested_assets):
         if asset not in asset_starts_absolute:
             # This is an unknown asset, leave its slot empty.
             continue
-        else:
-            asset_start_data = asset_starts_absolute[asset]
 
+        any_hits = 1
+
+        asset_start_data = asset_starts_absolute[asset]
         asset_end_data = asset_ends_absolute[asset]
         asset_start_calendar = asset_starts_calendar[asset]
         asset_end_calendar = (
@@ -130,6 +135,9 @@ cpdef _compute_row_slices(dict asset_starts_absolute,
         # Otherwise, offset by the number of rows in the query in which the
         # asset did not yet exist.
         offset_a[i] = max(0, asset_start_calendar - query_start)
+
+    if not any_hits:
+        raise ValueError('At least one valid asset id is required.')
 
     return first_row_a, last_row_a, offset_a
 
--- a/zipline/data/_equities.pyx
+++ b/zipline/data/_equities.pyx
@@ -19,6 +19,7 @@ from cpython cimport bool
 from numpy import (
     array,
     float64,
+    full,
     intp,
     uint32,
     zeros,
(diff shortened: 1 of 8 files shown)


zipline/data/hdf5_daily_bars.py
```diff
@@ -466,41 +466,63 @@ def load_raw_arrays(self,
...
 
-        sid_selector = self.sids.searchsorted(assets)
+        sid_selector, out_buf_indexer = self._make_sid_indexers(assets)
+
         date_slice = self._compute_date_range_slice(start, end)
 
-        nrows = date_slice.stop - date_slice.start
-        ncols = len(self.sids)
+        n_dates = date_slice.stop - date_slice.start
+        n_query_sids = len(assets)
```
ssanderson: The fact that the order of these two variables is transposed from the order in which they're used to initialize `read_buf`/`out_buf` below tripped me up a bit when trying to keep these variables straight.

zipline/pipeline/loaders/synthetic.py
```diff
@@ -341,13 +341,20 @@ def expected_bar_value_with_holes(asset_id,
...
+                           holes=None):
     """
     Return an 2D array containing cls.expected_value(asset_id, date,
     colname) for each date/asset pair in the inputs.
 
-    Values before/after an assets lifetime, as well as the locs defined,
-    in `holes` are filled with 0 for volume and NaN for price columns.
+    Missing locs are filled with 0 for volume and NaN for price columns:
+
+        - Values before/after an assets lifetime.
```
ssanderson: asset's

zipline/data/hdf5_daily_bars.py
```diff
@@ -466,41 +466,63 @@ def load_raw_arrays(self,
...
+            # Zero the buffers to prepare to receive new data.
+            read_buf.fill(0)
+            out_buf.fill(0)
 
             dataset = self._country_group[DATA][column]
 
             dataset.read_direct(
-                buf,
+                read_buf,
                 np.s_[:, date_slice.start:date_slice.stop],
```
ssanderson: Not changed here, but reading this again I think this might just be `np.s_[:, date_slice]`?

zipline/data/hdf5_daily_bars.py
```diff
@@ -466,41 +466,63 @@ def load_raw_arrays(self,
...
 
-        nrows = date_slice.stop - date_slice.start
-        ncols = len(self.sids)
+        n_dates = date_slice.stop - date_slice.start
+        n_query_sids = len(assets)
+        n_valid_sids = len(self.sids)
 
-        buf = np.zeros((ncols, nrows), dtype=np.uint32)
+        read_buf = np.zeros((n_valid_sids, n_dates), dtype=np.uint32)
+        out_buf = np.zeros((n_query_sids, n_dates), dtype=np.uint32)
```
ssanderson: I'm not sure pre-allocating the out_buf is helping us much here if we're making a copy of it each time anyway.

zipline/data/hdf5_daily_bars.py
```diff
@@ -726,21 +748,17 @@ def countries(self):
...
-        except KeyError as exc:
-            raise_from(
-                NoDataForSid(
-                    'Assets not contained in daily pricing file: {}'.format(
-                        set(assets) - set(self._country_map)
-                    )
-                ),
-                exc
-            )
+        country_codes = self._country_map.get(assets)
```
ssanderson: Why the switch from `.loc` to `.get` here? It looks like they have the same behavior:

```
In [6]: s
Out[6]:
a    1
b    2
dtype: int64

In [7]: s.get(['a', 'c'])
Out[7]:
a    1.0
c    NaN
dtype: float64

In [8]: s.loc[['a', 'c']]
Out[8]:
a    1.0
c    NaN
dtype: float64
```

ssanderson: Oh, hmm, they differ when none of the keys are in the series...`loc` raises a KeyError, but `get` returns an all-nan series.

zipline/data/hdf5_daily_bars.py
```diff
@@ -726,21 +748,17 @@ def countries(self):
...
 
-        unique_country_codes = country_codes.unique()
+        if country_codes is not None:
```
ssanderson: When does pandas return None? I couldn't recreate that example in pandas 18 at least.

zipline/data/hdf5_daily_bars.py
```diff
@@ -726,21 +748,17 @@ def countries(self):
...
+            unique_country_codes = country_codes.dropna().unique()
+            num_countries = len(unique_country_codes)
+        else:
+            num_countries = 0
 
-        if len(unique_country_codes) > 1:
+        if num_countries == 0:
+            return self._country_map.iloc[0]
```
ssanderson: Hmm...it feels a little dangerous to return a value here. As a consumer, I'm not sure I'd expect to get a random country back if no assets were found. Another slight worry I have is that this would break in a surprising way if we don't have any readers (we'd KeyError on the .iloc[0] call). 

One alternative would be to raise an error or return a sentinel value here and have the caller decide what to do in this case. For `load_raw_arrays`, for example, we could return an array containing all missing values without ever delegating to any underlying readers (which would work even if we don't have any such readers). I'm 50/50 on whether that compositionality improvement is with the extra complexity in call-sites though. 

Another option would be to just make this an error case, since most likely it's a bug if we hit this path in real-world code.

 Thoughts?

zipline/data/hdf5_daily_bars.py
yankees714: And in newer pandas, `get` returns None!

zipline/data/hdf5_daily_bars.py
yankees714: It's happening in 22, which one of the travis builds uses.

zipline/data/hdf5_daily_bars.py
yankees714: Yeah, my thinking here was that we just want to route this request to a reader that can give us back an empty array. But these are good points, I think I'll just have this raise a `ValueError`.

zipline/data/hdf5_daily_bars.py
yankees714: 🤔

zipline/data/_equities.pyx
```diff
@@ -191,8 +197,13 @@ cpdef _read_bcolz_data(ctable_t table,
 
             for asset in range(nassets):
                 first_row = first_rows[asset]
+                if first_row == -1:
```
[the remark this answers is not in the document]
yankees714: Yeah you're right...I'm not sure why this isn't getting caught. Looking into it now.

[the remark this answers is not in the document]
yankees714: Hah, turns out this actually works without the check. Since `first_row`, `last_row`, and `offset` are all -1, we end up trying set the slice `[-1:0]`, which is empty. I much prefer exiting early though, so I'll add the check in both places.

zipline/data/bcolz_daily_bars.py
```diff
@@ -570,6 +570,9 @@ def _compute_slices(self, start_idx, end_idx, assets):
         )
 
     def load_raw_arrays(self, columns, start_date, end_date, assets):
+        if not any(asset in self._calendar_offsets for asset in assets):
```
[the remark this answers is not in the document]
ssanderson: Implemented this suggestion here: https://github.com/quantopian/zipline/pull/2348.

yankees714:
For anyone reading this, #2352 has the suggestions mentioned above (#2348 is this PR).

zipline/data/bcolz_daily_bars.py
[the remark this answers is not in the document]
ssanderson: woops: https://github.com/quantopian/zipline/pull/2348#issuecomment-434435433

PR #2349, opened 2018-10-26
PERF: Avoid a copy on reads from h5.

ssanderson:
Instead of maintaining separate read and output buffers, allocate the read
buffer with one extra row, to which we never write, and put -1s in the indexer
at the locations of invalid sids.

### DIFF
--- a/zipline/data/hdf5_daily_bars.py
+++ b/zipline/data/hdf5_daily_bars.py
@@ -471,42 +471,51 @@ class HDF5DailyBarReader(SessionBarReader):
 
         start = start_date.asm8
         end = end_date.asm8
-
-        sid_selector, out_buf_indexer = self._make_sid_indexers(assets)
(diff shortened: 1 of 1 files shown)


zipline/data/hdf5_daily_bars.py
```diff
@@ -471,56 +471,63 @@ def load_raw_arrays(self,
...
-            )
+            # Select data from the **full buffer**. Unknown assets will pull
+            # from the last row, which is always empty.
+            out.append(self._postprocessors[column](full_buf[sid_selector].T))
 
         return out
 
     def _make_sid_indexers(self, assets):
         """
+        Build an indexer that maps an array parallel with ``self.sids`` to an
```
ssanderson: This docstring is out of date. Will fix assuming we decide we want this.

PR #2351, opened 2018-10-30
MAINT: update symbol lookup errors in AssetFinder

vikram-narayan:
- raise a `SameSymbolUsedAcrossCountries` if multiple assets across countries share the same symbol
- raise a `MultipleSymbolsFoundForFuzzySymbol` if multiple symbols are found when `fuzzy=True` is passed

### DIFF
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -887,7 +887,7 @@ class AssetFinder(object):
                 # there was only one exact match
                 return options[0]
 
-             # there is more than one exact match for this fuzzy symbol
+            # there is more than one exact match for this fuzzy symbol
             raise MultipleSymbolsFoundForFuzzySymbol(
                 symbol=symbol,
                 options=self.retrieve_all(owner.sid for owner in owners),
--- a/zipline/assets/assets.py
+++ b/zipline/assets/assets.py
@@ -822,11 +822,12 @@ class AssetFinder(object):
             options = {self.retrieve_asset(owner.sid) for owner in owners}
 
             if multi_country:
-                country_codes = set(map(attrgetter('country_code'), options))
+                country_codes = map(attrgetter('country_code'), options)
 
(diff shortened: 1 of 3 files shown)


zipline/errors.py
```diff
@@ -315,6 +315,33 @@ class MultipleSymbolsFound(ZiplineError):
     """.strip()
 
 
+class MultipleSymbolsFoundForFuzzySymbol(ZiplineError):
```
yankees714: Could a case be made for having this subclass `MultipleSymbolsFound`? I'm not sure from this PR what the motivation for separating these is.

vikram-narayan: the motivation is the error message being different, so I'm not sure if there is a benefit to subclassing `MultipleSymbolsFound`

zipline/assets/assets.py
```diff
@@ -812,39 +814,50 @@ def _lookup_symbol_strict(self,
...
-
             # exactly one equity has ever held this symbol, we may resolve
             # without the date
-            return self.retrieve_asset(owners[0].sid)
+            if len(owners) == 1:
+                return self.retrieve_asset(owners[0].sid)
+
+            options = set(map(
+                compose(self.retrieve_asset, attrgetter('sid')), owners
+            ))
```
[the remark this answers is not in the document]
vikram-narayan: 👍 fixed

zipline/errors.py
yankees714: Yeah, one thing I was thinking of is that if any consumer is catching this, the behavior there will change. Probably not a huge deal though.

vikram-narayan: ah that's true, my initial thought was to update consumers that catch this with the more specific exceptions, but they are all the same family of error. Updated  `SameSymbolUsedAcrossCountries` to also be subclassed from `MultipleSymbolsFound`

PR #2354, opened 2018-10-31
MAINT: add adjust_date to pipeline engine

vikram-narayan:


zipline/pipeline/engine.py
```diff
@@ -751,6 +751,23 @@ def _validate_compute_chunk_params(self,
                     )
                 )
 
+    def adjust_date(self, pipeline, dt):
```
ssanderson: Talked about this in person a bit. I think my preference here would be to split this into two piece of functionality:

1. `resolve_domain` (which already exists, but needs to be made public and tested)
2. `adjust_dates`, which can be part of the `Domain` interface.

The rationale for this split is twofold:

1. We might have consumers that have a domain and want to align dates to the domain's calendar, but don't want to construct a full-blown pipeline engine.
2. The domain is the logical "owner" of date information, so it feels like a more clean separation to have it be responsible for the rolling logic.

vikram-narayan: made these changes, and split up the tests into test_domain.py and test_engine.py

zipline/pipeline/domain.py
```diff
@@ -71,6 +71,27 @@ def data_query_cutoff_for_sessions(self, sessions):
             "available" on each session.
         """
 
+    @default
+    def adjust_date(self, dt):
```
ssanderson: Thoughts on renaming this to `adjust_forward` or `roll_forward`? `adjust` doesn't tell me as a reader which way the adjustment goes.

tests/pipeline/test_engine.py
```diff
@@ -1562,3 +1567,43 @@ def test_no_groupby_maximum(self):
...
+    def test_resolve_domain(self):
+        engine_generic = SimplePipelineEngine(
+            None, None, default_domain=GENERIC
+        )
+        engine_jp = SimplePipelineEngine(
+            None, None, default_domain=JP_EQUITIES
+        )
+
+        pipe_generic = Pipeline()
+        pipe_us = Pipeline(domain=US_EQUITIES)
```
ssanderson: One more case worth testing here would be a pipeline with no explicit domain, but with a US-specialized column, to make sure that we're properly using the pipeline's inference.

tests/pipeline/test_domain.py
```diff
@@ -479,3 +479,30 @@ def test_equity_session_domain(self, parameters):
...
+
+    def test_adjust_date(self):
+        #     January 2017
+        # Su Mo Tu We Th Fr Sa
+        #  1  2  3  4  5  6  7
+
+        # the first three days of the year are holidays on the Tokyo exchange,
+        # so the first trading day should be the fourth
+        self.assertEqual(
+            JP_EQUITIES.adjust_date('01-01-2017'),
```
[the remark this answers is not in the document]
vikram-narayan: I usually write yyyy-mm-dd, not sure why I wrote it this way here. fixed

zipline/pipeline/domain.py
vikram-narayan: true, changed to `roll_forward`

tests/pipeline/test_engine.py
vikram-narayan: added

tests/pipeline/test_engine.py
```diff
@@ -1562,3 +1567,51 @@ def test_no_groupby_maximum(self):
                         .reset_index(level=1, drop=True))
 
         assert_equal(groupby_max, pipeline_max)
+
+
+class ResolveDomainTestCase(zf.ZiplineTestCase):
+
+    def test_resolve_domain(self):
+        engine_generic = SimplePipelineEngine(
+            None, None, default_domain=GENERIC
```
ssanderson: I'd probably use `NamedExplodingObject` for these so that we raise a clear error if something in the engine starts depending on the asset finder or countries in loader dispatcher at construction time.

zipline/pipeline/domain.py
```diff
@@ -71,6 +71,32 @@ def data_query_cutoff_for_sessions(self, sessions):
...
+        dt : pd.Timestamp
+
+        Returns
+        -------
+        pd.Timestamp
+        """
+        dt = pd.Timestamp(dt, tz='UTC')
+
+        trading_days = self.all_sessions()
+        if dt in trading_days:
```
ssanderson: I think we can remove this branch.

zipline/pipeline/domain.py
```diff
@@ -71,6 +71,32 @@ def data_query_cutoff_for_sessions(self, sessions):
...
+            return dt
+        try:
+            return trading_days[trading_days.searchsorted(dt)]
+        except IndexError:
+            raise IndexError(
```
ssanderson: This should be a ValueError. From the caller's perspective, an IndexError is a strange error, because they're not interacting with an array/sequence.

zipline/pipeline/domain.py
```diff
@@ -71,6 +71,34 @@ def data_query_cutoff_for_sessions(self, sessions):
...
+        pd.Timestamp
+        """
+        dt = pd.Timestamp(dt, tz='UTC')
+
+        trading_days = self.all_sessions()
+        try:
+            return trading_days[trading_days.searchsorted(dt)]
+        except IndexError:
+            raise ValueError(
+                "Date {} was past the last session for domain {}. "
```
ssanderson: :+1: this is a very clear error!

PR #2355, opened 2018-11-01
ENH: Add India, Singapore, China domains.

jbredeche:


pvssvikas:
2018-10-18 <-- at India calendar missed a holiday

jbredeche:
thanks @pvssvikas.  You are right, this was an oversight.  I'll fix.

PR #2379, opened 2018-12-05
MAINT: Pin transitive dependencies for `requests`

cmichelQT:


etc/requirements.txt
```diff
@@ -14,6 +14,12 @@ numpy==1.11.3
 # for pandas-datareader
 requests-file==1.4.1
 
+# for requests
+certifi==2018.8.24
```
richafrank: Do we need to fix our build to handle pinning this new requirement, like we saw with https://github.com/quantopian/zipline/pull/2331 ?

cmichelQT:
@richafrank While I wait for test to past, do you see anything obviously wrong in the commits?

cmichelQT:
Did not push the requirements file, fixing.

.travis.yml
```diff
@@ -73,7 +74,10 @@ install:
   # Make sure stdout is in blocking mode. If we don't, then conda create will barf during downloads.
   # See https://github.com/travis-ci/travis-ci/issues/4704#issuecomment-348435959 for details.
   - python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'
-  - conda create -n testenv --use-local --yes -c quantopian -c quantopian/label/ci pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION ta-lib=$TALIB_VERSION libgfortran=3.0
+  - conda create -n testenv --use-local --yes -c quantopian -c quantopian/label/ci pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION ta-lib=$TALIB_VERSION libgfortran=3.0 certifi=$CERTIFI_VERSION
+  - source activate testenv
```
richafrank: Looks like duplicate commands were added here.

etc/requirements_certifi.txt
```diff
@@ -0,0 +1 @@
+certifi==certifi=2018.8.24
```
richafrank: Is there a reason to install `certifi` separately, or could we move this into requirements.txt? I think it's not optional, while TA-lib is.

.appveyor.yml
```diff
@@ -94,7 +94,8 @@ install:
   - ps: $env:PYTABLES_VERSION=(sls "tables==(.*)" .\etc\requirements.txt -ca).matches.groups[1].value
   - ps: $env:H5PY_VERSION=(sls "h5py==(.*)" .\etc\requirements.txt -ca).matches.groups[1].value
   - ps: $env:TALIB_VERSION=(sls "TA-Lib==(.*)" .\etc\requirements_talib.txt -ca).matches.groups[1].value
-  - conda create -n testenv --yes -q --use-local pip python=%PYTHON_VERSION% numpy=%NUMPY_VERSION% pandas=%PANDAS_VERSION% scipy=%SCIPY_VERSION% ta-lib=%TALIB_VERSION% bcolz=%BCOLZ_VERSION% numexpr=%NUMEXPR_VERSION% pytables=%PYTABLES_VERSION% h5py=%H5PY_VERSION% -c quantopian -c https://conda.anaconda.org/quantopian/label/ci
+  - ps: $env:CERTIFI_VERSION=(sls "certifi==(.*)" .\etc\requirements_certifi.txt -ca).matches.groups[1].value
```
richafrank: Probably worth adding comments around why we need to install `certifi` like this (since it's different from most of the requirements).

cmichelQT:
@richafrank Can you explain why appveyor is failing here? 

It seems to exit with `conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform win-64: set(['python 2.7.*'])` 
running the command `ps: copy .\ci\appveyor\vcvars64.bat "C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\bin\amd64"`

After some bit of googling, it may be a bug? Suggestions are to 
`conda config --add channels conda-forge` or `conda build -c conda-forge your_pacakge_name`

But I do not believe we want to do that here? So I am stuck and this issue seems unrelated to my changes?

richafrank:
Looks like a transient anaconda issue. (Build is passing now.)

.appveyor.yml
```diff
@@ -94,7 +94,9 @@ install:
   - ps: $env:PYTABLES_VERSION=(sls "tables==(.*)" .\etc\requirements.txt -ca).matches.groups[1].value
   - ps: $env:H5PY_VERSION=(sls "h5py==(.*)" .\etc\requirements.txt -ca).matches.groups[1].value
   - ps: $env:TALIB_VERSION=(sls "TA-Lib==(.*)" .\etc\requirements_talib.txt -ca).matches.groups[1].value
-  - conda create -n testenv --yes -q --use-local pip python=%PYTHON_VERSION% numpy=%NUMPY_VERSION% pandas=%PANDAS_VERSION% scipy=%SCIPY_VERSION% ta-lib=%TALIB_VERSION% bcolz=%BCOLZ_VERSION% numexpr=%NUMEXPR_VERSION% pytables=%PYTABLES_VERSION% h5py=%H5PY_VERSION% -c quantopian -c https://conda.anaconda.org/quantopian/label/ci
+  # conda install `certifi` but its a dependency of `requests`, so read from requirements so versions do not diverege.
```
richafrank: I think instead we should try to explain why we need to install `certifi` differently from even other dependencies of `requests`. How about:

```suggestion
  # We conda install certifi at the pinned exact version because it is a transitive dependency of zipline via requests and uses distutils for packaging. 
  # Since conda installs latest certifi by default, we would fail to uninstall that new version when trying to install the pinned version using pip later in the build:
  # "Cannot uninstall 'certifi'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall."
```

.travis.yml
```diff
@@ -73,7 +74,7 @@ install:
   # Make sure stdout is in blocking mode. If we don't, then conda create will barf during downloads.
   # See https://github.com/travis-ci/travis-ci/issues/4704#issuecomment-348435959 for details.
   - python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'
-  - conda create -n testenv --use-local --yes -c quantopian -c quantopian/label/ci pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION ta-lib=$TALIB_VERSION libgfortran=3.0
+  - conda create -n testenv --use-local --yes -c quantopian -c quantopian/label/ci pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION ta-lib=$TALIB_VERSION libgfortran=3.0 certifi=$CERTIFI_VERSION
```
richafrank:
```suggestion
  # We conda install certifi at the pinned exact version because it is a transitive dependency of zipline via requests and uses distutils for packaging. 
  # Since conda installs latest certifi by default, we would fail to uninstall that new version when trying to install the pinned version using pip later in the build:
  # "Cannot uninstall 'certifi'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall."
  - conda create -n testenv --use-local --yes -c quantopian -c quantopian/label/ci pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION ta-lib=$TALIB_VERSION libgfortran=3.0 certifi=$CERTIFI_VERSION
```

cmichelQT:
@richafrank Thanks for the suggestions, I applied them, I am going to merge since the only diff was the comment.

PR #2380, opened 2018-12-05
MAINT: Update python-interface version

quantophred:
This update eliminates some DeprecationWarnings on newer versions of
Python.


cmichelQT:
@fwagner-quantopian Looks like `python-interface` pulled in `funcsigs==1.0.2` and `typing==3.6.2`, do we want to pin those here as well?

PR #2394, opened 2019-01-02
BUG: Fix `week_start` when Monday precedes the New Year.

ehebert:
The `week_start` date rule was incorrectly returning `False` for
`2018-12-31` on `NYSE` calendar.

The root cause of the bug was that the groupby used in
`TradingDayOfWeekRule.execution_period_values` was using the pairing
of `year` and `weekofyear`. This pairing grouped `2018-12-31` with
the first week of January 2018 (`2018-01-02`, `2018-01-03`, ...),
instead of with `2019-01-02`, `2019-01-03`, ...

This misgrouping occurred because the values of (year, weekofyear)
for `2018-12-31` are (2018, 1); however the other dates that are in the
same week as `2018-12-31` have a value of (2019, 1).

Since `2018-12-31` was misgrouped, the call to `nth(0)` would not select
it as the first of the week; since it appeared to be the last day of the
first week of `2018`

Instead, use the `isocalendar` method of the datetime object.
https://docs.python.org/2/library/datetime.html#datetime.date.isocalendar
`2018-12-31` is part of the ISO Week `2019-W01-1`, as are `2019-01-02`, etc.
(ISO Weeks location is determined by where the Thursday falls.)

Also, change the edge case tests so that a truth table of date and
expected 'should trigger' result is defined.

In the previous version of the test, moving the start date before
the New Year was hitting edge cases within the test construction around
offsets and calendar days.
Moving the first "sim date" would have been required to make sure that
a date like 2016-12-30 or 2018-12-31 was covered.

The change to use of the truth tables also takes the tests time down from 3.5
seconds to around 100ms on my dev machine.

### DIFF
--- a/tests/events/test_events_nyse.py
+++ b/tests/events/test_events_nyse.py
@@ -18,7 +18,6 @@ from datetime import timedelta
 import pandas as pd
 from nose_parameterized import parameterized
 
-from zipline.testing import parameter_space
 from zipline.utils.events import NDaysBeforeLastTradingDayOfWeek, AfterOpen, \
     BeforeClose
 from zipline.utils.events import NthTradingDayOfWeek
--- a/tests/events/test_events_nyse.py
+++ b/tests/events/test_events_nyse.py
@@ -145,7 +145,7 @@ class TestStatelessRulesNYSE(StatelessRulesTests, TestCase):
         self.assertEquals(expected, results)
 
     @parameterized.expand([('week_start',), ('week_end',)])
-    def test_week_and_time_composed_rule(self, taype):
+    def test_week_and_time_composed_rule(self, type):
         week_rule = NthTradingDayOfWeek(0) if type == 'week_start' else \
             NDaysBeforeLastTradingDayOfWeek(4)
         time_rule = AfterOpen(minutes=60)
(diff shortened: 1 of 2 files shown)


tests/events/test_events_nyse.py
```diff
@@ -49,70 +44,108 @@ def test_edge_cases_for_TradingDayOfWeek(self,
...
+        results = {
+            x: rule.should_trigger(self.cal.next_open(T(x)))
+            for x in expected.keys()
+        }
+
+        self.assertEquals(expected, results)
 
     @parameterized.expand([('week_start',), ('week_end',)])
-    def test_week_and_time_composed_rule(self, type):
+    def test_week_and_time_composed_rule(self, taype):
```
jbredeche: typo?

jbredeche:
Besides minor typo, LGTM.

tests/events/test_events_nyse.py
ehebert: This snuck in somehow while making the edit. Will fix.

However, `nosetests -x tests/events/test_events_nyse.py`  does pass even with the typo.
It is passing because we are using same name as the the built-in function `type` in this line, https://github.com/quantopian/zipline/blob/4cac769beebb22cadff84e42f93c37a1fcd2a554/tests/events/test_events_nyse.py#L116; resulting in always evaluating to the the `week_end` branch.

jmccorriston:
Double checked with @ehebert. He said we're good to merge.

PR #2398, opened 2019-01-07
MAINT: Use a param name that is not the type builtin.

ehebert:
# Notes

Follows up on https://github.com/quantopian/zipline/pull/2394#discussion_r244758588


jbredeche:
❤️

PR #2399, opened 2019-01-08
MAINT: Upgrade flake8 to 3.6.0

quantophred:
The new version catches some new errors, and supports nicer syntax in configuration files.

### DIFF
--- a/zipline/testing/core.py
+++ b/zipline/testing/core.py
@@ -1451,7 +1451,7 @@ class _TmpBarReader(with_metaclass(ABCMeta, tmp_dir)):
                 self._data,
             )
             return self._reader_cls(tmpdir.path)
-        except:
+        except BaseException:  # Clean up even on KeyboardInterrupt
             self.__exit__(None, None, None)
             raise
 
--- a/zipline/testing/core.py
+++ b/zipline/testing/core.py
@@ -893,7 +893,7 @@ class SubTestFailures(AssertionError):
             '\n    '.join((
                 ', '.join('%s=%r' % item for item in scope.items()),
                 self._format_exc(exc_info),
-            )) for scope, exc_info in self.failures,
+            )) for scope, exc_info in self.failures
         )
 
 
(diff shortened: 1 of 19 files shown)


quantophred:
Note that the only commit which should change any behavior at all is 7d853af: everything else should just be syntax.

zipline/testing/core.py
```diff
@@ -1451,7 +1451,7 @@ def __enter__(self):
                 self._data,
             )
             return self._reader_cls(tmpdir.path)
-        except:
+        except BaseException:  # Gotta catch 'em all
```
llllllllll: I almost don't even think this is a reasonable change. If you actually want to catch all exceptions `except:` does that and is pretty clear about it.

llllllllll:
I think I would prefer using a noqa and `except:` over `except BaseException`, how do other people feel?

Other than that lgtm

zipline/testing/core.py
richafrank: Could we move to `except Exception:` and assume `MemoryError` and the like are unrecoverable?

llllllllll: This code doesn't recover, it calls calls an `__exit__` and then re-raises. If you are re-raising I think it is fine to catch everything.

llllllllll: You can get a memory error for a lot of reasons, `np.arange(2 ** 40)` will memory error but you still have plenty of memory when you need to clean up. Even recursion error is likely fine because once you make it back to the except handler you have unwound all the frames created in the body of the try so you have spare frames to cleanup.

quantophred: It looks like catching only `Exception` would result in lingering temp files if sigint or sigterm is received here, so I think `BaseException` is the right choice. (`MemoryError` is actually a subclass of `Exception`.)

I agree that it is abundantly clear to you and me what `except` does, but whenever I encounter it in someone else's code, I wonder whether they really meant it (as we seem to not have at two other locations in this PR).

I personally think `# noqa` looks/feels a little grosser than `BaseException`, but they're basically equivalent to me.

quantophred:
I tried switching the `BaseException`s over to `# noqa`, but there seems to be some weirdness in flake8 around E722 (bare except clause) and `noqa` comments. Not sure exactly what it's doing yet, but I'm pretty sure there's a bug in either flake8 or pycodestyle that only affects that particular error code. I'd recommend just sticking with `BaseException` for now.

llllllllll:
LGTM

PR #2415, opened 2019-02-06
Add sphinx-sitemap extension

jdillard:
This adds a sphinx extension that creates a sitemap.xml at https://www.zipline.io/sitemap.xml to improve SEO etc and help people find the information they need.


ssanderson:
@jdillard I don't think we currently have much in the way of SEO issues with the Zipline docs, and I'm hesitant to take on another third party dependency for this. Going to close this for now.

PR #2419, opened 2019-02-11
BUG: group exprdata by the expressions, not identity

llllllllll:
When comparing exprdata, check `isidentical` on the constituent expressions instead of identity on the exprdata tuple itself. There are cases where we don't register the whole underlying table in a single call to `register_dataset` because it is difficult to group the columns this way. The equality of the exprdata object determines how we batch queries to the underlying, so we want to group them up as much as possible or we will be querying for the same `sid`, `asof_date`, and `timestamp` columns and re-aligning them for multiple columns.

### DIFF
--- a/zipline/pipeline/loaders/blaze/core.py
+++ b/zipline/pipeline/loaders/blaze/core.py
@@ -729,25 +729,42 @@ class ExprData(_expr_data_base):
             expr,
             deltas,
             checkpoints,
-            odo_kwargs or {},
+            frozenset((odo_kwargs or {}).items()),
         )
 
     def __repr__(self):
         # If the expressions have _resources() then the repr will
         # drive computation so we take the str here.
-        cls = type(self)
-        return super(ExprData, cls).__repr__(cls(
+        return repr(_expr_data_base(
             str(self.expr),
             str(self.deltas),
             str(self.checkpoints),
-            self.odo_kwargs,
+            dict(self.odo_kwargs),
         ))
 
+    @staticmethod
+    def _expr_eq(a, b):
+        return a is b is None or a.isidentical(b)
+
     def __hash__(self):
-        return id(self)
+        return super(ExprData, self).__hash__()
 
     def __eq__(self, other):
-        return self is other
+        if not isinstance(other, ExprData):
+            return NotImplemented
+
+        return (
+            self._expr_eq(self.expr, other.expr) and
+            self._expr_eq(self.deltas, other.deltas) and
+            self._expr_eq(self.checkpoints, other.checkpoints) and
+            self.odo_kwargs == other.odo_kwargs
+        )
+
+    def __ne__(self, other):
+        # note: ``tuple`` (inherited from ``namedtuple``) adds a ``__ne__``,
+        # but we want to rely on ``__eq__` to use ``isidentical`` to compare
+        # expressions
+        return not (self == other)
 
 
 class BlazeLoader(object):
@@ -907,6 +924,7 @@ class BlazeLoader(object):
             )
 
         expr, deltas, checkpoints, odo_kwargs = expr_data
+        odo_kwargs = dict(odo_kwargs)
 
         have_sids = (first(columns).dataset.ndim == 2)
         added_query_fields = {AD_FIELD_NAME, TS_FIELD_NAME} | (
--- a/tests/pipeline/test_blaze.py
+++ b/tests/pipeline/test_blaze.py
@@ -2415,6 +2415,38 @@ class MiscTestCase(ZiplineTestCase):
             " checkpoints='checkpoints', odo_kwargs={'a': 'b'})",
         )
 
+    def test_exprdata_eq(self):
+        dshape = 'var * {sid: int64, asof_date: datetime, value: float64}'
+        base_expr = bz.symbol('base', dshape)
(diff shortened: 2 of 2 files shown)


zipline/pipeline/loaders/blaze/core.py
```diff
@@ -729,25 +729,36 @@ def __new__(cls,
...
     def __eq__(self, other):
-        return self is other
+        if not isinstance(other, ExprData):
+            return NotImplemented
+
+        return (
+            self.expr.isidentical(other.expr) and
+            self._expr_eq(self.deltas, other.deltas) and
+            self._expr_eq(self.checkpoints, other.checkpoints) and
+            self.odo_kwargs == other.odo_kwargs
```
ssanderson: Any worry that `odo_kwargs` might contain object that aren't comparable and/or objects that don't have the equality semantics we want?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -729,25 +729,36 @@ def __new__(cls,
...
+        )
+
+    def __ne__(self, other):
```
ssanderson: I think this is the default behavior. Is there a reason to implement this explicitly?

```
In [16]: class Foo(object):
    ...:
    ...:     def __eq__(self, other):
    ...:         print("calling eq")
    ...:         return True
    ...:
    ...:

In [17]: Foo() == Foo()
calling eq
Out[17]: True

In [18]: Foo() != Foo()
calling eq
Out[18]: False
```

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -729,25 +729,36 @@ def __new__(cls,
...
-            odo_kwargs or {},
+            frozenset((odo_kwargs or {}).items()),
         )
 
     def __repr__(self):
         # If the expressions have _resources() then the repr will
         # drive computation so we take the str here.
-        cls = type(self)
-        return super(ExprData, cls).__repr__(cls(
+        return repr(_expr_data_base(
```
ssanderson: Why this change over the `super`-based implementation?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -729,25 +729,36 @@ def __new__(cls,
...
+    def _expr_eq(a, b):
+        return a is b is None or a.isidentical(b)
 
     def __eq__(self, other):
-        return self is other
+        if not isinstance(other, ExprData):
+            return NotImplemented
+
+        return (
+            self.expr.isidentical(other.expr) and
```
ssanderson: Are we not using `_expr_eq` here to avoid the extra cost of the two `is None` checks? That seems unlikely to be worth the cost over the improved clarity of making it clear that we want to do the logical comparison for all three blaze expressions. At the very least, we should comment here that these are the same logical check (assuming that's the case and I'm not missing something). If the perf does matter that much, should we be binding `_expr_eq` as a local?

zipline/pipeline/loaders/blaze/core.py
```diff
@@ -729,25 +729,36 @@ def __new__(cls,
             expr,
             deltas,
             checkpoints,
-            odo_kwargs or {},
+            frozenset((odo_kwargs or {}).items()),
```
ssanderson: Why the change from dict to frozenset? Is this just being paranoid about mutation? Are all the objects we pass as values to `odo_kwargs` hashable?

zipline/pipeline/loaders/blaze/core.py
llllllllll: `expr` can't be None so I didn't consider using that. I could use the method to be more clear though.

zipline/pipeline/loaders/blaze/core.py
llllllllll: They are currently hashable (bytes values, unicode keys). This is because we want the `__hash__` of `ExprData` to be by value, not identity.

zipline/pipeline/loaders/blaze/core.py
llllllllll: No, these are mostly simple objects like strings, bytes, or ints. I think the only real use site we have for this is `bytes`.

zipline/pipeline/loaders/blaze/core.py
llllllllll: I believe `tuple` changes this:
```python
In [9]: class T(tuple):
   ...:     def __eq__(self, other):
   ...:         print('eq')
   ...:         return super(T, self).__eq__(other)
   ...:     
   ...:     

In [10]: t = T((1, 2, 3))

In [11]: t == t
eq
Out[11]: True

In [12]: t != t
Out[12]: False

```

zipline/pipeline/loaders/blaze/core.py
llllllllll: I actually want to call the super class's constructor and then repr that. I don't want to do the `dict` to `frozenset` conversion. This seemed like the most straightforward way to do that. I am not that worried about some complex hierarchy below this type.

zipline/pipeline/loaders/blaze/core.py
ssanderson: Ah. Can we add a comment here noting that we need a `__ne__` to override the one we get from being a tuple subclass?

PR #2420, opened 2019-02-13
ENH: Add AdjustedArray.map_labels

yankees714:
To allow mapping the baseline and adjustment values when backed by a `LabelArray`.

This also make `value` attribute on `Adjustment` objects public, so that `map_labels()` can perform the mapping in place.

### DIFF
--- a/zipline/lib/adjusted_array.py
+++ b/zipline/lib/adjusted_array.py
@@ -13,6 +13,7 @@ from numpy import (
     uint32,
     uint8,
 )
+from six import iteritems
 from zipline.errors import (
     WindowLengthNotPositive,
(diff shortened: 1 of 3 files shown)


zipline/lib/adjusted_array.py
```diff
@@ -237,6 +238,25 @@ def inspect(self):
             adjustments=self.adjustments,
         )
 
+    def map_labels(self, func):
```
llllllllll: I might call this something like `update_labels` to imply that this operates in place.

yankees714: Ah nice, I'll update

PR #2421, opened 2019-02-13
use identity of odo_kwargs for id/eq

llllllllll:
We use a nested dict in some places, this seems safer in the general case.


quantophred:
The S3 bucket Travis uses to install Python 3.5 is currently returning 403 for all requests, causing our builds to fail. The Appveyor 3.5 build passed, though, so we are going ahead and merging this, and will re-run the Travis build later just in case.

quantophred:
Update: Travis is working and master builds are re-running!

PR #2424, opened 2019-02-14
Fix exprdata comparison again

llllllllll:
When `odo_kwargs` was None we would eagerly construct a new empty dict which would have a unique ID. This change does 2 things:

1. ExprData no longer inherits from named tuple, it was getting to be more than a simple struct.
2. ExprData now lazily constructs an empty dict for `odo_kwargs` when requested, but preserves the None value for id comparisons.

### DIFF
--- a/zipline/pipeline/loaders/blaze/core.py
+++ b/zipline/pipeline/loaders/blaze/core.py
@@ -138,7 +138,6 @@ www.quantopian.com/help#pipeline-api
 from __future__ import division, absolute_import
 
 from abc import ABCMeta, abstractproperty
-from collections import namedtuple
 from functools import partial
 from itertools import count
 import warnings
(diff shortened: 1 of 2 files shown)


zipline/pipeline/loaders/blaze/core.py
```diff
@@ -719,37 +713,66 @@ class ExprData(_expr_data_base):
...
+        }
+        invalid_kwargs = set(kwargs) - set(base_kwargs)
+        if invalid_kwargs:
+            raise TypeError('invalid param(s): %s' % sorted(invalid_kwargs))
+
+        base_kwargs.update(kwargs)
+        return type(self)(**base_kwargs)
+
+    def __iter__(self):
+        yield self.expr
```
ssanderson: Is this for backwards compat with `*`-unpacking?

llllllllll: we do `expr, deltas, checkpoints, odo_kwargs = ...` in a few places.

PR #2438, opened 2019-03-04
ENH: Parametrize load_adjustments_from_sqlite to selectively load adjustments tables.

ernestoeperez88:
Allows us to exclude any adjustment tables when loading `Adjustment` objects from adjustments.db.

### DIFF
--- a/zipline/data/_adjustments.pyx
+++ b/zipline/data/_adjustments.pyx
@@ -164,7 +164,8 @@ cpdef load_adjustments_from_sqlite(object adjustments_db,
                                    Int64Index_t assets,
                                    bool should_include_splits,
                                    bool should_include_mergers,
-                                   bool should_include_dividends):
+                                   bool should_include_dividends,
+                                   str adjustment_type):
     """
     Load a dictionary of Adjustment objects from adjustments_db.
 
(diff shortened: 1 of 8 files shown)


zipline/data/_adjustments.pyx
```diff
@@ -136,60 +145,85 @@ cdef _adjustments(object adjustments_db,
...
+        Whether merger adjustments should be included.
+    should_include_dividends : bool
+        Whether dividend adjustments should be included.
 
     Returns
     -------
-    adjustments : list[dict[int -> Adjustment]]
-        A list of mappings from index to adjustment objects to apply at that
-        index.
+    adjustments : tuple[dict[int -> Adjustment]]
```
llllllllll: Can we add a flag to control whether this returns price-type, volume-type, or both? If we load all columns except for `volume`, we don't need to construct this dict.

zipline/data/_adjustments.pyx
```diff
@@ -234,64 +291,50 @@ cpdef load_adjustments_from_sqlite(object adjustments_db,  # sqlite3.Connection
...
+            volume_adj = Float64Multiply(
+                0, date_loc, asset_ix, asset_ix, 1.0 / ratio
+            )
+            volume_adjustments.setdefault(date_loc, []).append(volume_adj)
 
-        if not PyDict_Contains(asset_ixs, sid):
-            asset_ixs[sid] = assets.get_loc(sid)
-        asset_ix = asset_ixs[sid]
+    # mergers and dividends affect prices only
+    if should_include_price_adjustments:
```
llllllllll: We can add this without an extra branch by setting `should_include_mergers/dividends = False` if `should_include_price_adjustments` is False.

zipline/data/_adjustments.pyx
```diff
@@ -234,64 +291,50 @@ cpdef load_adjustments_from_sqlite(object adjustments_db,  # sqlite3.Connection
...
+                asset_ixs[sid] = assets.get_loc(sid)
+            asset_ix = asset_ixs[sid]
 
-        if not PyDict_Contains(asset_ixs, sid):
-            asset_ixs[sid] = assets.get_loc(sid)
-        asset_ix = asset_ixs[sid]
+            price_adj = Float64Multiply(0, date_loc, asset_ix, asset_ix, ratio)
+            price_adjustments.setdefault(date_loc, []).append(price_adj)
+
+        for sid, ratio, eff_date in dividends:
```
llllllllll: should we just chain the two sets together with `itertools.chain` to have the same loop body for both sets?

zipline/data/adjustments.py
```diff
@@ -112,13 +112,73 @@ def __exit__(self, *exc_info):
...
+            should_include_mergers,
+            should_include_dividends,
+            adjustment_type,
+        )
+
+    def load_pricing_adjustments(self, columns, dates, assets):
+        if 'volume' not in set(columns):
+            adjustment_type = 'P'
+        else:
+            if len(set(columns)) == 1:
```
llllllllll: this can turn into an `elif` and be unindented.

zipline/data/_adjustments.pyx
```diff
@@ -234,64 +291,50 @@ cpdef load_adjustments_from_sqlite(object adjustments_db,  # sqlite3.Connection
...
-        adj = Float64Multiply(0, date_loc, asset_ix, asset_ix, ratio)
-        for i, column in enumerate(columns):
-            col_adjustments = results[i]
-            if column != 'volume':
-                try:
-                    col_adjustments[date_loc].append(adj)
-                except KeyError:
-                    col_adjustments[date_loc] = [adj]
+    if should_include_price_adjustments:
+        result['price_adjustments'] = price_adjustments
```
llllllllll: should these names match the input arguments like `P` or `V`, or should we use these names as the inputs? Also, we could probably drop the `_adjustments` suffix because it is clear from context.

ernestoeperez88: I think using `price` and `volume` would be better, I'll make that change.

llllllllll:
LGTM!

PR #2440, opened 2019-03-06
ENH: Add append_adjustments method to AdjustedArray.

ernestoeperez88:
Allows us to append new sets of adjustments to `AdjustedArray.adjustments`.

### DIFF
--- a/tests/pipeline/test_adjusted_array.py
+++ b/tests/pipeline/test_adjusted_array.py
@@ -745,3 +745,42 @@ last_col=0, value=4.000000)]}
         # values and adjustment values.
         check_arrays(adj_array.data, expected_adj_array.data)
         self.assertEqual(adj_array.adjustments, expected_adj_array.adjustments)
+
+    @parameterized.expand([
+        # Append new adjustment to back of the list if index already contains
+        # adjustments.
+        (
+            True,
+            {
+                1: [Float64Multiply(0, 3, 0, 0, 4.0),
+                    Float64Overwrite(2, 3, 0, 0, 1.0)],
+                3: [Float64Multiply(0, 3, 0, 0, 4.0)],
+            },
+        ),
+        # Append new adjustment to front of the list if index already contains
+        # adjustments.
+        (
+            False,
+            {
+                1: [Float64Overwrite(2, 3, 0, 0, 1.0),
+                    Float64Multiply(0, 3, 0, 0, 4.0)],
+                3: [Float64Multiply(0, 3, 0, 0, 4.0)],
+            },
+        )
+    ])
+    def test_append_adjustments(self, append_back, expected_adjustments):
+        adjustments = {
+            1: [Float64Multiply(0, 3, 0, 0, 4.0)],
+        }
+        adjustments_to_append = {
+            1: [Float64Overwrite(2, 3, 0, 0, 1.0)],
+            3: [Float64Multiply(0, 3, 0, 0, 4.0)],
+        }
+
+        data = arange(30, dtype=float).reshape(6, 5)
+        adjusted_array = AdjustedArray(data, adjustments, float('nan'))
+
+        adjusted_array.append_adjustments(
+            adjustments_to_append, append_back=append_back
+        )
+        self.assertEqual(adjusted_array.adjustments, expected_adjustments)
--- a/tests/pipeline/test_adjusted_array.py
+++ b/tests/pipeline/test_adjusted_array.py
@@ -746,41 +746,62 @@ last_col=0, value=4.000000)]}
         check_arrays(adj_array.data, expected_adj_array.data)
         self.assertEqual(adj_array.adjustments, expected_adj_array.adjustments)
 
-    @parameterized.expand([
-        # Append new adjustment to back of the list if index already contains
-        # adjustments.
-        (
-            True,
-            {
-                1: [Float64Multiply(0, 3, 0, 0, 4.0),
-                    Float64Overwrite(2, 3, 0, 0, 1.0)],
-                3: [Float64Multiply(0, 3, 0, 0, 4.0)],
-            },
-        ),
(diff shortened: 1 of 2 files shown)


zipline/lib/adjusted_array.py
```diff
@@ -165,6 +166,46 @@ def __init__(self, data, adjustments, missing_value):
         self.adjustments = adjustments
         self.missing_value = missing_value
 
+    def append_adjustments(self, adjustments_to_append, append_back):
```
ssanderson: Two minor naming notes:

1. I would rename `adjustments_to_append` to just `adjustments`. It's clear from context what's happening to them.
2. The name `append` here is a bit confusing. "append" usually means "add to the end", so having a flag to say "append to front" or "append to back" sounds off to me. Something like `add`/`merge`/`update_adjustments` would make more sense to me here.

zipline/lib/adjusted_array.py
```diff
@@ -165,6 +166,46 @@ def __init__(self, data, adjustments, missing_value):
...
+        """
+        Adds ``adjustments_to_append`` to existing adjustments, handling
+        index collisions according to ``append_back``.
+
+        Parameters
+        ----------
+        adjustments_to_append : dict[int -> list[Adjustment]]
+            The mapping of row indices to lists of adjustments that should be
+            appended to existing adjustments.
+        append_back : bool
```
ssanderson: Related to the note above, `append_back` reads slightly off to me. One option would be to call this `append` and have `append=False` mean "prepend", though that's somewhat confusing b/c it could be read as "don't append at all". Another option might do something like `np.searchsorted` and have `side='front'` or `side='back'`.

tests/pipeline/test_adjusted_array.py
```diff
@@ -745,3 +745,42 @@ def test_update_labels(self):
...
+            },
+        )
+    ])
+    def test_append_adjustments(self, append_back, expected_adjustments):
+        adjustments = {
+            1: [Float64Multiply(0, 3, 0, 0, 4.0)],
+        }
+        adjustments_to_append = {
+            1: [Float64Overwrite(2, 3, 0, 0, 1.0)],
+            3: [Float64Multiply(0, 3, 0, 0, 4.0)],
```
ssanderson: Is there a reason we're using the same values for the multiplies here? That makes it theoretically possible for us to write a bug where the multiplies get swapped, but we don't notice because they're the same.

tests/pipeline/test_adjusted_array.py
```diff
@@ -745,3 +745,42 @@ def test_update_labels(self):
         # values and adjustment values.
         check_arrays(adj_array.data, expected_adj_array.data)
         self.assertEqual(adj_array.adjustments, expected_adj_array.adjustments)
+
+    @parameterized.expand([
```
ssanderson: This is a good start, but I think there are a few important cases we should probably cover:

1. Initial adjustments have a key not present in `adjustments_to_append`. It should still show up in the output.
2. Initial adjustments has more than one adjustment at a given index.
3. `adjustments_to_append` has more than one adjustment at a given index.

tests/pipeline/test_adjusted_array.py
```diff
@@ -745,3 +745,42 @@ def test_update_labels(self):
...
+        # Append new adjustment to back of the list if index already contains
+        # adjustments.
+        (
```
ssanderson: This test is pretty readable because the inputs are small and the behavior under test is pretty simple, but IMO this usage of `parameterized` makes it harder to follow, because `parameterized` adds more space between the expectation and the assertion, and it puts space between the expectation and the assertion.

If I were writing this, I would probably structure it as something like:

```
@parameterized.expand([
    (initial_adjustments, added_adjustments, expected_result_when_prepending, expected_result_when_appending)
])
```

That structure would put the inputs and expected outputs closer together in the file, and would also make it easier to test more complex cases without making everything more complex.

tests/pipeline/test_adjusted_array.py
```diff
@@ -745,3 +745,66 @@ def test_update_labels(self):
         # values and adjustment values.
         check_arrays(adj_array.data, expected_adj_array.data)
         self.assertEqual(adj_array.adjustments, expected_adj_array.adjustments)
+
+    @parameterized.expand([(
+        # Initial adjustments
+        {
+            1: [Float64Multiply(0, 4, 1, 1, 0.5),
```
ssanderson: One thing that might help clarify this further would be to bind these adjustments to variables, e.g.:

```
A = Float64Multiply(...)
B = Float64Overwrite(...)
```

Then your parameterized setup would look like:

```
# Initial
{1 : [A, B], 2: [C], 3: [D]},
# To add
{1: [E], 2: [F, G], 3: [H]},
# Expected with append.
...
# Expected with prepend.
```

The reason this might help is that the values in the adjustments themselves aren't really all that important here. What we care is just that, whatever the values in the adjustments, they get rearranged in the right way after merging.

tests/pipeline/test_adjusted_array.py
```diff
@@ -745,3 +745,63 @@ def test_update_labels(self):
...
+    B = Float64Overwrite(3, 3, 4, 4, 4.2)
+    C = Float64Multiply(0, 2, 0, 0, 0.14)
+    D = Float64Overwrite(0, 3, 0, 0, 4.0)
+    E = Float64Overwrite(0, 0, 1, 1, 3.7)
+    F = Float64Multiply(0, 4, 3, 3, 10.0)
+    G = Float64Overwrite(5, 5, 4, 4, 1.7)
+    H = Float64Multiply(0, 4, 2, 2, 0.99)
+    S = Float64Multiply(0, 1, 4, 4, 5.06)
+
+    @parameterized.expand([(
```
ssanderson: :+1: this is much clearer now!

PR #2443, opened 2019-03-12
Add recent version of Python to test matrix.

ehebert:


ssanderson:
This needs more work to happen. xref: https://github.com/quantopian/zipline/issues/2616

PR #2444, opened 2019-03-18
BUG: Ensure correct dtypes when unpacking empty dataframes.

ssanderson:
Fixes a bug where columns of adjustment db dataframes would come back as object
dtype when the table is empty.

### DIFF
--- a/zipline/__init__.py
+++ b/zipline/__init__.py
@@ -97,10 +97,9 @@ def setup(self,
           new_pandas=new_pandas):
     """Lives in zipline.__init__ for doctests."""
 
-    legacy_version = '1.13'
-    if numpy_version > StrictVersion(legacy_version):
+    if numpy_version >= StrictVersion('1.14'):
         self.old_opts = np.get_printoptions()
-        np.set_printoptions(legacy=legacy_version)
+        np.set_printoptions(legacy='1.13')
     else:
         self.old_opts = None
 
--- a/zipline/data/adjustments.py
+++ b/zipline/data/adjustments.py
@@ -74,6 +74,16 @@ SQLITE_STOCK_DIVIDEND_PAYOUT_COLUMN_DTYPES = {
 }
 
 
+def specialize_any_integer(d):
+    out = {}
+    for k, v in six.iteritems(d):
+        if v is any_integer:
+            out[k] = int64_dtype
+        else:
+            out[k] = v
+    return out
+
+
 class SQLiteAdjustmentReader(object):
     """
     Loads adjustments based on corporate actions from a SQLite database.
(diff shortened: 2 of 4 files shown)


zipline/__init__.py
```diff
@@ -97,10 +97,9 @@ def setup(self,
           new_pandas=new_pandas):
     """Lives in zipline.__init__ for doctests."""
 
-    legacy_version = '1.13'
-    if numpy_version > StrictVersion(legacy_version):
+    if numpy_version >= StrictVersion('1.14'):
         self.old_opts = np.get_printoptions()
-        np.set_printoptions(legacy=legacy_version)
+        np.set_printoptions(legacy='1.13')
```
ssanderson: The `legacy=` keyword arg doesn't appear until 1.14, so this test was failing for me when trying to test with 1.13.3.

zipline/data/adjustments.py
```diff
@@ -283,13 +292,29 @@ def get_df_from_table(self, table_name, convert_dates=False):
...
+    def _df_dtypes(self, table_name, convert_dates):
+        """Get dtypes to use when unpacking sqlite tables as dataframes.
+        """
+        out = self._raw_table_dtypes[table_name]
+        if convert_dates:
+            out = out.copy()
+            for date_column in self._datetime_int_cols[table_name]:
+                out[date_column] = datetime64ns_dtype
+
+        return out
```
ernestoeperez88: linter expects an extra blank line below.

zipline/data/adjustments.py
```diff
@@ -283,13 +292,29 @@ def get_df_from_table(self, table_name, convert_dates=False):
...
 
-        return pd.read_sql(
+        result = pd.read_sql(
             'select * from "{}"'.format(table_name),
             self.conn,
             index_col='index',
             **kwargs
         ).rename_axis(None)
 
+        if not len(result):
```
ernestoeperez88: TIL `len(result)` is faster than `result.empty`. It seems `result.empty` performs extra checks we might not care about in this case.

Slightly faster, we could do `len(result.index)` instead.

ssanderson: `empty` will be true if either rows or columns are empty. In this case, we only care about empty rows, so I think `len` is preferable (we never expect no columns, so they should have the same behavior here, but `len` expresses what we actually expect here).

re: `len(df.index)`, the implementation of `DataFrame.__len__` is just:

```
In [5]: pd.DataFrame.__len__??
Signature: pd.DataFrame.__len__(self)
Source:
    def __len__(self):
        """Returns length of info axis, but here we use the index """
        return len(self.index)
```

Adding an extra function call here isn't going to make a meaningful difference relative to the other things happening in this function, so I'd prefer the simpler expression here. If we were concerned with that level of micro performance here, much more important would be things like caching the column order produced by `keysorted`.

PR #2450, opened 2019-04-05
ENH: Update assert_equal(ndarray) to handle datetime arrays.

ernestoeperez88:
`numpy.testing.assert_array_equal` (`numpy==1.11.3`) doesn't handle datetime arrays correctly if either of the inputs contain `NaT`s.

To get around that, we can cast the arrays to ints using `view('int64')` and compare them. This is possible because 'NaT' maps to a reserved int.

### DIFF
--- a/zipline/testing/predicates.py
+++ b/zipline/testing/predicates.py
@@ -54,6 +54,7 @@ from zipline.utils.compat import getargspec, mappingproxy
 from zipline.utils.formatting import s
 from zipline.utils.functional import dzip_exact, instance
 from zipline.utils.math_utils import tolerant_equals
+from zipline.utils.numpy_utils import compare_datetime_arrays
 
 
 @instance
@@ -547,11 +548,29 @@ def assert_array_equal(result,
                        array_verbose=True,
                        array_decimal=None,
                        **kwargs):
-    f = (
-        np.testing.assert_array_equal
-        if array_decimal is None else
-        partial(np.testing.assert_array_almost_equal, decimal=array_decimal)
-    )
+    result_dtype = result.dtype
+    expected_dtype = expected.dtype
+
+    if result_dtype.kind in 'mM' and expected_dtype.kind in 'mM':
+        assert result_dtype == expected_dtype, (
+            "\nType mismatch:\n\n"
+            "result dtype: %s\n"
+            "expected dtype: %s\n%s"
+            % (result_dtype, expected_dtype, _fmt_path(path))
+        )
+        f = partial(
+            np.testing.utils.assert_array_compare,
+            compare_datetime_arrays,
+            header='Arrays are not equal',
+        )
+    elif array_decimal is not None:
+        f = partial(
+            np.testing.assert_array_almost_equal,
+            decimal=array_decimal,
+        )
+    else:
+        f = np.testing.assert_array_equal
+
     try:
         f(
             result,
--- a/zipline/utils/numpy_utils.py
+++ b/zipline/utils/numpy_utils.py
@@ -498,7 +498,7 @@ def changed_locations(a, include_first):
 
 def compare_datetime_arrays(x, y):
     """
-    Compare datetime64 ndarrays in a NaT-aware fashion.
+    Compare datetime64 ndarrays, treating NaT values as equal.
     """
 
     return array_equal(x.view('int64'), y.view('int64'))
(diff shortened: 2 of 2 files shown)


zipline/testing/predicates.py
```diff
@@ -552,6 +552,14 @@ def assert_array_equal(result,
         if array_decimal is None else
         partial(np.testing.assert_array_almost_equal, decimal=array_decimal)
     )
+
+    if result.dtype.kind == 'M' and expected.dtype.kind == 'M':
```
ernestoeperez88: if only one of the arrays has 'M' dtype, we can let `assert_array_equal` handle the mismatch.

zipline/testing/predicates.py
```diff
@@ -560,6 +568,18 @@ def assert_array_equal(result,
             err_msg=msg,
         )
     except AssertionError as e:
+        if is_datetime:
+            raise AssertionError(
```
ernestoeperez88: if the arrays have datetime dtype and their values do not match, we need to cast them back to datetime64 to provide a coherent assertion error.

zipline/testing/predicates.py
ssanderson: I would either use `.view` here or call `astype` with `copy=False`. The default behavior for `astype` is to copy the data, which is unnecessary here.

zipline/testing/predicates.py
```diff
@@ -552,6 +552,13 @@ def assert_array_equal(result,
         if array_decimal is None else
         partial(np.testing.assert_array_almost_equal, decimal=array_decimal)
     )
+
+    is_datetime = False
+    if result.dtype.kind == 'M' and expected.dtype.kind == 'M':
```
ssanderson: You probably also want to handle `'m'` (timedelta) the same way. I'd write that check as:

```
if result.dtype.kind in 'mM' and expected.dtype == result.dtype:
```

zipline/testing/predicates.py
```diff
@@ -552,6 +552,13 @@ def assert_array_equal(result,
...
+        is_datetime = True
+        result = result.astype('int64')
+        expected = expected.astype('int64')
```
ssanderson: Hmm. One thing we might want to handle here is if result and expected are both datetime dtype, but have different units. I'm not sure if we'd want those to compare the same.

zipline/testing/predicates.py
```diff
@@ -560,6 +567,18 @@ def assert_array_equal(result,
             err_msg=msg,
         )
     except AssertionError as e:
+        if is_datetime:
```
ssanderson: Hmm. this feels kinda gross. I feel like it might be cleaner to just delegate to an `assert_datetime_array_equal` method so that we don't have to keep the flag around.

ssanderson: (I think this comment is also moot if we use `assert_array_compare`)

zipline/utils/numpy_utils.py
```diff
@@ -493,3 +494,11 @@ def changed_locations(a, include_first):
...
 
     return hstack([[0], indices])
+
+
+def compare_datetime_arrays(x, y):
+    """
+    Compare datetime64 ndarrays in a NaT-aware fashion.
+    """
+
+    return array_equal(x.view('int64'), y.view('int64'))
```
ssanderson: Is there a reason to use `array_equal` over `==` here?

zipline/utils/numpy_utils.py
```diff
@@ -493,3 +494,11 @@ def changed_locations(a, include_first):
         return indices
 
     return hstack([[0], indices])
+
+
+def compare_datetime_arrays(x, y):
+    """
+    Compare datetime64 ndarrays in a NaT-aware fashion.
```
ssanderson: I would probably note specifically that this treats `NaT` values as equal.

zipline/utils/numpy_utils.py
ernestoeperez88: `array_equal` also checks for matching shapes. If we use `==` we could run into an issue where arrays of different length containing the same value compare as equal:
```
In [18]: a = np.array(['2014-01-01', '2014-01-01'], dtype='M8[ns]')

In [19]: b = np.array(['2014-01-01'], dtype='M8[ns]')

In [20]: a.view('int64') == b.view('int64')
Out[20]: array([ True,  True], dtype=bool)
```

PR #2456, opened 2019-04-17
ENH: Enable abstract DataSetFamily subclasses

quantophred:
This change enables us to define "abstract" subclasses of DataSetFamily
which implement some set of attributes, but don't represent tangible
families of DataSets, and thus don't have meaningful values for
`extra_dims`.

Previously, all subclasses of DataSetFamily had to define non-empty
`extra_dims`. Subclasses were detected in DataSetFamilyMeta by checking
if the base classes of the class under construction were specified as
anything other than the designated sentinel value: if so, `extra_dims`
was transformed to an OrderedDict and checked for non-emptiness.

With this change, the metaclass instead checks for an `_abstract`
attribute of the class under construction: if it is found and set to a
truthy value, `extra_dims` (and `_SliceType`) is left unchecked and
unmodified. Regardless of its value, the `_abstract` attribute is
removed from the constructed class' body: each subclass which wishes to
remain abstract must specify this attribute again.

### DIFF
--- a/tests/pipeline/test_multidimensional_dataset.py
+++ b/tests/pipeline/test_multidimensional_dataset.py
@@ -251,7 +251,7 @@ class TestDataSetFamily(ZiplineTestCase):
             ),
         )
 
-    def test_inheritence(self):
+    def test_inheritance(self):
         class Parent(DataSetFamily):
             extra_dims = [
                 ('dim_0', {'a', 'b', 'c'}),
--- a/tests/pipeline/test_multidimensional_dataset.py
+++ b/tests/pipeline/test_multidimensional_dataset.py
@@ -62,7 +62,7 @@ class TestDataSetFamily(ZiplineTestCase):
     def test_empty_extra_dims(self):
         msg = (
             "DataSetFamily must be defined with non-empty extra_dims,"
-            " or with `__abstract__ = True`"
+            " or with `_abstract = True`"
         )
         with assert_raises_str(ValueError, msg):
             class NoExtraDims(DataSetFamily):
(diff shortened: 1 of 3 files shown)


zipline/pipeline/data/dataset.py
```diff
@@ -609,25 +608,24 @@ def __new__(cls, name, bases, dict_):
                 columns[k] = v
                 dict_[k] = _DataSetFamilyColumn(k)
 
-        is_base_class = bases == (cls._base_marker,)
-        if is_base_class:
-            bases = (object,)
+        is_abstract = dict_.pop('__abstract__', False)
```
llllllllll: `__{}__` is reserved for the implementation of Python. I would assume this interacts with `ABCMeta, not part of our code. We should use something like `_abstract`.

tests/pipeline/test_multidimensional_dataset.py
```diff
@@ -60,13 +60,38 @@ class MD2(DataSetFamily):
...
     def test_empty_extra_dims(self):
-        expected_msg = (
-            'DataSetFamily must be defined with non-empty extra_dims'
+        msg = (
+            "^DataSetFamily must be defined with non-empty extra_dims,"
+            " or with `__abstract__ = True`$"
         )
-        with assert_raises_str(ValueError, expected_msg):
-            class MD(DataSetFamily):
+        with self.assertRaisesRegexp(ValueError, msg):
```
llllllllll: why the change to a regex with no pattern?

quantophred: Didn't see that we had an `assert_raises_str`. I'll use that instead. Thanks!

PR #2459, opened 2019-04-19
ENH: DataSetFamily namedtuple coords

quantophred:
This PR adds a `_CoordType` attribute to DataSetFamily classes, which is
a namedtuple whose fields are the extra_dims.

namedtuples have a number of advantages over OrderedDicts for this
purpose:

- Default equality comparison is based on values, but may be easily
  extended to compare types as well;
- They are hashable, removing the need for a separate hash key;
- The constructor provides well-tested argument validation;
- Instances' repr includes the class name, giving some indication of the
  object's intended purpose;
- The fields are stored on the type, rather than redundantly on every
  instance;
- getattr syntax (if you're into that).

Moreover, this change also lets us factor out the `_validate_coords`
method. Currently, subclasses which override `slice` may choose to spell
out the expected extra_dims in the signature, but this will replace our
helpful dimension-specific error messages with Python's less helpful
built-in ones, such as "slice() takes 3 positional arguments but 4 were
given".

Factoring out `_validate_coords` enables subclasses to define their own
validation logic for coord objects while maintaining the original slice
signature, and without having to re-parse `(*args, **kwargs)`.

### DIFF
--- a/zipline/pipeline/data/dataset.py
+++ b/zipline/pipeline/data/dataset.py
@@ -638,9 +638,9 @@ class DataSetFamilyMeta(abc.ABCMeta):
 
             BaseSlice.__name__ = '%sBaseSlice' % self.__name__
             self._SliceType = BaseSlice
+            # each non-abstract subclass gets a unique cache
+            self._slice_cache = {}
 
-        # each type gets a unique cache
-        self._slice_cache = {}
         return self
 
     def __repr__(self):
--- a/zipline/pipeline/data/dataset.py
+++ b/zipline/pipeline/data/dataset.py
@@ -1,7 +1,6 @@
 import abc
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 from functools import total_ordering
-from itertools import repeat
 from textwrap import dedent
 from weakref import WeakKeyDictionary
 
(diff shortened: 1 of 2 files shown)


quantophred:
Per discussion with Scott, we might not actually care about dimension-specific error messages: it might even be better to give more standard, searchable messages.

In that case, we could either close this PR, or could remove `_diagnose_slice_error` (which contains most of the code formerly from `_canonical_key`) and rely on the namedtuple constructor to raise, e.g., `TypeError: __new__() takes exactly 3 arguments (4 given)`.

In either case, our expectation is that most base classes will probably redefine `slice` anyway in order to spell out the expected arguments and add a docstring. If that expectation is not correct, we should probably keep our argument parsing code, either by closing this PR or merging it as-is.

zipline/pipeline/data/dataset.py
```diff
@@ -597,7 +596,7 @@ def __get__(self, instance, owner):
 
 class DataSetFamilyMeta(abc.ABCMeta):
 
-    def __new__(cls, name, bases, dict_):
+    def __new__(meta, name, bases, dict_):
```
llllllllll: I think this rename makes it less clear what "cls" is. `cls` is normally the name of the argument passed to a classmethod or `__new__`, I don't think being a metaclass should change that. At least in this part of the code, and with pipeline datasets in general, the `cls` value is acting more like an instance than a class.

quantophred:
Gotcha. If you don't have any strong feelings about subclasses' slice methods reverting to Python's built-in TypeError messages, I'd say we should keep this code as is.

PR #2475, opened 2019-06-03
Fix bad reprs

ssanderson:


zipline/pipeline/hooks/progress.py
```diff
@@ -460,3 +460,16 @@ def publish(self, model):
...
+
+def repr_htmlsafe(t):
+    """Repr a value and html-escape the result.
+
+    If an error is thrown by the repr, show a placeholder.
+    """
+    try:
+        r = repr(t)
+    except Exception:
+        r = "(Error Displaying {})".format(type(t).__name__)
```
quantophred: Looks like you fixed a bunch of possibilities for this in the previous commit. Is it worth logging an error here so we can catch new ones if/when they come up?

ssanderson: This could be triggered by user code in enough ways that I don't think we'd want to be alerted if we hit this code path.

PR #2486, opened 2019-06-24
PERF: Speed up pipeline output index construction.

ssanderson:
For a 3 year US Equities pipeline, this shaves about .5 seconds off a 4 second
pipeline execution for me.

Before:

```
        1    0.203    0.203    0.728    0.728 engine.py:716(_to_narrow)
```

After:

```
        1    0.128    0.128    0.283    0.283 engine.py:716(_to_narrow)
```

### DIFF
--- a/zipline/pipeline/engine.py
+++ b/zipline/pipeline/engine.py
@@ -59,7 +59,7 @@ from abc import ABCMeta, abstractmethod
 from functools import partial
 
 from six import iteritems, with_metaclass, viewkeys
-from numpy import array
+from numpy import array, arange
 from pandas import DataFrame, MultiIndex
 from toolz import groupby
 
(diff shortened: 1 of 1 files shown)


zipline/pipeline/engine.py
```diff
@@ -881,3 +877,34 @@ def _resolve_hooks(self, hooks):
...
+    index : pd.MultiIndex
+        MultiIndex  containing (date,  asset) pairs  corresponding to  ``True``
+        values in ``mask``.
+    """
+    date_labels = repeat_last_axis(arange(len(dates)), len(assets))[mask]
+    asset_labels = repeat_first_axis(arange(len(assets)), len(dates))[mask]
+    return MultiIndex(
+        levels=[dates, assets],
+        labels=[date_labels, asset_labels],
+        # TODO: We should probably add names for these.
```
ssanderson: We might want to consider putting `assets` and `dates` as the column labels here, but I didn't want to do that in this change to minimize functional changes. (We also have some test suites that break with that change.)

PR #2506, opened 2019-07-18
ENH: Added a PercentChange factor with unit test

peterhbromley:


zipline/pipeline/factors/basic.py
```diff
@@ -54,6 +54,28 @@ def compute(self, today, assets, out, close):
         out[:] = (close[-1] - close[0]) / close[0]
 
 
+class PercentChange(CustomFactor):
+    """
+    Calculates the percent change in the input value over the given window_length.
+    Note: Uses the formula "new - old / abs(old)" to handle negative input values
+
+    **Default Inputs:** None
```
dmichalowicz: Looks like other factors mention `**Default Window Length**: None`, might be worth adding that here too.

tests/pipeline/test_factor.py
```diff
@@ -17,6 +17,7 @@
     eye,
     log1p,
     nan,
+    inf,
```
dmichalowicz: Small note, it's helpful to keep imports in alphabetical order.

zipline/pipeline/factors/basic.py
```diff
@@ -54,6 +54,28 @@ def compute(self, today, assets, out, close):
...
+    """
+    window_safe = True
+
+    def _validate(self):
+        super(PercentChange, self)._validate()
+        if self.window_length < 2:
+            raise ValueError(
+                "'PercentChange' expected a window length of at least 2, but was "
+                "given {window_length}. For daily percent change, use a window "
```
dmichalowicz: Looks like the [build failed ](https://ci.appveyor.com/project/Quantopian/zipline/builds/26082715/job/i0ylb6hyd414mcxm) due to this line and a couple others being too long. If you run flake8 on this file it will tell you which ones, but it requires a maximum of 80 characters per line. In this case you can just shorten these strings in each line.

tests/pipeline/test_factor.py
```diff
@@ -585,6 +586,34 @@ def test_returns(self, seed_value, window_length):
...
+            window_length=window_length
+        )
+
+        today = datetime64(1, 'ns')
+        assets = arange(7)
+
+        seed(seed_value)  # Seed so we get deterministic results.
+        test_data = randn(window_length, 7)
+        test_data[0] = array([1, 2, 1, -1, -1, 0, nan])
+        test_data[-1] = array([2, 1, -2, 2, -2, 1, 1])
```
dmichalowicz: Maybe add a 0% change case?

tests/pipeline/test_factor.py
```diff
@@ -585,6 +586,34 @@ def test_returns(self, seed_value, window_length):
...
+
+        # Calculate the expected percent change
+        expected = array([1, -0.5, -3, 3, -1, inf, nan])
+
+        out = empty((7,), dtype=float)
+        pct_change.compute(today, assets, out, test_data)
+
+        check_allclose(expected, out)
+
```
dmichalowicz: Another case you could test for is passing a `window_length` of 1 and making sure an exception is raised. This can be done with something like:

```python
with self.assertRaises(ValueError):
    PercentChange(inputs=(), window_length=1)
```

tests/pipeline/test_factor.py
```diff
@@ -585,6 +586,34 @@ def test_returns(self, seed_value, window_length):
...
+    @parameterized.expand([
+        (100, 15),
+        (101, 4),
+        (102, 100),
+        ])
+    def test_percentchange(self, seed_value, window_length):
+
+        pct_change = PercentChange(
+            inputs=(), 
+            window_length=window_length
```
dmichalowicz: Another small style note, I like to leave trailing commas at the end of multiline calls of functions, import lines, or lists so that if a new parameter ever got added the git diff would only be +1 line instead of -1/+2 lines. I also tend to think it looks nicer 🙂

tests/pipeline/test_factor.py
```diff
@@ -585,6 +586,37 @@ def test_returns(self, seed_value, window_length):
...
 
+    @parameterized.expand([
+        (100, 15),
+        (101, 4),
+        (102, 100),
+        ])
+    def test_percentchange(self, seed_value, window_length):
+
+        pct_change = PercentChange(
+            inputs=(),
```
ssanderson: I think we still probably want a length-1 list of inputs here rather than an empty tuple. Ideally, we'd use `SingleInputMixin` to require exactly one input.

peterhbromley: Oh right, I forgot about that.  Will make that change now

tests/pipeline/test_factor.py
```diff
@@ -585,6 +586,37 @@ def test_returns(self, seed_value, window_length):
...
+        # Calculate the expected percent change
+        expected = array([1, -0.5, 0, -3, 3, -1, inf, nan])
+
+        out = empty((8,), dtype=float)
+        pct_change.compute(today, assets, out, test_data)
+
+        check_allclose(expected, out)
+
+        with self.assertRaises(ValueError):
+            PercentChange(inputs=(), window_length=1)
```
ssanderson: If we do https://github.com/quantopian/zipline/pull/2506/files#r305114758, we'll probably want to update this check to separately check empty inputs and too-short window_length (as-is, this check will pass, but will only verify whichever of the checks happens to happen first).

zipline/pipeline/factors/basic.py
```diff
@@ -54,6 +54,30 @@ def compute(self, today, assets, out, close):
         out[:] = (close[-1] - close[0]) / close[0]
 
 
+class PercentChange(SingleInputMixin, CustomFactor):
+    """
+    Calculates the percent change over the given window_length.
+    Note: Uses formula "new - old / abs(old)" for negative input values
```
ssanderson: Minor point, but this sentence could be read as saying that this is only the formula for negative input values (leaving it unclear what the formula is otherwise). I would probably just put something like. "Percent change is calculated as ``(new - old) / abs(old)``

ssanderson: Regarding the "Note:" qualifier on this, we generally follow https://numpydoc.readthedocs.io/en/latest/format.html for our docstring convention, which among other things allows us to auto-generate relatively nice-looking documentation from these docstrings using [Sphinx](https://www.sphinx-doc.org/en/master/). If you want to add "notes", to a docstring, the usual convention would be to add a section that looks like:

```
Notes
-----
Your content here.
```

Normally we would also document with a `Parameters` section (you'll see lots of examples of those around elsewhere in the codebase), but we don't do so for pipeline expressions b/c they have a bit of a weird construction model (related to all the `__new__`s floating around in the base classes).

tests/pipeline/test_factor.py
```diff
@@ -585,6 +586,40 @@ def test_returns(self, seed_value, window_length):
...
+        pct_change = PercentChange(
+            inputs=[EquityPricing.close, ],
+            window_length=window_length,
+        )
+
+        today = datetime64(1, 'ns')
+        assets = arange(8)
+
+        seed(seed_value)  # Seed so we get deterministic results.
+        test_data = randn(window_length, 8)
```
ssanderson: Another, more "functional" way to write this would be to use [np.concatenate](https://docs.scipy.org/doc/numpy/reference/generated/numpy.concatenate.html) or [np.vstack](https://docs.scipy.org/doc/numpy/reference/generated/numpy.vstack.html#numpy.vstack) to stack the first row, the middle row, and the end row together. (This is fine as is, but in other contexts concat might be clearer than updating the array in place.)

zipline/pipeline/factors/basic.py
```diff
@@ -54,6 +54,34 @@ def compute(self, today, assets, out, close):
...
+    """
+    Calculates the percent change over the given window_length.
+
+    **Default Inputs:** None
+
+    **Default Window Length:** None
+
+    Notes
+    -----
+    Percent change is calculated as (new - old) / abs(old).
```
ssanderson:
```suggestion
    Percent change is calculated as ``(new - old) / abs(old)``.
```

This will make the formula render in monospace in our generated docs.

PR #2507, opened 2019-07-19
BUG: Fixes bug with the mixins that was preventing validation from occuring

peterhbromley:


peterhbromley:
Adding most recent commit to previous PR

PR #2509, opened 2019-07-25
ENH: Added PeerCount custom factor returned by classifier method peer_count()

peterhbromley:


zipline/pipeline/factors/basic.py
```diff
@@ -489,6 +491,38 @@ def compute(self, today, assets, out, returns, annualization_factor):
...
+        super(PeerCount, self)._validate()
+        if self.window_length != 1:
+            raise ValueError(
+                "'PeerCount' expected a window length of 1, but was given"
+                "{window_length}.".format(window_length=self.window_length)
+            )
+
+    def compute(self, today, assets, out, arrays):
+        # Convert arrays to group label int arrays
+        group_labels, _ = self.inputs[0]._to_integral(arrays[0])
```
dmichalowicz: Not your doing, but I'm surprised to see `_to_integral` as a private method. I suppose we don't expect users to ever call it on classifiers in an actual algorithm.

zipline/pipeline/factors/basic.py
```diff
@@ -489,6 +491,38 @@ def compute(self, today, assets, out, returns, annualization_factor):
...
+
+    def _validate(self):
+        super(PeerCount, self)._validate()
+        if self.window_length != 1:
+            raise ValueError(
+                "'PeerCount' expected a window length of 1, but was given"
+                "{window_length}.".format(window_length=self.window_length)
+            )
+
+    def compute(self, today, assets, out, arrays):
```
dmichalowicz: What do you think of calling `arrays` something more like `classifier_values` or `labels`?

zipline/pipeline/factors/basic.py
```diff
@@ -489,6 +491,38 @@ def compute(self, today, assets, out, returns, annualization_factor):
...
+
+    **Default Window Length:** 1
+    """
+    window_length = 1
+    missing_value = -1
+    dtype = int64_dtype
+
+    def _validate(self):
+        super(PeerCount, self)._validate()
+        if self.window_length != 1:
```
dmichalowicz: Should we also enforce that `inputs` is length 1?

zipline/pipeline/factors/basic.py
```diff
@@ -489,6 +491,38 @@ def compute(self, today, assets, out, returns, annualization_factor):
...
+            raise ValueError(
+                "'PeerCount' expected a window length of 1, but was given"
+                "{window_length}.".format(window_length=self.window_length)
+            )
+
+    def compute(self, today, assets, out, arrays):
+        # Convert arrays to group label int arrays
+        group_labels, _ = self.inputs[0]._to_integral(arrays[0])
+        _, inverse, counts = unique(  # Get counts, idx of unique groups
+            group_labels,
```
dmichalowicz: Does `unique` not just work on the input array itself?

dmichalowicz: Oh I see, looks like it doesn't work for LabelArray.

tests/pipeline/test_classifier.py
```diff
@@ -607,6 +607,48 @@ class C(Classifier):
...
+                missing_value=None,
+            )
+
+        expected = np.array(
+            [[1, 3, 3, 2, 2, 3, 2, 1, 2]],
+            dtype=int64_dtype
+        )
+
+        terms = {
+            'peer_counts': c.peer_count(),
```
dmichalowicz: I wonder if it's also worth asserting that the result of constructing a `PeerCount` explicitly also matches the expected results? Probably not, but figured I'd throw that idea out there.

zipline/pipeline/factors/basic.py
peterhbromley: Yeah, I've only seen it used two other places and both were in ._compute() functions.  Not sure if it is bad practice for me to call a private method in CustomFactor's .compute()?

peterhbromley: I could also hand compute _to_integral() pretty easily in the compute() function

zipline/pipeline/factors/basic.py
peterhbromley: Definitely, changing it to classifier_values

zipline/pipeline/factors/basic.py
peterhbromley: I think this is currently enforced because it inherits from SingleInputMixin

tests/pipeline/test_classifier.py
peterhbromley: Good idea, I think I could do that pretty easily

zipline/pipeline/factors/basic.py
dmichalowicz: Oh cool, I missed that.

zipline/pipeline/factors/basic.py
dmichalowicz: I think given its uses elsewhere, what you have here is fine.

tests/pipeline/test_classifier.py
peterhbromley: Decided to not incorporate that in the test since atm I'm thinking PeerCount should only get accessed by a .peer_count() call.  Could very easily add the test, probably would want to put it in test_factor though.

dmichalowicz: Ok, considering `peer_count` is just a thin wrapper around `PeerCount` anyway, the coverage is already here.

PR #2515, opened 2019-07-30
BUG: Fixed behavior of peer_count() when called on missing data

peterhbromley:


zipline/pipeline/factors/basic.py
```diff
@@ -501,8 +502,7 @@ class PeerCount(SingleInputMixin, CustomFactor):
     **Default Window Length:** 1
     """
     window_length = 1
-    missing_value = -1
-    dtype = int64_dtype
+    missing_value = nan
```
ssanderson: missing value of nan is the default for factors with dtype float64 (which is the default dtype for factors), so we shouldn't need this.

zipline/pipeline/classifiers/classifier.py
```diff
@@ -400,14 +400,36 @@ def _to_integral(self, output_array):
 
     def peer_count(self):
         """
-        Construct a PeerCount factor that gives the number of occurrences of
-        each distinct category in a classifier
+        Construct a factor that gives the number of occurrences of
```
ssanderson: :+1: this reads very clearly.

zipline/pipeline/factors/basic.py
```diff
@@ -514,13 +514,19 @@ def _validate(self):
...
+        group_labels, null_label = self.inputs[0]._to_integral(
+            classifier_values[0]
+        )
         _, inverse, counts = unique(  # Get counts, idx of unique groups
             group_labels,
             return_counts=True,
             return_inverse=True,
         )
-        out[:] = counts[inverse]
+        out[:] = where(
```
ssanderson: I think slightly more efficient here would be:

```python
np.copyto(out, counts[inverse], group_labels != null_label)
```

```
In [9]: %timeit a[mask] = cp[mask]
The slowest run took 5.88 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 44.6 µs per loop

In [10]: %timeit np.copyto(a, cp, where=mask)
The slowest run took 5.96 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 24.2 µs per loop
```

PR #2537, opened 2019-09-11
ENH: BoundColumn comparisons now result in an error

peterhbromley:
Added rich comparison ordering methods to BoundColumn class to prevent users from comparing BoundColumn with anything.  Mainly to prevent the case where users accidentally write `EquityPricing.volume > 1000` instead of `EquityPricing.volume.latest > 1000`.

### DIFF
--- a/tests/pipeline/test_column.py
+++ b/tests/pipeline/test_column.py
@@ -99,5 +99,3 @@ class LatestTestCase(WithSeededRandomPipelineEngine,
 
             with self.assertRaises(BoundColumnInvalidCompare):
                 compare(column, 'test')
-
-            
--- a/tests/pipeline/test_column.py
+++ b/tests/pipeline/test_column.py
@@ -10,6 +10,7 @@ from pandas.util.testing import assert_frame_equal
 
 from zipline.lib.labelarray import LabelArray
 from zipline.pipeline import Pipeline
+from zipline.pipeline.data import USEquityPricing
 from zipline.pipeline.data.testing import TestingDataSet as TDS
 from zipline.pipeline.domain import US_EQUITIES
 from zipline.testing.fixtures import (
@@ -91,23 +92,33 @@ class LatestTestCase(WithSeededRandomPipelineEngine,
         (operator.le,),
     ])
     def test_comparison_errors(self, op):
-        columns = TDS.columns
-
-        def get_error_msg(op, compare_type):
-            return (
-                "'{op}' not supported between instance of "
-                "'{compare_type}' and '.*'. "
-                "Did you mean use '.latest' with '.*'?"
-            ).format(op=op, compare_type=compare_type)
-
-        ops = {'gt': '>', 'ge': '>=', 'lt': '<', 'le': '<='}
-
-        for column in columns:
-            int_err_msg = get_error_msg(ops[op.__name__], 'int')
-            str_err_msg = get_error_msg(ops[op.__name__], 'str')
-
-            with self.assertRaisesRegexp(TypeError, int_err_msg):
+        for column in TDS.columns:
+            with self.assertRaises(TypeError):
                 op(column, 1000)
-
-            with self.assertRaisesRegexp(TypeError, str_err_msg):
+            with self.assertRaises(TypeError):
+                op(1000, column)
+            with self.assertRaises(TypeError):
                 op(column, 'test')
+            with self.assertRaises(TypeError):
+                op('test', column)
+
+    def test_comparison_error_message(self):
+        column = USEquityPricing.volume
+        err_msg = (
+            "'<' not supported between instance of"
+            " 'int' and 'EquityPricing<US>.volume'."
+            " Did you mean to use 'EquityPricing<US>.volume.latest'?"
+        )
+
+        with self.assertRaises(TypeError) as e:
+            column < 1000
+            self.assertEqual(str(e), err_msg)
+
+        with self.assertRaises(TypeError) as e:
+            1000 < column
+            self.assertEqual(str(e), err_msg)
+
+        try:
+            column.latest < 1000
+        except Exception:
+            self.fail()
--- a/tests/pipeline/test_column.py
+++ b/tests/pipeline/test_column.py
@@ -112,11 +112,11 @@ class LatestTestCase(WithSeededRandomPipelineEngine,
 
         with self.assertRaises(TypeError) as e:
             column < 1000
-            self.assertEqual(str(e), err_msg)
+        self.assertEqual(str(e.exception), err_msg)
 
         with self.assertRaises(TypeError) as e:
-            1000 < column
-            self.assertEqual(str(e), err_msg)
+            1000 > column
+        self.assertEqual(str(e.exception), err_msg)
 
         try:
             column.latest < 1000
--- a/tests/pipeline/test_column.py
+++ b/tests/pipeline/test_column.py
@@ -1,11 +1,14 @@
 """
 Tests BoundColumn attributes and methods.
 """
+import operator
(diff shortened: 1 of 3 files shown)


zipline/errors.py
```diff
@@ -858,3 +858,14 @@ class IncompatibleTerms(ZiplineError):
...
     )
+
+
+class BoundColumnInvalidCompare(ZiplineError):
+    """
+    Raised when trying to use BoundColumn with a comparison operator.
+    Frequently occurs when users forget to use '.latest'.
+    """
+    msg = (
+        "Can't compare BoundColumn with {other}.  Did you mean to append "
```
ssanderson: Two notes on this:

1. I would probably render `type(other)` rather than `other` here. Objects, especially complex pipeline expressions, can have very large string representations, which tend to be less helpful in errors like this.
2. This message should probably include the `__str__` of the column being compared. Saying "You can't compare `EquityPricing.close` to an `int` is more likely to be helpful for a user than "You can't compare BoundColumns to ints." The existing error we raise on invalid inputs to `Pipeline` does a good job of this IMO.

peterhbromley: Sounds good.  Regarding 1., right now I am passing `type(other)` as the kwarg value to `BoundColumnInvalidCompare` when I raise the exception.  Is there a better way to do that?

Agreed on 2., fixing now.

ssanderson: > Is there a better way to do that?

You could override the `__init__` of your error to forward `type(other)` rather than requiring it at each call-site? 

In general, I'm not really sure the implicit formatting behavior of the `ZiplineError` hierarchy here is worth the confusion it causes, but given all the stuff that's using it I don't think we should change it.

peterhbromley: Oh nvm, I found {x.type} in python string formatting

peterhbromley: Ended up going with {other.__class__} so I don't have to call type(other) at each call-site

tests/pipeline/test_column.py
```diff
@@ -81,3 +84,18 @@ def test_latest(self):
...
+        (operator.le,),
+    ])
+    def test_comparison_errors(self, compare):
+        columns = TDS.columns
+        for column in columns:
+            with self.assertRaises(BoundColumnInvalidCompare):
+                compare(column, 1000)
+
+            with self.assertRaises(BoundColumnInvalidCompare):
+                compare(column, 'test')
```
quantophred: These are good tests; but I think it's worth also testing the specific error message they produce. You can probably just replace `assertRaises` with `assertRaisesRegexp`: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp

Testing the message will ensure that we know exactly how `other.__class__` gets rendered -- maybe `{other.__class__.__name__}` would be better? Maybe `{other.__class__} instance` would be more readable? You may have already experimented with variations in the REPL, but you can save future readers from having to do the same.

Testing the message also helps avoid a very common error, which I don't _think_ you've made here, but a future editor of your code might add. Can you spot the bug below?

```python
msg = (
    "Can't compare {column!s} with {other.__class__}.  Did you mean"
    "to append .latest?"
)
```

zipline/errors.py
```diff
@@ -858,3 +858,14 @@ class IncompatibleTerms(ZiplineError):
...
+
+
+class BoundColumnInvalidCompare(ZiplineError):
+    """
+    Raised when trying to use BoundColumn with a comparison operator.
+    Frequently occurs when users forget to use '.latest'.
+    """
+    msg = (
+        "Can't compare {column!s} with {other.__class__}.  Did you mean to "
+        "append .latest?"
```
quantophred: "append" was a little unclear to me at first. How about:

```
Can't compare {other.__class__} with {column!s}. Did you mean to compare with `{column!r}.latest`?
```

(Would depend on what the column's repr looks like -- might have to omit the `{column!r}`.)

quantophred: btw the `!s` is the default and technically not needed there

quantophred: On second thought, it might be worth modelling the error message after the default in Python 3.7:

```python
>>> 1 < None
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

So maybe something like:

```
'{op}' not supported between instance of '{other.__class__.__name__}' and '{column}'. Did you mean to compare with '{column}.latest'?
```

(Looks like `BoundColumn.__str__` isn't defined, so it just calls `__repr__`, and so its `!r` and `!s` are equivalent.)

quantophred: Also, might be good to either make `BoundColumnInvalidCompare` inherit from `TypeError`, or just nix the custom class and raise a regular `TypeError` with the custom message.

tests/pipeline/test_column.py
peterhbromley: In the updated test I use `assertRaisesRegexp`, the only thing I don't explicitly check for is the column name because I was running into regular expression comparison issues with some of them.  But I do explicitly check for `other.__class__.__name__`, so we know exactly how it gets rendered.

zipline/errors.py
peterhbromley: I like your updated error message a lot, but due to the column repr i changed it a little to:
```
"'{op}' not supported between instance of '{compare_type}' and '{column}'. Did you mean use '.latest' with '{column}'?"
```
because putting `.latest` directly on the column repr was a bit confusing.

peterhbromley: And I got rid of `BoundColumnInvalidCompare` in favor of `TypeError`

tests/pipeline/test_column.py
peterhbromley: Oh, and bug below is "meanto" which I have certainly done many of times!

tests/pipeline/test_column.py
```diff
@@ -91,23 +92,33 @@ def test_latest(self):
...
+            " 'int' and 'EquityPricing<US>.volume'."
+            " Did you mean to use 'EquityPricing<US>.volume.latest'?"
+        )
 
-        for column in columns:
-            int_err_msg = get_error_msg(ops[op.__name__], 'int')
-            str_err_msg = get_error_msg(ops[op.__name__], 'str')
+        with self.assertRaises(TypeError) as e:
+            column < 1000
+            self.assertEqual(str(e), err_msg)
```
quantophred: This needs to happen outside the `with` block, otherwise it won't get hit.

peterhbromley: Good find, think I fixed it by unindenting and then doing `self.assertEqual(str(e.exception), err_msg)`

tests/pipeline/test_column.py
```diff
@@ -81,3 +84,41 @@ def test_latest(self):
...
+            " 'int' and 'EquityPricing<US>.volume'."
+            " Did you mean to use 'EquityPricing<US>.volume.latest'?"
+        )
+
+        with self.assertRaises(TypeError) as e:
+            column < 1000
+        self.assertEqual(str(e.exception), err_msg)
+
+        with self.assertRaises(TypeError) as e:
+            1000 > column
```
quantophred: Hmm... I was hoping we'd always be able to display the correct operator symbol in the error message; but I guess we can't actually know which one it is without inspecting the stack, since Python just uses `__lt__` instead of `__rgt__`. (It seems reasonable that the interpreter would make use of its own stack to get that information when generating the built-in errors in Python 3.7+, though.)

Since the `<` symbol in this error message is not the same as the `>` symbol in the code that caused it, I think it's probably better to not display a symbol at all, rather than one that's only correct in half the cases. On the plus side, it should simplify the code a bit. I'll go ahead and make the change; let me know if you have other ideas, though.

PR #2550, opened 2019-09-25
Domains for Chile, Colombia, Czechia, Mexico, Peru, Poland

dmichalowicz:
Incorporate some newly released trading calendars for emerging markets.


zipline/country.py
```diff
@@ -15,7 +15,10 @@ class CountryCode(object):
     BELGIUM = code('BELGIUM')
     BRAZIL = code('BRAZIL')
     CANADA = code('CANADA')
+    CHILE = code('CHILE')
     CHINA = code('CHINA')
+    COLOMBIA = code('COLOMBIA')
+    CZECH_REPUBLIC = code('CZECHIA')
```
jmccorriston: Can we change this to 'CZECH REPUBLIC'?

dmichalowicz: This is controlled by the ISO python package we use, so unfortunately I can't change it:

```python
In [11]: from iso3166 import countries_by_name

In [12]: countries_by_name['CZECHIA']
Out[12]: Country(name='Czechia', alpha2='CZ', alpha3='CZE', numeric='203', apolitical_name='Czechia')

In [13]: countries_by_name['CZECH REPUBLIC']

KeyError: 'CZECH REPUBLIC'

In [14]: countries_by_name['CZECH_REPUBLIC']

KeyError: 'CZECH_REPUBLIC'
```

PR #2580, opened 2019-11-13
BUG: Fix bug concatenating empty chunks with strings.

ssanderson:
Fixes a bug where running chunked pipelines containing string columns would
fail if some of the chunks were completely empty. We failed because the empty
chunks would have different dtypes (all `object`) than the non-empty chunks,
because pandas can't determine a dtype for an empty DataFrame.

We fix the bug by filtering out empty chunks before concatenation.

### DIFF
--- a/tests/pipeline/test_engine.py
+++ b/tests/pipeline/test_engine.py
@@ -61,9 +61,9 @@ from zipline.pipeline.factors import (
     ExponentialWeightedMovingAverage,
     ExponentialWeightedMovingStdDev,
     MaxDrawdown,
-    Returns,
     SimpleMovingAverage,
 )
+from zipline.pipeline.filters import CustomFilter
 from zipline.pipeline.loaders.equity_pricing_loader import (
     EquityPricingLoader,
 )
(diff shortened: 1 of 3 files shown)


tests/pipeline/test_engine.py
```diff
@@ -1502,7 +1503,7 @@ def dispatcher(c):
         )
 
 
-class ChunkedPipelineTestCase(zf.WithUSEquityPricingPipelineEngine,
+class ChunkedPipelineTestCase(zf.WithSeededRandomPipelineEngine,
```
ssanderson: This test didn't actually care about using `EquityPricing` specifically; it just used the default values from the fixture and asserted that we got the same results running with or without chunks.

I switched to using `WithSeededRandomPipelineEngine` both to make that more clear, and to be get access to a string-dtype column.

zipline/pipeline/engine.py
```diff
@@ -348,7 +348,10 @@ def run_chunked_pipeline(self,
             # if we don't have to.
             return chunks[0]
 
-        return categorical_df_concat(chunks, inplace=True)
+        # Filter out empty chunks. Empty dataframes lose dtype information,
+        # which makes concatenation fail.
+        nonempty_chunks = [c for c in chunks if len(c)]
+        return categorical_df_concat(nonempty_chunks, inplace=True)
```
ssanderson: One argument I could see here is that `categorical_df_concat` should just always do this filtering. I'm not confident that that's always the right behavior for other consumers though, so I left this here. If we end up adding similar filtering for other consumers, we should probably move this into concat.

Issue #2501, opened 2019-07-09
Python syntax errors in asset code


cclauss:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment 

<details>

https://github.com/quantopian/zipline on the master branch

</details>

Now that you know a little about me, let me tell you about the issue I am
having:

# Description of Issue

It is not clear why the CI tests are missing these Python syntax errors but...

[flake8](http://flake8.pycqa.org) testing of https://github.com/quantopian/zipline on Python 3.7.1

$ __flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics__
```
./zipline/assets/assets.py:656:16: E999 SyntaxError: Generator expression must be parenthesized
                self.engine.execute(
               ^
./zipline/assets/asset_writer.py:320:20: E999 SyntaxError: Generator expression must be parenthesized
                    '%s (%s):\n  intersections: %s\n  %s' % (
                   ^
2     E999 SyntaxError: Generator expression must be parenthesized
2
```
__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. These 5 are different from most other flake8 issues which are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

evyn120:
this problem has been solved？

cclauss:
Nope.

richafrank:
Ok, the `SyntaxError`s should be fixed on master now.

PR #2582, opened 2019-11-15
MAINT: Fix SyntaxError in python 3.7+

richafrank:
as per https://bugs.python.org/issue32012
"Disallow ambiguous syntax f(x for x in [1],)"


richafrank:
For https://github.com/quantopian/zipline/issues/2501

PR #2586, opened 2019-11-18
ENH: Prototype currency conversions for EquityPricing.

ssanderson:
Initial work on adding support for currency data.

This adds the basic scaffolding for supporting currency conversions in Pipeline. Currently, we only support conversions for EquityPricing, and the only source of currency data is via an `InMemoryFXRatesReader`. Subsequent updates will provide alternative interfaces to fx rates data.

### DIFF
--- a/tests/pipeline/test_international_markets.py
+++ b/tests/pipeline/test_international_markets.py
@@ -1,5 +1,7 @@
 """Tests for pipelines on international markets.
 """
+from itertools import cycle, islice
+
 import numpy as np
 import pandas as pd
 
@@ -43,6 +45,18 @@ class WithInternationalDailyBarData(zf.WithAssetFinder):
         'XTSE': 50,   # Toronto Stock Exchange
         'XLON': 25,   # London Stock Exchange
     }
+    # Assets in these countries will be quoted in one of the listed currencies.
+    INTERNATIONAL_PRICING_CURRENCIES = {
+        'XNYS': ['USD'],
+        'XTSE': ['CAD'],
+        'XLON': ['GBP', 'EUR', 'USD'],
+    }
+    assert (
+        INTERNATIONAL_PRICING_STARTING_PRICES.keys()
+        == INTERNATIONAL_PRICING_CURRENCIES.keys()
+    )
+
+    EXCHANGE_RATES_CURRENCIES = ["USD", "CAD", "GBP", "EUR"]
 
     @classmethod
     def make_daily_bar_data(cls, assets, calendar, sessions):
@@ -67,6 +81,14 @@ class WithInternationalDailyBarData(zf.WithAssetFinder):
         for asset in assets:
             sid = asset.sid
             yield sid, base_frame + sid
+
+    @classmethod
+    def make_currency_codes(cls, calendar, assets):
+        currencies = cls.INTERNATIONAL_PRICING_CURRENCIES[calendar.name]
+        return pd.Series(
+            index=assets,
+            data=list(islice(cycle(currencies), len(assets)))
+        )
 
     @classmethod
     def init_class_fixtures(cls):
@@ -95,6 +117,7 @@ class WithInternationalDailyBarData(zf.WithAssetFinder):
             cls.daily_bar_readers[name] = InMemoryDailyBarReader.from_panel(
                 panel,
                 calendar,
+                currency_codes=cls.make_currency_codes(calendar, assets)
             )
 
 
--- a/tests/pipeline/test_international_markets.py
+++ b/tests/pipeline/test_international_markets.py
@@ -290,7 +290,10 @@ class InternationalEquityTestCase(WithInternationalPricingPipelineEngine,
 
         sessions = self.daily_bar_sessions[calendar_name]
 
-        start, end = sessions[[-17, -10]]
+        # The dates here are arbitrary. We're just running a couple days in the
+        # middle of the data that's configured in this suite.
+        execution_sessions = sessions[-17:-9]
+        start, end = execution_sessions[[0, -1]]
         result = self.run_pipeline(pipe, start, end)
 
         # Raw closes as a (dates, assets) dataframe.
@@ -321,7 +324,8 @@ class InternationalEquityTestCase(WithInternationalPricingPipelineEngine,
                 quote=target,
                 bases=np.array(currency_codes, dtype='S3'),
                 # Exchange rates used for pipeline output with label N should
-                # be from day N - 1.
+                # be from day N - 1, so shift back from `execution_sessions` by
+                # a day.
                 dates=sessions[-18:-10],
             )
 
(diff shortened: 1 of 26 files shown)


zipline/data/hdf5_daily_bars.py
```diff
@@ -240,75 +267,54 @@ def write(self, country_code, frames, scaling_factors=None):
         if scaling_factors is None:
             scaling_factors = DEFAULT_SCALING_FACTORS
 
-        with self.h5_file(mode='a') as h5_file:
```
ssanderson: This block was getting long enough that I broke it out into subroutines that write each subgroup.

zipline/data/hdf5_daily_bars.py
```diff
@@ -132,6 +146,10 @@
 END_DATE = 'end_date'
 
 
+# XXX is reserved for "transactions involving no currency".
+MISSING_CURRENCY = 'XXX'
```
peterhbromley: Is there a distinction here between these "transactions involving no currency" and what we have been using for the missing currency value (`b'\0\0\0'`), or should we maybe unify those?

zipline/data/hdf5_daily_bars.py
```diff
@@ -602,6 +690,40 @@ def asset_start_dates(self):
     def asset_end_dates(self):
         return self.dates[self._country_group[LIFETIMES][END_DATE][:]]
 
+    @lazyval
+    def _currency_codes(self):
+        return self._country_group[CURRENCY][CODE][:]
+
+    def currency_codes(self, sids):
```
peterhbromley: Mainly just for my understanding of the code: where is this `self.currency_codes()` being used (or going to be used)?

zipline/data/hdf5_daily_bars.py
ssanderson: Hmm. That's a good question. My reason for using `'XXX'` here is that it felt like a slightly more "user-friendly" missing value than `b'\0\0\0'` (which numpy displays as `b''`, which can often create confusion for users). But I agree it seems not great to have multiple "missing"/"unknown" representations for currency data.

zipline/data/hdf5_daily_bars.py
ssanderson: It's not currently being used by any of the code on this branch. I'm working on updating the equity pricing pipeline loader (https://github.com/quantopian/zipline/blob/master/zipline/pipeline/loaders/equity_pricing_loader.py#L39-L69) to be able to read this data when it's requested.

zipline/data/hdf5_daily_bars.py
ssanderson: I punted on this for the time being, but we should circle back on this with @jmccorriston before we finalize this feature.

tests/pipeline/test_international_markets.py
```diff
@@ -236,6 +271,64 @@ def test_generic_pipeline_with_explicit_domain(self, domain):
...
+            'close': EquityPricing.close.latest,
+            'close_USD': EquityPricing.close.fx('USD').latest,
+            'close_CAD': EquityPricing.close.fx('CAD').latest,
+            'close_EUR': EquityPricing.close.fx('EUR').latest,
+            'close_GBP': EquityPricing.close.fx('GBP').latest,
+        }, domain=domain)
+
+        sessions = self.daily_bar_sessions[calendar_name]
+
+        start, end = sessions[[-17, -10]]
```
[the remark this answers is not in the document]
ssanderson: This is just taking a semi-arbitrary chunk of the dates for which data has been set up in this suite. It's the same chunk we use in https://github.com/quantopian/zipline/blob/master/tests/pipeline/test_international_markets.py#L193-L198, though that test cares a bit more about the choice than this one does.

I can add some comments here to clarify the date choices.

zipline/currency.py
```diff
@@ -0,0 +1,70 @@
...
+    Parameters
+    ----------
+    code : str
+        ISO-4217 code for the currency.
+    """
+    def __new__(cls, code):
+        try:
+            return _ALL_CURRENCIES[code]
+        except KeyError:
+            iso_currency = ISO4217Currency(code)
```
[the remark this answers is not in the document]
ssanderson: So, at the moment, this type isn't actually being used very much. I had originally added it as part of a larger branch that included a "currency domain" that allowed you to run pipelines that output currency data. On that branch, this type took took the place of the `Equity` object as part of the index of pipeline outputs.

On this branch, I believe the only place that `Currency` actually gets constructed is via `BoundColumn.fx`, which accepts either a `Currency` or a string and converts the string into a currency. For that functionality we could honestly probably get by with just using the string code, but think we're going to want the more complete object eventually, so it seemed cleaner to just use it from the start.

As far as the `bytes -> Currency` conversion goes, my expectation is that once we support currency domain outputs (and I think we'll want to do so relatively soon), the conversion will happen in roughly the same places where we currently convert from sid to Equity. For performance reasons, we tend to try to stay with the "low-level" representation (i.e., arrays of int64 or S3) inside the pipeline machinery, only converting to the object represnetation at the very end of the pipeline run. Concretely, in pipeline, this currently happens [here](https://github.com/quantopian/zipline/blob/master/zipline/pipeline/engine.py#L788). I expect we'll want to do something similar with currencies.

All of which is to say that I don't think there will be much value in being lenient w/r/t the input types here. There will be relatively few construction sites for Currency objects, and they'll have well-defined inputs and outputs. I'd rather have this type be strict about what it expects and have clients be responsible for the conversion. We can always make this more lenient in the future if we end up having a mix of call sites with different convenient input types, but it's relatively hard to go the other way (once you have an object that "helps" callers, it's harder to go back and fix all the callers that are doing the wrong thing).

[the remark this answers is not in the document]
ssanderson: > Just wondering because of all of the bytes that are floating around for currency values in non-zipline places.

Also worth noting, I did end up having to add a conversion in: the `InMemoryExchangeRatesReader`: https://github.com/quantopian/zipline/pull/2586/commits/8d7cc53a02e7480166cb6540a5d713108869785f.

zipline/data/hdf5_daily_bars.py
jmccorriston: What is the context in which this would be visible to a user?

ssanderson: Currently, this value wouldn't be user-visible, but there are other contexts where we need a notion of a "missing" currency which might be user-visible, and it'd be nice to be consistent.

I think where we landed on this for discussion the other day is that any user-facing currency columns should have a missing value of `None` (like other pipeline string columns). `None` isn't an option in this particular case (we're storing arrays of fixed-length strings, which can't store python objects like `None`), so no matter what we're going to have to convert from an internal representation to a user-facing representation, which means we've got a good amount of room to be flexible here.

PR #2587, opened 2019-11-21
MAINT: Update egg names

quantophred:


richafrank:
LGTM. Seems logical to me, and based on your fixed TODO, matches what pip expects.

PR #2613, opened 2020-01-13
BUG: Fix crash in pipeline with all currency-converted data.

ssanderson:
Fixes a bug where we would crash when trying to run a pipeline that contained
only currency-converted data.

### DIFF
--- a/tests/pipeline/test_international_markets.py
+++ b/tests/pipeline/test_international_markets.py
@@ -333,6 +333,51 @@ class InternationalEquityTestCase(WithInternationalPricingPipelineEngine,
 
             assert_equal(result_2d, expected_result_2d)
 
+    @parameterized.expand([
+        ('US', US_EQUITIES, 'XNYS'),
+        ('CA', CA_EQUITIES, 'XTSE'),
+        ('GB', GB_EQUITIES, 'XLON'),
+    ])
+    def test_only_currency_converted_data(self, name, domain, calendar_name):
+        # Test running a pipeline on a domain whose assets are all denominated
+        # in the same currency.
+        pipe = Pipeline({
+            'close_USD': EquityPricing.close.fx('USD').latest,
+            'close_EUR': EquityPricing.close.fx('EUR').latest,
+        }, domain=domain)
+
+        start, end = self.daily_bar_sessions[calendar_name][-2:]
+        result = self.run_pipeline(pipe, start, end)
+
+        calendar = get_calendar(calendar_name)
+        daily_bars = self.daily_bar_data[calendar_name]
+        currency_codes = self.daily_bar_currency_codes[calendar_name]
+
+        for (dt, asset), row in result.iterrows():
+            # Subtract a day b/c pipeline output on day N should have prior
+            # day's price.
+            price_date = dt - calendar.day
+            expected_close = daily_bars[asset].loc[price_date, 'close']
+            expected_base = currency_codes.loc[asset]
+
+            expected_rate_USD = self.in_memory_fx_rate_reader.get_rate_scalar(
+                rate='mid',
+                quote='USD',
+                base=expected_base,
+                dt=price_date.asm8,
+            )
+            expected_price = expected_close * expected_rate_USD
+            assert_equal(row.close_USD, expected_price)
+
+            expected_rate_EUR = self.in_memory_fx_rate_reader.get_rate_scalar(
+                rate='mid',
+                quote='EUR',
+                base=expected_base,
+                dt=price_date.asm8,
+            )
+            expected_price = expected_close * expected_rate_EUR
+            assert_equal(row.close_EUR, expected_price)
+
     def test_explicit_specialization_matches_implicit(self):
         pipeline_specialized = Pipeline({
             'open': EquityPricing.open.latest,
--- a/zipline/data/fx/base.py
+++ b/zipline/data/fx/base.py
@@ -28,7 +28,7 @@ class FXRateReader(Interface):
             may appear multiple times.
         dts : pd.DatetimeIndex
             Datetimes for which to load rates. Must be sorted in ascending
-            order.
+            order and localized to UTC.
 
         Returns
         -------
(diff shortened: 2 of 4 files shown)


zipline/pipeline/loaders/equity_pricing_loader.py
```diff
@@ -166,7 +166,7 @@ def _inplace_currency_convert(self, columns, arrays, dates, sids):
             by_spec[column.currency_conversion].append(array)
 
         # Nothing to do for terms with no currency conversion.
-        by_spec.pop(None)
+        by_spec.pop(None, None)
```
ssanderson: This is the bugfix. The code above this groups the columns by "conversion spec", which describes what currency conversion needs to happen on the column. `None` is used for columns that don't need conversion, and we want to remove those before proceeding, since we don't need to do anything for those. All my previous tests that exercised currency conversion also included at least one non-converted column, so the `pop(None)` always succeeded. When running with all converted columns, however, we crash here because there's no `None` key.

zipline/data/fx/base.py
```diff
@@ -19,9 +22,9 @@ def get_rates(self, rate, quote, bases, dts):
             will be used by default for Pipeline API terms that don't specify a
             specific rate.
         quote : str
-            Currency code of the currency into to convert.
+            Currency code of the currency to convert into.
         bases : np.array[object]
-            Array of codes of the currencies from which to convert. A single
+            Array of codes of the currencies to convert from from. A single
```
ssanderson:
```suggestion
            Array of codes of the currencies to convert from. A single
```

zipline/data/fx/base.py
```diff
@@ -36,3 +39,34 @@ def get_rates(self, rate, quote, bases, dts):
...
+        Returns
+        -------
+        rate : np.float64
+            Exchange rate from base -> quote on dt.
+        """
+        rates_array = self.get_rates(
+            rate,
+            quote,
+            bases=np.array([base], dtype=object),
+            dts=pd.DatetimeIndex([dt], tz='UTC'),
```
quantophred: I'm not seeing much in the Pandas docs about what the `tz` argument to `DatetimeIndex` actually does -- is it just used to convert its inputs? If so, should we document its type as "convertible to datetime64", or do you want to limit what this interface officially supports?

Also, some failure modes we may or may not want to account for:

- Someone in Japan provides `dt='2020-01-01'` (or a timezone-naive Timestamp or datetime64 object, intended to represent local time) and gets results for a different market day than they expected.
- They try again with `dt=pd.Timestamp('2020-01-01', tz='Japan')` and get`TypeError: Already tz-aware, use tz_convert to convert.`

Not sure what happens (or should happen) if non-UTC timezone-aware datetimes are passed directly to `get_rates`. Looks like the `test_fx` tests all use UTC Timestamps: should that just be the requirement in the interface?

tests/pipeline/test_international_markets.py
```diff
@@ -333,6 +333,51 @@ def test_currency_convert_prices(self, name, domain, calendar_name):
 
             assert_equal(result_2d, expected_result_2d)
 
+    @parameterized.expand([
+        ('US', US_EQUITIES, 'XNYS'),
+        ('CA', CA_EQUITIES, 'XTSE'),
+        ('GB', GB_EQUITIES, 'XLON'),
+    ])
+    def test_only_currency_converted_data(self, name, domain, calendar_name):
```
quantophred: We _might_ could use a direct test for `get_rates` without any non-converted columns, in case `InMemoryFXRateReader` ever sprouts its own implementation of `get_rate_scalar`; but maybe coverage measurement is sufficient to catch that if it happens.

ssanderson: > We might could use a direct test for get_rates without any non-converted columns

This doesn't really make sense. `get_rates` doesn't know anything about pipeline columns. The interface it provides is for reading a dates-by-currencies block of of fx rate data.

That said, I think a reasonable check to add is to ensure that `get_rates_scalar` is always "just" syntactic sugar for calling `get_rates` with a single date and currency and extracting the value of the 1 x 1 array. I've added that check to an existing test here: https://github.com/quantopian/zipline/pull/2613/commits/b842b87e8ed1d88604de92bb8959a06dff559d07

zipline/data/fx/base.py
ssanderson: >  If so, should we document its type as "convertible to datetime64", or do you want to limit what this interface officially supports?

In general, `DatetimeIndex` will accept anything that pandas can convert to a datetime. For example:

```python
In [8]: pd.DatetimeIndex(['2014-01-02', 5])
Out[8]: DatetimeIndex(['2014-01-02 00:00:00', '1970-01-01 00:00:00.000000005'], dtype='datetime64[ns]', freq=None)
```

So, if the goal here is to be maximally accurate, the signature of this is probably "anything that pandas can convert to a datetime". In practice, however, this class is a relatively low-level interface. I expect its primary clients to be pipeline loaders, not end users, so I'd rather advertise the interface I actually expect, which is Timestamp or datetime64.

> Looks like the test_fx tests all use UTC Timestamps: should that just be the requirement in the interface?

One of the relatively bad mistakes we made in Zipline a long time ago is that we uniformly represent "dates" as pandas `Timestamp`s at midnight, localized to utc, of the represented date. This is an objectively bad representation for a date for a variety of reasons, but it's been our representation for a long time, and it's pretty deeply ingrained into the codebase. That's the expected representation here, and I updated the docs on `get_rates` to reflect that here: https://github.com/quantopian/zipline/pull/2613/commits/8a361e7613984645f4d5bab1a6fff2b1bea48f0b.

> Someone in Japan provides dt='2020-01-01' (or a timezone-naive Timestamp or datetime64 object, intended to represent local time) and gets results for a different market day than they expected.

Both of these cases would result in loading the rates for the date requested, which is probably the expected behavior.

> They try again with dt=pd.Timestamp('2020-01-01', tz='Japan') and getTypeError: Already tz-aware, use tz_convert to convert.

As you note, this would be an error, which I think is reasonable. We never expect to see non-utc-localized dates in this function (or, really, anywhere in zipline).

PR #2617, opened 2020-01-15
ENH: Add get_rates_columnar method to FXRatesReader.

ssanderson:
Add a new method, `get_rates_columnar` to `FXRateReader`, with a default implementation given in terms of `get_rates`.

The behavior of `get_rates_columnar` should be logically equivalent to:

```python
def get_rates_columnar(self, rate, quote, bases, dts):
    out = []
    for base, dt in zip(bases, dts):
        out.append(self.get_rate_scalar(rate, quote, base, dt)
    return np.array(out)
```

### DIFF
--- a/zipline/data/fx/base.py
+++ b/zipline/data/fx/base.py
@@ -9,22 +9,64 @@ DEFAULT_FX_RATE = sentinel('DEFAULT_FX_RATE')
 
 
 class FXRateReader(Interface):
+    """
+    Interface for reading foreign exchange (fx) rates.
+
+    An FX rate reader contains one or more distinct "rates", each of which
+    corresponds to a collection of mappings from (quote, base, dt) ->
+    float. The value produced for a given (quote, base, dt) triple is the
+    exchange rate to use when converting from ``base`` to ``quote`` on ``dt``.
+
+    The specific set of rates contained in a particular reader is
+    user-defined. We infer no particular semantics from their names, other than
+    that they are distinct rates. Examples of possible rate names might be
+    things like "bid", "mid", and "ask", or "london_close", "tokyo_close",
+    "nyse_close".
+
+    Implementations of :class:`FXRateReader` must provide at least one method::
+
+        def get_rates(self, rate, quote, bases, dts):
+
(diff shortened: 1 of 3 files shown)


zipline/data/fx/base.py
```diff
@@ -63,10 +103,44 @@ def get_rate_scalar(self, rate, quote, base, dt):
...
+
+        unique_bases, bases_ix = np.unique(bases, return_inverse=True)
+        unique_dts, dts_ix = np.unique(dts.values, return_inverse=True)
+        rates_2d = self.get_rates(
+            rate,
+            quote,
+            unique_bases,
+            pd.DatetimeIndex(unique_dts, tz='utc')
+        )
+        return rates_2d[dts_ix, bases_ix]
```
peterhbromley: I could be wrong about this, but it seems like in the columnar case we might load a bunch more data than we need since `get_rates` computes the cartesian product but the column is only a subset of that.  I assume this doesn't really matter, but was curious if that is a concern at all.

ssanderson: > it seems like in the columnar case we might load a bunch more data than we need since get_rates computes the cartesian product but the column is only a subset of that.

Yup, this is definitely a possibility. The worst case scenario here would be something like a columnar query where every day has a different currency in it. We'd end up loading all the currencies for every day, even though we only needed one day of each currency. I'm not too worried about this because:

1. In our expected workloads, I expect that if we need a currency on day N, we'll probably need it for nearby days. In particular, most datasets will have static currencies, which means the most common case will be that if we ever need a currency we'll need it every day.
2. Even in the pathological case, it's not clear that we can do much better than what we're doing here, because we're going to want to use compression and chunking in the HDF5 files. What that means in practice is that the data arrays get broken up into regular sized pieces, and each piece gets compressed and stored separately. If you load any value out of the array, you have to decompress and load the entire chunk holding the value, so what dominates performance isn't so much the number of **values** read: it's the number of distinct **chunks** read. So, most likely, even if we tried to more precise reads here, we'd end up doing the same amount of IO, and we'd likely have higher overheads from trying to read smaller chunks of the file.

PR #2621, opened 2020-01-18
MAINT: The function get_stock_dividends shouldn't read from tuple 

samatix:
The function get_stock_dividends shouldn't read from tuple as the order of the columns is not known. This lead to errors when upgrading the dependencies to higher releases. 
Adding a dictionary (fast) to get the id for the each datum is safer.

### DIFF
--- a/zipline/data/data_portal.py
+++ b/zipline/data/data_portal.py
@@ -297,6 +297,13 @@ class DataPortal(object):
             if self._first_trading_day is not None else None
         )
 
+        # Store the location of the dividends table fields
+        if self._adjustment_reader is not None:
+            stock_dividend_payouts_fields = self._adjustment_reader.conn.execute(
+                "PRAGMA table_info(stock_dividend_payouts);").fetchall()
+
+            self._dividends_fields = {field[1]: field[0] for field in stock_dividend_payouts_fields}
+
     def _ensure_reader_aligned(self, reader):
         if reader is None:
             return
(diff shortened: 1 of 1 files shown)


zipline/data/data_portal.py
```diff
@@ -297,6 +297,17 @@ def __init__(self,
             if self._first_trading_day is not None else None
         )
 
+        # Store the location of the dividends table fields
+        if self._adjustment_reader is not None:
+            stock_dividend_payouts_fields = self._adjustment_reader.conn.\
```
ssanderson: Hmm. I'm not excited about adding more places in `DataPortal` where we're depending on using `adjustment_reader.conn`. The intent of the `SQLiteAdjustmentReader` class is to provide an interface to split and dividend data that hides the implementation details of how those values are actually stored. The fact that we're making SQL queries directly here is a bit awkward in that regard.

It looks to me like the only call site for `get_stock_dividends` is in the [BenchmarkSource class](https://github.com/quantopian/zipline/blob/c825927d6e34e70659e543b6bc1564b9ec9de54d/zipline/sources/benchmark_source.py#L148-L159), where we call it to check if a stock has any stock dividends (and if so, we raise an error). My guess is we're doing that because the benchmark source doesn't properly account for stock dividends.

So, given the above concerns, I think my ideal solution here would be to remove `DataPortal.get_stock_dividends` entirely in favor of a narrower method on `SQLiteAdjustmentReader` that can be used to query for whether or not a given sid has stock dividends. Thoughts?

PR #2622, opened 2020-01-19
DEP: Replacing the assertRaisesRegexp which is deprecated by assertRaisesRegex

samatix:
This deprecation will enable the upgrade to add support for python 3.7 and higher
https://docs.python.org/3/library/unittest.html


ssanderson:
@samatix are you still planning to come back to this? If not, I or one of the other Zipline maintainers can push this over the line.

samatix:
> @samatix are you still planning to come back to this? If not, I or one of the other Zipline maintainers can push this over the line.

Definitely @ssanderson, I'll do the amendment this evening.

PR #2631, opened 2020-01-23
Introducing pre-commit hooks

willianpaixao:
In a intention to increase code quality and reducing faulty commits,
pre-commit help checking the staged code for simple static code
analysis.

The hooks added in `.pre-commit-config.yaml` are pretty much basic and [many more](https://pre-commit.com/hooks.html) can be added.


.pre-commit-config.yaml
```diff
@@ -6,3 +6,15 @@ repos:
     -   id: end-of-file-fixer
     -   id: check-merge-conflict
     -   id: flake8
+
+-   repo: https://github.com/PyCQA/bandit
```
willianpaixao: One useful hook that will perform static code analysis, searching for bug and vulnerabilities. More information on the [official website](https://bandit.readthedocs.io/en/latest/).

.pre-commit-config.yaml
```diff
@@ -6,3 +6,15 @@ repos:
...
+    rev: '1.6.2'
+    hooks:
+    -   id: bandit
+
+-   repo: local
```
willianpaixao: Another suggested hook that runs the existing test suite before applying the commit, possibly catching a build failure in TravisCI

.pre-commit-config.yaml
```diff
@@ -0,0 +1,20 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: 'v2.4.0'
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-merge-conflict
+    -   id: flake8
```
ssanderson: Elsewhere internally at Quantopian we're using:

```
  - repo: https://gitlab.com/pycqa/flake8
    rev: 3.7.7
    hooks:
    - id: flake8
```

which has the benefit that it allows us to pin a specific version of flake8. I'm not sure if there's a way to do the using the flake8 plugin supplied by `pre-commit`.

.pre-commit-config.yaml
ssanderson: Hmm. From looking at the bandit site, it seems like the checks it performs are mostly geared toward web applications. I'm not sure the checks it performs make sense for a library like zipline. Is there a clear kind of error that you think this would catch/prevent in zipline? I think I'd prefer not to add this hook for now.

.pre-commit-config.yaml
ssanderson: Given how long the zipline test suite takes to run right now, I don't think I'd want to require people to run the test suite before every commit. I'd rather work toward speeding up the test suite and/or our CI first.

.pre-commit-config.yaml
willianpaixao: Yeah, I actually used that way before too, but since it was added to the default list of default hooks I switched for simplicity.

But sure, I can revert o that way too.

.pre-commit-config.yaml
willianpaixao: I see, I think the majority of the tests are web security so I understand your point. I can remove this hook.

PR #2634, opened 2020-01-24
Github Actions CI Setup

danisim:
Setup for GH actions to replace Travis/Appveyor config. 

Note:
- Did not setup coveralls which was part of the travis build


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,120 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Pip cache
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/.pip/pip_np$NUMPY_VERSION
```
ssanderson: Given that we're using a `key`, I don't think we should need to encode the `numpy_version` into the pip cache directory anymore.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,120 @@
...
+          if [[ "$RUNNER_OS" == "Linux" ]]; then
+              sudo apt-get install python3-venv
+          fi
+          python3 -m venv testenv
+          source ./testenv/bin/activate
+          conda install -c anaconda libgfortran=3.0
+          pip install pip cython numpy pandas scipy ta-lib certifi -c etc/requirements.txt --cache-dir="$HOME/.cache/.pip/pip_np$NUMPY_VERSION"
+          etc/dev-install --cache-dir="$HOME/.cache/.pip/pip_np$NUMPY_VERSION"
+          pip freeze | sort
+          echo "::set-env name=PATH::$PATH"
```
ssanderson: This seems like a lot more complexity than I would have expected. A couple quick suggestions:

1. Given that we're using the `setup-python` action to make a fresh python for each of these builds, I don't think it should be necessary to create a separate virtualenv for the build.
2. I don't think we should need to use conda for anything here anymore. In particular, installing `anaconda` brings in a **lot** of stuff we don't want (anaconda is python + all the packages that anaconda the company curates, c.f. https://stackoverflow.com/questions/30034840/what-are-the-differences-between-conda-and-anaconda).
3. I don't think we want to be pre-installing all the packages you currently have listed in the top-level `pip install`. In particular, we probably **don't** want to be installing `scipy` and `pandas` at the same time as we install `numpy`, since the reason we pre-install numpy and Cython is so that those packages are available at build time for scipy and pandas (and other packages that depend on them at build time). I'd expect that the right sequence of installs here should look something like: `pip install pip==<pinned pip version> setuptools=<pinned setuptools version>`, `pip install -r etc/requirements_build.in -c etc/requirements.txt`, `pip install -e .[all] -c etc/requirements.txt`.

danisim: @ssanderson Sorry you caught this PR in a wip state but appreciate the feedback!
I set up a virtualenv to get around this error that I was running into when I tried to pip install directly to the actions host.
```ERROR: Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/lib/python2.7/dist-packages/cython.py'
Consider using the `--user` option or check the permissions.```

I tried to strip away the conda installs and as you warned me last week, pip doesn't handle the dependencies as well  and I was running into a bunch of dependency constraint errors.

PR #2646, opened 2020-02-05
MAINT: Transpose arrays in fx artifact for better compression

peterhbromley:


zipline/data/fx/hdf5.py
```diff
@@ -203,56 +201,55 @@ def get_rates(self, rate, quote, bases, dts):
...
+        # Transpose everything to maintain dts as row labels, currencies as col
+        # labels which is expected everywhere else.
+        return out.transpose()
 
 
 class HDF5FXRateWriter(object):
     """Writer class for HDF5 files consumed by HDF5FXRateReader.
     """
-    def __init__(self, group):
+    def __init__(self, group, date_chunk_size):
```
ssanderson: I think it would probably be reasonable to give this a default value so that zipline users don't need to do the same tuning work that we did on this.

zipline/data/fx/hdf5.py
```diff
@@ -203,56 +201,55 @@ def get_rates(self, rate, quote, bases, dts):
...
+            (len(self.currencies), slice_end - slice_begin + 1),
             np.nan,
         )
 
-        # Read data into all but the last row/column of the buffer.
-        dataset.read_direct(
-            buf[:-1],
-            np.s_[slice_begin:slice_end],
-        )
+        buf[:, :-1] = dataset[:, slice_begin:slice_end]
```
ssanderson: If we're not using `read_direct` anymore, we might as well allocate both an extra row and an extra column and use the extra row/column trick for handling both cases rather than using it for just one of them.

zipline/data/fx/hdf5.py
```diff
@@ -203,56 +201,55 @@ def get_rates(self, rate, quote, bases, dts):
...
+        #   2. col_ixs contains -1 for dts before the start of self.dts.
         #
         # If either of the above cases obtains, we want to return NaN for the
         # corresponding output locations.
 
         # We handle (1) by reading raw data into a buffer with one extra
-        # row. When we then apply the row index to permute the raw data into
-        # the correct order, any rows with values of -1 will pull from the
-        # extra row, which will always contain NaN>
+        # column. When we then apply the column index to permute the raw data
```
ssanderson: I think this comment got a bit scrambled? `(1)` still refers to the possibility of nonexistent rows, but then we talk about columns here.

zipline/data/fx/hdf5.py
```diff
@@ -203,56 +201,55 @@ def get_rates(self, rate, quote, bases, dts):
...
+        # into the correct order, any columnss with values of -1 will pull from
```
ssanderson:
```suggestion
        # into the correct order, any columns with values of -1 will pull from
```

peterhbromley: :+1:

zipline/data/fx/hdf5.py
peterhbromley: Now that we use the extra row/column trick for both cases, I reworded the comment.

zipline/data/fx/hdf5.py
peterhbromley: :+1:

zipline/data/fx/hdf5.py
peterhbromley: I added the default as a global defined at the beginning of the file.

zipline/data/fx/hdf5.py
```diff
@@ -267,11 +260,22 @@ def write(self, dts, currencies, data):
             Iterator of (rate, quote_currency, array) tuples. Each array
             should be of shape ``(len(dts), len(currencies))``, and should
             contain a table of rates where each column is a timeseries of rates
-            mapping its column label's currency to ``quote_currency``.
+            mapping its column label's currency to ``quote_currency``. The
+            arrays that are actually written to the HDF5 file will be
+            transposed to have shape ``(len(currencies), len(dts))`` so that
+            similar values are in C-contiguous order, which improves overall
+            compression.
```
ssanderson: This is an implementation detail of the file format. I wouldn't include that in this docstring.

PR #2659, opened 2020-02-19
Inform Pipeline Columns of Currency Awareness

dmichalowicz:
Add the ability to instantiate pipeline `Column`s with a `currency_aware` flag, indicating whether or not it makes sense to call `.fx`/convert the column's values to another currency. For example, OHLC data is currency-aware, but Volume is not.

### DIFF
--- a/tests/pipeline/test_international_markets.py
+++ b/tests/pipeline/test_international_markets.py
@@ -402,6 +402,16 @@ class InternationalEquityTestCase(WithInternationalPricingPipelineEngine,
             sessions[-1],
         )
 
+    def test_cannot_convert_volume_data(self):
+        with self.assertRaises(TypeError) as exc:
+            EquityPricing.volume.fx('EUR')
+
+        assert_equal(
+            str(exc.exception),
+            'The .fx() method cannot be called on EquityPricing.volume '
+            'because it does not produce currency-denominated data.',
+        )
+
     def check_expected_latest_value(self, calendar, col, date, asset, value):
         """Check the expected result of column.latest from a pipeline.
         """
--- a/zipline/pipeline/data/dataset.py
+++ b/zipline/pipeline/data/dataset.py
@@ -28,7 +28,7 @@ from zipline.utils.input_validation import (
     ensure_dtype,
     expect_types,
 )
-from zipline.utils.numpy_utils import NoDefaultMissingValue
+from zipline.utils.numpy_utils import float64_dtype, NoDefaultMissingValue
 from zipline.utils.preprocess import preprocess
 from zipline.utils.string_formatting import bulleted_list
 
@@ -45,11 +45,20 @@ class Column(object):
                  dtype,
                  missing_value=NotSpecified,
                  doc=None,
-                 metadata=None):
+                 metadata=None,
+                 currency_aware=False):
+        if currency_aware and dtype != float64_dtype:
+            raise ValueError(
+                'Columns cannot be constructed with currency_aware={}, '
+                'dtype={}. Currency aware columns must have a float64 dtype.'
+                .format(currency_aware, dtype)
+            )
+
         self.dtype = dtype
         self.missing_value = missing_value
         self.doc = doc
         self.metadata = metadata.copy() if metadata is not None else {}
+        self.currency_aware = currency_aware
 
     def bind(self, name):
         """
@@ -61,6 +70,7 @@ class Column(object):
             name=name,
             doc=self.doc,
             metadata=self.metadata,
+            currency_aware=self.currency_aware,
         )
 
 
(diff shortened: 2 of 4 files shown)


zipline/pipeline/data/dataset.py
```diff
@@ -104,14 +115,19 @@ def __get__(self, instance, owner):
         We don't bind to datasets at class creation time so that subclasses of
         DataSets produce different BoundColumns.
         """
+        if self.currency_aware:
+            currency_conversion = None
+        else:
+            currency_conversion = Disallowed
```
dmichalowicz: The reason I went with using a sentinel here is because I thought it might be confusing/excessive to have both a `currency_conversion` parameter and a new `currency_aware` parameter, so I tried to combine the two.

zipline/pipeline/data/dataset.py
```diff
@@ -250,8 +266,24 @@ def fx(self, currency):
...
+            raise TypeError(
+                'The .fx method cannot be called on {} because {} does not '
+                'have a float64 dtype.'.format(self, name)
+            )
+
         conversion = self._currency_conversion
-        if conversion is not None and conversion.currency == currency:
+        if conversion is Disallowed:
+            raise TypeError(
+                'The .fx method cannot be called on {} because {} is not '
```
ssanderson: Thinking about the wording here:

- I'm not sure why we're including both `self` and a variant on `self.name` in this message. I think `self.qualname` should be sufficient to tell the reader what column they used.
- Do we want "is not currency-aware" to be the user-facing description of the problem here? That's a very mechanical description of the issue vs. something more conceptual like "does not produce currency-denominated data". cc @jmccorriston for thoughts on the language here.

tests/pipeline/test_international_markets.py
```diff
@@ -402,6 +402,16 @@ def test_explicit_specialization_matches_implicit(self):
             sessions[-1],
         )
 
+    def test_cannot_convert_volume_data(self):
+        with self.assertRaises(TypeError) as exc:
+            Pipeline({'volume': EquityPricing.volume.fx('EUR').latest})
```
ssanderson:
```suggestion
             EquityPricing.volume.fx('EUR')
```

We don't need to construct a Pipeline around this for the error to trigger, so no reason to include that in the code under test.

tests/pipeline/test_international_markets.py
```diff
@@ -402,6 +402,16 @@ def test_explicit_specialization_matches_implicit(self):
...
+
+        assert_equal(
+            str(exc.exception),
+            'The .fx method cannot be called on EquityPricing.volume because '
```
ssanderson: Elsewhere in the API for errors like this we write `.fx()` rather than `.fx`. See, for instance, [float64_only](https://github.com/quantopian/zipline/blob/master/zipline/pipeline/factors/factor.py#L368-L374)

zipline/pipeline/data/dataset.py
```diff
@@ -153,7 +166,18 @@ def __new__(cls,
                 name,
                 doc,
                 metadata,
-                currency_conversion):
+                currency_conversion,
+                currency_aware):
+        if currency_aware and dtype != float64_dtype:
+            raise ValueError(
+                '{} cannot be constructed with currency_aware={}, dtype={}. '
```
ssanderson: Might be useful to include the dataset in this message too. `name` here will just be the column name, so if there are multiple datasets with similar column names, the name won't be helpful.

zipline/pipeline/data/dataset.py
```diff
@@ -251,7 +281,14 @@ def fx(self, currency):
             into ``currency``.
         """
         conversion = self._currency_conversion
-        if conversion is not None and conversion.currency == currency:
+        currency_aware = self._currency_aware
+
+        if not currency_aware:
```
ssanderson: This can just be `if not self._currency_aware`. The reason I'm binding `conversion` as a local variable here is because the if-statement below exceeds 80 chars if we use `self._currency_conversion` twice.

zipline/pipeline/data/dataset.py
```diff
@@ -267,6 +304,13 @@ def currency_conversion(self):
         """
         return self._currency_conversion
 
+    @property
+    def currency_aware(self):
+        """
+        Whether or not this column produces currency-denominated data.
+        """
+        return self._currency_aware
```
ssanderson: Are there clients that need to be able to access this attribute?

zipline/pipeline/data/dataset.py
ssanderson: We should probably also add a quick test case for this.

zipline/pipeline/data/dataset.py
dmichalowicz: Yes I was using this downstream.

dmichalowicz:
@ssanderson This is good to go unless you had anything else?

tests/pipeline/test_dataset.py
```diff
@@ -23,6 +24,10 @@ class LargeDataSet(DataSet):
     })
 
 
+class InvalidColumnDataset(DataSet):
+    currency_aware_date = Column(dtype=datetime64ns_dtype, currency_aware=True)
```
ssanderson: Talked about this in person, but I think we want this statement to error rather than the attribute access.

PR #2673, opened 2020-03-24
MAINT: removing dead code

### LINKED ISSUE #2321: Unused variables in several methods
Dear Zipline Maintainers,

while working on a non-related issue, I stumbled upon several functions that have an unused argument. Is that for some backward compatibility? Is it safe to remove? Follow some examples:

`zipline/data/bundles/core.py:`
``` python
def adjustment_db_relative(bundle_name, timestr, environ=None):
    return bundle_name, timestr, 'adjustments.sqlite'

def adjustment_db_relative(bundle_name, timestr, environ=None):
    return bundle_name, timestr, 'adjustments.sqlite'
```

Here the variable `environ` will never be used.

willianpaixao:
Fix #2321

### DIFF
--- a/zipline/data/bundles/core.py
+++ b/zipline/data/bundles/core.py
@@ -403,8 +403,7 @@ def _make_bundle_core():
 
                 daily_bar_writer.write(())
                 minute_bar_writer = BcolzMinuteBarWriter(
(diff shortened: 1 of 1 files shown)


richafrank:
Thanks @willianpaixao !

PR #2678, opened 2020-03-31
Conda forge prep

richafrank:
- Conda packaging is much easier if we allow newer bcolz.
- Bump pip-tools and bcolz in CI requirements lockfile


richafrank:
> I'm assuming the updates the dependency sources are caused by the bump to pip-tools?

Yep, that's right.

Issue #2474, opened 2019-06-01
cannot install zipline


jaykim82:
Dear Zipline Maintainers,

Before I tell you about my issue, let me describe my environment:

# Environment

<details>

* Operating System: Windows 10
* Python Version: 3.4.4
* How did you install Zipline: through pycharm 
* Python packages:
C:\Users\jay_k>pip freeze
DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).
certifi==2019.3.9
chardet==3.0.4
cycler==0.10.0
idna==2.8
kiwisolver==1.1.0
lxml==4.3.3
matplotlib==2.2.4
numpy==1.16.4
pandas==0.22.0
pandas-datareader==0.7.0
pyparsing==2.4.0
python-dateutil==2.8.0
pytz==2019.1
pywin32==221
requests==2.21.0
six==1.12.0
urllib3==1.24.3
wrapt==1.11.1
</details>

Now that you know a little about me, let me tell you about the issue I am
having problem with installing zipline

# Description of Issue
When I try to install zipline through pycharm, an error occured and can't install zipline. You can see the detail message from pycharm below.

------------------------------------------------------------------------------------------------
Executed command:
  pip install zipline
Error occured:
  ERROR: Complete output from command python setup.py egg_info:
Proposed solution:
  Try to run this command from the system terminal. Make sure that you use the correct version of 'pip' installed for your Python interpreter located at 'C:\Program Files (x86)\Python34\python.exe'.
Command output:
Collecting zipline
  Using cached https://files.pythonhosted.org/packages/be/59/8c5802a7897c1095fdc409fb557f04df8f75c37174e80d2ba58c8d8a6488/zipline-1.3.0.tar.gz
Requirement already satisfied: pip>=7.1.0 in c:\program files (x86)\python34\lib\site-packages (from zipline) (19.1.1)
Requirement already satisfied: setuptools>18.0 in c:\program files (x86)\python34\lib\site-packages (from zipline) (18.2)
Collecting Logbook>=0.12.5 (from zipline)
  Using cached https://files.pythonhosted.org/packages/f6/83/20fc0270614919cb799f76e32cf143a54c58ce2fa45c19fd38ac2e4f9977/Logbook-1.4.3.tar.gz
[...]

PR #2679, opened 2020-03-31
List hdf5 as a dependency

daraul:
Possibly related to #2474, this package was a requirement for me to install zipline on arch. I'm not sure that the package name for RHEL and ubuntu is correct, however.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -44,7 +44,7 @@ binary dependencies from ``apt`` by running:
 
 .. code-block:: bash
 
-   $ sudo apt-get install libatlas-base-dev python-dev gfortran pkg-config libfreetype6-dev hdf5
+   $ sudo apt-get install libatlas-base-dev python-dev gfortran pkg-config libfreetype6-dev hdf5-tools
 
 On recent `RHEL-derived`_ derived Linux distributions (e.g. Fedora), the
 following should be sufficient to acquire the necessary additional
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -44,7 +44,7 @@ binary dependencies from ``apt`` by running:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -44,26 +44,27 @@ binary dependencies from ``apt`` by running:
 
 .. code-block:: bash
 
-   $ sudo apt-get install libatlas-base-dev python-dev gfortran pkg-config libfreetype6-dev
+   $ sudo apt-get install libatlas-base-dev python-dev gfortran pkg-config libfreetype6-dev hdf5
```
richafrank: Is this the right package name? `apt-get install hdf5` fails for me with `Unable to locate package hdf5`.

We've been `apt-get` installing `hdf5-tools`.

docs/source/install.rst
```diff
@@ -44,26 +44,27 @@ binary dependencies from ``apt`` by running:
...
 .. code-block:: bash
 
-   $ pacman -S lapack gcc gcc-fortran pkg-config
+   $ pacman -S lapack gcc gcc-fortran pkg-config hdf5
 
 There are also AUR packages available for installing `ta-lib
 <https://aur.archlinux.org/packages/ta-lib/>`_, an optional Zipline dependency.
 Python 2 is also installable via:
 
+
```
richafrank: Would you update the macOS instructions to include brew installing `hdf5` too?

docs/source/install.rst
daraul: Ah, I figured it might be wrong. I'll update.

docs/source/install.rst
daraul: Sure!

richafrank:
Thanks for the updates! I also don't know the correct package name for RHEL-derived linux distros, but if it's wrong, we can fix it.

PR #2680, opened 2020-04-01
H5 Minute Reader Bug in Python 3

dmichalowicz:
If an equities or futures update file was written in python 2 but read in python 3, `pandas.read_hdf` would error trying to read the tz attr in the file.

### DIFF
--- a/zipline/data/minute_bars.py
+++ b/zipline/data/minute_bars.py
@@ -21,6 +21,7 @@ from textwrap import dedent
 from lru import LRU
 import bcolz
 from bcolz import ctable
+import h5py
 from intervaltree import IntervalTree
 import logbook
 import numpy as np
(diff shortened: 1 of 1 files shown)


etc/requirements_locked.txt
```diff
@@ -35,7 +35,7 @@ flask-cors==2.1.3
 flask==1.1.1              # via flask-cors
 funcsigs==1.0.2           # via mock, python-interface
 futures==3.2.0            # via tornado
-h5py==2.7.1
+h5py==2.10.0
```
dmichalowicz, line 38: This update was necessary to read the file's attr in the test...`2.7.1` was failing with a strange numpy c error.

zipline/data/minute_bars.py
```diff
@@ -1373,7 +1374,36 @@ class H5MinuteBarUpdateReader(MinuteBarUpdateReader):
...
+        with h5py.File(path, 'r') as f:
+            updates = f['updates']
+            values = updates['block0_values']
+            items = updates['axis0']
+            major = updates['axis1']
+            minor = updates['axis2']
+
+            # Our current version of h5py is unable to read the tz attr in the
+            # tests as it was written by HDFStore. This is fixed in version
+            # 2.10.0 of h5py, but that requires >=Python3.7, so until then we
```
ssanderson, line 1394: I think 2.10 only requires python 3.7 on conda.

dmichalowicz:
Right, thanks, I'll fix and then merge.

PR #2691, opened 2020-04-24
ENH: Add Filter.if_else and ComputableTerm.fillna.

ssanderson:
This PR implements several improvements to "universal terms", and implements several new such terms to enable conditional-based terms.

- Refines our existing machinery for implementing behaviors common to multiple `ComputableTerm` subclasses. The existing `DownsampledMixin` and `AliasedMixin` classes no longer require separate boilerplate in each of `Factor`, `Filter`, and `Classifier`. Instead, the boilerplate has been moved into `ComputableTerm`, where it can appear just once. Factor, Filter, and Classifier now provide a method called `_principal_computable_term_type`, which is used to implement the necessary boilerplate. The existing `Slice` class has been refactored to use this machinery as well, which means `Factor[asset]` is now a proper `Factor` instance (same for `Filter` and `Classifier`). We might also consider refactoring `NumericalExpression` (which implements all our binary operators) to use the new machinery, but that's a bigger refactor than I wanted to do here.

  In the code, I've taken to calling these shared mixins "universal mixins". In addition to the refactored existing mixins, this PR adds two new such mixin, `IfElseMixin` and `ConstantMixin`. See below for details

- Adds a new method, `Filter.if_else()`, that wraps `np.where`. This allows
  users to create expressions that conditionally draw from the outputs of one
  of two terms. This is implemented using the newly-standardized "universal"
  term machinery, similarly to downsample() and alias(), and Slice.

  This is implemented using the new universal `IfElseMixin`.

- Adds a new method, `ComputableTerm.fillna()`, implemented in terms of
  `if_else()`. `fillna` allows users to fill missing data with either a
  constant value, or values from another term. It supports both 2d and 1d
  terms. The constant case is implemented using a new universal `ConstantMixin`.

### DIFF
--- a/tests/pipeline/test_computable_term.py
+++ b/tests/pipeline/test_computable_term.py
@@ -244,17 +244,17 @@ class FillNATestCase(BaseUSEquityPipelineTestCase):
         self.assertEqual(message, expected_message)
 
     def test_bad_inputs(self):
-        def dtype_for_str(s):
-            return str(np.array([s]).dtype)
+        def dtype_for(o):
+            return np.array([o]).dtype
 
         self.should_error(
             lambda: Floats().fillna('3.0'),
             TypeError,
             "Fill value '3.0' is not a valid choice for term Floats with"
             " dtype float64.\n\n"
-            "Coercion attempt failed with: Cannot cast array from dtype('{}')"
-            " to dtype('float64') according to the rule 'same_kind'"
-            .format(dtype_for_str('3.0'))
+            "Coercion attempt failed with: Cannot cast array from {!r}"
+            " to {!r} according to the rule 'same_kind'"
+            .format(dtype_for('3.0'), np.dtype(float))
         )
 
         self.should_error(
@@ -262,9 +262,9 @@ class FillNATestCase(BaseUSEquityPipelineTestCase):
             TypeError,
             "Fill value '2014-01-02' is not a valid choice for term Dates with"
             " dtype datetime64[ns].\n\n"
-            "Coercion attempt failed with: Cannot cast array from dtype('{}')"
-            " to dtype('<M8[ns]') according to the rule 'same_kind'"
-            .format(dtype_for_str('2014-01-02'))
+            "Coercion attempt failed with: Cannot cast array from {!r}"
+            " to {!r} according to the rule 'same_kind'"
+            .format(dtype_for('2014-01-02'), np.dtype('M8[ns]'))
         )
 
         self.should_error(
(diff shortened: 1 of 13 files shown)


zipline/pipeline/mixins.py
```diff
@@ -548,26 +529,149 @@ def make_downsampled_type(cls, other_base):
...
+            mask=term.mask,
+            dtype=term.dtype,
+            missing_value=term.missing_value,
+            window_safe=term.window_safe,
+            ndim=1,
+        )
+
+    def __repr__(self):
+        return "{parent_term}[{asset}]".format(
+            type=type(self).__name__,
```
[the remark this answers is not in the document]
ssanderson: This is copy/pasted from the old location, but yeah, looks like this isn't being used

tests/pipeline/test_computable_term.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+class AltInts(Classifier):
+    inputs = ()
+    window_length = 0
+    dtype = int64_dtype
+    missing_value = -1
+
+
+class FillNATestCase(BaseUSEquityPipelineTestCase):
```
peterhbromley, line 81: Are there any tests for `.if_else()` directly?  I know `.fillna` uses `.if_else()` but would it be worth testing for `.if_else()` specifically?

zipline/pipeline/mixins.py
```diff
@@ -543,31 +567,144 @@ def skip_this_input():
...
+    ----------
+    term : zipline.pipeline.Term
+        The term from which to extract a column of data.
+    asset : zipline.assets.Asset
+        The asset corresponding to the column of `term` to be extracted.
+
+    Notes
+    -----
+    Users should rarely construct instances of `Slice` directly. Instead, they
+    should construct instances via indexing, e.g. `MyFactor()[Asset(24)]`.
```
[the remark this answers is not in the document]
ssanderson: Yeah. This is a feature we added to enable computations like `beta`, where you want to run a linear regression of each asset's returns against the regression of a specific asset (e.g., SPY).

tests/pipeline/test_computable_term.py
ssanderson: I hadn't added these b/c I wasn't sure about the API for `if_then` (in an earlier draft it was `Factor.if_else` rathern thatn `Filter.if_then`. I think I'm pretty happy with this though, so I can add some standalone tests.

ssanderson: Added a suite of direct tests for `if_else` here: https://github.com/quantopian/zipline/pull/2691/files#diff-a6c64e76bf199a87f0d8fd09291d8db1R1128.

PR #2697, opened 2020-04-29
ENH: Add daily summary methods to Factor.

ssanderson:
Adds several methods to Factor for producing daily summary statistics. The
following methods are supported:

- mean
- median
- stddev
- max
- min
- sum
- notnull_count

Each of these methods produces a term with ndim=1, meaning that it produces a
single scalar value per day.

### DIFF
--- a/tests/pipeline/base.py
+++ b/tests/pipeline/base.py
@@ -123,6 +123,17 @@ class BaseUSEquityPipelineTestCase(WithTradingSessions,
             hooks=NoHooks(),
         )
 
+    def run_terms(self, terms, initial_workspace, mask):
+        start_date, end_date = mask.index[[0, -1]]
+        graph = ExecutionPlan(
+            domain=US_EQUITIES,
+            terms=terms,
+            start_date=start_date,
+            end_date=end_date,
+        )
+
+        return self.run_graph(graph, initial_workspace, mask)
+
     def check_terms(self,
                     terms,
                     expected,
@@ -133,15 +144,8 @@ class BaseUSEquityPipelineTestCase(WithTradingSessions,
         Compile the given terms into a TermGraph, compute it with
         initial_workspace, and compare the results with ``expected``.
         """
-        start_date, end_date = mask.index[[0, -1]]
-        graph = ExecutionPlan(
-            domain=US_EQUITIES,
-            terms=terms,
-            start_date=start_date,
-            end_date=end_date,
-        )
+        results = self.run_terms(terms, initial_workspace, mask)
 
-        results = self.run_graph(graph, initial_workspace, mask)
         for key, (res, exp) in dzip_exact(results, expected).items():
             check(res, exp)
 
--- a/zipline/pipeline/factors/factor.py
+++ b/zipline/pipeline/factors/factor.py
@@ -52,11 +52,20 @@ from zipline.pipeline.sentinels import NotSpecified, NotSpecifiedType
 from zipline.pipeline.term import AssetExists, ComputableTerm, Term
 from zipline.utils.functional import with_doc, with_name
 from zipline.utils.input_validation import expect_types
-from zipline.utils.math_utils import nanmean, nanstd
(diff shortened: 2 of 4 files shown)


zipline/pipeline/factors/factor.py
```diff
@@ -382,6 +391,73 @@ def mathfunc(self):
...
+        return nansum(a, axis=1)
+
+    @staticmethod
+    def notnull_count(a, missing_value):
+        return (~is_missing(a, missing_value)).sum(axis=1)
+
+    names = {k for k in locals() if not k.startswith('_')}
+
+
+def summary_method(name):
```
peterhbromley, line 429: I'm a bit confused what's going on in this function, but that might be because of my lack of understanding of how `clsdict` eventually gets used.  So this method returns a function that returns a `DailySummary` factor?  I guess my question is why do we have to return the function rather than just the class directly.

tests/pipeline/base.py
```diff
@@ -133,15 +144,8 @@ def check_terms(self,
...
         initial_workspace, and compare the results with ``expected``.
         """
-        start_date, end_date = mask.index[[0, -1]]
-        graph = ExecutionPlan(
-            domain=US_EQUITIES,
-            terms=terms,
-            start_date=start_date,
-            end_date=end_date,
-        )
+        results = self.run_terms(terms, initial_workspace, mask)
```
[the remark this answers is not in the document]
ssanderson: I'm using the factored out function [here](https://github.com/quantopian/zipline/pull/2697/files#diff-df725806e27aae338bf6ebf3b67b8942R1595) to be able to just run the terms without the default logic of making assertions about the values. Specifically, at that use-site, I'm asserting that two different ways of constructing an expression produce the same result.

zipline/pipeline/factors/factor.py
ssanderson: This function defines a method that gets added to the `Factor` class, and that method constructs a `DailySummary` instance.

My original draft of this code had a bunch of new `Factor` methods written conventionally, which looked like:

```python

    @expect_types(mask=(Filter, NotSpecifiedType))
    @float64_only
    def mean(self, mask=NotSpecified):
        """Create 1D factor computing the average value of ``self``.

        Returns
        -------
        mean : zipline.pipeline.Factor
        """
        return DailySummary(
            summary_funcs.mean,
            self,
            self.dtype,
            self.missing_value,
            mask=mask,
        )

    @expect_types(mask=(Filter, NotSpecifiedType))
    @float64_only
    def median(self, mask=NotSpecified):
        """Make a factor that computes the median value of ``self``.

        Returns
        -------
        median : zipline.pipeline.Factor
        """
        return DailySummary(
            summary_funcs.median,
            self,
            self.dtype,
            self.missing_value,
            mask=mask
        )

    @expect_types(mask=(Filter, NotSpecifiedType))
    @float64_only
    def sum(self, mask=NotSpecified):
        """Create 1D factor that sums the values of ``self``.

        Returns
        -------
        sum : zipline.pipeline.Factor
        """
        return DailySummary(
            summary_funcs.sum,
            self,
            self.dtype,
            self.missing_value,
            mask=mask,
        )
```
etc.

If you squint at all of those though, you can see that they all have the same structure, which is a method that returns `DailySummary({function}, self, self.dtype, self.missing_value, mask=mask)`. Rather than have that structure duplicated 7 times, I instead wrote a function constructs the method as a function of the name of the method it calls.

In the code that manipulates `clsdict` below, `clsdict` is the namespace of attributes of the `Factor` class, which includes all the function objects that act as methods of Factor. When we do

```python
    clsdict.update(
        {name: summary_method(name) for name in summary_funcs.names},
    )
```

that's equivalent to doing

```
mean = summary_method('mean')
median = summary_method('median')
max = summary_method('max')
```

which is in turn equivalent to doing:

```
def max(self, mask):
    ... body of function emitted by summary_method
```

PR #2717, opened 2020-05-29
Fixed syntax in Clip docstring.

jmccorriston:


llllllllll:
I think we also have this in the `clip` method, can you update that also?

jmccorriston:
done. look ok?

llllllllll:
lgtm

PR #2723, opened 2020-06-11
Window length 0 fails to copy input

llllllllll:
Some terms mutate their inputs, but by not copying the data before passing them
to a term, the input could be corrupted and then used in other terms that didn't
expect the mutated input.

This change uses the same reference counting mechanism to only copy when needed
that is applied to windowed terms.


jmccorriston:
lgtm

PR #2726, opened 2020-06-18
Assert fixes

richafrank:
- Assert that float pipeline outputs are equal with a tolerance
- Don't assert_almost_equal with string arrays


gerrymanoim:
LGTM

PR #2727, opened 2020-06-19
ENH: Support Pipeline arithmetic that combines more than 32 terms

richafrank:


richafrank:
@llllllllll I'm curious your thoughts on the resulting tree - not very balanced...

llllllllll:
As an experiment, can you see what the performance looks like if we _never_ merge nodes? I personally believe that arithmetic folding should happen in the engine so that we can make high level scheduling decisions, for example: memory vs speed. I think that folding in the engine would also make it easier to balance, since we have the whole tree at that point.

richafrank:
Local benchmarking of never-merging experiment with following pipeline:
```python
from quantopian.pipeline import Pipeline
from quantopian.pipeline.factors import Returns, CustomFactor
terms_list = [Returns(window_length=i) for i in range(2, 33)]  # 31 terms
def make_pipeline():
    combined_factor = sum(terms_list)
    return Pipeline(
        columns={
            'combined_factor': combined_factor
        }
    )
```
31 terms: merging: 10 years: ~127s
31 terms: no merging: 10 years: ~127s

I tried also with 2 terms:
2 terms: merging: 10 years: 12s
2 terms: no merging: 10 years: 11s

From talking with @llllllllll about further optimization, there's more work to be done, so going to merge this for now, since it unlocks some larger pipelines.

Issue #2627, opened 2020-01-21
Remove Implicit Dependency on Benchmarks and Treasury Returns


ssanderson:
## Background

Zipline currently requires two special data inputs for simulations: "Benchmark Returns", which are used to calculate the "Alpha" and "Beta" metrics, among other things, and "Treasury Curves", which were at one time used as the "Risk Free Rate", which was part of the Sharpe Ratio calculation. 

Since these inputs are required by all simulations, we implicitly fetch them from third party API sources if they're not provided by users. We get treasury data from the US Federal Reserve's API, and we get benchmarks from IEX.

## Problems

Implicitly fetching benchmarks and treasuries causes many problems:

- Implicitly fetching means that running simulations requires an internet connection. We try to make this less painful by caching downloaded results and re-using them when possible, but this is only a partial fix, and it means that many users don't notice the implicit download until it starts causing mysterious problems.

- The APIs we fetch from sometimes fail, leading to confusing behavior for users and spurious bug reports for Zipline maintainers.

- The APIs we fetch from sometimes change in incompatible ways, which breaks older versions of Zipline. This is currently the case for the IEX API we use to fetch benchmarks, resulting in issues like:

  - https://github.com/quantopian/zipline/issues/2480
  - https://github.com/quantopian/zipline/issues/2607
  - https://github.com/quantopian/zipline/issues/2445

- Our default benchmark is US-centric. We default to using SPY as the benchmark, which only makes sense in the US (and even then, only makes sense if you have also historical dividends for SPY, which many users don't have).

## Proposed Solution

I think we should remove these implicit dependencies from Zipline. Treasuries we should just remove, since they're not actually used anymore. Figuring out what to do with benchmarks is a bit trickier.

### Treasuries

Removing treasuries is relatively straightforward because we no longer actually use them. A quick scan of our GitHub issues turns up these issues that should be fixed by the removal:

- https://github.com/quantopian/zipline/issues/324
- https://github.com/quantopian/zipline/issues/119
- https://github.com/quantopian/zipline/issues/144
- https://github.com/quantopian/zipline/issues/2422

I've opened a PR at https://github.com/quantopian/zipline/pull/2626 to finally remove all traces of the treasury subsystem.

### Benchmarks

[...]

PR #2734, opened 2020-07-16
Benchmark cleanup

richafrank:
Test is to ensure we're not broken as described in https://github.com/quantopian/zipline/issues/2627#issuecomment-608987852

### DIFF
--- a/tests/finance/test_risk.py
+++ b/tests/finance/test_risk.py
@@ -317,6 +317,24 @@ class TestRisk(zf.WithBenchmarkReturns, zf.ZiplineTestCase):
         )
         self.assertEqual(test_period['sharpe'], 0.0)
 
+    def test_sharpe_value_when_benchmark_null(self):
+        # Sharpe is displayed as '0.0' instead of np.nan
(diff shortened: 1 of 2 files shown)


tests/finance/test_risk.py
```diff
@@ -317,6 +317,24 @@ def test_sharpe_value_when_null(self):
         )
         self.assertEqual(test_period['sharpe'], 0.0)
 
+    def test_sharpe_value_when_benchmark_null(self):
+        # Sharpe is displayed as '0.0' instead of np.nan
+        null_returns = factory.create_returns_from_list(
+            [0.0]*251,
```
gerrymanoim, line 323: Why 251? Aren't there 252 trading days?

gerrymanoim:
Small question, but LGTM

tests/finance/test_risk.py
richafrank: I didn't investigate further, but seems like not on NYSE in 2006. This came from `trading_calendar.sessions_in_range(...)`.

Issue #2322, opened 2018-10-04
Zipline "latest" build has conflicting requirements


marketneutral:
I am having trouble getting Zipline and Alphalens to work together in the same environment. I am filing this issue here, because I believe the problem is that the Zipline **latest** build has inconsistent requirements not caught/tested.

First, the **stable** build for Zipline and the current Alphalens package do indeed play well together. I can see that this is what is installed in Quantopian Research prod.

Concretely, following the zipline [travis CI yml file](https://github.com/quantopian/zipline/blob/master/.travis.yml), the **stable** build is:

```
conda create -n testenv -y python=3.5 numpy=1.11.3 pandas=0.18.1 scipy=0.17.1 libgfortran=3.0 pip
source activate testenv
python -m pip install -r requirements_stable.txt --no-cache-dir
python -m pip install -r requirements_blaze_stable.txt --no-cache-dir
pip install zipline==1.3.0 --no-cache-dir
pip install alphalens==0.3.2 --no-cache-dir
```

where the requirements files are those in `zipline/etc` (I renamed them here because they are [modified in place via](https://github.com/quantopian/zipline/blob/master/ci/travis/overwrite_requirements.sh) `sed` depending on stable/latest).

This works well

```
Python 3.5.5 | packaged by conda-forge | (default, Jul 23 2018, 23:45:11) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import zipline
>>> import alphalens
>>> import blaze
>>> import pandas
```

However, when I try the **latest** zipline build,

```
conda create -n testenv_latest -y python=3.5 numpy=1.14.1 pandas=0.22.0 scipy=1.0.0 libgfortran=3.0 pip
source activate testenv_latest
python -m pip install -r requirements_latest.txt --no-cache-dir
python -m pip install -r requirements_blaze_latest.txt --no-cache-dir
pip install zipline==1.3.0 --no-cache-dir
pip install alphalens==0.3.2 --no-cache-dir
```

Where the `requirements_*latest.txt` files are those modified in place.

This does **not** work

```
Python 3.5.5 | packaged by conda-forge | (default, Jul 23 2018, 23:45:11) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import zipline
[...]

PR #2739, opened 2020-07-21
BLD: Use newer statsmodels with newer scipy in the test matrix

richafrank:
Let's see if tests pass.

See https://github.com/quantopian/zipline/issues/2322


llllllllll:
lgtm

PR #2755, opened 2020-08-14
MAINT: Use loc instead of deprecated ix

richafrank:


gerrymanoim:
Seems reasonable, thanks!

Issue #2761, opened 2020-08-20
 Issue (and solution) for running run_algorithm without benchmark_returns


ghost:
Dear Zipline Maintainers,

Running run_algorithm without benchmark_returns raises the following NoBenchmark exception:

NoBenchmark: Must specify either benchmark_sid or benchmark_returns.

The issue can be resolved by fixing a bug/removing 'not' in  the from_returns method of the BenchmarkSpec class:

    @classmethod
    def from_returns(cls, benchmark_returns):
        return cls(
            benchmark_returns=benchmark_returns,
            benchmark_file=None,
            benchmark_sid=None,
            benchmark_symbol=None,
            no_benchmark=benchmark_returns is not None,
        )

Sincerely,
Flounder

flounderhost:
Great, glad to hear.

arisliang:
In which version the bug is fixed? I'm running zipline 1.4.1, and still got this error from the beginner tutorial https://www.zipline.io/beginner-tutorial.html.

PR #2762, opened 2020-08-23
More benchmark fixes

richafrank:
Fixes https://github.com/quantopian/zipline/issues/2761 and adds the docstring missing in https://github.com/quantopian/zipline/issues/2627#issuecomment-675127746.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -17,12 +17,11 @@ from operator import itemgetter
 import tarfile
 
 import matplotlib
-from nose_parameterized import parameterized
 import pandas as pd
 
 from zipline import examples
 from zipline.data.bundles import register, unregister
-from zipline.testing import test_resource_path
+from zipline.testing import test_resource_path, parameter_space
 from zipline.testing.fixtures import (
     WithTmpDir,
     ZiplineTestCase,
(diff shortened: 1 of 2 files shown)


tests/test_examples.py
```diff
@@ -73,9 +73,20 @@ def test_example(self, example_name):
...
                 'ZIPLINE_ROOT': self.tmpdir.getpath('example_data/root'),
             },
-            benchmark_returns=self.benchmark_returns,
+            benchmark_returns=benchmark_returns,
         )
-        expected_perf = self.expected_perf[example_name]
+        if benchmark_returns is not None:
+            expected_perf = self.expected_perf[example_name]
+        else:
+            # With no benchmark, expect zero results for these metrics:
```
llllllllll, line 81: should we precompute this in the init like `self.no_benchmark_expected_perf`?

zipline/utils/run_algo.py
```diff
@@ -513,7 +515,8 @@ def resolve(self, asset_finder, start_date, end_date):
                 benchmark_returns = None
             except SymbolNotFound:
                 raise _RunAlgoError(
-                    "Symbol %s as a benchmark not found in this bundle."
+                    "Symbol %r as a benchmark not found in this bundle."
+                    % self.benchmark_symbol
```
llllllllll, line 519: +1

llllllllll:
lgtm

tests/test_examples.py
richafrank: Good idea. Done.

RichardDale:
Given this is such a common-use case I would suggest this probably warrants a v1.41 release?  It's certainly been a common topic from our user base.

richafrank:
> Given this is such a common-use case I would suggest this probably warrants a v1.41 release? It's certainly been a common topic from our user base.

👍 I was thinking the same thing.

PR #2763, opened 2020-08-24
Documentation and bundles fixes

richafrank:
Before:
image

After:
image

Before:
image

After:
image


zipline/data/bundles/core.py
```diff
@@ -586,8 +586,8 @@ def clean(name,
             def should_clean(name):
                 dt = from_bundle_ingest_dirname(name)
                 return (
-                    (before is not None and dt < before) or
-                    (after is not None and dt > after)
+                    (before is None or dt < before) and
+                    (after is None or dt > after)
```
richafrank, line 590: @llllllllll What is the intended way to clean out all the ingestions for a bundle? I tried with no args, but nothing happened. However, my change looks to fail the relevant test...

llllllllll: You need to pass either `--before`, `--after` or `--keep-last` to specify which ingests you would like to delete.

llllllllll, line 590: why invert all of these conditions?

docs/source/conf.py
```diff
@@ -72,7 +72,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['.static']
+html_static_path = []
```
llllllllll, line 75: do we not have any images or anything?

zipline/__main__.py
```diff
@@ -317,13 +320,6 @@ def run(ctx,
         benchmark_spec=benchmark_spec,
     )
 
-    if output == '-':
```
llllllllll, line 320: why can we remove this logic?

zipline/__main__.py
```diff
@@ -112,6 +112,9 @@ def _(*args, **kwargs):
     return d
 
 
+DEFAULT_BUNDLE = 'quantopian-quandl'
```
llllllllll, line 115: :+1:

zipline/data/bundles/core.py
richafrank: Ah, ok, that was my confusion. I expected that passing none of them would delete everything.

richafrank: This was my attempt to fix the noop `clean` commands (when I passed no args). I'll instead raise `BadClean` in that case.

docs/source/conf.py
richafrank: Not there at least. We have 2 pngs in docs/tutorial_files.

```
Running Sphinx v3.1.2
making output directory... done
WARNING: html_static_path entry '.static' does not exist
```

zipline/__main__.py
richafrank: The final lines of `_run` (called above) are:
```python
    if output == '-':
        click.echo(str(perf))
    elif output != os.devnull:  # make the zipline magic not write any data
        perf.to_pickle(output)

    return perf
```

zipline/data/bundles/core.py
```diff
@@ -578,6 +578,9 @@ def clean(name,
             if e.errno != errno.ENOENT:
                 raise
             raise UnknownBundle(name)
+
+        if before is after is keep_last is None:
```
llllllllll, line 582: :+1:

PR #2767, opened 2020-09-08
BLD: Fix caching in Github Actions builds

gerrymanoim:
Makes a few changes to fix our usage of `actions/cache` work:

1. Installs `wheel` as part of CI so that we actually build wheels and they get cached properly
2. Fixes the environment variable passed to `hashFiles` so that we're actually hashing the right requirements file (this was failing silently before)
3. Fixes `restore-keys` to be more specific

While we are now actually caching/restoring wheels correctly, this does not speed up builds all that much. Most of our time in install is spent building/installing zipline itself.


richafrank:
Looks like an improvement to me! But I do still hope that we can cut out the 3-5 minutes of pip install time...

PR #2776, opened 2020-09-22
DOC: Update source activate command to conda activate

leonarduschen:
Zipline installation docs currently uses `source activate` and `source deactivate` - which only work on Linux and MacOS. So I thought we might want to change that to `conda activate` and `conda deactivate` as they are supported on all OS (available on Conda 4.4 onwards https://www.anaconda.com/blog/conda-4-6-release)

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -152,7 +152,7 @@ by using the command
 
 .. code-block:: bash
 
-    $ source activate env_zipline
+    $ conda activate env_zipline
 
 You can install Zipline by running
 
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -164,4 +164,10 @@ To deactivate the ``conda`` environment:
 
 .. code-block:: bash
 
-    (env_zipline) $ source deactivate
+    (env_zipline) $ conda deactivate
+
+.. note::
```
leonarduschen, line 169: Actually I'm not entirely sure whether we want to use the `note` directive or just a normal text here.

leonarduschen:
I'm not sure why the test fails. I thought it was the Miniconda version used in AppVeyor test, but it doesn't explain why the test fails only in this particular environment:

Environment: PYTHON_VERSION=3.6, PANDAS_VERSION=0.22.0, NUMPY_VERSION=1.14.1, SCIPY_VERSION=1.0.0, STATSMODELS_VERSION=0.9.0, PANDAS_DATAREADER_VERSION=0.4.0, DASK_VERSION=0.17.1

richafrank:
Thanks @leonarduschen ! I restarted the failed AppVeyor build, and it succeeded. And I think the `note` directive looks nice.

PR #2834, opened 2022-01-20
Update data_portal.py

idiomaticrefactoring:
refactoring code with var unpack which is more pythonic, concise, readable and efficient; how do think this change which has practical value？


andycwang:
***@***.***邮箱联系我，谢谢！

adamLutPhi (approved):
File Changes _reviewed_

adamLutPhi (commented):
File Changes _reviewed_

PR #2855, opened 2023-12-26
Use GITHUB_OUTPUT envvar instead of set-output command as the latter is deprecated

arunsathiya:
`save-state` and `set-output` commands used in GitHub Actions are deprecated and [GitHub recommends using environment files](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/).

This PR updates the usage of `set-output` to `$GITHUB_OUTPUT`

Instructions for envvar usage from GitHub docs:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,6 @@ on:
 
 jobs:
   build-and-test:
-
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -10,59 +10,58 @@ on:
...
+        run: |
+          echo ::set-env name=PIP_CONSTRAINT::etc/requirements_locked.txt
+      - name: Set Lockfile py36
+        if: matrix.python-version == 3.6
+        run: |
+          echo ::set-env name=PIP_CONSTRAINT::etc/requirements_py36_locked.txt
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          echo "dir=$(pip cache dir)" >> $GITHUB_OUTPUT
```
arunsathiya, line 52: This is the only change being proposed in this PR. The rest of the changes are formatting corrections made by my IDE.

PR #2861, opened 2025-06-17
Diagram representation of the zipline's codebase

ivanmilevtues:
This PR contains high-level diagram representation of the zipline codebase. You can see how the diagrams render in Github here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/spleeter/on_boarding.md

These diagrams are designed to help new people get up-to-speed with the zipline codebase. I would love to hear what do you think of diagram first documentation. Do you have a current on-boarding process at the moment with which we can help? We believe that diagrams give a good high-level context of how the project works, then a person can focus on a component of interest.

Any feedback is more than welcome! We've just released a free GitHub Action that can automatically update the diagrams and this way keep them always up-to-date.

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

I would usually start by opening a discussion, but they are not enabled for this repository so I opened a PR instead!


ivanmilevtues:
Hey all, a quick update on our side. The generation engine for codeboarding is now open-source. You can check it our here: https://github.com/CodeBoarding/CodeBoarding

